PR #28, opened 2016-12-03
Add Member.colour

fwrs:


src/model/guild.rs
```diff
@@ -853,6 +853,17 @@ impl Member {
         self.nick.as_ref().unwrap_or(&self.user.name)
     }
 
+    /// Calculates the member's colour.
+    pub fn colour(&self) -> Option<::utils::Colour> {
+        let roles = self.roles.clone();
```
zeyla: The roles is a Vec of `RoleId`s which derive Copy, they don't need to be cloned

src/model/guild.rs
```diff
@@ -853,6 +853,17 @@ impl Member {
...
+        for n in &roles {
+            if let Some(r) = n.find() {
+                return Some(r.colour);
+            }
+        }
+        None
+    }
```
zeyla: The problem is that roles are stored in ascending order by Id. If a user has 5 roles, and the first role exists in the cache, its colour value (say 0xff0000) will be returned. Roles with a higher hierarchy that have their own colour to override the first colour's value won't be taken into consideration.

So what should be done is reversing the vec, check if it has a _non-default_ colour (`Colour::default()`), and if so, return the role's colour. If none have a unique colour, _then_ return a `Colour::default()`. If doing this, then there is no way that this function can return None, so the signature can be changed to just `-> Colour`.

src/model/guild.rs
```diff
@@ -853,6 +853,17 @@ impl Member {
         self.nick.as_ref().unwrap_or(&self.user.name)
     }
 
+    /// Calculates the member's colour.
+    pub fn colour(&self) -> Option<::utils::Colour> {
```
zeyla: Let's import `Colour` for direct use at the top of the file

PR #50, opened 2016-12-31
Implement context message queueing

fwrs:
WIP


src/client/context.rs
```diff
@@ -1445,6 +1450,17 @@ impl Context {
...
+    /// Adds a string to message queue, which is sent joined by a newline
+    /// when context goes out of scope.
+    ///
+    /// **Note**: Only works if the `say` method works.
+    pub fn queue(&mut self, content: &str) -> &mut Self {
+        let to_push = "\n".to_owned() + content;
+        self.queue.push_str(&to_push);
+
+        self
+    }
```
zeyla: This can be faster by pushing the `\n` char and then pushing `content`, avoiding producing a temporary String

src/client/context.rs
```diff
@@ -1445,6 +1450,17 @@ impl Context {
         }
     }
 
+    /// Adds a string to message queue, which is sent joined by a newline
+    /// when context goes out of scope.
+    ///
+    /// **Note**: Only works if the `say` method works.
```
zeyla: This comment should be clarified to:
```rust
/// **Note**: Only works in a context where a channel is present. Refer to
/// [`say`] for a list of events where this is applicable.
///
/// [`say`]: #method.say
```

src/client/context.rs
```diff
@@ -3,6 +3,8 @@ use std::collections::HashMap;
 use std::fmt::Write as FmtWrite;
 use std::io::Read;
 use std::sync::{Arc, Mutex};
+#[cfg(feature="extras")]
+use std::ops::ShlAssign;
```
zeyla: Place this before the `use super::CACHE` below for consistency (grouping conditional uses together)

src/client/context.rs
```diff
@@ -115,6 +119,7 @@ impl Context {
             data: data,
             shard: shard,
             login_type: login_type,
+            queue: String::new()
```
zeyla: Trailing comma

src/client/context.rs
```diff
@@ -1880,3 +1896,20 @@ impl Context {
...
+    fn drop(&mut self) {
+        if !self.queue.is_empty() {
+            let _ = self.say(&self.queue);
+        }
+    }
+}
+
+/// Allows the `<<=` operator to be used to queue messages.
+#[cfg(feature="extras")]
+impl<'a> ShlAssign<&'static str> for &'a mut Context {
```
zeyla: This being static will only allow static strings to be passed in. This should be `ShlAssign<&'a str>`, so that users can pass slices of strings: `ctx <<= &string`

zeyla: May as well make this `<<`?

src/utils/builder/search.rs
```diff
@@ -129,9 +129,10 @@ impl SortingOrder {
 ///         return Ok(());
 ///     }
 ///
-///     let guild = message.guild().unwrap();
+///     let guild = message.guild_id().unwrap().find().unwrap();
```
zeyla: What's wrong with `Message::guild()` here?

PR #55, opened 2017-01-06
Add rohrkatze_blue to colour.rs

bippum:
hello


hsiW:
I was too ded to add the labels :<

PR #71, opened 2017-03-26
Change the framework to use attributes

acdenisSK:
(edit: Changed the formatting heavily,)
This's a very unfinished pr and as such needs all of these checked:
### main things to do first
- [x] Getting the right data out of the function provided by the user and make the appropriate output function for the `Command` trait (this is due to the reason that i want to mostly mimic the functionality of the current `command!` macro, and one of it is the choice of ignoring the message/args arguments)
- [x] Arg parsing (again, a functionality from the `command!` macro)
- [x] Splitting out the whole `Command` trait with the right data from the attributes and from the function.
- [ ] Change the framework to adjust it to the new design.
### minor
- [x] Change tabs to spaces from my lazyness.
- [x] Change the wildcard versions defined in `Cargo.toml` for `serenity_attributes` to actual versions so the crate can be uploaded to [crates.io](https://crates.io).

### Discarded ideas
- [x] Possibly change the _whole_ framework usage to attributes, but the question lies, where should the attribute for configuring the framework go? (Well, i think just the commands part of it is fine to use attributes, so i'm checking this off and the one below)
- [x] Add support for `Before`/`After` hooks. (Look up)
- [x] Add support for rest args, e.g `fn command(..., arg1: String, rest_after_arg1: (String|Vec<String>)) { ... }` (Code was becoming too complex, TOO complex, and wasn't really easy to do, so i'm just discarding this.)


SirDoctors:
the seventh item on this list sounds like something that could have been fixed with an editor setting change,

acdenisSK:
Oh actually it seems vscode was smart enough to not put tab characters in places of spaces.
Guess i was a bit lucky, but thanks anyway.

SirDoctors:
#ConflictingFiles
src/ext/framework/mod.rs

acdenisSK:
I know, but i still have to change many things in the framework to see if all of this'd work as expected.
So it's low priority. But i guess, it won't hurt to do it now

acdenisSK:
Okay i accidentally fucked up a bit, oh well.

acdenisSK:
Okay well, the branch's pretty much messed up now (what i get for rebasing wrong) but also, i can't really figure out the way i *should* be editing the framework to adapt it to attributes. As no matter what i try (4 times if you mind), it still won't work.
So i'm just gonna make a new pr, however, with just the `serenity_attributes` and a bare bones `command.rs` and let it as a up for grabs for those who could manage to change the framework in a way i was intending to have in this pr (mainly the `Framework::register` method), and'd work.

For those who can't edit the repo directly, because of them not being "maintainers", you can just go to my fork, make a local repo out of it, do the edits, then by doing a pr to my fork. Everything else about this'll be handled by me afterwards.

(And yes, i'll credit you.)

PR #72, opened 2017-04-16
Update the way errors are handled in dispatch

fwrs:


src/model/channel/message.rs
```diff
@@ -195,7 +195,7 @@ impl Message {
 
         // First replace all user mentions.
         for u in &self.mentions {
-            result = result.replace(&u.mention(), &u.distinct());
+            result = result.replace(&u.mention(), &format!("@{}", u.distinct()));
```
zeyla: If we're going to change this, let's make `format!("@{}", u.distinct())` faster:

```rust
let mut at_distinct = String::with_capacity(38); // 1 + max length of username (32) + 1 + length of discrim (4)
at_distinct.push('@');
at_distinct.push_str(&u.name);
at_distinct.push('#');
at_distinct.push_str(&u.discriminator);
```

compared to the committed code, this is nearly 90% faster:
```
test tests::one ... bench:          20 ns/iter (+/- 0)
test tests::two ... bench:         180 ns/iter (+/- 2)
```

src/ext/framework/mod.rs
```diff
@@ -265,29 +300,112 @@ impl Framework {
...
+                   args: usize,
+                   to_check: String,
+                   built: &String) -> Option<DispatchError> {
+        if self.configuration.ignore_bots && message.author.bot {
+            Some(DispatchError::IgnoredBot)
+        } else if self.configuration.ignore_webhooks && message.webhook_id.is_some() {
+            Some(DispatchError::WebhookAuthor)
+        } else if !self.configuration.owners.contains(&message.author.id) {
+            let ref bucket = command.bucket;
+            if let Some(rate_limit) = bucket.clone().map(|x| self.ratelimit_time(x.as_str(), message.author.id.0)) {
```
zeyla: Empty line before code blocks

src/ext/framework/mod.rs
```diff
@@ -265,29 +300,112 @@ impl Framework {
...
+    }
+
+    #[doc(hidden)]
+    fn should_fail(&mut self,
+                   mut context: &mut Context,
+                   message: &Message,
+                   command: &Arc<Command>,
+                   args: usize,
+                   to_check: String,
+                   built: &String) -> Option<DispatchError> {
```
zeyla: `&str` instead?

src/ext/framework/mod.rs
```diff
@@ -265,29 +300,112 @@ impl Framework {
...
+                if args < x as usize {
+                    return Some(DispatchError::NotEnoughArguments { min: x, given: args });
                 }
-            },
-            AccountType::Any => {}
+            }
+
+            if let Some(x) = command.max_args {
+                if args > x as usize {
+                    return Some(DispatchError::TooManyArguments { max: x, given: args });
```
zeyla: Let's put this across multiple lines:
```rust
return Some(DispatchError::TooManyArguments {
    max: x,
    given: args,
});
```

PR #77, opened 2017-05-11
Trim content before parsing args

### LINKED ISSUE #76: Spaces cause part of command to be sent as an argument.
After typing a prefix it seems that any amount of whitespace is cut out while checking for the command to invoke. However, when parsing arguments the whitespace still remains and the command is cut up.

For example, given the following command: 

```rust
command!(cmd(_ctx, msg, args) {
    if args.len() == 1 {
        println!("Got argument '{}'", args[0]);
    } else {
        println!("No arguments.");
    }
});
```

I get the following behavior, assuming a prefix of + (With _ being spaces because I don't know how to make spaces not compact)

`+cmd` => "No arguments"
`+_cmd` => "No arguments"
`+__cmd` => "Got argument 'd'"
`+___cmd` => "Got argument 'md'"
etc.

Roughsketch:
Fixes #76 according to my tests. However, if there is a better way to do this feel free to change it.

### DIFF
--- a/src/ext/framework/mod.rs
+++ b/src/ext/framework/mod.rs
@@ -477,10 +477,13 @@ impl Framework {
                         let after = self.after.clone();
                         let groups = self.groups.clone();
 
+                        let content = message.content[position..].to_owned();
+                        let content = content.trim();
+
                         let args = if command.use_quotes {
-                            utils::parse_quotes(&message.content[position + command_length..])
+                            utils::parse_quotes(&content[command_length..])
                         } else {
-                            message.content[position + command_length..]
+                            content[command_length..]
                                 .split_whitespace()
                                 .map(|arg| arg.to_owned())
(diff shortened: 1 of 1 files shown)


src/ext/framework/mod.rs
```diff
@@ -477,10 +477,13 @@ impl Framework {
                         let after = self.after.clone();
                         let groups = self.groups.clone();
 
+                        let content = message.content[position..].to_owned();
+                        let content = content.trim();
+
                         let args = if command.use_quotes {
```
zeyla: If you move the `content` binding to within a new scope within `args`, then you can avoid `to_owned()`.

e.g.:
```rust
let args = {
    let content = ....

    if command.use_quotes {
        //
    } else {
        //
    }
};
```

zeyla:
Thanks

PR #79, opened 2017-05-13
Added the faded_purple colour

hsiW:
`faded_purple` was added as a colour and the typo in `meibe_pink` being `meibi_pink` has been corrected.


zeyla:
Thanks

PR #80, opened 2017-05-14
Don't skip everyone role when checking channel overwrites.

Roughsketch:
I came across this issue while trying to check permissions using `permissions_for`. I found out that if I made the **everyone** role disable attaching images, my bot would still attempt to attach them even with the check. This seemed to be because `permissions_for` skips the **everyone** role explicitly. All this PR does is make it check that role as well.


zeyla:
Thanks. Tests for permissions would be good, I'll get around to them sometime unless someone else does

PR #81, opened 2017-05-14
handle member join messages (type 7)

fwrs:


src/client/dispatch.rs
```diff
@@ -111,7 +151,7 @@ fn dispatch_message(context: Context,
                     message: Message,
                     event_store: &Arc<RwLock<EventStore>>) {
     if let Some(handler) = handler!(on_message, event_store) {
-        thread::spawn(move || (handler)(context, message));
+        thread::spawn(move || (handler)(context, improve_content(message)));
```
zeyla: See my general review comment for more info on this.

src/client/rest/mod.rs
```diff
@@ -1182,6 +1182,7 @@ pub fn get_messages(channel_id: u64, query: &str)
                            || client.get(&url))?;
 
     serde_json::from_reader::<HyperResponse, Vec<Message>>(response).map_err(From::from)
+        .map(|mut x| x.iter_mut().map(|x| x.clone().improved_content()).collect::<Vec<Message>>())
```
zeyla: This clone feels unnecessary. You should be able to change the `iter_mut` into an `into_iter` to remove the clone.

zeyla:
Thanks for the PR.

I changed `Message::improved_content` to accept `&mut self`, and
renamed it to `Message::transform_content`. I've added a test for
deserializing message type 7.

Originally I did say that the content transformations can be done in
the `rest` module, but due to an incoming commit changing the purpose
of the module I've moved this to `ChannelId::{message, messages}`.

PR #90, opened 2017-05-18
Add support for retrieving invites with counts

hsiW:
Previously retrieving an invite with the `?with_counts=true` wasn't possible.

Added support to `get_invite` for retrieving an invite with the counts, this requires the user to pass true to the function.
The counts include `approximate_presence_count` and `approximate_member_count` which have been added to the `Invite` stuct.

Done in response to: https://github.com/zeyla/serenity/issues/83


zeyla:
Thanks

PR #92, opened 2017-05-19
Fix call to VoiceManager.join in examples/06_voice

barzamin:
`VoiceManager.join(&mut self, guild_id: Option<GuildId>, ...)`
has been changed to
`VoiceManager.join<G>(&mut self, guild_id: G, ...) where G: Into<GuildId>`.
 
We should no longer wrap `guild_id` with `Some`.


zeyla:
Thanks

PR #93, opened 2017-05-22
Sort help by group and command names

Roughsketch:
Tested both with_embeds and plain and they seem to work. Let me know if something needs to be changed or reworked.

On a side note, I thought about manually moving the "Ungrouped" section to be the last in the list but am not sure if this would be desired behavior.


src/framework/help_commands.rs
```diff
@@ -146,7 +146,11 @@ pub fn with_embeds(ctx: &mut Context,
                 .description("To get help with an individual command, pass its \
                               name as an argument to this command.");
 
-            for (group_name, group) in groups {
+            let mut group_names: Vec<_> = groups.keys().collect();
```
zeyla: Can we update this to turbofish syntax?

src/framework/help_commands.rs
```diff
@@ -155,9 +159,14 @@ pub fn with_embeds(ctx: &mut Context,
 
                 let mut no_commands = true;
 
-                for (n, cmd) in remove_aliases(&group.commands) {
+                let commands = remove_aliases(&group.commands);
+                let mut command_names: Vec<_> = commands.keys().collect();
```
zeyla: And here.

src/framework/help_commands.rs
```diff
@@ -255,7 +264,11 @@ pub fn plain(ctx: &mut Context,
                       name as an argument to this command.\n\n"
         .to_string();
 
-    for (group_name, group) in groups {
+    let mut group_names: Vec<_> = groups.keys().collect();
```
zeyla: Here.

src/framework/help_commands.rs
```diff
@@ -264,9 +277,14 @@ pub fn plain(ctx: &mut Context,
 
         let mut no_commands = true;
 
-        for (n, cmd) in remove_aliases(&group.commands) {
+        let commands = remove_aliases(&group.commands);
+        let mut command_names: Vec<_> = commands.keys().collect();
```
zeyla: And lastly here.

zeyla:
Thanks

Issue #88, opened 2017-05-17
Pre-configure reactions to send with message


zeyla:
Add a `reactions` method to `CreateMessage` to configure reactions to place on the message after being sent.

hsiW:
Closed by https://github.com/zeyla/serenity/commit/77b5b480d67e747908f8f4fb9f910bab23b761b5.

PR #94, opened 2017-05-24
Fix issue #88

acdenisSK:
Will try to do #84 tomorrow.


zeyla:
Thanks. This could use examples, but I'll make a general issue for tracking what still needs example code instead.

PR #95, opened 2017-05-25
Fix issue 65

acdenisSK:


zeyla:
I assume thumbnails and icon URLs can reference attachments too, do we know if that's the case?

Flat:
According to https://discordapp.com/developers/docs/resources/channel#using-attachments-within-embeds you should be able to use them anywhere in the embed (that supports images).

PR #98, opened 2017-05-28
Add _line and _line_safe variants to MessageBuilder

Roughsketch:
Tested it with every variant without an explicit newline and it worked well. They all utilize the older methods to do the actual formatting, so it should be correct.

### DIFF
--- a/src/utils/message_builder.rs
+++ b/src/utils/message_builder.rs
@@ -341,6 +341,54 @@ impl MessageBuilder {
         self
     }
 
+    /// Pushes the given text with a newline appended to the content.
+    pub fn push_line(mut self, content: &str) -> Self {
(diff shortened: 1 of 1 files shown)


zeyla:
Could we add examples with assertions as doctests for these?

For example, `push_line` would look like:
```rust
/// Pushes the given text with a newline appended to the content.
///
/// # Examples
///
/// Push content and then append a newline:
///
/// ```rust,no_run
/// use serenity::utils::MessageBuilder;
///
/// let content = MessageBuilder::new().push_line("hello").push("world").build();
///
/// assert_eq!(content, "hello\nworld");
/// ```
```

zeyla:
Thanks!

PR #100, opened 2017-05-28
Add Content for MessageBuilder

eLunate:
This adds a new Content type which `push`, `push_safe`, `push_line`, and `push_line_safe` have been modified to take. This allows for flexible formatting without the duplicated `push_mono`, `push_bold_italic` etc.

```rs
let message_content = MessageBuilder::new()
  .push(Bold + Italic + Code + "text")
  .build()

assert_eq!(message_content, "***`text`***")
```

This should be fairly flexible, and makes sure to apply formatting in the correct order regardless of which order the modifiers are provided. Provides coercions for `&str` and `String`

### DIFF
--- a/src/utils/message_builder.rs
+++ b/src/utils/message_builder.rs
@@ -1,8 +1,7 @@
 use std::default::Default;
 use std::fmt::{self, Write};
-use ::model::{ChannelId, Emoji, Mentionable, RoleId, UserId};
 use std::ops::Add;
-use std::convert::{From, Into};
+use ::model::{ChannelId, Emoji, Mentionable, RoleId, UserId};
 
 /// The Message Builder is an ergonomic utility to easily build a message,
 /// by adding text and mentioning mentionable structs.
@@ -728,7 +727,7 @@ impl fmt::Display for MessageBuilder {
 }
 
 
-/// Content modifier enum
+/// Formatting modifiers for MessageBuilder content pushes
 ///
 /// Provides an enum of formatting modifiers for a string, for combination with
 /// string types and Content types.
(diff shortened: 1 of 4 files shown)


src/utils/message_builder.rs
```diff
@@ -1,6 +1,8 @@
 use std::default::Default;
 use std::fmt::{self, Write};
 use ::model::{ChannelId, Emoji, Mentionable, RoleId, UserId};
+use std::ops::Add;
+use std::convert::{From, Into};
```
zeyla: These are a part of the [stdlib prelude](https://doc.rust-lang.org/std/prelude/), explicitly bringing them into scope is redundant.

src/utils/message_builder.rs
```diff
@@ -192,8 +194,9 @@ impl MessageBuilder {
     ///
     /// assert_eq!(message.push("ing").0, "testing");
     /// ```
-    pub fn push(mut self, content: &str) -> Self {
-        self.0.push_str(content);
+    pub fn push<T: Into<Content>>(mut self, content: T) -> Self {
+        let content: Content = content.into();
```
zeyla: The project prefers not to type annotate where unnecessary.

src/utils/message_builder.rs
```diff
@@ -192,8 +194,9 @@ impl MessageBuilder {
     ///
     /// assert_eq!(message.push("ing").0, "testing");
     /// ```
-    pub fn push(mut self, content: &str) -> Self {
-        self.0.push_str(content);
+    pub fn push<T: Into<Content>>(mut self, content: T) -> Self {
```
zeyla: See my comments below on the Content struct about this.

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+/// # Examples
+///
+/// Create a new Content type which describes a bold-italic "text":
+///
+/// ```rust
+/// use serenity::utils::{Bold, Italic, Content};
+/// let content: Content = Bold + Italic + "text";
+/// ```
+pub enum ContentModifier {
+  Italic,
```
zeyla: Serenity uses 4-space indentation.

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
     }
 }
 
+
+/// Content modifier enum
```
zeyla: This should be a bit more descriptive; maybe "Modifiers for message builder content pushes."?

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+      },
+      ContentModifier::Code => {
+        self.code = true;
+      },
+      ContentModifier::Underline => {
+        self.underline = true;
+      }
+    }
+  }
+  pub fn to_string(&self) -> String {
```
zeyla: What we could do is `impl ::std::string::ToString for Content` (bringing `ToString` into scope above). This would allow the methods on `MessageBuilder` to take a `ToString` bound, allowing for a bit more generic usage out of the functions.

For example, the `push_safe` signature could look like `pub fn push_safe<T: ToString>(mut self, content: T) -> Self`.

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+    let mut newstr = String::with_capacity(
+      self.inner.len()
+      + if self.bold {4} else {0}
+      + if self.italic {2} else {0}
+      + if self.strikethrough {4} else {0}
+      + if self.underline {4} else {0}
+      + if self.code {2} else {0}
+    );
```
zeyla: Could we format this as:

```rust
let capacity = self.inner.len()
    + if self.bold { 4 } else { 0 }
    + ...;

let mut new_str = String::with_capacity(capacity);
```
Note that the `+` is indented 4, there's a space before and after the curly braces, and the `newstr` binding is renamed to `new_str`.

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+    if self.bold {
+      newstr.push_str("**");
+    }
+    if self.italic {
```
zeyla: An empty line before and after blocks please.

```rust
if self.bold {
    new_str.push("**");
}

if self.italic {
  newstr.push('*');
}

if self.strikethrough {
// ...
```

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+/// Content modifier enum
+///
+/// Provides an enum of formatting modifiers for a string, for combination with
+/// string types and Content types.
+///
+/// # Examples
+///
+/// Create a new Content type which describes a bold-italic "text":
+///
+/// ```rust
```
zeyla: This doesn't assert anything, so it doesn't need to actually run, it only needs to compile. This can be done by appending `,no_run`.

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+
+/// Describes formatting on string content
+#[derive(Default, Clone)]
+pub struct Content {
+  italic: bool,
+  bold: bool,
+  strikethrough: bool,
+  pub inner: String,
+  code: bool,
+  underline: bool
```
zeyla: Although there's never a reason to mutate or read these fields directly, we should expose them to users. Unlike other things like the framework, users mutating it won't mess with internals.

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+}
+
+impl Add<String> for Content {
+  type Output = Content;
+  fn add(self, rhs: String) -> Content {
+    let mut nc = self.clone();
+    nc.inner = nc.inner + &rhs;
+    nc
```
zeyla: An empty line before the return (as well as the others) please -- it helps with readability.

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+  bold: bool,
+  strikethrough: bool,
+  pub inner: String,
+  code: bool,
+  underline: bool
+}
+
+impl Add<String> for Content {
+  type Output = Content;
+  fn add(self, rhs: String) -> Content {
```
zeyla: Empty line before and after functions.

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+      + if self.italic {2} else {0}
+      + if self.strikethrough {4} else {0}
+      + if self.underline {4} else {0}
+      + if self.code {2} else {0}
+    );
+    if self.bold {
+      newstr.push_str("**");
+    }
+    if self.italic {
+      newstr.push_str("*");
```
zeyla: Since this is pushing only a char, [`String::push`](https://doc.rust-lang.org/std/string/struct.String.html#method.push) can be used.

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+    }
+    if self.strikethrough {
+      newstr.push_str("~~");
+    }
+    if self.underline {
+      newstr.push_str("__");
+    }
+    if self.code {
+      newstr.push_str("`");
```
zeyla: Same here.

src/utils/message_builder.rs
```diff
@@ -723,6 +727,190 @@ impl fmt::Display for MessageBuilder {
...
+    }
+    newstr.push_str(&self.inner);
```
zeyla: Empty line before and after this please, partially also for the same reason as above.

src/utils/mod.rs
```diff
@@ -19,7 +19,8 @@ use std::path::Path;
 use ::internal::prelude::*;
 use ::model::{EmojiIdentifier, EmojiId};
 
-pub use self::message_builder::MessageBuilder;
+pub use self::message_builder::{MessageBuilder, Content, ContentModifier};
+pub use self::message_builder::ContentModifier::*;
```
zeyla: I'm not sure this re-export is necessary, users can do this themselves if they'd like to bring all variants into scope. It's also cleaner, as if users do `use serenity::utils::*;` it'll bring the variants into scope, the other `MessageBuilder` structs+variants, `Colour`, _and_ the other functions below into scope. Instead users should do `use serenity::utils::ContentModifier::*;` if they'd like.

src/utils/mod.rs
```diff
@@ -19,7 +19,8 @@ use std::path::Path;
 use ::internal::prelude::*;
 use ::model::{EmojiIdentifier, EmojiId};
 
-pub use self::message_builder::MessageBuilder;
+pub use self::message_builder::{MessageBuilder, Content, ContentModifier};
```
zeyla: Can these re-exports be moved below line 9 to organize them together?

src/utils/message_builder.rs
```diff
@@ -1,6 +1,8 @@
 use std::default::Default;
 use std::fmt::{self, Write};
 use ::model::{ChannelId, Emoji, Mentionable, RoleId, UserId};
+use std::ops::Add;
```
zeyla: Can this be moved above line 3 to be alphabetical?

src/utils/message_builder.rs
eLunate: I don't understand this comment.

eLunate:
Changes made for the style guide. Included an editorconfig in the root to automatically enforce tab guides for editors which respect it.

zeyla:
Looks good, thanks

PR #101, opened 2017-05-28
Add documentation and tests for framework

Roughsketch:
Work in progress:

- [x] framework
  - [x] help_commands::{plain, with_embeds}
  - [x] Configuration (others optional)
    - [x] blocked_guilds
    - [x] blocked_users
    - [x] disabled_commands
    - [x] owners
    - [x] prefix
    - [x] prefixes
  - [x] Framework (struct-level example needed)
    - [x] bucket
    - [x] simple_bucket
    - [x] on
    - [x] group
    - [x] on_dispatch_error
    - [x] before
    - [x] after

### DIFF
--- a/src/framework/mod.rs
+++ b/src/framework/mod.rs
@@ -700,6 +700,41 @@ impl Framework {
 
     /// Specify the function to be called prior to every command's execution.
     /// If that function returns true, the command will be executed.
+    ///
+    /// # Examples
+    ///
+    /// Using before to log command usage:
+    ///
+    /// ```rust
+    /// # use serenity::Client;
+    /// # let mut client = Client::login("token");
+    ///
+    /// client.with_framework(|f| f
+    ///     .before(|ctx, msg, cmd_name| {
+    ///         println!("Running command {}", cmd_name);
+    ///         true
+    ///     }));
+    /// ```
+    ///
+    /// Using before to prevent command usage:
+    ///
+    /// ```rust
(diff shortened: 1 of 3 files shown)


src/framework/mod.rs
```diff
@@ -644,6 +767,23 @@ impl Framework {
...
     /// Fourth argument exists if command returned an error which you can handle.
+    ///
+    /// # Examples
+    ///
+    /// Using after to log command usage:
+    ///
+    /// ```rust
+    /// # use serenity::Client;
+    /// # let mut client = Client::login("token");
+    ///
```
zeyla: As with above, if I'm not mistaken this empty line will cause extra padding at the top of the codeblock -- I think it'd look cleaner if this were `/// #`, instead of removing the line altogether.

src/framework/mod.rs
```diff
@@ -635,6 +723,41 @@ impl Framework {
...
+    /// client.with_framework(|f| f
+    ///     .before(|ctx, msg, cmd_name| {
+    ///         if let Ok(channel) = msg.channel_id.get() {
+    ///             if !channel.is_nsfw() {
+    ///                 return false;   //  Don't run unless in nsfw channel
```
zeyla: I believe all but two areas of the documentation prefer the comment to be above the code, rather than on the same line. Can this be moved above the `if !channel.is_nsfw()`?

PR #102, opened 2017-05-30
Don't show a group in help if there are no commands to show.

Roughsketch:
This will prevent stuff like the following:

> **Group**: `ping`, `pong`
> **Other**: _[No Commands]_
> **Ungrouped**: `help`

If no commands are found, it will simply drop the group like so: 

> **Group**: `ping`, `pong`
> **Ungrouped**: `help`


zeyla:
Thanks for the PRs

PR #105, opened 2017-05-31
More docs

Roughsketch:
Some more docs and some miscellaneous fixes.

Added docs:

- [x] Embed
  - [x] fake
- [x] Emoji
  - [x] delete
  - [x] edit
  - [x] find_guild_id
  - [x] url
- [x] Game
  - [x] playing
  - [x] streaming

Miscellaneous fix:

- Updated references to things like `Context::send_message` in several `builder` docs.
- Deprecated `Role::edit_role` in favor of `Role::edit`

### DIFF
--- a/src/model/guild/role.rs
+++ b/src/model/guild/role.rs
@@ -93,7 +93,7 @@ impl Role {
     ///
     /// [`Role`]: struct.Role.html
     /// [Manage Roles]: permissions/constant.MANAGE_ROLES.html
-    #[cfg(feature="cache")]
+    #[cfg(all(feature="builder", feature="cache"))]
     pub fn edit<F: FnOnce(EditRole) -> EditRole>(&self, f: F) -> Result<Role> {
         match self.find_guild() {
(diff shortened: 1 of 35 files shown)


src/model/guild/role.rs
```diff
@@ -83,15 +83,44 @@ impl Role {
...
+    /// # use serenity::model::RoleId;
+    /// # let role = RoleId(7).find().unwrap();
+    /// // assuming a `role` has been bound
+    //
+    /// role.edit(|r| r.hoist(true));
+    /// ```
+    ///
+    /// [`Role`]: struct.Role.html
+    /// [Manage Roles]: permissions/constant.MANAGE_ROLES.html
+    #[deprecated(since="0.2.1", note="Please use `edit` instead.")]
```
zeyla: Since this is deprecated, can we replace the function body with just `self.edit(f)`?

zeyla: Also since this is deprecated, we can just replace these docs and link to the new function saying ```/// Alias of [`edit`].```

zeyla:
Thanks

PR #110, opened 2017-06-09
Fix issue 109

acdenisSK:


src/model/guild/member.rs
```diff
@@ -352,6 +352,27 @@ impl Member {
...
+    /// println!("The permission bits for the member are: {}", member.permissions().expect("permissions").bits);
+    /// ```
+
+    /// # Errors
+    ///
+    /// Returns a [`ModelError::GuildNotFound`] if the guild the member's in could not be
+    /// found in the cache.
+    ///
+    /// [`ModelError::GuildNotFound`]: enum.ModelError.html#variant.GuildNotFound
+    pub fn permissions(&self) -> Result<Permissions> {
```
zeyla: This needs to be gated behind the `cache` feature.

src/model/guild/member.rs
```diff
@@ -352,6 +352,27 @@ impl Member {
...
+        let dummy = self.find_guild()?.find().unwrap();
```
zeyla: Needs updating for b4bd771, needs to use the structfield `guild_id` instead of the method `find_guild` now.

acdenisSK:
Regarding the two "merge branch ...", i messed up a bit in the first one since i forgot to update my local master.

zeyla:
Thanks

PR #111, opened 2017-06-10
Fix invite_url

Flat:
Use Client Id from Application instead of UserId

### DIFF
--- a/src/model/user.rs
+++ b/src/model/user.rs
@@ -160,7 +160,7 @@ impl CurrentUser {
     ///
     /// Get the invite url with no permissions set:
     ///
-    /// ```rust
+    /// ```rust,no_run
     /// # use serenity::client::CACHE;
     /// #
     /// # let mut cache = CACHE.write().unwrap();
@@ -169,14 +169,14 @@ impl CurrentUser {
     /// use serenity::model::permissions::Permissions;
     ///
     /// // assuming the cache has been unlocked
-    /// let url = cache.user.invite_url(Permissions::empty());
+    /// let url = cache.user.invite_url(Permissions::empty()).unwrap();
     ///
     /// assert_eq!(url, "https://discordapp.com/api/oauth2/authorize?client_id=249608697955745802&scope=bot");
     /// ```
     ///
     /// Get the invite url with some basic permissions set:
     ///
-    /// ```rust
+    /// ```rust,no_run
     /// # use serenity::client::CACHE;
     /// #
     /// # let mut cache = CACHE.write().unwrap();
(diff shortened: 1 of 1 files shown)


src/model/user.rs
```diff
@@ -185,17 +185,21 @@ impl CurrentUser {
     /// use serenity::model::permissions::*;
     ///
     /// // assuming the cache has been unlocked
-    /// let url = cache.user.invite_url(READ_MESSAGES | SEND_MESSAGES | EMBED_LINKS);
+    /// let url = cache.user.invite_url(READ_MESSAGES | SEND_MESSAGES | EMBED_LINKS).unwrap();
     ///
     /// assert_eq!(url, "https://discordapp.com/api/oauth2/authorize?client_id=249608697955745802&scope=bot&permissions=19456");
     /// ```
-    pub fn invite_url(&self, permissions: Permissions) -> String {
+    pub fn invite_url(&self, permissions: Permissions) -> Result<String> {
```
zeyla: Can we make a note in the primary description that this hits REST, and list in an errors section the possible errors (`Unauthorized`; see `ChannelId::send_file` for an example on how to format it)?

src/model/user.rs
```diff
@@ -185,17 +185,21 @@ impl CurrentUser {
...
+            Ok(v) => v.id,
+            Err(e) => return Err(e),
+        };
 
         if bits == 0 {
-            format!("https://discordapp.com/api/oauth2/authorize?client_id={}&scope=bot", self.id)
+            Ok(format!("https://discordapp.com/api/oauth2/authorize?client_id={}&scope=bot", client_id))
         } else {
-            format!("https://discordapp.com/api/oauth2/authorize?client_id={}&scope=bot&permissions={}", self.id, bits)
+            Ok(format!("https://discordapp.com/api/oauth2/authorize?client_id={}&scope=bot&permissions={}", client_id, bits))
```
zeyla: This can be simplified by wrapping the if-else in the Ok variant. I didn't think it important enough the first time, but while we're at it, since there's a fair bit of duplication here we could move the initial format out and bind it to a string. If `bits != 0`, `write!` the permissions query params to it, and then return the string afterwards.

src/model/user.rs
```diff
@@ -407,7 +411,7 @@ impl User {
     ///
     /// client.on_message(|_, msg| {
     ///     if msg.content == "~help" {
-    ///         let url = CACHE.read().unwrap().user.invite_url(Permissions::empty());
+    ///         let url = CACHE.read().unwrap().user.invite_url(Permissions::empty()).unwrap();
```
zeyla: I'd like to avoid non-lock unwraps in future documentation (and slowly fixing them in old documentation). Maybe we could match this.. or push the invite text in an if-let only if this is Ok? It's unfortunate event handlers don't return a Result right now to make use of `?`.

src/model/user.rs
```diff
@@ -169,14 +169,14 @@ impl CurrentUser {
     /// use serenity::model::permissions::Permissions;
     ///
     /// // assuming the cache has been unlocked
-    /// let url = cache.user.invite_url(Permissions::empty());
+    /// let url = cache.user.invite_url(Permissions::empty()).unwrap();
```
zeyla: See my bottom comment on unwraps, and for the example below.

zeyla: Also since these tests now have to be `no_run`d, the CurrentUser Id mutation can be removed from both doctests.

zeyla:
Thanks

PR #112, opened 2017-06-11
Change from #[doc(hidden)] to pub(crate)

acdenisSK:
The only exception i made was to `src/voice/error.rs`, as all of the cases of an enum are _always_ public; but the actual enum can or can't be public.

### DIFF
--- a/src/framework/configuration.rs
+++ b/src/framework/configuration.rs
@@ -28,18 +28,30 @@ use ::model::{GuildId, Message, UserId};
 /// [`Client`]: ../../client/struct.Client.html
 /// [`Framework`]: struct.Framework.html
 pub struct Configuration {
-    pub(crate) allow_dm: bool,
-    pub(crate) allow_whitespace: bool,
-    pub(crate) blocked_guilds: HashSet<GuildId>,
-    pub(crate) blocked_users: HashSet<UserId>,
-    pub(crate) depth: usize,
-    pub(crate) disabled_commands: HashSet<String>,
-    pub(crate) dynamic_prefix: Option<Box<PrefixCheck>>,
-    pub(crate) ignore_bots: bool,
-    pub(crate) ignore_webhooks: bool,
-    pub(crate) on_mention: Option<Vec<String>>,
-    pub(crate) owners: HashSet<UserId>,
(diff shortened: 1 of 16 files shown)


src/framework/configuration.rs
```diff
@@ -28,30 +28,18 @@ use ::model::{GuildId, UserId};
...
+    pub(crate) blocked_guilds: HashSet<GuildId>,
+    pub(crate) blocked_users: HashSet<UserId>,
+    pub(crate) depth: usize,
+    pub(crate) disabled_commands: HashSet<String>,
+    pub(crate) dynamic_prefix: Option<Box<PrefixCheck>>,
+    pub(crate) ignore_bots: bool,
+    pub(crate) ignore_webhooks: bool,
+    pub(crate) on_mention: Option<Vec<String>>,
+    pub(crate) owners: HashSet<UserId>,
+    pub(crate) prefixes: Vec<String>,
```
zeyla: All of these are going to be made public in a meaningful way in the future, for now they can just still be hidden from the docs.

src/gateway/shard.rs
```diff
@@ -522,8 +521,8 @@ impl Shard {
 
     /// Shuts down the receiver by attempting to cleanly close the
     /// connection.
-    #[doc(hidden)]
-    pub fn shutdown_clean(client: &mut WsClient) -> Result<()> {
+    #[allow(dead_code)]
+    pub(crate) fn shutdown_clean(client: &mut WsClient) -> Result<()> {
```
zeyla: The websocket client is now available to the user, they should be able to call this if they'd like -- the `#[doc(hidden)]` can be removed.

src/gateway/shard.rs
```diff
@@ -544,8 +543,7 @@ impl Shard {
     }
 
     /// Uncleanly shuts down the receiver by not sending a close code.
-    #[doc(hidden)]
-    pub fn shutdown(&mut self) -> Result<()> {
+    pub(crate) fn shutdown(&mut self) -> Result<()> {
```
zeyla: Same here

zeyla:
Thanks

PR #113, opened 2017-06-14
Use native-tls stuff when making clients and requests.

Roughsketch:
Ran into this issue with `http::send_files` a few days ago, but found out that the problem was also in `http::get_messages`. I decided to just change all occurrences that I could find.


zeyla:
Thanks

PR #114, opened 2017-06-18
Add an `on_cached` event

### LINKED ISSUE #89: Add an `on_cached` Client method
The method will fire when all `GuildCreate`s have been received.

Questions:

- Should this fire when each shard is fully cached (providing the shard ID), or when all shards the Client is responsible for are fully cached?
- What information should be provided in the event fire?

acdenisSK:
Fixes #89


acdenisSK:
Okay, did i adapt this right?

PR #118, opened 2017-06-28
Fixed clippy warnings

imnotbad:
Fixed some clippy warnings, didn't touch tests/test_prelude.rs:7:25 because i'm not exactly sure what it does right now.


acdenisSK:
It's more or less to see if the re-exports work, but nontheless, thanks.

PR #120, opened 2017-07-10
Fixed clippy warnings

imnotbad:
Fixed some clippy warnings, made code more concise.
Clippy also complained about src/client/dispatch.rs:446:36 but i'm not sure if these branches are required for the bot framework so i didn't modify them.


acdenisSK:
Yikes, if only i were able to run clippy properly on my machine. Oh well, thanks!

PR #133, opened 2017-07-31
Fix failing tests in framework and utils code

xentec:
Looks like the previous `rustfmt` commit broke some code.

### DIFF
--- a/src/framework/help_commands.rs
+++ b/src/framework/help_commands.rs
@@ -6,7 +6,7 @@
 //! embeds:
 //!
 //! ```rs,no_run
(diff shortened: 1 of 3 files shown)


acdenisSK:
Always forgetting about the tests and alike, dammit.
But anyway, thanks.

xentec:
@acdenisSK I really appreciate your daily work on serenity, but I also wish you would run `cargo test` once in a while. ;)

acdenisSK:
Yeah i know, but i hate running it, since it takes so much time for it to compile serenity and do all of the (doc-)tests.

PR #134, opened 2017-08-13
Fixed string-delimiters, added Regex- and Itertools-crates, ...

Lakelezz:
... changed Delimeter-typo to Delimiter across the library

### DIFF
--- a/src/framework/mod.rs
+++ b/src/framework/mod.rs
@@ -78,6 +78,9 @@ use model::{ChannelId, GuildId, Message, UserId};
 use model::permissions::Permissions;
 use utils;
 use tokio_core::reactor::Handle;
+use itertools::Itertools;
+use regex::Regex;
+use regex::escape;
 
 #[cfg(feature = "cache")]
 use client::CACHE;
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -128,6 +128,9 @@ extern crate tokio_core;
 extern crate typemap;
 #[cfg(feature = "gateway")]
 extern crate websocket;
+#[cfg(feature = "framework")]
+extern crate regex;
+extern crate itertools;
```
acdenisSK: Feature-flag this as well

src/framework/mod.rs
```diff
@@ -937,16 +940,19 @@ impl ::Framework for BuiltinFramework {
...
-                                    .split(delimeter)
-                                    .map(|arg| arg.to_owned())
-                                    .collect::<Vec<String>>()
+                                let delimiters = &self.configuration.delimiters;
+                                let regular_expression = delimiters.iter().
+                                    map(|delimiter| escape(delimiter)).join("|");
+
+                                let regex = Regex::new(&regular_expression).unwrap();
+
+                                let mut result: Vec<String> = vec![];
```
acdenisSK: This and the for loop can be changed to 
```rust
regex.split(content)
    .filter_map(|p| if !p.is_empty() { Some(p.to_string()) } else { None })
    .collect::<Vec<_>>()
```

src/framework/mod.rs
```diff
@@ -937,16 +940,19 @@ impl ::Framework for BuiltinFramework {
...
-                                    .find(|d| content.contains(d.as_str()))
-                                    .map(|s| s.as_str())
-                                    .unwrap_or(" ");
-
-                                content
-                                    .split(delimeter)
-                                    .map(|arg| arg.to_owned())
-                                    .collect::<Vec<String>>()
+                                let delimiters = &self.configuration.delimiters;
+                                let regular_expression = delimiters.iter().
```
acdenisSK: It'd look better (or less awkward) if you put this dot on the same line as `map`.

PR #144, opened 2017-08-22
Fix presence updates

Roughsketch:
Due to the changes made, type and url are now required parameters. Without them the presence will not update.


zeyla:
Aaand I put my reply in the commit message, that was clumsy. Anyway, copy-pasting:

```
Was waiting a couple more days on #139, but since this is broken functionality it's probably best to fix it sooner.

Travis failures are all from unrelated commits, thanks.
```

PR #146, opened 2017-08-22
Use cache when possible in UserId's get method

Roughsketch:
I'm not sure if unwrapping is fine here, but I assumed since unwraps are used for the cache that it'd be fine for the lock as well. If this won't work, then I'll try to find another way to handle it.


acdenisSK:
And yes it is, as if the unwrap fails, then there's some serious dead-lock.

PR #147, opened 2017-08-23
Allow FromStr for User to use REST

Roughsketch:
If a user is invisible or not in the cache, the current implementation will fail to get their information. This change will make a REST call if the user is not found in the cache.

Also fixed lint warnings.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -85,7 +85,8 @@
 //! [examples]: https://github.com/zeyla/serenity/tree/master/examples
 //! [gateway docs]: gateway/index.html
 #![doc(html_root_url = "https://docs.rs/serenity/*")]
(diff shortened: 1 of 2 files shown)


acdenisSK:
cc @zeyla, are you fine with this change?

zeyla:
👍 

I'm going to assume the mac tests pass because there's no reason they shouldn't. Looks like Travis is having issues on their Mac build infrastructure.

Can this search the cache first and then fallback to REST?

Roughsketch:
In #146  `UserId.get` was changed to try the cache before making a rest call. So as far as I'm aware, this should check the cache first because the cache feature will have to be on to even use User's FromStr.

zeyla:
Good point, wasn't aware of that change. LGTM.

PR #154, opened 2017-08-28
Role references can now be obtained via their name.

Lakelezz:


src/model/guild/partial_guild.rs
```diff
@@ -450,4 +450,16 @@ impl PartialGuild {
     /// [Manage Webhooks]: permissions/constant.MANAGE_WEBHOOKS.html
     #[inline]
     pub fn webhooks(&self) -> Result<Vec<Webhook>> { self.id.webhooks() }
+
+    /// Obtain a reference to a role by its name.
+    ///
+    /// **Note**: If two or more roles have the same name, obtained reference will be one of them.
```
zeyla: I prefer functions to have examples where possible, could we make one here? Refer [here](https://github.com/zeyla/serenity/blob/master/src/model/channel/guild_channel.rs#L420-L516) for a good example of this.

src/model/guild/partial_guild.rs
```diff
@@ -450,4 +450,16 @@ impl PartialGuild {
...
+    pub fn role_by_name(&self, role_name: &str) -> Option<&Role> {
+        for (_, role) in &self.roles {
```
acdenisSK: `self.roles.values().find(|role| role_name == role.name)`

src/model/guild/partial_guild.rs
```diff
@@ -450,4 +450,33 @@ impl PartialGuild {
...
+    /// ```rust,no_run
+    /// use serenity::prelude::*;
+    /// use serenity::model::*;
+    /// struct Handler;
+    ///
+    /// use serenity::CACHE;
+    ///
+    /// impl EventHandler for Handler {
+    ///     fn on_message(&self, _: Context, msg: Message) {
+    ///         match message.guild_id().unwrap().get().unwrap().role_by_name("roll_name") {
```
zeyla: This can be simplified to an if-let, `if let Some(role) = message.guild_id()...`

src/model/guild/partial_guild.rs
```diff
@@ -450,4 +450,33 @@ impl PartialGuild {
...
+    ///             Some(role) => println!("Obtained role's reference: {:?}", role),
+    ///             None => return,
+    ///         };
+    ///     }
+    /// }
+    /// let mut client = Client::new("token", Handler); client.start().unwrap();
```
zeyla: Can this be made into 2 lines?

src/model/guild/partial_guild.rs
acdenisSK: I'd also favour to use `find` instead of `get` here.

src/model/guild/partial_guild.rs
```diff
@@ -470,7 +470,8 @@ impl PartialGuild {
     ///
     /// impl EventHandler for Handler {
     ///     fn on_message(&self, _: Context, msg: Message) {
-    ///         if let Some(role) = msg.guild_id().unwrap().get().unwrap().role_by_name("role_name") {
+    ///         if let Some(role) =
+    ///            msg.guild_id().unwrap().get().unwrap().role_by_name("role_name") {
```
zeyla: Just checking, does rustfmt want this to be indented 3 spaces?

Lakelezz: Every indentation - I tried - that features one space from the comment-start is accepted.

zeyla:
Thanks!

PR #157, opened 2017-09-05
Commands can now be limited to certain roles

Lakelezz:
Also, `help_commands::with_embeds` and `help_commands::plain` check if user has the right role to use them.

Both will be bypassed by a user with `permissions::ADMINISTRATOR` but not with being an owner. I'm not sure if the bot-owner should be able to do bypass, as that might enable security backdoors... someone uses your bot, you join their server and then the owner abuses admin-features and so on...

I called the field of `Command` `allowed_roles`, because having one of these roles gains permission to use it. If `Vec<String>` stays empty, no special role-checks will be done.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -29,7 +29,7 @@ use std::fmt::Write;
 use super::command::InternalCommand;
 use super::{Args, Command, CommandGroup, CommandOrAlias};
 use client::Context;
-use model::{ChannelId, Message};
+use model::{ChannelId, Guild, Member, Message};
 use utils::Colour;
 
 fn error_embed(channel_id: &ChannelId, input: &str) {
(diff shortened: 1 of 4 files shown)


src/framework/standard/help_commands.rs
```diff
@@ -245,6 +276,20 @@ pub fn plain(_: &mut Context,
...
+                                        if let Ok(permissions) = member.permissions() {
+                                            if !permissions.administrator() &&
+                                               !right_roles(&cmd, &guild, &member) {
+                                                break;
+                                            }
+                                        }
+                                    }
+                                }
+                            }
+                            println!("found");
```
acdenisSK: Remove this.

src/framework/standard/help_commands.rs
```diff
@@ -50,6 +50,16 @@ fn remove_aliases(cmds: &HashMap<String, CommandOrAlias>) -> HashMap<&String, &I
...
+fn right_roles(cmd: &Command, guild: &Guild, member: &Member) -> bool {
+    if cmd.allowed_roles.len() > 0 {
+        return cmd.allowed_roles
+            .iter()
+            .flat_map(|r| guild.role_by_name(&r))
+            .any(|g| member.roles.contains(&g.id));
+    }
+    true
+}
+
```
acdenisSK:
```rust
fn right_roles(cmd: &Command, guild: &Guild, member: &Member) -> bool {
    if cmd.allowed_roles.len() > 0 {
        cmd.allowed_roles
            .iter()
            .flat_map(|r| guild.role_by_name(&r))
            .any(|g| member.roles.contains(&g.id))
    } else {
        true
    }
}
```

(This style is usually preferred when the function is as simple as this.)

PR #175, opened 2017-10-03
Fix most clippy warnings

Roughsketch:
I went through and fixed as many as I could. There are still a few left, but they are things I wasn't sure whether to change/remove.


zeyla:
Thanks

Issue #174, opened 2017-10-03
Functions should accept IntoIterator instead of a slice, vector, etc.


acdenisSK:
- [x] Vectors
- [x] Slices

PR #176, opened 2017-10-03
Replace Vec parameters by IntoIterator

ftriquet:
Part of #174 

Maybe I should use `where` clauses instead of `fn foo<It: IntoIterator>(_: It)`. What do you think?

### DIFF
--- a/src/client/event_handler.rs
+++ b/src/client/event_handler.rs
@@ -9,7 +9,7 @@ use std::sync::RwLock;
 
 pub trait EventHandler {
     #[cfg(feature = "cache")]
-    fn on_cached<It: IntoIterator<Item=GuildId>>(&self, _: Context, _: It) {}
+    fn on_cached(&self, _: Context, _: Vec<GuildId>) {}
     fn on_channel_create(&self, _: Context, _: Arc<RwLock<GuildChannel>>) {}
     fn on_category_create(&self, _: Context, _: Arc<RwLock<ChannelCategory>>) {}
     fn on_category_delete(&self, _: Context, _: Arc<RwLock<ChannelCategory>>) {}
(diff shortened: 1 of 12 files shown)


src/client/event_handler.rs
```diff
@@ -9,7 +9,7 @@ use std::sync::RwLock;
 
 pub trait EventHandler {
     #[cfg(feature = "cache")]
-    fn on_cached(&self, _: Context, _: Vec<GuildId>) {}
```
acdenisSK: I should've specified that functions that users pass data unto us needed to be changed to use `IntoIterator` (since these're like the users' callbacks, just smashed into a trait), but oh well, revert these that you've done here.

src/framework/standard/configuration.rs
```diff
@@ -401,7 +402,7 @@ impl Configuration {
     /// client.with_framework(StandardFramework::new().configure(|c| c
     ///     .delimiters(vec![", ", " "])));
     /// ```
-    pub fn delimiters(mut self, delimiters: Vec<&str>) -> Self {
+    pub fn delimiters<T: ::std::string::ToString, It: ::std::iter::IntoIterator<Item=T>>(mut self, delimiters: It) -> Self {
```
acdenisSK: Particular reason you didn't use just `ToString`/`IntoIterator` instead of their absolute paths here?

src/http/mod.rs
```diff
@@ -1611,7 +1611,7 @@ pub fn remove_group_recipient(group_id: u64, user_id: u64) -> Result<()> {
 /// if the file is too large to send.
 ///
 /// [`HttpError::InvalidRequest`]: enum.HttpError.html#variant.InvalidRequest
-pub fn send_files<'a, T>(channel_id: u64, files: Vec<T>, map: JsonMap) -> Result<Message>
+pub fn send_files<'a, T, It: ::std::iter::IntoIterator<Item=T>>(channel_id: u64, files: It, map: JsonMap) -> Result<Message>
```
acdenisSK: Same here.

src/model/channel/channel_id.rs
```diff
@@ -439,7 +439,7 @@ impl ChannelId {
     /// [`GuildChannel`]: struct.GuildChannel.html
     /// [Attach Files]: permissions/constant.ATTACH_FILES.html
     /// [Send Messages]: permissions/constant.SEND_MESSAGES.html
-    pub fn send_files<'a, F, T>(&self, files: Vec<T>, f: F) -> Result<Message>
+    pub fn send_files<'a, F, T, It: ::std::iter::IntoIterator<Item=T>>(&self, files: It, f: F) -> Result<Message>
```
acdenisSK: Same.

src/model/channel/group.rs
```diff
@@ -295,7 +295,7 @@ impl Group {
     /// [Attach Files]: permissions/constant.ATTACH_FILES.html
     /// [Send Messages]: permissions/constant.SEND_MESSAGES.html
     #[inline]
-    pub fn send_files<'a, F, T>(&self, files: Vec<T>, f: F) -> Result<Message>
+    pub fn send_files<'a, F, T, It: ::std::iter::IntoIterator<Item=T>>(&self, files: It, f: F) -> Result<Message>
```
acdenisSK: Same.

acdenisSK:
(Forgot to add my response to your `where` clause question in the review fuss, so i'm just gonna do it like this.)

The bounds aren't necessarily long so nah, this is fine.

src/framework/standard/configuration.rs
ftriquet: I forgot to use directly after noticing `ToString` and `IntoIterator` where in the prelude. I'll fix it.

src/client/event_handler.rs
ftriquet: Oh right, so this will break users' code?

acdenisSK: Yes, since they'd have to rewrite their impl for that handler with the newly added trait bound; or else their code won't compile.

ftriquet: Ok I see.

Issue #181, opened 2017-10-06
Manually setting command argument delimiter only respected if delimiter is found in command


jhelwig:
As part of exploring what was going on with #180 I tried setting various argument delimiter settings. I was surprised to see that the manually specified argument delimiter setting was only used if the string was found in the arguments to the command. If the manually specified delimiter wasn't found, it would always use a single space as the delimiter. https://github.com/zeyla/serenity/blob/caf69d66893c2688f0856cc33f03702071d1314a/src/framework/standard/mod.rs#L909-L910

I was wondering what the intent behind this behavior was as it was rather surprising, and didn't seem to be documented (https://docs.rs/serenity/0.4.0/serenity/framework/standard/struct.Configuration.html#method.delimiter).

acdenisSK:
It defaults to a space since that's the most used delimiter in bots. But i agree that i should've added some docs regarding that behaviour.

jhelwig:
I completely understand the general default of using spaces. https://github.com/zeyla/serenity/blob/598252365c189b5b8b83a1a131a2d45df29516fd/src/framework/standard/configuration.rs#L446

That seems to be a separate discussion though. I'm surprised by overriding the manually specified delimiter with a single space when the user's specified delimiter isn't found.

acdenisSK:
Oops, should've added more context on that part, welp.

Let's use [this](https://play.rust-lang.org/?gist=8af7b42bea6c05efc02f47227179709e&version=stable) as a start on explaining with the "if the manually put delimiter of the user isn't found in the message, default to a space" stuff.

So suppose with the first case, we'd indeed use the passed delimiter; but we can't do much with the input in regards to `Args`'s api, since the moment you do `Args::single::<String>` or something, it's over.
But with the second case, we _can_ at least get some input to mess with, parse, etc.

Though, i chose to default to a space for another reason, if multiple delimiters are passed (`[", ", ". "]`), then which delimiter should be used for splitting the message if either of them isn't even found in the message? With the first one probably, but as i mentioned earlier, we'd get the useless input.
So a space was a reasonable solution to keep things at least nice.

jhelwig:
So, I don't really think the "usefulness" of the result of`"42. 55".split(", ")` is really the point. If a user configured the bot with `c.delimiters(vec![", "])` then they asked that arguments only be split using `", "`.

[...]

PR #183, opened 2017-10-06
Do not default the command argument delimiter to " " if none of the provided delimiters are found

jhelwig:
By no longer forcing a last-resort argument delimiter of " ", we now allow a command configured with `min_args(2)`, and a bot configured with `delimiters(vec![", "])` to have `!command first argument` fail, instead of it being run with `["first", "argument"]`.

If a bot wishes to retain the previous behavior of falling back to `" "` for the argument parsing if no other delimiters are found, then `" "` will need to be added to the end of the delimiters list (Eg: `delimiters(vec![", ", " "])`).

This also moves the delimiter detection into Args itself in anticipation of future refactoring of the argument parsing.

This  should address #181.

### DIFF
--- a/src/framework/standard/args.rs
+++ b/src/framework/standard/args.rs
@@ -60,7 +60,12 @@ pub struct Args {
 }
 
 impl Args {
-    pub fn new(message: &str, delimiter: &str) -> Self {
+    pub fn new(message: &str, possible_delimiters: Vec<String>) -> Self {
+        let delimiter = possible_delimiters
+            .iter()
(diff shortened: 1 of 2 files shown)


hsiW:
You're currently PR'ing this to master which you really shouldn't do so you should probably change it to PR'ing into the `v0.5.0` branch as this is technically a breaking change.

jhelwig:
I'll update it.

jhelwig:
Rebased onto the `v0.5.0` branch, and updated the base in the PR to reflect this.

PR #187, opened 2017-10-09
Variety of methods to search for `Member` with a substring/prefix in user- or nickname.

Lakelezz:
`members_containing` finds `Member` with a given substring in either their user- or nickname.
`members_nick_containing` finds `Member` with a given substring in their nickname.
`members_username_containing` finds `Member` with a given substring in their username.
`members_starting_with` got updated and supports sorting the result now.

Additionally, all of these use `closest_to_origin()` in order to compare two `str&` against a `str&` representing the `origin` and informs whether the first `str&` is closer or further from the `origin`.

### DIFF
--- a/src/model/guild/mod.rs
+++ b/src/model/guild/mod.rs
@@ -21,6 +21,7 @@ use serde::de::Error as DeError;
 use serde_json;
 use super::utils::*;
 use model::*;
+use std;
 
 #[cfg(all(feature = "cache", feature = "model"))]
 use CACHE;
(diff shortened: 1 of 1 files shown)


src/model/guild/mod.rs
```diff
@@ -767,19 +775,65 @@ impl Guild {
...
+                            } else {
+                                nick.to_owned()
+                            }
+                        },
+                        None => b.user.read().unwrap().name.to_owned(),
+                    };
+
+                    closest_to_origin(&prefix, &name_a[..], &name_b[..])
+                });
+            return members;
```
acdenisSK: Unnecessary `return`.

src/model/guild/mod.rs
```diff
@@ -790,13 +844,144 @@ impl Guild {
...
+                            } else {
+                                nick.to_owned()
+                            }
+                        },
+                        None => b.user.read().unwrap().name.to_owned(),
+                    };
+
+                    closest_to_origin(&substring, &name_a[..], &name_b[..])
+                });
+            return members;
```
acdenisSK: Same here.

src/model/guild/mod.rs
```diff
@@ -767,19 +775,65 @@ impl Guild {
...
-                    })).collect()
+                    })).collect();
+
+        if sorted {
+            members 
+                .sort_by(|a, b| {
+                    let name_a = match a.nick {
+                        Some(ref nick) => {
+                            if contains_case_insensitive(&a.user.read().unwrap().name[..], &prefix) {
+                                a.user.read().unwrap().name.to_owned()
```
acdenisSK: Replace `to_owned` from these bits of code with `clone`.

src/model/guild/mod.rs
```diff
@@ -790,13 +844,144 @@ impl Guild {
...
+                    let name_b = match b.nick {
+                        Some(ref nick) => {
+                                nick.to_owned()
+                            },
+                        None => b.user.read().unwrap().name.to_owned(),
+                    };
+
+                    closest_to_origin(&substring, &name_a[..], &name_b[..])
+                });
+            return members;
```
acdenisSK: Same here.

src/model/guild/mod.rs
```diff
@@ -790,13 +844,144 @@ impl Guild {
...
+            }).collect();
+
+        if sorted {
+            members 
+                .sort_by(|a, b| {
+                    let name_a = &a.user.read().unwrap().name;
+                    let name_b = &b.user.read().unwrap().name;
+                    closest_to_origin(&substring, &name_a[..], &name_b[..])
+                });
+            return members;
```
acdenisSK: Same here.

PR #190, opened 2017-10-12
Help-features display `Aliases` and list information for `Aliases`.

Lakelezz:
Old help-features gave the real name of a command if the user sent an alias (but did not actually provide any information about it), now the help-features will search for the real command via an alias and use that result instead.

Additionally, Help-features will display a command's aliases as well.

Last but not least, moved duplicated code into `has_all_requirements`.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -29,7 +29,7 @@ use std::fmt::Write;
 use super::command::InternalCommand;
 use super::{Args, Command, CommandGroup, CommandOrAlias, CommandError};
 use client::Context;
-use model::{ChannelId, Guild, Member, Message};
+use model::{ChannelId, Message};
 use utils::Colour;
 use framework::standard::{has_correct_roles, has_correct_permissions};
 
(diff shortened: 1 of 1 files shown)


src/framework/standard/help_commands.rs
```diff
@@ -100,25 +111,31 @@ pub fn with_embeds(_: &mut Context,
...
+                                found = Some((command_name, cmd));
                             }
-                            found = Some((command_name, cmd));
+                            else {
+                                break;
+                            }                 
                         },
                         CommandOrAlias::Alias(ref name) => {
-                            error_embed(&msg.channel_id, &format!("Did you mean \"{}\"?", name));
+                            let actual_command = &group.commands.get(name).unwrap();
```
acdenisSK: Remove the borrow here, and in the pattern matches below.

src/framework/standard/help_commands.rs
```diff
@@ -274,24 +289,31 @@ pub fn plain(_: &mut Context,
...
+                            }
+                            else {
+                                break;
                             }
-                            found = Some((command_name, cmd));
                         },
                         CommandOrAlias::Alias(ref name) => {
-                            let _ = msg.channel_id.say(&format!("Did you mean {:?}?", name));
-                            return Ok(());
+                            let actual_command = &group.commands.get(name).unwrap();
```
acdenisSK: Same here.

PR #197, opened 2017-10-16
Take into account the Date Header when ratelimiting

hsiW:
This makes the internal ratelimiter take into account the date header to calculate an offset from the date header provided in the response from discord. A fair bit of the ratelimting code had to be rewritten + re-arranged to accommodate this change.

This addresses issue https://github.com/zeyla/serenity/issues/84


hsiW:
Going to close this PR and redo do it

hsiW:
This wouldn't have worked properly in the lib so all work on it has been scrapped

PR #198, opened 2017-10-17
Fix clippy warnings

hsiW:


src/framework/standard/help_commands.rs
```diff
@@ -93,7 +93,7 @@ pub fn has_all_requirements(cmd: &Command, msg: &Message) -> bool {
 pub fn with_embeds(_: &mut Context,
                    msg: &Message,
                    groups: HashMap<String, Arc<CommandGroup>>,
-                   args: Args)
+                   args: &Args)
```
zeyla: This is an API change, so this is a breaking change.

src/framework/standard/help_commands.rs
```diff
@@ -271,7 +268,7 @@ pub fn with_embeds(_: &mut Context,
 pub fn plain(_: &mut Context,
              msg: &Message,
              groups: HashMap<String, Arc<CommandGroup>>,
-             args: Args)
+             args: &Args)
```
zeyla: Same here.

zeyla:
Thanks

Issue #193, opened 2017-10-15
setting position in create_role doesn't work


danxensen:
using `guild.create_role` and setting position with the role builder will not actually set the role's position.

zeyla:
Should be closed by 5a0b8a6.

PR #201, opened 2017-10-18
Implements change role position

Flat:
related to #193, the position on create seems to have been removed from the API. Should be based on the right branch now

### DIFF
--- a/src/model/guild/guild_id.rs
+++ b/src/model/guild/guild_id.rs
@@ -304,7 +304,7 @@ impl GuildId {
     }
 
     /// Edits the order of [`Role`]s
-    /// Requires the [Mange Roles] permission.
+    /// Requires the [Manage Roles] permission.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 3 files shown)


src/model/guild/guild_id.rs
```diff
@@ -303,6 +303,26 @@ impl GuildId {
...
+    ///
+    /// ```rust,ignore
+    /// use serenity::model::{GuildId, RoleId};
+    /// GuildId(7).edit_role_position(RoleId(8), 2);
+    /// ```
+    ///
+    /// [`Role`]: struct.Role.html
+    /// [Manage Roles]: permissions/constant.MANAGE_ROLES.html
+    #[inline]
+    pub fn edit_role_position(&self, role_id: u64, position: u64) -> Result<Vec<Role>> {
```
acdenisSK: Add an `Into<RoleId>` bound and use it instead of `u64` here.
(In order to be consistent with other methods on `GuildId` and other parts of the lib)

src/model/guild/guild_id.rs
```diff
@@ -303,6 +303,26 @@ impl GuildId {
         http::edit_role(self.0, role_id.into().0, &f(EditRole::default()).0)
     }
 
+    /// Edits the order of [`Role`]s
+    /// Requires the [Mange Roles] permission.
```
acdenisSK: "Mange" -> "Manage"

src/model/guild/guild_id.rs
acdenisSK: And also add a shortcut to this on `Guild`.

PR #202, opened 2017-10-19
Emojis `HashMap` will be updated, Serenity will retry connecting, cleaned up code

Lakelezz:
When a guild updates its Emojis, Serenity will `clone()` the new `HashMap` instead of just `extend()`.

If `connect()` returns an `Err()`, it will retry connecting.

Cleaned up `help_command.rs`.

### DIFF
--- a/src/gateway/shard.rs
+++ b/src/gateway/shard.rs
@@ -127,7 +127,7 @@ impl Shard {
                token: Arc<Mutex<String>>,
                shard_info: [u64; 2])
                -> Result<Shard> {
-        let client = connect(&*ws_url.lock().unwrap())?;
+        let client = connecting(&*ws_url.lock().unwrap());
 
         let current_presence = (None, OnlineStatus::Online, false);
         let heartbeat_instants = (None, None);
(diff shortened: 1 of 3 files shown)


src/gateway/shard.rs
```diff
@@ -1024,6 +1025,19 @@ fn build_gateway_url(base: &str) -> Result<Url> {
         .map_err(|_| Error::Gateway(GatewayError::BuildingUrl))
 }
 
+/// Tries to connect and upon failure, retries.
+fn connecting() -> WsClient {
+    loop {
+        match connect(&String::new()) {
```
acdenisSK: replace `&String::new()` with just `""`

src/gateway/shard.rs
```diff
@@ -1024,6 +1025,21 @@ fn build_gateway_url(base: &str) -> Result<Url> {
         .map_err(|_| Error::Gateway(GatewayError::BuildingUrl))
 }
 
+/// Tries to connect and upon failure, retries.
+fn connecting() -> WsClient {
+    let waiting_time = 120;
+
+    loop {
+        match connect("") {
```
zeyla: This won't work, because a base URL to the gateway is needed.

src/gateway/shard.rs
```diff
@@ -1024,6 +1025,21 @@ fn build_gateway_url(base: &str) -> Result<Url> {
         .map_err(|_| Error::Gateway(GatewayError::BuildingUrl))
 }
 
+/// Tries to connect and upon failure, retries.
+fn connecting() -> WsClient {
+    let waiting_time = 120;
```
zeyla: Let's make this smaller, like 30.

src/gateway/shard.rs
```diff
@@ -1024,6 +1025,21 @@ fn build_gateway_url(base: &str) -> Result<Url> {
...
+
+    loop {
+        match connect("") {
+            Ok(client) => return client,
+            Err(_) => {
```
zeyla: Might be nice to log what this error is.

src/model/event.rs
```diff
@@ -404,7 +404,7 @@ impl CacheUpdate for GuildEmojisUpdateEvent {
 
     fn update(&mut self, cache: &mut Cache) -> Option<()> {
         cache.guilds.get_mut(&self.guild_id).map(|guild| {
-            guild.with_mut(|g| g.emojis.extend(self.emojis.clone()))
+            guild.with_mut(|g| g.emojis = self.emojis.clone())
```
zeyla: I don't like `(|foo| foo = bar)` syntax, can we make this:
```rust
guild.with_mut(|g| {
    g.emojis = ...;
});
```

Additionally, please use `HashMap::clone_from`.

zeyla:
Thanks

PR #215, opened 2017-11-12
Fixed strange behaviour for when the prefix has spaces

UninterestinAcc:
The last time I was in this section of code, I splitted the string by *whitespaces* then took the second element, ignoring whether the prefix has whitespaces in it or not. I re-implemented the function to prevent strange behaviours for prefixes with whitespaces, and also added cases for non-space whitespaces.

### DIFF
--- a/src/framework/standard/command.rs
+++ b/src/framework/standard/command.rs
@@ -208,12 +208,9 @@ fn find_end_of_prefix_with_whitespace(content: &str, position: usize) -> Option<
 
     let slice = unsafe { content.slice_unchecked(position, content_len) }.as_bytes();
     for i in 0..slice.len() {
-        // 0x20 is ASCII for space
         match slice[i] {
-            0x09 => {} // \t
-            0x0a => {} // \n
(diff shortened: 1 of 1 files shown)


src/framework/standard/command.rs
```diff
@@ -203,20 +203,19 @@ fn find_mention_end(content: &str, conf: &Configuration) -> Option<usize> {
...
-                return Some(index_of_cmd);
-            }
+    let content_len = content.len();
+    if position >= content_len { return None; }
+
+    let slice = unsafe { content.slice_unchecked(position, content_len) }.as_bytes();
+    for i in 0..slice.len() {
+        // 0x20 is ASCII for space
+        match slice[i] {
+            0x09 => {} // \t
```
acdenisSK: Use `|` to group these up.
e. g
`0x09 | 0x0a | 0x0d | 0x20 => {}`

UninterestinAcc: Sure.

PR #226, opened 2017-11-23
Convert from try_opt macro to ?

hsiW:
Now that `?` is stable on `Option` replace the macro that mimicked it.


Coding-Doctors:
Is it really only used twice?

hsiW:
Yes :c

Coding-Doctors:
nice.

PR #227, opened 2017-11-30
Help-commands are now customisable

Lakelezz:
Using `customised_help()` on `StandardFramework` can alter the help-menu's text, labels and behaviour.
`05_command_framework` got updated (works with `v0.5.0`) and includes `customised_help()`.

### DIFF
--- a/src/framework/standard/command.rs
+++ b/src/framework/standard/command.rs
@@ -129,9 +129,9 @@ pub struct HelpOptions {
     pub suggestion_text: String,
     /// If no help is available, this text will be displayed.
     pub no_help_available_text: String,
-    /// How to use a command, `{usage_embed_label}: {command_name} {args}`
+    /// How to use a command, `{usage_label}: {command_name} {args}`
     pub usage_label: String,
-    /// Actual sample label, `{sample_embed_label}: {command_name} {args}`
+    /// Actual sample label, `{usage_sample_label}: {command_name} {args}`
     pub usage_sample_label: String,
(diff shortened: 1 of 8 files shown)


src/framework/standard/command.rs
```diff
@@ -116,6 +123,128 @@ pub struct CommandOptions {
...
+
+pub trait HelpCommand: Send + Sync + 'static {
+    fn execute(&self, &mut Context, &Message, &HelpOptions, HashMap<String, Arc<CommandGroup>>, Args) -> Result<(), Error>;
+
+    fn options(&self) -> Arc<CommandOptions> {
+        Arc::clone(&DEFAULT_OPTIONS)
+    }
+}
+
+impl HelpCommand for Arc<Help> {
```
acdenisSK: i'd favour for this to be an impl for `Arc<HelpCommand>`, as then anything that implements `HelpCommand` and is inside `Arc` will automatically impl `HelpCommand` as well.

src/framework/standard/command.rs
```diff
@@ -116,6 +123,128 @@ pub struct CommandOptions {
...
+    fn execute(&self, c: &mut Context, m: &Message, ho: &HelpOptions, hm: HashMap<String, Arc<CommandGroup>>, a: Args) -> Result<(), Error> {
+        (self.0)(c, m, ho, hm, a)
+    }
+}
+
+impl fmt::Debug for HelpOptions {
```
acdenisSK: This should be replaced with `#[derive(Debug)]`.

PR #232, opened 2017-12-12
Fix Guild::member_named

tahahawa:
Use rfind, in case there's more '#' in username than just the discriminator

Split at `pos+1` instead of `pos` and remove the trailing '#' in the split.0 (Name)

### DIFF
--- a/src/model/guild/mod.rs
+++ b/src/model/guild/mod.rs
@@ -725,11 +725,13 @@ impl Guild {
     ///
     /// [`Member`]: struct.Member.html
     pub fn member_named(&self, name: &str) -> Option<&Member> {
-        let (name, discrim) = if let Some(pos) = name.find('#') {
-            let split = name.split_at(pos);
+        let (name, discrim) = if let Some(pos) = name.rfind('#') {
+            let split = name.split_at(pos+1);
+            
+            let split2 = (split.0.get(0..split.0.len()-1).unwrap() , split.1);
 
-            match split.1.parse::<u16>() {
-                Ok(discrim_int) => (split.0, Some(discrim_int)),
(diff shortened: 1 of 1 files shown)


tahahawa:
If given a formatted string of "username#discrim", would split at '#' and end up with `("username", "#discrim")` instead of the needed `("username", "discrim")` and attempt to parse "#discrim" as a `u16` which would always fail

src/model/guild/mod.rs
```diff
@@ -725,11 +725,13 @@ impl Guild {
     ///
     /// [`Member`]: struct.Member.html
     pub fn member_named(&self, name: &str) -> Option<&Member> {
-        let (name, discrim) = if let Some(pos) = name.find('#') {
-            let split = name.split_at(pos);
+        let (name, discrim) = if let Some(pos) = name.rfind('#') {
+            let split = name.split_at(pos+1);
+            
+            let split2 = (split.0.get(0..split.0.len()-1).unwrap() , split.1);
```
zeyla: Could we take advantage of this Option and instead of unwrapping, map over it instead? Even though it's in theory safe to unwrap, it'd still be nice not to.

zeyla:
Also please re-target this for the v0.5.0 branch, where development on the next version is currently happening.

tahahawa:
Sure! Will do, as soon as I get around to it tonight. I just used the `unwrap()` because I was doing some quick testing on it.

I ran into the bug I mentioned in my previous comment while replacing a workaround I was using for the same functionality. Thought it was simple enough of a fix, that I might do it myself.

As for naming conventions, would the test be under `test_guild.rs`?

tahahawa:
Superseded by #233

PR #233, opened 2017-12-13
Fix Guild::member_named V0.5.0

tahahawa:
Made changes as requested, tests might be a bit jank, but they seem to work. Targeting the 5.0 branch


tahahawa:
Looking at this comment is how I designed the tests:
```
The following are valid types of searches:
**username**: "zey"
**username and discriminator**: "zey#5479"
**nickname**: "zeyla" or "zeylas#nick"
```
It seems the last case (nickname explicit search) wasn't implemented. That's what's causing the test fail. Should we remove that from the docs/comment, and remove the test? Or, implement it?

zeyla:
Let's remove the mention in the documentation and tests and leave the implementation for a different commit.

tahahawa:
Looks like the tests passed! 🙌

zeyla:
There seems to be a lot of random code style changes here, can we cut those out?

tahahawa:
Done

tahahawa:
Oops, forgot to fix the new testing stuff

tahahawa:
I think this is ready to merge, unless there's anything else to do

zeyla:
Thanks

PR #239, opened 2017-12-22
Multiple delimiter fix

Lakelezz:
This PR fixes `Args`-methods for multiple string delimiters, adds tests for `Args`-methods and adds benchmarking to the `Cargo.toml` (and benchmarks for mentioned methods).

Previous multiple-delimiter-solution found the first delimiter in the message and used that for every `Args`-operation.
This lead to the wrong behaviour, that if a quoted message contains a delimiter in the first quote (or prior the first actual and different delimiter in front of the next quote), `Args` would work with that instead of the intended delimiter.
It also failed on messages that contained multiple delimiters, imagine a set of delimiters used in a less abstract sentence (" and ", " or commas with and without trailing whitespace).

`Args::new()` checks whether delimiters are actually used inside the message, my local benchmarks showed me performance-increments, even with a reduction of only one or two delimiters.
If one delimiter is added to the framework-configuration or could be found in the message, `Args`-methods fallback to the previous procedure and save performance.

### DIFF
--- a/src/framework/standard/args.rs
+++ b/src/framework/standard/args.rs
@@ -80,20 +80,13 @@ fn parse_quotes<T: FromStr>(s: &mut String, delimiters: &[String]) -> Result<T, 
     let mut pos = second_quote_occurence(s).unwrap_or_else(|| s.len());
     let res = (&s[1..pos]).parse::<T>().map_err(Error::Parse);
 
-    pos += "\"".len();
+    pos += '"'.len_utf8();
 
-    if delimiters.len() == 1 {
-        if s[pos..].starts_with(&delimiters[0]) {
-            pos += delimiters[0].len();
-        }
+    for delimiter in delimiters {
 
-    } else {
-        for delimiter in delimiters {
-
-            if s[pos..].starts_with(delimiter) {
-                pos += delimiter.len();
-                break;
-            }
+        if s[pos..].starts_with(delimiter) {
+            pos += delimiter.len();
+            break;
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/framework/standard/args.rs
```diff
@@ -219,10 +234,79 @@ impl Args {
     ///
     /// assert_eq!(args.len(), 2); // `2` because `["42", "69"]`
     /// ```
-    pub fn len(&self) -> usize {
-        // TODO: Handle quotes too.
+    pub fn len(&mut self) -> usize {
+
```
acdenisSK: Unneeded newline.

src/framework/standard/args.rs
```diff
@@ -219,10 +234,79 @@ impl Args {
...
+    ///
+    /// ```rust
+    /// use serenity::framework::standard::Args;
+    ///
+    /// let mut args = Args::new(r#""42" "69""#, &[" ".to_string()]);
+    ///
+    /// assert_eq!(args.len_quoted(), 2); // `2` because `["42", "69"]`
+    /// ```
+    pub fn len_quoted(&mut self) -> usize {
+
```
acdenisSK: Again, the newline.

src/framework/standard/args.rs
```diff
@@ -235,9 +319,13 @@ impl Args {
     /// let mut args = Args::new("42 69", &[" ".to_string()]);
     ///
     /// assert_eq!(args.skip().unwrap(), "42");
-    /// assert_eq!(args, "69");
+    /// assert_eq!(args.full(), "69");
     /// ```
-    pub fn skip(&mut self) -> Option<String> { parse::<String>(&mut self.message, &self.delimiter).ok() }
+    pub fn skip(&mut self) -> Option<String> {
+
```
acdenisSK: Again.

PR #242, opened 2018-01-02
Fix docs for User::has_role

indiv0:
Update the docs for `User::has_role` to reflect that
`Into<GuildContainer>` is only implemented for `PartialGuild`,
`GuildId`, and `u64`, not `Guild`.


zeyla:
Thanks

PR #245, opened 2018-01-05
Fix `has_all_requirements`

Lakelezz:
This is supposed to be a fix for serenity's current help-commands on `v0.4.6`.

Bug: Help-command-features used in a DM-channel caused no help-commands to be displayed at all, due to `has_all_requirements` returning _false_ if the current channel is no guild
Solution: If `has_all_requirements` cannot find a guild, it will test whether the command is `dm_only` and not `guild_only`.

@Roughsketch shall test it before it gets merged! : )


Roughsketch:
Tested newest change and it works.

PR #248, opened 2018-01-08
Allow channels to be moved into and out of categories

jkcclemens:
This adds channel categories to the channel creation and channel editing methods.

Let me know if I need to change anything!

### DIFF
--- a/src/builder/edit_channel.rs
+++ b/src/builder/edit_channel.rs
@@ -1,5 +1,6 @@
 use internal::prelude::*;
 use utils::VecMap;
+use model::id::ChannelId;
 
 /// A builder to edit a [`GuildChannel`] for use via [`GuildChannel::edit`]
 ///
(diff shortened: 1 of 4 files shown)


src/builder/edit_channel.rs
```diff
@@ -72,4 +73,21 @@ impl EditChannel {
...
     }
+
+    /// The parent category of the channel.
+    ///
+    /// This is for [text] and [voice] channels only.
+    ///
+    /// [text]: ../model/enum.ChannelType.html#variant.Text
+    /// [voice]: ../model/enum.ChannelType.html#variant.Voice
+    pub fn category<C>(mut self, category: C) -> Self
+        where C: Into<Option<ChannelId>> {
```
acdenisSK: This should be just `ChannelId`, as since this is a builder the `Option` part becomes more or less implicit.

jkcclemens: How would you take a channel out of a category, then? As far as I understand it, since the PATCH method has all parameters optional, leaving the `parent_id` field out would not change anything.

acdenisSK: Oh, misunderstood the reasoning for the option 
(Since it seemed to almost do the same thing as serde, but explicitly). 
i'll merge it as is then.

PR #251, opened 2018-01-14
Added package.metadata.docs.rs at Cargo.toml to enable voice feature docs generation

MOZGIII:
According to this code it should work: https://github.com/onur/docs.rs/blob/13da7b2cd11fd6ea954c1d0145490586411700a4/src/docbuilder/metadata.rs#L19

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -117,3 +117,6 @@ standard_framework = ["framework", "vec_shift"]
 utils = ["base64"]
 voice = ["byteorder", "gateway", "opus", "sodiumoxide"]
 websocket = ["evzht9h3nznqzwl"]
+
+[package.metadata.docs.rs]
+features = ["default", "voice"]


acdenisSK:
This is already done in [v0.5.0](https://github.com/zeyla/serenity/blob/v0.5.0/Cargo.toml#L116-L117).

MOZGIII:
I see. Why it's not in master?

acdenisSK:
Because it's pointless, since docs.rs does not generate docs for unreleased code (not on crates.io).
Plus, master isn't the right branch for pr'ing stuff like this; nor are we using it as a dev branch.

MOZGIII:
Well, that's sad; there's no indication that master is not the dev branch.
I noticed you also seem to have release branches - what's the point of having a master branch?
Where can I learn more about the conventions this project is using?

acdenisSK:
`what's the point of having a master branch?`
The point is for linking to serenity's examples correctly and such; or for a more appropriate explanation: https://github.com/zeyla/serenity/issues/230#issuecomment-350499993

`Where can I learn more about the conventions this project is using?`
We don't a list for that yet, but is soon to arrive after v0.5.0 is released.

PR #255, opened 2018-01-21
Safe skipping for emote prefixes 

UninterestinAcc:
`[emoji prefix] [emoji]` caused an attempt to slice the second emote in the middle and not at the character boundary

* Can cause crashes, because I found it in the wild. 😓


src/framework/standard/mod.rs
```diff
@@ -971,7 +971,7 @@ impl Framework for StandardFramework {
 
                     let mut args = {
                         let mut content = message.content.chars().skip(position).collect::<String>();
-                        content = content[command_length..].trim().to_string();
+                        content = content.chars().skip(built.chars().count()).collect::<String>().trim().to_string();
```
acdenisSK: Change this to:
```rust
let content = message.content.chars().skip(position + built.chars().count()).collect::<String>();

Args::new(content.trim(), &self.configuration.delimiters)
```
(To reduce heap allocations)

PR #258, opened 2018-01-25
Strip RTP header extensions from voice stream

indiv0:
Add code to strip RTP header extensions from an incoming voice stream, if they are present.
See https://tools.ietf.org/html/rfc5285 for more info on RTP header extensions, and see https://github.com/discordjs/discord.js/issues/1310 for an issue in Discord.js which discussed this problem.

This PR is based on the fix implemented in https://github.com/discordjs/discord.js/commit/99419a36706a52ac2372712a86229c3e0a2c0666 by hydrabolt.


zeyla:
Thanks

PR #265, opened 2018-01-31
Multiple audio stream playback, volume control, pausing

FelixMcFelix:
Changes `Handler::play(...)` to return a shared memory object allowing some ongoing control of active audio clips. This should allow for more detailed control in future e.g. events tied to sound playback, termination, etc.

Crucially, users can now simultaneously play multiple streams over one voice connection.
Accordingly, this changes the PCM pipeline to combine all incoming streams in `f32` space.
At present, I am not decoding explicit Opus frames to join them in this fashion.
Old functionality is recreated in `Handler::play_only(...)`.

I expect some of the struct names/organisation may not be the best, feedback would be much appreciated.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -5,6 +5,8 @@ pub const EMBED_MAX_LENGTH: u16 = 6000;
 /// The gateway version used by the library. The gateway URI is retrieved via
 /// the REST API.
 pub const GATEWAY_VERSION: u8 = 6;
+/// The voice gateway version used by the library.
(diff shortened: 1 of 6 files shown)


zeyla:
Looks good, thanks for the PR! I cleaned up a few minour things, but the biggest change was moving the new `Handler::play` code to `Handler::play_returning`. Changing the return type is a breaking change, as someone could rely on a function returning `()`.

PR #273, opened 2018-02-10
Add a way to reorder roles

jkcclemens:
Adds [this](https://discordapp.com/developers/docs/resources/guild#modify-guild-role-positions) route, used to reorder roles in a guild.

The Discord docs say that this route only requires the `MANAGE_ROLES` permission, but my tests were consistently giving me permission errors. This makes me wonder if bots can even use this route, if my server was just messed up, or if the code is bad.

A review would be nice.


acdenisSK:
After testing the route as well, it seems bots *are* forbidden from using it. Which is strange as i don't see a reason why they shouldn't be able to.
Oh

PR #274, opened 2018-02-13
Suggest similar commands upon failing to find a matching name

Lakelezz:
Serenity's help-feature originally had a text telling the user that their attempted help-argument was an alias and recommended the command-name.
I removed that behaviour already but the field `suggestion_text` became nothing but a piece of legacy without any purpose.

With this PR, `suggestion_text` will provide _similar_ command-names upon failing to find a matching command-name.
E.g. `test command` would be a valid name, but `test comman` (without `d`) is being entered, the help-feature will still suggest `test command` if the `max_edit_distance`-option is set to `1`. If multiple commands are considered similar, they will be displayed as well.

One can completely opt-out of this behaviour by setting `find_similar_commands` to `false` or on certain commands by setting their `help_suggested`-field to `false`.


src/framework/standard/help_commands.rs
```diff
@@ -81,6 +82,95 @@ pub fn has_all_requirements(cmd: &Arc<CommandOptions>, msg: &Message) -> bool {
...
+            let guild = guild.read();
+
+            if let Some(member) = guild.members.get(&msg.author.id) {
+
+                if command_options.help_available && has_correct_permissions(command_options, msg) {
+
+                    if has_correct_roles(command_options, &guild, &member) {
+                        result.push(format!("`{}`", command_name.clone()));
+                    } else {
+                        match help_options.lacking_role {
```
acdenisSK: These matches are unnecessarily repeated.
It'd be more appropriate to put this logic into a macro:
```rust
macro_rules! format_help {
    ($behaviour:expr) => {
        match $behaviour {
            HelpBehaviour::Strike => result.push(format!("~~`{}`~~", command_name)),
            HelpBehaviour::Nothing => result.push(format!("`{}`", command_name)),
            HelpBehaviour::Hide => {},
        }
    };
}
```
then it's plainly just
```rust
format_help!(help_options.lacking_role);
```
and so on.

src/framework/standard/help_commands.rs
```diff
@@ -81,6 +82,95 @@ pub fn has_all_requirements(cmd: &Arc<CommandOptions>, msg: &Message) -> bool {
...
+
+        if let Some(guild) = msg.guild() {
+            let guild = guild.read();
+
+            if let Some(member) = guild.members.get(&msg.author.id) {
+
+                if command_options.help_available && has_correct_permissions(command_options, msg) {
+
+                    if has_correct_roles(command_options, &guild, &member) {
+                        result.push(format!("`{}`", command_name.clone()));
```
acdenisSK: redundant `.clone()`

src/framework/standard/help_commands.rs
```diff
@@ -81,6 +82,95 @@ pub fn has_all_requirements(cmd: &Arc<CommandOptions>, msg: &Message) -> bool {
...
+                }
+            } else {
+                match help_options.lacking_permissions {
+                    HelpBehaviour::Strike => result.push(format!("~~`{}`~~", command_name.clone())),
+                    HelpBehaviour::Nothing => result.push(format!("`{}`", command_name.clone())),
+                    HelpBehaviour::Hide => (),
+                }
+            }
+        } else {
+            result.push(format!("`{}`", command_name.clone()));
```
acdenisSK: same here

Issue #277, opened 2018-02-15
`ffmpeg` sources leak zombie processes


FelixMcFelix:
Noticed this after running [felyne](https://github.com/felixmcfelix/felyne-bot) last night, and returning to find ~30 000 zombie processes. The bot I'm working with plays many short-lived sounds to completion, but keeps at most 2 active sounds. This is verified by monitoring the amount of sounds queued up in the audio thread.

The issue seems to stem from [an implementation detail of Child](https://doc.rust-lang.org/std/process/struct.Child.html):
> There is no implementation of Drop for child processes, so if you do not ensure the Child has exited then it will continue to run, even after the Child handle to the child process has gone out of scope.

Naturally all we'd need would be a custom `impl Drop` for `ChildContainer`, but this would also be alleviated by moving to e.g. [rust-ffmpeg](https://github.com/meh/rust-ffmpeg).

I'd be happy to work on this, although I'm not sure which avenue you'd prefer.

zeyla:
Closed by #278

PR #278, opened 2018-02-16
Drop implementation for ffmpeg container (#277)

FelixMcFelix:
Simple edit, should prevent zombie ffmpeg processes.
Unsure if this should be prepended by a `.kill()` call, but this seems to work in practice.


zeyla:
Thanks

PR #280, opened 2018-02-24
Pad user discriminators in content_safe to 4 digits

megumisonoda:
This brings the function in line with the 'tag' function for User models, and with the official Discord app experience and other libraries.

I'm not sure whether or not this constitutes a 'breaking change', but this improves consistency.


zeyla:
Thanks

Issue #167, opened 2017-09-28
Add `Role::guild_id`


acdenisSK:
Rather than enforcing users to search each guild that has that role for the guild's id, grant a simple `guild_id` field to `Role` directly to avoid this irritation.

The effort for this is relatively small:
- When deserializing a [`Guild`](https://github.com/serenity-rs/serenity/blob/master/src/model/guild/mod.rs#L1520-L1678): Fill each role pertaining to the guild with its id (obtained at line #1524)
- When deserializing a [`GuildRoleCreateEvent`](https://github.com/serenity-rs/serenity/blob/master/src/model/event.rs#L637): Use the `guild_id` in the struct to accomplish this.

It is advisable to NOT modify the json and let serde do the rest; this should be our job.
(Related to this are the `members`, `channels` fields in `Guild`'s deserialize code.)

Lakelezz:
Added as of #790.

PR #291, opened 2018-03-16
GuildId changes

Roughsketch:
Attempting to address issue #167 

`Role`, `GuildChannel`, and `Member` will no longer deserialize with a guild id. Instead, they will now get their guild id's assigned from the `Guild` deserializer after the rest of the struct is deserialized.

Also GuildRoleCreateEvent has a new custom deserializer that assigns the guild id to the associated role.

### DIFF
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -212,6 +212,10 @@ pub fn create_channel(guild_id: u64, map: &Value) -> Result<GuildChannel> {
 
     serde_json::from_reader::<HyperResponse, GuildChannel>(response)
         .map_err(From::from)
+        .and_then(|mut channel| {
+            channel.guild_id = guild_id;
+            Ok(channel)
+        })
 }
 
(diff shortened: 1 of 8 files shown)


Roughsketch:
I'm not sure if I've covered all the cases, or even if this setup will actually work. In particular, it looks like there are certain endpoints in the http section that deserialize a GuildChannel directly and don't take a guild_id parameter. This doesn't matter if the payloads are still missing guild ids, but if they are included sporadically it is possible that some will be missed.

In particular I'm worried about the following sections:
`GuildChannel` being deserialized within `Channel`.
https://github.com/zeyla/serenity/blob/v0.6.x/src/model/channel/mod.rs#L491

`edit_channel` in http not taking a guild_id parameter leaves the resulting `GuildChannel` without one
https://github.com/zeyla/serenity/blob/v0.6.x/src/http/mod.rs#L669

And there may be more. I was unaware of these sections and the ones from the latest commit until very recently.

PR #295, opened 2018-03-21
Make some macros indexable #294

### LINKED ISSUE #294: Make `ChannelId` indexable
I can't even find it manually, would it be possible to just define the base struct without whatever (I'm assuming) macro?

ForsakenHarmony:
You can reject it if you don't like it

This also includes some shorthands for struct creation

Closes #294

### DIFF
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -26,6 +26,7 @@ use utils::VecMap;
 
 #[cfg(feature = "utils")]
 use utils::Colour;
+use std::u64;
 
 /// A builder to create a fake [`Embed`] object, for use with the
 /// [`ChannelId::send_message`] and [`ExecuteWebhook::embeds`] methods.
@@ -451,7 +452,7 @@ pub struct Timestamp {
 impl From<String> for Timestamp {
     fn from(ts: String) -> Self {
         Timestamp {
-            ts: ts,
+            ts,
         }
     }
 }
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -26,7 +26,6 @@ use utils::VecMap;
 
 #[cfg(feature = "utils")]
 use utils::Colour;
-use std::u64;
 
 /// A builder to create a fake [`Embed`] object, for use with the
 /// [`ChannelId::send_message`] and [`ExecuteWebhook::embeds`] methods.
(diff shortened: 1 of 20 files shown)


src/voice/audio.rs
```diff
@@ -7,7 +7,7 @@ pub const SAMPLE_RATE: u32 = 48_000;
 
 /// A readable audio source.
 pub trait AudioSource: Send {
-    fn is_stereo(&mut self) -> bool;
```
acdenisSK: Irrelevant [to this pr] and (possibly?) an unnecessary breaking change. So revert this please.

src/model/channel/message.rs
```diff
@@ -14,6 +14,7 @@ use std::fmt::Write;
 use std::mem;
 #[cfg(feature = "model")]
 use {constants, http, utils as serenity_utils};
+use std::{i64, u64};
```
acdenisSK: Unnecessary imports.

src/internal/timer.rs
```diff
@@ -1,6 +1,7 @@
 use chrono::{DateTime, Duration, Utc};
 use std::time::Duration as StdDuration;
 use std::thread;
+use std::{i64};
```
acdenisSK: Unnecessary import.

src/client/bridge/gateway/shard_messenger.rs
```diff
@@ -3,6 +3,7 @@ use model::prelude::*;
 use super::{ShardClientMessage, ShardRunnerMessage};
 use std::sync::mpsc::{SendError, Sender};
 use websocket::message::OwnedMessage;
+use std::result::Result;
```
acdenisSK: Unnecessary import.

src/builder/create_embed.rs
```diff
@@ -26,6 +26,7 @@ use utils::VecMap;
 
 #[cfg(feature = "utils")]
 use utils::Colour;
+use std::u64;
```
acdenisSK: Unnecessary import.

src/internal/timer.rs
ForsakenHarmony: Idk why but my ide was complaining when it wasn't imported, but I can remove it again

src/voice/audio.rs
ForsakenHarmony: I mean it should just read a value, and the only use does that, but you're right it's unrelated

ForsakenHarmony: that's why I did it

ForsakenHarmony:
Something seems to be wrong with travis?

PR #297, opened 2018-03-28
Remove user account relations

Lakelezz:
Serenity had some functions and comment-segments left that related to user accounts or were only usable by them, hence I removed these bits.

Nonetheless, the function `create` in `src/model/guild/mod.rs` is still left, as it only mentions `This endpoint is usually only available for user accounts.`, but does not explicitly exclude bot accounts.

By the way, I added a link to https://discordapp.com/developers/docs/resources/user#modify-current-user on the `edit_profile`-function, could be useful for someone : )


src/model/guild/emoji.rs
```diff
@@ -41,51 +41,47 @@ pub struct Emoji {
...
-    /// }
-    /// ```
-    #[cfg(feature = "cache")]
-    pub fn delete(&self) -> Result<()> {
-        match self.find_guild_id() {
-            Some(guild_id) => http::delete_emoji(guild_id.0, self.id.0),
-            None => Err(Error::Model(ModelError::ItemMissing)),
-        }
-    }
+   /// Deletes the emoji.
```
zeyla: Looks like all of this is now only 3-space indented

zeyla:
Thanks

Issue #304, opened 2018-04-11
Audit logs can never deserialize correctly if populated


jkcclemens:
Once #303 is merged, attempting to deserialize the audit logs produces an error on the `target_id` field, since it's a `u64` in the struct but a `String` in the received data.

See [code](https://github.com/zeyla/serenity/blob/a6eda12980d7410c8be41a4f2748bd57d65fcb52/src/model/guild/audit_log.rs#L214).

jkcclemens:
`target_id` and `count` are fixed, but all the fields in `Options` are optional. They're not wrapped in `Option`, though, so serialization fails.

PR #305, opened 2018-04-11
fix: make the options in Options optional

jkcclemens:
See #304.

### DIFF
--- a/src/model/guild/audit_log.rs
+++ b/src/model/guild/audit_log.rs
@@ -6,7 +6,10 @@ use serde::de::{
     MapAccess,
     Visitor
 };
-use serde::ser::Serializer;
+use serde::ser::{
+    Serialize,
+    Serializer
+};
 use super::super::prelude::*;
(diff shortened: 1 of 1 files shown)


src/model/guild/audit_log.rs
```diff
@@ -285,6 +293,26 @@ mod u64_handler {
...
+        if let Some(s) = s {
+            return Ok(s.parse().map(Some).map_err(|e| de::Error::custom(e))?);
+        }
+
+        Ok(None)
+    }
+
+    pub fn serialize<S: Serializer>(num: &Option<u64>, s: S) -> StdResult<S::Ok, S::Error> {
+        match num {
+            Some(n) => s.serialize_u64(*n),
```
jkcclemens: Changing this to `Option::serialize(num, s)` compiles, and I'm pretty sure it does the same thing. Should I change it?

jkcclemens:
help

PR #306, opened 2018-04-13
Send silence frames upon connection (Fix #301)

### LINKED ISSUE #301: Voice receive not initially working after connecting to voice channel
With the latest master (a6eda12 at the time of writing) voice receive appears to be wonky — the example 10_voice_receive doesn’t work as advertised.

We narrowed it down to an issue with discord and how they appear to handle (or not handle) initialization of voice connections. Essentially an implementation of `AudioReceiver`s `speaking_update` will fire if you add print statements or other work to it, but the `voice_packet` function doesn’t seem to do anything.

The workaround that was figured out for now is to send audio using the bot on connection to a voice channel; after that, incoming packets seem to log as you’d expect.

It sounds like JDA has a similar bug logged so I don’t think it’s necessarily a serenity thing, but we may want an answer for this, even if it’s just “This is known, send a packet first”.

FelixMcFelix:
* Sends 100 empty frames at discord upon connection.
* Moves silent frame def out into the constants file (named).

Works for both Example 10 and the modified version of my own bot.

### DIFF
--- a/src/voice/audio.rs
+++ b/src/voice/audio.rs
@@ -6,6 +6,7 @@ use std::{
 
 pub const HEADER_LEN: usize = 12;
 pub const SAMPLE_RATE: u32 = 48_000;
+pub static SILENT_FRAME: [u8; 3] = [0xf8,0xff,0xfe];
 
 /// A readable audio source.
 pub trait AudioSource: Send {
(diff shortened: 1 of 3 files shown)


src/voice/audio.rs
```diff
@@ -6,6 +6,7 @@ use std::{
 
 pub const HEADER_LEN: usize = 12;
 pub const SAMPLE_RATE: u32 = 48_000;
+pub static SILENT_FRAME: [u8; 3] = [0xf8,0xff,0xfe];
```
zeyla: Let's put a space after each of these

zeyla:
Thanks

PR #310, opened 2018-04-17
Fix voice hang on Audio `.stop()`.

FelixMcFelix:
Turns out `wait` was a little *too* cooperative. Waiting on an audio stream that hadn't yet finished (i.e., killing all existing sounds) caused the voice thread to hang since ffmpeg would never terminate...

Holdover until we change to real ffmpeg bindings.


zeyla:
Thanks

Issue #313, opened 2018-05-22
Make MessageBuilder use &mut self instead of self


PvdBerg1998:
`MessageBuilder` takes ownership of `self` in each function, which makes is unusable in for example an iterator. Replacing `self` with `&mut self` would fix this problem.

PvdBerg1998:
I see this is fixed in the 0.6.x branch.

PvdBerg1998:
Sorry I'm blind

zeyla:
Closed by https://github.com/serenity-rs/serenity/commit/67155c71af37c6e4056b153ab63a11424cf78334 and #314.

PR #314, opened 2018-05-22
Change message builder to use &mut self instead of self

PvdBerg1998:
This way the builder can be used where taking ownership is impractical.
Solves #313 .


PvdBerg1998:
Note this requires a clone upon building. Maybe there should be a consuming alternative build method to prevent unnecessary clones? Or people could just access builder.0 if they want to optimize their code.

zeyla:
Why change `MessageBuilder::build` to take `&self` instead of `self`?

I can't think of too many use cases where it's better for it to clone, the user should have to explicitly clone if they want that.

zeyla:
Due to some branch rebasing your PR is mostly unmergeable now, I'll clean it up and commit under your name.

zeyla:
Committed as https://github.com/serenity-rs/serenity/commit/67155c71af37c6e4056b153ab63a11424cf78334, thanks.

PR #323, opened 2018-05-29
Add methods to check whether `Message` is mentioning `User`

Lakelezz:
Two methods, one that accepts `u64` Ids and the other accepting the `User`, to check whether a `Message` contains a mention.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -490,6 +490,18 @@ impl Message {
         http::send_message(self.channel_id.0, &map)
     }
 
+    /// Checks whether the message mentions the user of given Id.
+    pub fn mentions_user_id(&self, id: u64) -> bool {
+        let search_for = vec![format!("<@{}>", id), format!("<@!{}>", id)];
+
+        search_for.iter().any(|ref mention| self.content.contains(mention.as_str()))
+    }
+
+    /// Checks whether the message mentions `User`.
+    pub fn mentions_user(&self, user: User) -> bool {
+        self.mentions_user_id(user.id.0)
+    }
+
     /// Unpins the message from its channel.
     ///
     /// **Note**: Requires the [Manage Messages] permission.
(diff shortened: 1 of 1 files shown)


src/model/channel/message.rs
```diff
@@ -490,6 +490,18 @@ impl Message {
         http::send_message(self.channel_id.0, &map)
     }
 
+    /// Checks whether the message mentions the user of given Id.
+    pub fn mentions_user_id(&self, id: u64) -> bool {
```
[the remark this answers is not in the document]
Lakelezz: The reason I picked `u64` was to allow very direct usage if the Id is already known (compile-time).
E.g. `message.mentions_user_id(123)`
For every other "higher level" case of usage, I thought `mentions_user` would be sufficient, since it takes the effort of accessing the `UserId`.

Nonetheless, it should be even more explicit and yet still simple to construct `UserId` if wanted for direct Id usage, so I agree and changed accordingly.

src/model/channel/message.rs
```diff
@@ -490,16 +490,18 @@ impl Message {
...
 
-    /// Checks whether the message mentions the user of given Id.
-    pub fn mentions_user_id(&self, id: u64) -> bool {
-        let search_for = vec![format!("<@{}>", id), format!("<@!{}>", id)];
-
-        search_for.iter().any(|ref mention| self.content.contains(mention.as_str()))
+    /// Checks whether the message mentions the user of given [`UserId`].
+    ///
+    /// [`UserId`]: ../model/user/struct.UserId.html
+    pub fn mentions_user_id(&self, user_id_to_find: &UserId) -> bool {
```
acdenisSK: It'd be better to genericify this.
```rust
fn mentions_user_id<I: Into<UserId>>(&self, id: I) -> bool {
    let user_id_to_find = id.into();
    ....
}
```

Lakelezz: Yes, I agree.

PR #325, opened 2018-05-31
Update voice example to join channel without passing channel id argument

drklee3:
This change checks the guild's voice states to find the command invoker's voice channel ID instead of requiring the user to pass the ID as an argument.

### DIFF
--- a/examples/06_voice/src/main.rs
+++ b/examples/06_voice/src/main.rs
@@ -19,7 +19,6 @@ use serenity::client::{CACHE, Client, Context, EventHandler};
 use serenity::framework::StandardFramework;
 use serenity::model::channel::Message;
 use serenity::model::gateway::Ready;
-use serenity::model::id::ChannelId;
 use serenity::model::misc::Mentionable;
 // Import the `Context` from the client and `parking_lot`'s `Mutex`.
 //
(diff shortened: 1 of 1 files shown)


examples/06_voice/src/main.rs
```diff
@@ -109,23 +108,32 @@ command!(deafen(ctx, msg) {
     }
 });
 
-command!(join(ctx, msg, args) {
-    let connect_to = match args.single::<u64>() {
-        Ok(id) => ChannelId(id),
-        Err(_) => {
-            check_msg(msg.reply("Requires a valid voice channel ID be given"));
+command!(join(ctx, msg) {
+    let guild_id = match CACHE.read().guild_channel(msg.channel_id) {
```
acdenisSK: I'd advise just using [`Message::guild`](https://docs.rs/serenity/0.5.3/serenity/model/channel/struct.Message.html#method.guild) here, as displayed by the usage of `Guild` below.

acdenisSK: No, i meant in this way to simplify fetching the channel id:
```rust
let guild = match msg.guild() {
    Some(g) => g,
    None => {
        check_msg(msg.channel_id.say("Groups and DMs not supported"));

        return Ok(());
    },
};

let channel_id = guild
    .read()
    .voice_states.get(&msg.author.id)
    .map(|vs| vs.channel_id);
```

PR #326, opened 2018-06-02
Modify `Args` internals to a lexer approach

acdenisSK:
This is a overhaul method-of-parsing things into "arguments" to an approach that's:
1. Better Unicode complaint
2. Easy to understand
3. Faster (by my measurements, around ~10k nanoseconds)

Also, for convenience, this adds 2 new methods:
- `rewind` - Goes one step behind
- `restore` - Goes back to the start

*(And tweaks docs as well)*

### DIFF
--- a/src/framework/standard/args.rs
+++ b/src/framework/standard/args.rs
@@ -452,7 +452,7 @@ impl Args {
     /// 
     /// let mut args = Args::new("42 69 95", &[" ".to_string()]);
     ///
-    /// // Lets parse 'em numbers!
+    /// // Let's parse 'em numbers!
     /// assert_eq!(args.single::<u32>().unwrap(), 42);
     /// assert_eq!(args.single::<u32>().unwrap(), 69);
     /// assert_eq!(args.single::<u32>().unwrap(), 95);
--- a/src/framework/standard/args.rs
+++ b/src/framework/standard/args.rs
@@ -677,8 +677,9 @@ impl Args {
     /// Returns the first argument that can be parsed and removes it from the message. The suitable argument 
     /// can be in an arbitrary position in the message.
     ///
-    /// **Note**: This replaces all delimiters within the message
-    /// by the first set in your framework-config to win performance.
+    /// **Note**: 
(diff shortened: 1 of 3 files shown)


acdenisSK:
@serenity-rs/ferris-elite-collection-of-epic-code please review

src/framework/standard/args.rs
```diff
@@ -299,10 +411,65 @@ impl Args {
...
+    /// Go back to the starting point.
+    /// 
+    /// # Examples
+    /// 
+    /// ```rust
+    /// use serenity::framework::standard::Args;
+    /// 
+    /// let mut args = Args::new("42 69 95", &[" ".to_string()]);
+    ///
+    /// // Lets parse 'em numbers!
```
Lakelezz: Should be `Let's`.

src/framework/standard/args.rs
```diff
@@ -299,10 +411,65 @@ impl Args {
...
+    /// assert_eq!(args.single::<u32>().unwrap(), 42);
+    /// assert_eq!(args.single::<u32>().unwrap(), 69);
+    /// assert_eq!(args.single::<u32>().unwrap(), 95);
+    /// 
+    /// // Oh, no! I actually wanted to multiply all of them by 2! 
+    /// // I don't want to call `rewind` 3 times manually....
+    /// // Wait, i could just go entirely back!
```
Lakelezz: Forgot to capitalise `i`.

src/framework/standard/args.rs
```diff
@@ -316,25 +483,12 @@ impl Args {
     ///
     /// assert_eq!(args.len_quoted(), 2); // `2` because `["42", "69"]`
     /// ```
+    #[deprecated(since = "0.5.3", note = "its task was merged with `len`, please use it instead.")]
```
Lakelezz: Should depreciation-notes start capitalised?

acdenisSK: Not really.
Though, i suppose i could.

PR #336, opened 2018-06-17
Add servermom methods to http module

TheUnitedStatesOfAmerica:
Originally we had `ban_zeyla` as an april fools joke. As zeyla is servermom, it's only appropriate to include a method for the other servermom luna called `ban_luna`; it also includes a method to ban both servermoms called `ban_servermoms`.

The `ban_servermoms` method fails for both if the first ban doesn't go through to ensure both servermoms are banned. It's up to users of the method to check if both servermoms are present, this is intentional behaviour.

### DIFF
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -33,18 +33,18 @@ pub use hyper::status::{StatusClass, StatusCode};
 use constants;
 use hyper::{
     client::{
-        Client as HyperClient, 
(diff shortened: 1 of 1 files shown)


perryprog:
Why hasn't this been merged yet? Is there something we are waiting for?

zeyla:
Thanks

PR #337, opened 2018-06-18
Provide help-details for all visible commands

Lakelezz:
Before, the help-features did not display further information about a command that was unable to use due to wrong channel (e.g. guild-only-command used via direct message).

From now the behaviour - whether a command is available for further details - will work as following:
When the command would be normally displayed or strikethrough, the command will be available for further details via `help command-name`.

I will need to add this behaviour to `v0.6.x` once the framework is working again and #274 is merged.
On another note, this might depreciate `has_all_requirements`, since Serenity will no longer use it - might keep it for the utility.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -28,13 +28,13 @@ use client::Context;
 use framework::standard::{has_correct_roles, has_correct_permissions};
 use model::{
     channel::Message,
-    id::ChannelId
+    id::ChannelId,
 };
 use std::{
     collections::HashMap,
     hash::BuildHasher,
     sync::Arc,
-    fmt::Write
+    fmt::Write,
 };
 use super::command::InternalCommand;
 use super::{
@@ -44,7 +44,7 @@ use super::{
     HelpOptions,
     CommandOptions,
     CommandError,
-    HelpBehaviour
+    HelpBehaviour,
 };
 use utils::Colour;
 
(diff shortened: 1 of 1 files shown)


src/framework/standard/help_commands.rs
```diff
@@ -88,6 +88,46 @@ pub fn has_all_requirements(cmd: &Arc<CommandOptions>, msg: &Message) -> bool {
...
+                } else {
+                    return help_options.lacking_permissions != HelpBehaviour::Hide;
+                }
+            }
+        }
+    } else {
+        return help_options.wrong_channel != HelpBehaviour::Hide;
+    }
+
+    return false;
```
acdenisSK: At end of function, so this could be changed to an implicit return (just `false`).

src/framework/standard/help_commands.rs
```diff
@@ -88,6 +88,46 @@ pub fn has_all_requirements(cmd: &Arc<CommandOptions>, msg: &Message) -> bool {
...
+    if !command_options.dm_only && !command_options.guild_only
+    || command_options.dm_only && msg.is_private()
+    || command_options.guild_only && !msg.is_private() {
+
+        if let Some(guild) = msg.guild() {
+            let guild = guild.read();
+
+            if let Some(member) = guild.members.get(&msg.author.id) {
+
+                if command_options.help_available {
```
acdenisSK: Since we return the same bool exp. (`help_options.lacking_role != HelpBehaviour::Hide`) in both this if and the `has_correct_permissions` if's else clauses, it'd make sense to merge these as a single if
```rust
if command_options.help_available && has_correct_permissions(command_options, msg) {
    if has_correct_roles(command_options, &guild, &member) {
        return true;
    } else {
        return help_options.lacking_role != HelpBehaviour::Hide;
    }
} else {
    return help_options.lacking_permissions != HelpBehaviour::Hide;
}
```

Lakelezz: That was my mistake, it should just return false if no help is available for a command : )

PR #338, opened 2018-06-24
Add CreateEmbed::Field/s to 11_create_message_builder example

drklee3:
Adds usage of both `CreateEmbed::Field` and `CreateEmbed::Fields`.


vityafx:
Are fields added to or replaced by each other in such case? Will there be 3 fields?

PR #349, opened 2018-07-12
Add checks for groups

Lakelezz:
Adds the abillity to create a check for an entire group, behaves similar to checks in commands.


acdenisSK:
Disregarding the conflict, could you update the [framework](https://github.com/serenity-rs/serenity/blob/master/examples/05_command_framework/src/main.rs) example too? As in, adding an `Admin` group with commands usable only by the owner of the bot, or similar.

src/framework/standard/command.rs
```diff
@@ -113,10 +115,29 @@ impl Default for CommandGroup {
             owners_only: false,
             allowed_roles: Vec::new(),
             help: None,
+            checks: Vec::new(),
         }
     }
 }
 
+impl fmt::Debug for CommandGroup {
```
acdenisSK: Since `checks` is using a struct with `Debug` impl'd now, this can be replaced with `#[derive(Debug)]`.

PR #352, opened 2018-07-23
Fix default-commands if dispatch started via a mention of the bot

Lakelezz:
If a bot gets mentioned, the default command creates unwanted behaviour and always triggers the default-command even if a command-name has been passed.

I slightly changed the `round`-creation. 
Before, `round` got created by skipping needed amount of `Char`s, then collected into a `String`. After that, a new `round` was created that includes trimming.
We combine the skipping and then trimming into one step, which allows us to compare the `original round` (but this time trimmed) with `built` (based on trimmed `round`) while its built up.

### DIFF
--- a/src/framework/standard/mod.rs
+++ b/src/framework/standard/mod.rs
@@ -961,6 +961,16 @@ impl StandardFramework {
     }
 }
 
+fn skip_chars_and_trim_to_new_string(str_to_transform_to_chars: &str, chars_to_skip: usize) -> String {
+    let mut chars = str_to_transform_to_chars.chars();
+
+    if chars_to_skip > 0 {
+        chars.nth(chars_to_skip - 1);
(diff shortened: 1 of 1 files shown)


src/framework/standard/mod.rs
```diff
@@ -961,6 +961,16 @@ impl StandardFramework {
     }
 }
 
+fn skip_chars_and_trim_to_new_string(str_to_transform_to_chars: &str, chars_to_skip: usize) -> String {
```
zeyla: This should be able to return a Cow.

zeyla:
Running travis and then will merge if it passes plus the nit.

src/framework/standard/mod.rs
perryprog: 🐄

PR #356, opened 2018-07-30
Make GuildId::member use the cache when possible

Roughsketch:


acdenisSK:
Could you also document the behaviour of `member`?
i.e: `Searches through the cache (if enabled). If it fails, resorts to a http request`

PR #357, opened 2018-08-01
Pass failed sub-command to default-command

Lakelezz:
If a user dispatches a command that shall be a sub-command but given sub-name does not exist, the default-command of given group shall receive the failed sub-name.
Prior this pull request, an empty argument was passed to the default-command, which was not intended.

`After` shall still receive the full length, including the group-prefix, in order to gain a bit more information about what group and what failed sub-command has been attempted to use.

### DIFF
--- a/src/framework/standard/mod.rs
+++ b/src/framework/standard/mod.rs
@@ -1147,7 +1147,7 @@ impl Framework for StandardFramework {
 
                     if check_contains_group_prefix {
 
-                        if let Some(CommandOrAlias::Command(ref command)) = &group.default_command {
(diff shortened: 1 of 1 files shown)


src/framework/standard/mod.rs
```diff
@@ -1146,12 +1146,13 @@ impl Framework for StandardFramework {
                     }
 
                     if check_contains_group_prefix {
-                        if let &Some(CommandOrAlias::Command(ref command)) = &group.default_command {
+
```
acdenisSK: Extraneous whitespace

PR #358, opened 2018-08-06
Fix default command upon shortcut prefix and passing sub-commands to default-command

Lakelezz:
This pull request makes sure that group-prefixes that are a substring of another group-prefix are never picked when they are not explicitly meant to be used.

On another note, this pull requests moves the creation of `Args` close to the actual used code-locations, causing three seperated creations.
One of these is modified to enable the default-command to gain information about what sub-command has been erroneously attempted to be used.

To finish it, the framework-example has been updated to showcase the possibilities with the default-command.

### DIFF
--- a/src/framework/standard/mod.rs
+++ b/src/framework/standard/mod.rs
@@ -1043,8 +1043,9 @@ impl Framework for StandardFramework {
                         // than the last matching one, this prevents picking a wrong prefix,
                         // e.g. "f" instead of "ferris" due to "f" having a lower index in the `Vec`.
                         let longest_matching_prefix_len = prefixes.iter().fold(0, |longest_prefix_len, prefix|
-                            if prefix.len() > longest_prefix_len && built.starts_with(prefix)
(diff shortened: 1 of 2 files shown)


src/framework/standard/mod.rs
```diff
@@ -1079,6 +1073,13 @@ impl Framework for StandardFramework {
 
                         if let Some(help) = help {
                             let groups = self.groups.clone();
+                            let mut args = {
+                                let content = message.content.chars().skip(position).skip_while(|x| x.is_whitespace())
```
acdenisSK: Could you add a comment stating that the `Args` creation at [#1108](https://github.com/serenity-rs/serenity/blob/fe3979a45732f96f5f3dc2d84479aac02e5f6b39/src/framework/standard/mod.rs#L1108) should also be updated (even reverse) when changing this?
Because as they're now, they're identical. I also doubt that the args for a help command should be different from a normal command.

PR #365, opened 2018-08-10
Use one macro instead of the same code twice, avoid seperated changes

Lakelezz:
Instead of having the exact same logic and code twice, this pull request uses a macro.

This has three advantages:
1. Code only needs to be changed inside the macro, thus no one can update one code-section and forget the other.
2. Less code repetition.
3. We do not have to rely on comments to describe that someone has to find and change both pieces of code.


Lymia:
why a macro and not a function

PR #367, opened 2018-08-11
Use `to_`- and `as_`-methods instead of `get` and `find` on `Id`-`newtype`s

Lakelezz:
Following [C-CONV](https://rust-lang-nursery.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) and [C-GETTER](https://rust-lang-nursery.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter), I updated all `Id`-`newtype`s to align with the guidelines.

`get` on its own did not clearly state what would be retrieved, which is made clear by using `to_`-equivalents.

`as_u64` and `as_mut_u64` have been implemented for all `Id`-`newtype`s as well, as it is clearer than accessing the 0th field, as in `some_id.0`.

On another note, I improved a little bit on the documentation and fixed one wrong syntax-issue that prevented links from being generated.

### DIFF
--- a/src/model/channel/channel_id.rs
+++ b/src/model/channel/channel_id.rs
@@ -312,7 +312,7 @@ impl ChannelId {
     /// upon failure requests it via the REST API.
     ///
     /// **Note**: If the cache is not enabled,
-    /// REST API will be used instead.
+    /// REST API will be used only.
     ///
     /// [`Channel`]: ../channel/enum.Channel.html
     pub fn to_channel(&self) -> Result<Channel> {
--- a/src/model/channel/channel_id.rs
+++ b/src/model/channel/channel_id.rs
@@ -390,7 +390,7 @@ impl ChannelId {
         use self::Channel::*;
 
         let finding = feature_cache! {{
-            Some(self.find())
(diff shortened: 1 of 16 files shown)


src/model/channel/channel_id.rs
```diff
@@ -285,10 +285,18 @@ impl ChannelId {
...
+    ///
+    /// [`Channel`]: ../channel/enum.Channel.html
+    #[cfg(feature = "cache")]
+    pub fn to_channel_cached(self) -> Option<Channel> { CACHE.read().channel(self) }
+
     /// Search the cache for the channel. If it can't be found, the channel is
     /// requested over REST.
+    #[deprecated(since = "0.5.8", note = "Use the `to_channel`-method instead.")]
     pub fn get(&self) -> Result<Channel> {
         #[cfg(feature = "cache")]
```
zeyla: We can remove the implementations of the implemented methods and have them call the new methods to remove code duplication.

zeyla:
Thanks

PR #368, opened 2018-08-12
Old message on update

Mishio595:
This PR is a breaking change to the way EventHandler::message_update works with cache feature enabled.

The method is changed to send the new message (as it did before) as well as the old message, optionally. I've also included a convenience method on Cache to bring the message cache up to par with the other cache contents.

Without cache enabled, message_update behavior is untouched.


src/client/dispatch.rs
```diff
@@ -497,13 +497,25 @@ fn handle_event<H: EventHandler + Send + Sync + 'static>(
             });
         },
         DispatchEvent::Model(Event::MessageUpdate(mut event)) => {
-            update!(event);
+            let _before = update!(event);
+
+            let _after: Option<Message> = feature_cache! {{
```
zeyla: Although most of the rest of this file uses this technique, we should be able to conditionally compile this binding, see line 264 for an example.

src/cache/mod.rs
```diff
@@ -576,6 +576,44 @@ impl Cache {
...
+    /// message's given Ids.
+    ///
+    /// **Note**: This will clone the entire message.
+    ///
+    /// # Examples
+    ///
+    /// Retrieving the message object from a channel, in a
+    /// [`EventHandler::message`] context:
+    ///
+    /// ```rust,ignore
```
zeyla: We don't need to make this ignored, we can instead do:

```rust
use serenity::CACHE;

# use serenity::model::id::MessageId;
# let message = MessageId(0).get().unwrap();

let cache = CACHE.read();
let fetched_message = ...
```

This way we can at least make it `no_run` and hide the code getting the "message" being used here.

Mishio595: Changed with latest commit.

zeyla:
Thanks

PR #369, opened 2018-08-15
Make is_command_visible work with DMs.

Roughsketch:
It seems that is_command_visible only handled cases where the message came from a guild. Otherwise it would default to the not found case. This PR separates the guild-specific role check so it works with guilds and DMs.


Lakelezz:
Cool, thanks : )

PR #378, opened 2018-09-02
Update some dependencies

Kroisse:
- `parking_lot` 0.5 -> 0.6
 - `multipart` 0.13 -> 0.15

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ version = "^1.0"
 default-features = false
 features = ["client", "hyper", "safemem"]
 optional = true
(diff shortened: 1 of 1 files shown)


Erk-:
We would have to bump our supported version to 1.26 if we wants to update these dependencies.

zeyla:
On top of the increased rustc 1.26 requirement, we expose parking_lot's types in our API, which contains breaking changes from parking_lot 0.5 to 0.6, meaning this won't be possible for 0.5.x (master). This change would be possible if you retarget for the 0.6.x branch.

PR #379, opened 2018-09-06
Fix to message_update dispatch

Mishio595:
Change signature of EventHandler::message_update and the dispatch call to account for situations where the new message is not cached

Previously, with small message cache sizes, the event would not dispatch because it left the cache prior to the read lock being opened. This commit changes the new message to an `Option<Message>` and always send the `MessageUpdateEvent` as an additional arg.


zeyla:
Rebased as bcd4474, thanks.

PR #385, opened 2018-09-10
Add single group help

Lakelezz:
This pull requests adds the ability to list a group in isolation by either writing its group-name or using any of its prefixes.

_Example_:
Group named `Ferris`, prefixes are `fer` and `f`.

Writing `help fer`, `help f`, and `help ferris` will all end in displaying the group `Ferris` only.

On another note, using `desc(&str)` on the group-builder will enable you to set a text headlining a single group.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -552,7 +552,7 @@ pub fn create_customised_help_data<'a, H: BuildHasher>(
                         );
 
                         return CustomisedHelpData::GroupedCommands {
-                            help_description: group.description.clone().unwrap_or_else(|| "".to_string()),
+                            help_description: group.description.clone().unwrap_or_default(),
                             groups: vec![single_group],
                         };
                     }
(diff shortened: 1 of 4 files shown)


src/framework/standard/help_commands.rs
```diff
@@ -517,6 +535,29 @@ pub fn create_customised_help_data<'a, H: BuildHasher>(
         return match fetch_single_command(&groups, &name, &help_options, &msg) {
             Ok(single_command) => single_command,
             Err(suggestions) => {
+                let searched_named_lowercase = name.to_lowercase();
+
+                for (key, group) in groups {
+
+                    if key.to_lowercase() == searched_named_lowercase
+                        || group.prefixes.as_ref()
+                            .map(|v| v.iter().any(|prefix|
```
acdenisSK: To shorten this a bit, use `map_or`:
```rust
group.prefixes.as_ref()
    .map_or(false, |v| v.iter().any(|prefix| *prefix == searched_named_lowercase))
```

PR #392, opened 2018-09-18
Make sure `delimiter` clears current and default delimiters

Lakelezz:
Currently, when you set `delimiter` it will just extend the default delimiter-vector holding space.

*Example*:
Setting: `.delimiter(",")`
`Hello I like Ferris` will yield following arguments: `"Hello", "I", "like", "Ferris"`.

This pull request clears the vector first and then adds the delimiter resulting in one delimiter only.
Which is the current behaviour on `delimiters`.

It might be up to discussion whether this is a breaking change as there might be bots that work perfectly with this behaviour and suddenly will break?
On the other hand... it's not really clear behaviour.


zeyla:
We decided this was a breaking change, so this just needs to be retargeted to the `v0.6.x` branch.

zeyla:
Rebased as 34b27af, thanks.

PR #393, opened 2018-09-18
Fix compiling with 'standard_framework', but without 'cache'

Bond-009:


zeyla:
Spurious travis errors, restarting the jobs just to make sure before merging.

PR #397, opened 2018-09-22
Remove mildly annoying, inconsistent braces

Proximyst:
Completely unnecessary but also completely necessary to remove because this hurts. A lot.


Lakelezz:
Thank you : )

PR #403, opened 2018-09-29
Underline command name and "Commands" in plain help

hyarsan:


hyarsan:
Before:
before
After:
after

Lakelezz:
@Roughsketch and @Mishio595, since I know you both use plain-help, what do you think of this change?

I am torn apart on accepting this for current-branch and v0.6.x, too.

Mishio595:
I think it helps create a contrast between the command name and the help information. Looks good.

acdenisSK:
Cherry-picked.

Issue #362, opened 2018-08-07
Error is not Send + Sync


FliegendeWurst:
The Rust API Guidelines (#171) suggest that it should be. It's also currently impossible to use failure with serenity's error type because of this. [Relevant commit](https://github.com/serenity-rs/serenity/commit/be43836)

Lakelezz:
Thanks for opening this issue! As of #532, this issue should be resolved.

PR #406, opened 2018-10-03
Add Mutex in HttpError to make Error sync again

djmcgill:
Closes https://github.com/serenity-rs/serenity/issues/362 and provides a significant usability boost when using `failure`.

This is an API breaking change so will need to wait for a major version bump. If you'd rather I place it on the `0.6.x` branch I can do that instead.

### DIFF
--- a/src/http/error.rs
+++ b/src/http/error.rs
@@ -5,13 +5,15 @@ use std::{
         Display,
         Formatter,
         Result as FmtResult
-    }
+    },
+    borrow::Borrow
 };
+use parking_lot::Mutex;
 
 #[derive(Debug)]
 pub enum Error {
     /// When a non-successful status code was received for a request.
-    UnsuccessfulRequest(Response),
+    UnsuccessfulRequest(Mutex<Response>),
     /// When the decoding of a ratelimit header could not be properly decoded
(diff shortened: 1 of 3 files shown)


troiganto:
I just found this PR because I stumbled over #362 myself. I looked a bit into this and ultimately, the error is not `Sync` because the `hyper::client::Response` contained in `HttpError` isn't. This is because reading the payload of the response is not thread-safe.

In my case, if such an error occurs, I'd like to print the reponse with status code and the response body. The solution in this PR, wrapping the entire struct in a mutex, would make this difficult because I'd be able to read the response only once, and it's gone afterwards.

What I'm wondering is: Would it make more sense to read the entire response into one or more `String`s and include that in the error type? The type would then roughly look like this:

```rust
#[derive(Debug)]
pub enum Error {
    // Maybe include Headers and HttpVersion, too.
    UnsuccessfulRequest{ status: StatusCode, url: Url, body: String, },
    RateLimitI64,
    RateLimitUtf8,
}
```

I could hack together a PR for this, but I wanted to check first if there's any interest for that.

Lakelezz:
Thanks for your pull request, but #532 ended up being a better solution to this problem!

PR #409, opened 2018-10-04
Implement watching presence status

### LINKED ISSUE #408: Support Watching status in serenity
You can't set the playing status to Watching for some reason. So like ``ctx.set_game(Game::watching(&name));``

Flat:
Closes #408

### DIFF
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -174,6 +174,36 @@ impl Activity {
             url: None,
         }
     }
+
+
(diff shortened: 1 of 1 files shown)


zeyla:
Since this isn't documented in the official API docs (and the issue about it being removed says it's unstable: <https://github.com/discordapp/discord-api-docs/issues/463>, we've decided to close this for now.

If the type ever becomes stable then we will cherry pick this commit in.

Thanks for the PR.

Issue #329, opened 2018-06-09
Make CreateMessage (and others) use &mut self instead of mut self


PvdBerg1998:
Followup issue on https://github.com/serenity-rs/serenity/issues/313 .

PR #410, opened 2018-10-08
Improve API of CreateGroup

chaosteil:
This improves the public API of `framework::standard::CreateGroup` to prevent stuttering when writing commands. As per #329


Erk-:
You should retarget this to the 0.6 branch, as it is a breaking change.

chaosteil:
Sure, sorry for the commit mess here.

PR #411, opened 2018-10-08
Updates rand, sodiumoxide and matches crates

Erk-:
Had to add std to sodiumoxide as no_std uses nightly features.


Lakelezz:
Thanks for your pull request : )

However the development of `v0.6.x` is not done yet and therefore updating dependencies prematurely will lead to outdated dependencies again.
We will make sure that `v0.6.x` ships with updated dependencies, thanks for your concerns! : )

PR #414, opened 2018-10-09
Add Function to neutralise Mentions

Lakelezz:
This function utilises the cache to replace user, channel, and role mentions with a safe alternative.

E.g. a mention of the user `@ferris` (internally represented as `<@!{id}>`) will be represented as: `@ferris#1234`, which does not alert the user.
`@everyone` and `@here` are replaced as well.

On another note, it's also possible to display users by their actual name or display name.

These questions are left:
- [x] Should the function-name stay the same? It's based on [content_safe](https://docs.rs/serenity/0.5.9/serenity/model/channel/struct.Message.html#method.content_safe). I would prefer `neutralise_content` etc.
- [x] What if an identifier is too short/long? They are currently marked as `deleted-user`/`deleted-role` respectively. 

  *Solution*: The function will display just like Discord would display. E.g. invalid identifiers on `<@{id}>` are displayed as `@invalid-user` on mobile - so it will display as safe non-mentioning `@invalid-user`.
- [x] Is the API acceptable?

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -540,6 +540,7 @@ pub fn with_cache_mut<T, F>(mut f: F) -> T
 ///
 /// [`content_safe`]: fn.content_safe.html
 #[cfg(feature = "cache")]
+#[derive(Clone, Debug)]
 pub struct ContentSafeOptions {
     clean_role: bool,
     clean_user: bool,
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -13,15 +13,22 @@ pub use self::{
 
 use base64;
 use internal::prelude::*;
-use model::id::EmojiId;
-use model::misc::EmojiIdentifier;
+use model::{
+    misc::EmojiIdentifier,
+    id::{
+        EmojiId,
+        RoleId,
+        UserId,
+    },
+};
 use std::{
     collections::HashMap,
     ffi::OsStr,
     fs::File,
     hash::{BuildHasher, Hash},
     io::Read,
-    path::Path
+    path::Path,
+    str::FromStr,
 };
 
 #[cfg(feature = "cache")]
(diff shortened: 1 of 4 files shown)


src/utils/mod.rs
```diff
@@ -526,6 +533,114 @@ pub fn with_cache_mut<T, F>(mut f: F) -> T
...
+
+            if let Ok(role_id) = RoleId::from_str(&s[mention_start..mention_end]) {
+
+                s = if let Some(role) = role_id.to_role_cached() {
+                    s.replace(&format!("<@&{}>", &role_id.as_u64()), &role.name)
+                } else {
+                    s.replace(&format!("<@&{}>", &role_id.as_u64()), &"deleted-role")
+                };
+
+                progress = mention_end;
```
acdenisSK: Should be outside the if-else:
```rust
if let Ok(role_id) = RoleId::from_str(&s[mention_start..mention_end]) {
    ....
}

progress = mention_end;
```

src/utils/mod.rs
```diff
@@ -526,6 +533,114 @@ pub fn with_cache_mut<T, F>(mut f: F) -> T
...
+                        Discriminator::Show => format!("@{}#{:04}", &user.name, user.discriminator),
+                        Discriminator::Hide => format!("@{}", &user.name),
+                    };
+
+                    s.replace(&format!("{}{}>", mention_start, id.as_u64()), &replacement)
+                } else {
+                    s.replace(&format!("{}{}>", mention_start, id.as_u64()), &"invalid-user")
+                };
+
+                progress = mention_end;
```
acdenisSK: Ditto.

src/utils/mod.rs
```diff
@@ -526,6 +533,114 @@ pub fn with_cache_mut<T, F>(mut f: F) -> T
...
+    while let Some(mut mention_start) = s[progress..].find("<@") {
+        mention_start += progress;
+
+        if let Some(mut mention_end) = s[mention_start..].find(">") {
+            mention_end += mention_start;
+            mention_start += "<@".len();
+            let mut has_exclamation = false;
+
+            if s[mention_start..].as_bytes().get(0)
+                .map_or_else(|| false, |c| *c == b'!') {
```
acdenisSK: Should be `.map_or`; construction of boolean values is very cheap.

src/utils/mod.rs
```diff
@@ -526,6 +537,327 @@ pub fn with_cache_mut<T, F>(mut f: F) -> T
...
+/// Sanitise an `@everyone` mention.
+///
+/// ```rust,ignore
+/// use serenity::utils::{
+///     content_safe,
+///     Discriminator,
+/// };
+///
+/// let everyone_mention = "@everyone";
+/// let defused_mention = content_safe(&everyone_mention, Discriminator::Hide);
```
acdenisSK: Should use the new `ContentSafeOptions` struct.

PR #415, opened 2018-10-11
Add configuration for the write lock the cache is using to update.

Erk-:


src/cache/settings.rs
```diff
@@ -70,16 +70,32 @@ impl Settings {
     ///
     /// Refer to [`cache_lock_time`] for more information.
     ///
+    /// **Note**: Should be set before the client gets started, as it can not be
```
Lakelezz: Add newline after `**Note**`.

src/cache/settings.rs
```diff
@@ -10,15 +12,35 @@
...
 /// ```
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct Settings {
     /// The maximum number of messages to store in a channel's message cache.
     ///
     /// Defaults to 0.
     pub max_messages: usize,
+
+    /// The duration that the serenity will and require a write lock on the
```
Lakelezz: Change to `The amount of milliseconds cache updates will try to acquire write-locks.`.

src/cache/settings.rs
```diff
@@ -10,15 +12,35 @@
...
+    /// cache for.
+    ///
+    /// Defaults to 10 milliseconds.
+    ///
+    /// **Note**:
+    /// If set to `None`, cache updates will request write-lock until available,
```
Lakelezz: Let's use `acquire` over `request` here too.

PR #418, opened 2018-10-15
Fix NSFW Checks

Lakelezz:
Whether a channel or category is NSFW no longer depends on whether either contains the term `nsfw` in it thus this PR deprecates `utils::is_nsfw` and removes its usage.
This is not breaking as it only corrects behaviour.

### DIFF
--- a/src/model/channel/channel_category.rs
+++ b/src/model/channel/channel_category.rs
@@ -128,7 +128,7 @@ impl ChannelCategory {
     #[cfg(feature = "utils")]
     #[inline]
     pub fn is_nsfw(&self) -> bool {
-        self.kind == ChannelType::Text && (self.nsfw || serenity_utils::is_nsfw(&self.name))
+        self.kind == ChannelType::Text && self.nsfw
     }
 
     /// Returns the name of the category.
--- a/src/model/channel/channel_category.rs
+++ b/src/model/channel/channel_category.rs
@@ -125,7 +125,6 @@ impl ChannelCategory {
         })
     }
 
-    #[cfg(feature = "utils")]
     #[inline]
     pub fn is_nsfw(&self) -> bool {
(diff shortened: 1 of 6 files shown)


src/model/channel/channel_category.rs
```diff
@@ -128,7 +128,7 @@ impl ChannelCategory {
     #[cfg(feature = "utils")]
```
acdenisSK: This `cfg` shall now be removed.

src/model/channel/guild_channel.rs
```diff
@@ -386,18 +386,15 @@ impl GuildChannel {
...
     ///
-    /// Refer to [`utils::is_nsfw`] for more details.
-    ///
     /// Only [text channels][`ChannelType::Text`] are taken into consideration
     /// as being NSFW. [voice channels][`ChannelType::Voice`] are never NSFW.
     ///
     /// [`ChannelType::Text`]: enum.ChannelType.html#variant.Text
     /// [`ChannelType::Voice`]: enum.ChannelType.html#variant.Voice
-    /// [`utils::is_nsfw`]: ../../utils/fn.is_nsfw.html
     #[cfg(feature = "utils")]
```
acdenisSK: Ditto.

src/model/channel/mod.rs
```diff
@@ -334,10 +334,6 @@ impl Channel {
     }
 
     /// Determines if the channel is NSFW.
-    ///
-    /// Refer to [`utils::is_nsfw`] for more details.
-    ///
-    /// [`utils::is_nsfw`]: ../../utils/fn.is_nsfw.html
     #[cfg(all(feature = "model", feature = "utils"))]
```
acdenisSK: Needs to be removed here too.
```suggestion
    #[cfg(feature = "model")]
```

PR #423, opened 2018-10-30
Fix of serenity always writing out the deadlock warning even though it got the write lock.

Erk-:


Erk-:
The reason this happens is because many of the implementations of `CacheUpdate` returned `None`
for example:
```rust
#[cfg(feature = "cache")]
impl CacheUpdate for PresencesReplaceEvent {
    type Output = ();

    fn update(&mut self, cache: &mut Cache) -> Option<()> {
        cache.presences.extend({
            let mut p: HashMap<UserId, Presence> = HashMap::default();

            for presence in &self.presences {
                p.insert(presence.user_id, presence.clone());
            }

            p
        });

        None
    }
}
```
so the `.or_else` would fire even if it was a success.

PR #424, opened 2018-10-30
Impl EventHandler for Vec<Box<EventHandler>>

fenhl:
This PR adds an implementation of `EventHandler` for `Vec<Box<EventHandler>>`, which allows splitting up handler code into multiple types implementing the trait. For example, I have multiple bots that serialize voice state to disk, this would allow me to reuse the same code for these and still have specific event handler code for each bot.

The implementation is generated using a macro, so that new trait methods of `EventHandler` are automatically implemented correctly when added.

### DIFF
--- a/examples/12_multiple_event_handlers/Cargo.toml
+++ b/examples/12_multiple_event_handlers/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "12_multiple_event_handlers"
+version = "0.1.0"
+authors = ["my name <my@email.address>"]
+
+[dependencies]
+serenity = { path = "../../" }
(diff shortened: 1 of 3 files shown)


acdenisSK:
Could you add an example in the [examples](https://github.com/serenity-rs/serenity/tree/current/examples) folder? Would make the intention of this `impl` clearer.

acdenisSK:
The implementation is made possible by `.clone()`-ing the arguments to each subsequent event handler.
However, this is bad as it can cause memory/performance inflation for events that are dispatched by discord rapidly (e.g `PRESENCE_UPDATE`). So, until the cloning can be bypassed or a better solution is discovered, I'm closing this.

PR #426, opened 2018-11-06
Update packages to allow building with OpenSSL 1.1.1

### LINKED ISSUE #394: Fails to build on systems with OpenSSL 1.1.1
See https://github.com/sfackler/rust-openssl/issues/994.

Serenity pulls in native-tls 0.1.5, which depends on an old version of openssl that doesn't build on systems with OpenSSL 1.1.1.

ColonelThirtyTwo:
native-tls v0.1.5 does not detect OpenSSL 1.1.1 when building, causing
compile errors on systems that have it installed (such as Arch).

This patch swaps out the outdated evzht9h3nznqzwl crate with its upstream
websocket crate.

Fixes #394 

`cargo test` passes, and my light testing of my bot confirms that both chat and voice work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ version = "~0.10"
 
 [dependencies.hyper-native-tls]
 optional = true
-version = "0.2.4"
+version = "0.3.0"
 
 [dependencies.lazy_static]
 optional = true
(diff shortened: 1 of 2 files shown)


Lakelezz:
Thanks for your pull request.

Nonetheless, as pointed out in #394, the issue in `websocket-rs` https://github.com/websockets-rs/rust-websocket/issues/137 is critical.

Therefore an update of the respective dependency would require our related fix to be integrated into `websocket-rs`. Although an alternative would be updating `evzht9h3nznqzwl` itself.

PR #432, opened 2018-11-11
Add Nick Methods for `Message` and `UserId`

Lakelezz:
This pullrequest implements methods to retrieve a user's nick on both the `Message`- and the `UserId`-struct.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -551,8 +551,8 @@ impl Message {
     /// **Note**:
     /// If message was sent in a private channel, then the function will return
     /// `None`.
-    pub fn author_nick(self) -> Option<String> {
-        self.guild_id.and_then(|guild_id| self.author.nick_in(guild_id))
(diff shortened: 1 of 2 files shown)


src/model/channel/message.rs
```diff
@@ -546,6 +546,15 @@ impl Message {
         http::unpin_message(self.channel_id.0, self.id.0)
     }
 
+    /// Tries to return author's nickname in the current channel's guild.
+    ///
+    /// **Note**:
+    /// If message was sent in a private channel, then the function will return
+    /// `None`.
+    pub fn author_nick(self) -> Option<String> {
```
acdenisSK: This shouldn't consume self.
Change it to borrow the message:
```rust
pub fn author_nick(&self) -> Option<String> {
    self.guild_id.as_ref().and_then(|guild_id| self.author.nick_in(*guild_id))
}
```

PR #440, opened 2018-11-19
Replace `hyper` with `reqwest`

Lakelezz:
This pull request replaces `hyper` with `request`. This also removes the necessity of other dependencies as `reqwest` bundles their own alternatives.

With `reqwest` we have a crate providing a high-level API around `hyper`. Therefore we indirectly update our `hyper` from `0.10.x` to `0.12.x` and can keep up to date by just adhering to `reqwest`.

`reqwest`s multipart is different than the actual `multipart`-crate.
E.g. while `multipart` accepts `AsRef<str>` on its [write_text](https://docs.rs/multipart/0.15.3/multipart/client/struct.Multipart.html#method.write_text), `reqwest` variation accepts `Into<Cow<'static, str>>`, forcing us to perform the clone or provide a `Into<Cow<'static, str>>`. In the end `multipart` clones from the passed `AsRef<str>`, too, so I assume there should not be much of a difference to before.

### DIFF
--- a/src/http/mod.rs
+++ b/src/http/mod.rs
@@ -77,13 +77,13 @@ pub enum LightMethod {
 }
 
 impl LightMethod {
-    pub fn hyper_method(&self) -> Method {
+    pub fn reqwest_method(&self) -> Method {
         match *self {
-            LightMethod::Delete => Method::Delete,
-            LightMethod::Get => Method::Get,
-            LightMethod::Patch => Method::Patch,
-            LightMethod::Post => Method::Post,
-            LightMethod::Put => Method::Put,
+            LightMethod::Delete => Method::DELETE,
+            LightMethod::Get => Method::GET,
+            LightMethod::Patch => Method::PATCH,
+            LightMethod::Post => Method::POST,
+            LightMethod::Put => Method::PUT,
         }
     }
 }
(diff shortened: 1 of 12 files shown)


src/http/mod.rs
```diff
@@ -51,17 +49,14 @@ use std::{
 };
 
 lazy_static! {
-    static ref CLIENT: HyperClient = {
-        let tc = NativeTlsClient::new().expect("Unable to make http client");
-        let connector = HttpsConnector::new(tc);
-
-        HyperClient::with_connector(connector)
+    static ref CLIENT: ReqwestClient = {
```
acdenisSK: Style nitpick: This is just a simple call to `ReqwestClient::new()` now, the braces the can be removed.
```suggestion
    static ref CLIENT: ReqwestClient = ReqwestClient::new();
```

src/http/request.rs
```diff
@@ -61,33 +63,35 @@ impl<'a> Request<'a> {
...
             route: ref route_info,
         } = *self;
+
         let (method, _, path) = route_info.deconstruct();
 
         let mut builder = CLIENT.request(
-            method.hyper_method(),
-            &path.into_owned(),
+            method.reqwest_method(),
+            Url::parse(&path.to_string())?,
```
acdenisSK: It should be possible to just do `&path` here, as `Cow` implements `Deref<Target=str>` for `Cow<str>`.
```suggestion
            Url::parse(&path)?,
```

PR #448, opened 2018-12-16
Remove global Cache

Lakelezz:
This pullrequests removes the global cache and restructures Serenity to locate the cache inside the client and then pass it around.

Updating the update-cache-timeout can be set via `new_with_cache_update_timeout`-constructor now.

Functions that require the cache will expect `&Arc<RwLock<Cache>>` as first argument now - it can be borrowed from the `Context` via `context.cache`.
Functions that need to send a request via HTTP in addition to use potentially use the cache (if the `cache`-feature is set) will expect an entire `&Context`.

On another note, examples, tests, and doctests have been adjusted to compile with changes.


src/client/dispatch.rs
```diff
@@ -69,7 +71,7 @@ pub(crate) enum DispatchEvent {
     Model(Event),
 }
 
-#[cfg(feature = "framework")]
+#[cfg(all(feature = "framework"), )]
```
acdenisSK:
```suggestion
 #[cfg(feature = "framework")]
```

PR #457, opened 2018-12-28
Rustls support

Erk-:
~~note: This does not touch the websocket
for voice, the reason for this is the pr #444~~

~~This is currently blocked by #456~~

- [x] Fix features so rustls is the default, and a `native_tls` feature to switch to that.

### DIFF
--- a/src/gateway/shard.rs
+++ b/src/gateway/shard.rs
@@ -23,9 +23,14 @@ use super::{
 use tungstenite::{
     self,
     error::Error as TungsteniteError,
-    handshake::client::Request,
 };
 use url::Url;
+
+#[cfg(feature = "rustls_support")]
+use std::net::TcpStream;
+
+#[cfg(any(not(feature = "rustls_support"), feature = "native_tls"))]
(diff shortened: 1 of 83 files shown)


src/gateway/shard.rs
```diff
@@ -829,6 +834,33 @@ impl Shard {
...
+#[cfg(feature = "rustls_support")]
+fn connect(base_url: &str) -> Result<WsClient> {
+    let mut config = rustls::ClientConfig::new();
+    config.root_store.add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);
+
+    let url = build_gateway_url(base_url)?;
+
+    let base_host = if let Some(h) = url.host_str() {
+        let (dot, _) = h.rmatch_indices('.').skip(1).next().unwrap_or((0, ""));
+        let (_, base) = h.split_at(dot + 1); //We do not want the leading '.'
```
acdenisSK: Nitpick, space after the double slashes:
```suggestion
        let (_, base) = h.split_at(dot + 1); // We do not want the leading '.'
```

PR #458, opened 2018-12-30
Rewrite the framework to attributes (WIP)

acdenisSK:
It is working, but it's still far from done.

- [x] Documentation
- [x] Further testing
- [x] Permissions


src/client/mod.rs
```diff
@@ -933,29 +933,11 @@ impl Client {
         #[cfg(feature = "voice")]
         self.voice_manager.lock().set_shard_count(shard_data[2]);
 
-        // This is kind of gross, but oh well.
-        //
-        // Both the framework and voice bridge need the user's ID, so we'll only
-        // retrieve it over REST if at least one of those are enabled.
-        #[cfg(any(all(feature = "standard_framework", feature = "framework"),
-                  feature = "voice"))]
+        #[cfg(feature = "voice")]
```
arqunis, line 960: I searched through the codebase if we still used `StandardFramework::id`, but other than its initialization in the `Framework` trait, there were none.

PR #461, opened 2019-01-07
Support OpenSSL 1.1.1

Laaas:
This PR updates to version 0.10 of the OpenSSL crate to support version 1.1.1 of OpenSSL.
To do this, I had to update some other crates across breaking changes, however serenity still compiles and all tests pass, but I can not assert that everything works alright.
In addition, I do not know if any of the changed crates are public dependencies, because if so, this is a breaking change too.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ version = "~0.10"
 
 [dependencies.hyper-native-tls]
 optional = true
-version = "0.2.4"
+version = "0.3"
 
 [dependencies.lazy_static]
 optional = true
@@ -54,7 +54,7 @@ version = "0.13"
 
 [dependencies.native-tls]
 optional = true
-version = "0.1"
+version = "0.2"
 
 [dependencies.opus]
 optional = true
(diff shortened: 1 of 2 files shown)


Erk-:
We have used our own branch because of a bug happening with the websocket crate. For more information about that read the #394 issue. This is already fixed on the next major version where it will also be possible to use rustls for tls.

Laaas:
Sounds great! Where can I find the next major version?

acdenisSK:
The next major version is [v.0.6.x](https://github.com/serenity-rs/serenity/tree/v0.6.x), albeit the [rustls support](https://github.com/serenity-rs/serenity/pull/457) hasn't been merged yet, but it will.

Erk-:
If you use v.0.6.x I would recommend setting a commit you use as there is still major changes there are waiting to be merged. For example:
```toml
[dependencies.serenity]
git = "https://github.com/serenity-rs/serenity"
rev = "3c166e3"
```

Laaas:
Thanks! I will close this PR now since it's obviously unneeded.

PR #465, opened 2019-01-18
Add a position method

Erk-:
This method can be used to get the position of a channel in the channel enum.

### DIFF
--- a/src/model/channel/mod.rs
+++ b/src/model/channel/mod.rs
@@ -425,6 +425,21 @@ impl Channel {
         }
     }
 
+    /// Retrieves the position of the inner [`GuildChannel`] or
+    /// [`ChannelCategory`].
+    ///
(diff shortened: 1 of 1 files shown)


src/model/channel/mod.rs
```diff
@@ -425,6 +425,21 @@ impl Channel {
...
 
+    /// Retrieves the position of the inner [`GuildChannel`] or
+    /// [`ChannelCategory`].
+    ///
+    /// If other channel types are used it will return None.
+    ///
+    /// [`GuildChannel`]: struct.GuildChannel.html
+    /// [`CatagoryChannel`]: struct.ChannelCategory.html
+    pub fn position(&self) -> Option<i64> {
+        match self {
```
acdenisSK: To let the `1.25` travis job to build, you need to deference `self` here:
```suggestion
        match *self {
```

PR #467, opened 2019-01-19
(branch: futures) Use `Arc` instead of `Rc` in the cache

Proximyst:
Seeing as tokio now is threadpooled by default and futures have to be Send, `Rc` is not suitable for the cache anymore.


zeyla:
Thanks

PR #471, opened 2019-01-27
Fix connection error being thrown on leaving voice

JellyWX:
One line rip

### DIFF
--- a/src/voice/handler.rs
+++ b/src/voice/handler.rs
@@ -212,7 +212,7 @@ impl Handler {
         // Only send an update if we were in a voice channel.
         if self.channel_id.is_some() {
             self.channel_id = None;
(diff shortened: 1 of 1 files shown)


src/voice/handler.rs
```diff
@@ -212,6 +212,7 @@ impl Handler {
         // Only send an update if we were in a voice channel.
         if self.channel_id.is_some() {
             self.channel_id = None;
+            self.send(VoiceState::Disconnect);
```
FelixMcFelix: Right idea, because this will stop the voice thread from trying to send over a broken socket. But this doesn't actually compile.

FelixMcFelix:
```suggestion
            self.send(VoiceStatus::Disconnect);
```

PR #477, opened 2019-02-12
Remove boxed futures from the gateway and http modules

NieDzejkob:


meithecatte:
I did a terrible job with this kitchen sink of a PR, I'll redo it if I'll get a lapse of passion for futures again.

PR #479, opened 2019-02-14
Add new Check System

Lakelezz:
The old check-system lacked the ability to explain why a check would have failed as it boiled down to `true` and `false`.
Hence offering `CheckResult` enabling bot-devs to pass reasons of failure and then log or return it to the user.
Additionally, `Check`s can be named now and the bot-dev can mark them to be checked inside the help-system and whether they shall be displayed in the related command's help-page.

### DIFF
--- a/src/framework/standard/check.rs
+++ b/src/framework/standard/check.rs
@@ -1,0 +1,221 @@
+use std::fmt::Debug;
+use std::fmt;
+use model::channel::Message;
+use client::Context;
+use framework::standard::{Args, CommandOptions};
+
+pub type CheckFunction = dyn Fn(&mut Context, &Message, &mut Args, &CommandOptions) -> CheckResult
+    + Send
+    + Sync
+    + 'static;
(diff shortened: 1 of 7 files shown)


src/framework/standard/check.rs
```diff
@@ -0,0 +1,221 @@
...
+            function: Box::new(function),
+            check_in_help,
+            display_in_help,
+        }
+    }
+}
+
+impl Debug for Check {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_tuple("Check")
```
acdenisSK: This redundantly allocates strings on every debug call. The code also looks ugly.
```rust
f.debug_struct("Check")
   .field("name", &self.name)
   .field("function", &"<fn>")
   .field("check_in_help", &self.check_in_help)
   .field("display_in_help", &self.display_in_help)
   .finish()
```

PR #484, opened 2019-02-17
client::Client doctest fixes

MOZGIII:
This PR makes some of the doctests at `client::Client` `no_run` instead of `rust,ignore`.
Not all of them, but the ones I could convert.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -73,23 +73,26 @@ use self::bridge::voice::ClientVoiceManager;
 /// Creating a Client instance and adding a handler on every message
 /// receive, acting as a "ping-pong" bot is simple:
 ///
-/// ```rust,ignore
+/// ```no_run
+/// # fn main() -> Result<(), Box<std::error::Error>> {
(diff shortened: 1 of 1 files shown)


MOZGIII:
Why does build fail on this? I think it's not relevant to the changes I made.

acdenisSK:
It is not. It's just the 1.25 Rust build failing because of a recent update to one of the [sub-dependencies](https://github.com/alicemaz/rust-base64) of our dependencies which changed to use [`slice::rotate_left`] and [`slice::rotate_right`] methods; apis that were stabilized in 1.26 (>_>).

You shouldn't mind that.

[`slice::rotate_left`]: https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.rotate_left
[`slice::rotate_right`]: https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.rotate_right

MOZGIII:
How about merging this then? I'd like to get this over with, conflicts may be introduced if we pointlessly hold on with merging due to "fake" build fails.

PR #488, opened 2019-02-19
Impl EventHandler for Vec<Box<EventHandler>>, take 2

fenhl:
This is a redo of #424, this time without the clone calls that @acdenisSK pointed out as a potential performance issue. To accomplish that, I have changed all arguments of `EventHandler` methods to types that implement `Copy`, which is a breaking change, so this PR is for 0.6.x.

The code compiles without the cache feature but hasn't been tested since the v0.6.x branch doesn't compile with `#[cfg(test)]`.


arqunis:
I have concluded that I will not merge this pull request, for three reasons I'm about to state, but before that, some history time:

Long time ago, when zeyla was actively maintaining the library, events used to be handled by assigning a "listener" closure to a respective event. The surface api looked all right, but internally it was a mess. There were boxes everywhere, huge amount of these event methods stuck on the client, filling up its generated documentation by quite a lot.

Then, when zey passed maintainer status to me, I immediately went to resolve this convoluted clutter.
One of my main ideas was a trait which opened the possibility of registering multiple listeners simultaneously, that you only needed to implement **once**. And of course, you can see where that got us.

Now for the reasoning:
- The intended purpose as I had outlined.
- References consuming every event signature, which for multiple event handlers isn't bad. However, for single event handlers, it's a needless limitation; what one might have simply moved for memory efficiency, now has to clone things, when the thing itself gets dropped *right after the call to the event*.
- Because of how I envisioned the new event handling, Serenity just wasn't built for multiple event handlers. That's why it's tricky to glue them to the library, it'd changing something fundamental. Consequently (and unfortunately), I don't have a solution to this either.

So dearest apologies, but mayhap we take a bite at this again in the future, if we ever come to a better answer to this.

PR #489, opened 2019-02-19
Fix building without `cache`-feature

Lakelezz:
Disabling the `cache`-feature caused serenity to fail building.

This pull request makes sure that building with default features excluding `http` and/or `cache` will still work.


fenhl:
Should the `CONTRIBUTING.md` and/or the `.travis.yml` be adjusted to make sure this doesn't happen in the future?

Lakelezz:
Hardly doable as it seems as only one feature-set can be built.
If it ever becomes possible we also need to examine how we will handle the already long build-times on Travis.

PR #490, opened 2019-02-19
Bumped sodiumoxide version to 0.2

MOZGIII:


Erk-:
We had this pr before #454, but we reverted it in [d6bd7ac](https://github.com/acdenisSK/serenity/commit/d6bd7ace1c46b35f98958beba4af1ab447c3e636) because it broke compilation for rustc 1.25, which is the lowest compiler version we support.

MOZGIII:
:(
Ok

PR #505, opened 2019-03-04
Make compilation to fail on the master branch

Erk-:


rukai:
I just spent a while extremely confused by the branch situation until I found the contributing.md file.
So I think it would be good to expand this message to something like:
```
This branch is no longer updated, please use the current branch or the 0.Z.x branch.
Please check this document for more information: https://github.com/serenity-rs/serenity/blob/current/CONTRIBUTING.md#pull-requests
```

PR #508, opened 2019-03-14
[Voice] AudioReceiver extension, event deserialisation fixes

FelixMcFelix:
* Ports the fixes introduced in #471 and #472 to be idiomatic Rust 2018 code.
* Adds handling for the (undocumented) `VoiceClientConnect` event which the voice thread receives over websocket.
* Adds ability to handle client (dis)connects in the `AudioReceiver`.
* Adds an extra field to `AudioReceiver::voice_packet`, `compressed_size`. This should allow better instrumentation of per-user stream bandwidth if needed without, e.g., re-encoding the raw audio data.
* Makes impls of functions on `AudioReceiver` optional.
* Fixes deserialisation of `VoiceHeartbeatAck`.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -151,6 +151,8 @@ pub enum VoiceOpCode {
     Hello = 8,
     /// Sent by the server if a session coulkd successfully be resumed.
     Resumed = 9,
+    /// Message indicating that another user has connected to the voice channel.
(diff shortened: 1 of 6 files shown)


FelixMcFelix:
Closing to further subdivide...

PR #515, opened 2019-03-19
switch to hashbrown

DarkKirb:
@Lakelezz has suggested on the discord to switch to a faster implementation of HashMap and HashSet (hashbrown). This pull request changes all uses of `std::collection::HashMap` and `std::collection::HashSet` with the corresponding `hashbrown` implementations.

This is a breaking change. Users of the library need to change their code to use hashbrown as well. Additionally it seems like that hashbrown might get merged into rust's libstd soon.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,6 +91,10 @@ version = "0.18.1"
 optional = true
 version = "0.15.0"
 
+[dependencies.hashbrown]
+version = "^0.1"
+features = ["serde"]
+
 [dev-dependencies.matches]
 version = "0.1.6"
 
--- a/src/cache/cache_update.rs
+++ b/src/cache/cache_update.rs
@@ -18,10 +18,8 @@ use super::Cache;
 ///     },
 ///     prelude::RwLock,
 /// };
-/// use std::{
-///     collections::hash_map::Entry,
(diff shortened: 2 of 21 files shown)


xacrimon:
@DarkKirb Hashbrown will go into `libstd` very soon. Consider closing this pull request.

acdenisSK:
As @xacrimon had mentioned that the hashmap implementation in std is being switched to hashbrown's, we will not merge this. @Lakelezz said that we can wait anyway.

However, since you already put effort into this, we could use this occasion to benchmark things for the sake of it: Is the cache faster? Does it use less memory? etc.
If you'd be willing to do it, could you, @DarkKirb, add `#[bench]` tests and post the before and after differences here as well? 
I'm asking because hashbrown claims to be 8x faster than the standard hashmap, so I'm interested if that's true for our case.

xacrimon:
@acdenisSK I've already done some beckhamrking. Performance is about 4x better in singlethreaded scenarios. Space utilization is about 10% better when benchmarking with a tracking allocator.

acdenisSK:
@xacrimon Ah, that's nice. Kirb won't have to bother.
Closing this pull request then.

PR #516, opened 2019-03-20
Update rusttls, sodiumoxide, webpki and webpki-roots

DarkKirb:
Currently building 0.6 with voice enabled pulls in native-tls, even on builds without the native_tls feature.
This is problematic on systems without a compatible version of openssl (like arch linux) where the crate simply cannot be built.
Updating these dependencies to a newer version results in libsodium-sys using rusttls instead of native_tls as a build-time dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,7 @@ version = "^0.15"
 default-features = false
 features = ["std"]
 optional = true
-version = "0.2.1"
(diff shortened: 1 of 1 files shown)


xacrimon:
This seems like a rather high priority issue. Consider merging this as soon as possible @acdenisSK

PR #517, opened 2019-03-21
Fix current_application_info

mattico:
API Reference: https://discordapp.com/developers/docs/topics/oauth2#get-current-application-information


Mishio595:
Why is description an Option? Per the docs, description is not nullable or optional.

mattico:
I vaguely remember encountering a null response for that, but we can stick to the docs.

Erk-:
This is also a breaking change so it should be retargeted to `v0.6.x`

mattico:
Done

Lakelezz:
Thank you : )

PR #519, opened 2019-03-23
Add support for raw events

DarkKirb:
Currently serenity does not provide any functionality to receive raw discord websocket events. This would be useful for distributed/"serverless" bots that want to feed discord events into a queue service.

This pull request adds support for another type of event handler, the RawEventHandler. It is possible to extend it to other raw event types if necessary.

As a side effect this pr makes it possible to create a Client instance without any Handler.

The code has a small performance impact on every event, however the events are only cloned if there is both a regular and a raw handler. In most cases you'll only want one of these, so this can be ignored. 

If you find a better way of implementing what this PR does, feel free to comment.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -139,6 +139,7 @@ native_tls = ["reqwest/default-tls", "tungstenite/tls"]
 standard_framework = ["framework", "uwl", "command_attr"]
 utils = ["base64"]
 voice = ["byteorder", "gateway", "audiopus", "rand", "sodiumoxide"]
+raw_event = []
 
 [package.metadata.docs.rs]
 all-features = true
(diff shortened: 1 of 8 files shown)


DarkKirb:
I reapplied the patches to the latest source tree

Lakelezz:
Why is the `gateway`-module not sufficient for this? 
Additionally, you mention a performance hit, please elaborate on what you mean for the record. There might be a confusion stem from the fact that you state every event will cause a performance impact but then cloning only appears if both handlers are used. Is the performance hit the cloning or is there something else?

DarkKirb:
As far as i can tell, using the gateway module requires you to handle and maintain (multiple) WS connections yourself, a feature which the client has natively.

The performance hit only exists if you are using both a normal and a raw handler and it results in each event being cloned. I am not sure how much that performance impact is or how to measure it, but it doesn't apply to 99% of users.

DarkKirb:
rebased my PR to the latest commit.

Lakelezz:
Thanks a lot for your contribution! : )

PR #532, opened 2019-04-14
Make the HttpError Send+Sync

### LINKED ISSUE #362: Error is not Send + Sync

Erk-:
This pr is a attempt to fix #362, and it supersedes #406, but compared to #406 it does not make use of a mutex for the response. It instead uses the solution by @troiganto which he proposed there.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,6 +97,10 @@ version = "^0.19"
 optional = true
 version = "^0.16"
 
+[dev-dependencies.http_crate]
+version = "~0.1"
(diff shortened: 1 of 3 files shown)


Erk-:
Ready for review.

Lakelezz (approved):

PR #535, opened 2019-04-15
Update `on_mention` to take a `UserId`

### LINKED ISSUE #530: `framework::standard::Configuration::on_mention` broken with removal of global Http in v0.6.x
`Configuration::on_mention` previously used the global Http instance to get the bot's current user ID, and currently (as of cc81e47d) constructs a new independent Http instance, but this new instance has no access to the token for authorisation.

The easiest way forward I can see is for `on_mention` to take a UserId instead of a `bool`, in a similar fashion to `owners`.

Celti:
With the removal of global `Http` objects, `on_mention` is no longer
authorised for the necessary routes to get the bot's `UserId`.

Directly pass it an `Option<UserId>` instead.

Fixes #530

### DIFF
--- a/src/framework/standard/configuration.rs
+++ b/src/framework/standard/configuration.rs
@@ -85,7 +85,7 @@ impl From<(bool, bool, bool)> for WithWhiteSpace {
 /// let mut client = Client::new(&token, Handler).unwrap();
 ///
 /// client.with_framework(StandardFramework::new()
-///     .configure(|c| c.on_mention(true).prefix("~")));
+///     .configure(|c| c.on_mention(Some(UserId(5))).prefix("~")));
 /// ```
 ///
 /// [`Client`]: ../../client/struct.Client.html
--- a/src/framework/standard/configuration.rs
+++ b/src/framework/standard/configuration.rs
@@ -273,14 +273,14 @@ impl Configuration {
         self
     }
 
-    /// Whether or not to respond to commands initiated with a mention. Note
-    /// that this can be used in conjunction with [`prefix`].
+    /// Whether or not to respond to commands initiated with the mentioned id.
+    /// Note that this can be used in conjunction with [`prefix`].
     ///
-    /// **Note**: Defaults to `false`.
+    /// **Note**: Defaults to ignore mentions.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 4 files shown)


arqunis (approved):

src/framework/standard/configuration.rs
```diff
@@ -273,14 +273,14 @@ impl Configuration {
         self
     }
 
-    /// Whether or not to respond to commands initiated with a mention. Note
-    /// that this can be used in conjunction with [`prefix`].
+    /// Whether or not to respond to commands initiated with the mentioned id.
```
Lakelezz, line 276:
```suggestion
    /// Whether or not to respond to commands initiated with `id_to_mention`.
```

src/framework/standard/configuration.rs
```diff
@@ -273,14 +273,14 @@ impl Configuration {
...
+    /// Note that this can be used in conjunction with [`prefix`].
```
Lakelezz, line 277: While we are at it, turn this into another `**Note**`-block just like the note about ignoring mentions by default.

src/framework/standard/configuration.rs
```diff
@@ -294,19 +294,8 @@ impl Configuration {
...
-
-        let http = Http::new(
-            reqwest::Client::builder().build().expect("Could not construct Reqwest-Client."),
-            "",
-        );
-
-        if let Ok(current_user) = http.get_current_user() {
-            self.on_mention = Some(current_user.id.to_string());
-        }
+    pub fn on_mention(&mut self, on_mention: Some(UserId)) -> &mut Self {
```
Lakelezz, line 297:
```suggestion
    pub fn on_mention(&mut self, id_to_mention: Some(UserId)) -> &mut Self {
```

Lakelezz (changes requested):
Thank you for your PR : )

I mentioned some small changes I want to see.
On another note, please change the lowercase word `id` to `ID` - to be consistent with most of Serenity's docs.

arqunis:
[Cherry-picked and squashed](https://github.com/serenity-rs/serenity/commit/d995fa0e08c67e1a7f217b427d4d9b4dcedaa45b).

PR #545, opened 2019-04-25
Remove 'static on `HelpOptions` for Help-Functions

Lakelezz:
As `with_embeds` and `plain` are only called from inside the actual `help`-function, they no longer need the `#[help]`-macro and thus no longer a static lifetime on `HelpOptions`. 

Additionally, I implemented `Clone` for `HelpOptions` and `HelpBehaviour`. In combination of mentioned above and `Clone`, bot-developers can now clone the received `HelpOptions` in the `help`-function, mutate it, and pass runtime-mutated `HelpOptions`, e.g. to provide randomness. 

Side-note: If you notice performance-issues due tons of help-commands being called every second, it might be beneficial to implement your own version of `with_embeds`/`plain` and simply provide runtime-mutated local variables instead of `HelpOptions`' fields.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -976,10 +976,26 @@ fn send_error_embed(
 /// # impl EventHandler for Handler {}
 /// # let mut client = Client::new("token", Handler).unwrap();
 /// #
-/// use serenity::framework::standard::{StandardFramework, help_commands::*};
+/// use std::{collections::HashSet, hash::BuildHasher};
+/// use serenity::{framework::standard::{Args, CommandGroup, CommandResult,
+///     StandardFramework, macros::help, HelpOptions,
+///     help_commands::*}, model::prelude::*,
+/// };
+///
+/// #[help]
+/// fn my_help(
+///     context: &mut Context,
+///     msg: &Message,
+///     args: Args,
+///     help_options: &'static HelpOptions,
+///     groups: &[&'static CommandGroup],
+///     owners: HashSet<UserId, impl BuildHasher>
+/// ) -> CommandResult {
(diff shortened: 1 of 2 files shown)


AregevDev:
Why not replace `&HelpOptions` with `&mut HelpOptions`, to allow modification inside the help function?

Lakelezz:
> Why not replace `&HelpOptions` with `&mut HelpOptions`, to allow modification inside the help function?

What do you mean by `help`-function? `with_embeds` and `plain` equivalents or the proc-macro one that dispatches to these?

AregevDev:
The user defined "help" function, annotated with `#[help]`

src/framework/standard/help_commands.rs
```diff
@@ -976,18 +976,33 @@ fn send_error_embed(
...
+///     StandardFramework, macros::help, HelpOptions,
+///     help_commands::*}, model::prelude::*,
+/// };
+///
+/// #[help]
+/// fn my_help(
+///     context: &mut Context,
+///     msg: &Message,
+///     args: Args,
+///     help_options: &'static HelpOptions,
```
acdenisSK:
```suggestion
///     help_options: &HelpOptions,
```

src/framework/standard/help_commands.rs
```diff
@@ -1129,18 +1144,33 @@ fn single_command_to_plain_string(help_options: &HelpOptions, command: &Command<
...
```
acdenisSK:
```suggestion
///     help_options: &HelpOptions,
```

arqunis:
@AregevDev It is not possible for us to give a mutable reference of the `HelpOptions` made by the `#[help]` macro, as `static`s by default are immutable.

PR #579, opened 2019-05-25
 Fix all Clippy-lints for Clippy `0.0.212`

Lakelezz:
This updates Serenity's code to adhere to lints triggered by Clippy version `0.0.212`.

Some lints got allowed that are not helpful or false-positives.

On another note, this removes the explicit *impl BuildHasher* on `owners`' `HashSet<UserId>` which was introduced to please Clippy in a prior commit.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -46,7 +46,6 @@ use std::{
     borrow::Borrow,
     collections::HashSet,
     fmt::Write,
-    hash::BuildHasher,
     ops::{Index, IndexMut},
 };
 #[cfg(all(feature = "cache", feature = "http"))]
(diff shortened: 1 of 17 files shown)


src/framework/standard/help_commands.rs
```diff
@@ -819,14 +822,14 @@ fn flatten_group_to_string(
         .join(&format!("\n{}", &repeated_indent_str));
 
 
-    if group.command_names.len() >= 1 {
+    if group.command_names.is_empty() {
```
nickelc: this looks wrong

Lakelezz: That's what I get for copy and pasting. Good catch!

PR #580, opened 2019-05-25
changed command identifier to match original function name

Raatty:
this changes the way the global varable is named for commands, currently it will take the $name from in 
#[command($name)] or if non existant it will use the name of the function, then if it starts with a number it will add an underscore to the front and will completely fail if it has an emoji in it.

this pr changes it to still use the name in the attribute for the command name in discord, but in code it remains as the name of the function, the only reason I can see someone putting a name in the #[command($name)] thing is because you want an invalid ident as the command name.

My fix makes emoji command names possible, removes the need for prepending the random underscore too

### DIFF
--- a/command_attr/src/lib.rs
+++ b/command_attr/src/lib.rs
@@ -252,12 +252,7 @@ pub fn command(attr: TokenStream, input: TokenStream) -> TokenStream {
 
     let name = _name.clone();
 
-    // If name starts with a number, prepend an underscore to make it a valid identifier.
-    let n = if _name.starts_with(|c: char| c.is_numeric()) {
-        format!("_{}", _name)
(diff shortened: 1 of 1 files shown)


arqunis:
(Replying to this [comment](https://github.com/serenity-rs/serenity/commit/05254c8376b6b198aff6734aa8c0b58560e3a756#commitcomment-33728125))
Sorry, I'm sure that must've come off as rude. 
I got too far ahead of myself in cleaning up dust and formalizing behaviour for the third release candidate of v0.6.x that I had totally forgotten to take your pull request into equation. 😟

But, since your change eventually did come into fruition, I think we can just close your pr now. 

And again, pardon for ignoring you.

PR #592, opened 2019-06-02
Add `GuildChannel::members`-method

Lakelezz:
Adds a new method to `GuildChannel` to retrieve connected members on voice-channels or members able to read a text/news-channel.

### DIFF
--- a/src/model/channel/guild_channel.rs
+++ b/src/model/channel/guild_channel.rs
@@ -721,40 +721,40 @@ impl GuildChannel {
     #[cfg(feature = "cache")]
     #[inline]
     pub fn members(&self, cache: impl AsRef<CacheRwLock>) -> Result<Vec<Member>> {
-           let cache = cache.as_ref();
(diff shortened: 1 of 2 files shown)


src/model/channel/guild_channel.rs
```diff
@@ -702,6 +702,60 @@ impl GuildChannel {
...
+           let cache = cache.as_ref();
+    let guild = cache.read().guild(self.guild_id).unwrap();
+
+    match self.kind {
+        ChannelType::Voice => Ok(guild
+            .read()
+            .voice_states
+            .values()
+            .filter_map(|v| {
+                v.channel_id.map_or_else(
```
arqunis: Unless it's expensive to compute the default value, using the `_else` variant is unnecessary here.
```suggestion
                v.channel_id.map_or(None, ...
```

arqunis (changes requested):

PR #616, opened 2019-06-16
Store inner HTTP-client in an Arc

ghost:
In the 'http' module's 'Http' struct, store the inner reqwest client in
an Arc. This doesn't have much use right now for serenity itself, but it
does allow users to pass in a reference to their own client when making
a client via 'http::Http::new', avoiding the need to have multiple
clients.

The primary use for this is to ease a slight rework of the
'http::ratelimiting' module.

Blocks on #614.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -28,7 +28,7 @@ use std::{
 };
 
 pub struct Http {
-    client: Client,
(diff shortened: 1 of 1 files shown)


ghost:
Rebased.

PR #624, opened 2019-06-21
Fix panic on deserializing partial guild with no Nitro boosters.

### LINKED ISSUE #623: http::get_guild panics on guilds without Nitro boosters
I have a guild without any Nitro boosters, and trying to http::get_guild it panics the JSON decoder. This is because the JSON object contains `"premium_subscription_count": null`, while the field in the Rust struct is defined as `pub premium_subscription_count: u64`.

The error itself is `Json(Error("invalid type: null, expected u64", line: 1, column: 239))`, and sure enough if I `curl` the endpoint myself, column 239 corresponds to `"premium_subscription_count": null`.

I'm working on a PR for this myself, but I thought I'd log the issue beforehand in case I forget/am not able to fix it.

EDIT: This is on Serenity version `0.6.0-rc.2.1`, rustc version `rustc 1.35.0 (3c235d560 2019-05-20)`

xSke:
Closes #623 .

### DIFF
--- a/src/model/guild/partial_guild.rs
+++ b/src/model/guild/partial_guild.rs
@@ -1,7 +1,7 @@
 #[cfg(feature = "http")]
 use crate::http::CacheHttp;
 use crate::{model::prelude::*};
-use super::super::utils::{deserialize_emojis, deserialize_roles};
+use super::super::utils::{deserialize_emojis, deserialize_roles, deserialize_u64_or_zero};
 
 #[cfg(feature = "model")]
 use crate::builder::{CreateChannel, EditGuild, EditMember, EditRole};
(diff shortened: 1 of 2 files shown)


xSke:
The solution here means the deserialization/serialization *technically* isn't round-trip (input null would get reserialized to 0), I'm not sure whether this is important to you or not. I didn't want to directly make the field an `Option<u64>`, since that's a breaking API change for no practical benefit (`None` and `Some(0)` would be equivalent in practice).

src/model/guild/partial_guild.rs
```diff
@@ -39,6 +39,8 @@ pub struct PartialGuild {
     pub verification_level: VerificationLevel,
     pub description: Option<String>,
     pub premium_tier: PremiumTier,
+    // In some cases Discord returns `null` rather than 0
```
Lakelezz: Missing fullstop at the end of the sentence.

src/model/utils.rs
```diff
@@ -231,6 +231,11 @@ pub fn deserialize_u64<'de, D: Deserializer<'de>>(deserializer: D) -> StdResult<
     deserializer.deserialize_any(U64Visitor)
 }
 
+/// Deserializes an u64, returning 0 if the deserialization failed
```
Lakelezz: Missing fullstop at the end of the sentence.

Lakelezz:
Thank you :)

PR #629, opened 2019-06-23
WIP: More efficient cache

xacrimon:
This is a pull request merging some patches from me to make the cache more effective and remove the global lock. This should improve latency and allows multiple threads to operate on the cache at the same time. Please note that this is a suggestion and that the patches are NOT finished yet.


xacrimon:
Closed since wrong target.

PR #632, opened 2019-06-25
Use the threadpool from uvth instead of the threadpool crate.

xacrimon:
The official `threadpool` crate is very unoptimized and somewhat slow. `uvth` implements a still very compact but much more efficient pool with less overhead. It is still not as optimized as it can be (far from) but it is better than the `threadpool` crate. This pull swaps threadpool for uvth.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,7 @@ version = "0.2"
 
 [dependencies.uvth]
 optional = true
-version = "1.0.1"
+version = "1.0.3"
 
 [dependencies.tungstenite]
 default-features = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,9 +69,9 @@ features = ["std"]
 optional = true
 version = "0.2"
 
-[dependencies.threadpool]
+[dependencies.uvth]
 optional = true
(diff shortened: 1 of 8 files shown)


ghost:
:-1: I'd rather use tested and community supported libraries for something like this, especially ones that have had peer review. The suggested library doesn't seem to actually be threadpool-y, and blocks its thread on every new task to run it to completion.

xacrimon:
I'll just add my comment here that zeyla and Erk missed some code and that it is infact a threadpool. I wont touch other opinions.

xacrimon:
Huh, CI is acting up on Linux. I think this might be due to uvth requiring rust 1.35. Fixes will be pushed shortly.

xacrimon:
Some benchmarks that compare uvth to the threadpool crate currently in use.

```
i7-7700HQ 4C/8T Clear Linux

Time taken to run 100k no-op (empty) jobs. Averages from 5050 iterations.

threadpool_crate        time:   [42.201 ms 42.373 ms 42.560 ms]

threadpool_uvth         time:   [4.4844 ms 4.5123 ms 4.5412 ms]

```

PR #634, opened 2019-06-26
Add the option to override the display name of a group

### LINKED ISSUE #622: Group name is limited to rust identifiers since v0.6
```
error: proc macro panicked
  --> src/main.rs:55:1
   |
55 | / group!({
56 | |     name: "Foo Bar",
57 | |     options: {
58 | |         // Sets multiple prefixes for a group.
...  |
68 | |     commands: [cat, dog],
69 | | });
   | |___^
   |
   = help: message: `"Foo Bar"` is not a valid identifier
```

arqunis:
Fixes #622


Erk-:
There should also be docs with it and maybe an example.

Lakelezz (approved):

PR #636, opened 2019-06-28
Args: Add `remains` and fix a typo in the `len` docs

anyputer:
Let me know if `option_rest` isn't a good name.

### DIFF
--- a/src/framework/standard/args.rs
+++ b/src/framework/standard/args.rs
@@ -750,7 +750,7 @@ impl Args {
     ///
     /// Returns `None` if there are no remaining arguments.
     #[inline]
-    pub fn option_rest(&self) -> Option<&str> {
+    pub fn remains(&self) -> Option<&str> {
         if self.is_empty() {
             return None;
         }
--- a/src/framework/standard/args.rs
+++ b/src/framework/standard/args.rs
@@ -737,13 +737,7 @@ impl Args {
     /// Starting from the offset, return the remainder of available arguments.
     #[inline]
     pub fn rest(&self) -> &str {
-        if self.is_empty() {
-            return "";
-        }
-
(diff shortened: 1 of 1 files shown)


src/framework/standard/args.rs
```diff
@@ -746,13 +746,27 @@ impl Args {
         &self.message[start..]
     }
 
+    /// Starting from the offset, return the remainder of available arguments.
+    ///
+    /// Returns `None` if there are no remaining arguments.
+    #[inline]
+    pub fn option_rest(&self) -> Option<&str> {
```
arqunis: I propose naming this a synonym of "rest":
```suggestion
    pub fn remains(&self) -> Option<&str> {
```

src/framework/standard/args.rs
```diff
@@ -746,13 +746,27 @@ impl Args {
         &self.message[start..]
```
arqunis: It's just barely out of reach, but could you reimplement `rest` in terms of `remains`? I.e.:
```rust
pub fn rest(&self) -> &str {
    self.remains().unwrap_or("")
}
```

arqunis (changes requested):

src/framework/standard/args.rs
anyputer: I have thought of that but I'm worried about any possible performance implications. Sure thing.

arqunis: Don't worry. If anything, rustc and, especially, llvm are very smart in inlining/optimizing unnecessary calls.
If you look at [this output from godbolt](https://godbolt.org/z/_XsW-r), you can see `rest` and `rest2` are pretty much identical in their assembly.

arqunis (approved):

PR #645, opened 2019-07-03
Add `secretbox` as `sodiumoxide`-backend alternative

### LINKED ISSUE #644: Add `secretbox` as alternative for the `sodiumoxide`-backend
The `sodiumoxide` crate is currently used to encrypt data for discord voice support. As a wrapper around a C-library it might not work on some platforms.

Additionally the upstream library used by `sodiumoxide` (`libsodium`) sometimes replaces some release tarballs, causing build failures for example on windows. (see  jedisct1/libsodium#813, sodiumoxide/sodiumoxide#326)

As a replacement i have recently written a crate called `secretbox`. It supports everything serenity uses `sodiumoxide` for currently and does so without a C-Library dependency.

DarkKirb:
This pull request fixes #644 

This adds two new features, `voice-secretbox` and `voice-sodiumoxide`. By default neither is chosen and currently results in a compilation error when just using voice.

The PR adds some small wrapper functions to abstract away the differences in API between the two libraries. This could be implemented more clearly with the voice reimplementation.

voice with secretbox has been tested and it maneges to successfully transmit audio.


Erk-:
For it to compile and work probably you have to make it work even if both are enabled, for example set
a default backend it will use if both are there. Also add a `compile_error!` to `build.rs` like the one there is for rustls and native-tls.

ghost:
The readme should mention these two new features, and say why you'd choose one over the other: https://github.com/serenity-rs/serenity/blob/1d797a20768dfb34502937d49ffd80d4ecf91299/README.md#features

This is a breaking change right now since there's no default with the `voice` feature. We'll need to think of what default we want, if we want one. The examples pick `voice-secretbox`. I'd lean to `voice-sodiumoxide` for 0.7 since it's more known and tested, and maybe switch to `voice-secretbox` later on after it's been used.

DarkKirb:
I would default to voice-secretbox on windows or when cross compiling. Should i automatically pick that in the build.rs file?

Erk-:
I think it would be better just to make secretbox the default, remember this default only comes into play if both features are enabled which is not likely in many cases, instead write that it is easier to cross compile with it in the readme for the feature.

ghost:
Clarifying: Erk's issue is about when *both* `secretbox` and `sodiumoxide` features are chosen while mine is about when *neither* are chosen

DarkKirb:
I was talking about the default when neither is chosen. When both are chosen I would use the faster implementation which is sodiumoxide

README.md
```diff
@@ -148,6 +148,13 @@ TLS implementation.
 - **native_tls_backend**: Uses SChannel on Windows, Secure Transport on macOS,
 and OpenSSL on other platforms.
 
+Serenity offers two Voice encryption backends. You need to pick one to use `voice`
```
Lakelezz: Missing fullstop at the end of the sentence.

README.md
```diff
@@ -148,6 +148,13 @@ TLS implementation.
...
+
+- **secretbox**: Uses secretbox for all platforms, a pure Rust Secretbox implementation. Use this if you are compiling for windows or cross-compiling
```
Lakelezz: Missing fullstop at the end of the sentence.

README.md
```diff
@@ -148,6 +148,13 @@ TLS implementation.
...
+- **sodiumoxide**: Uses sodiumoxide for all platforms, a wrapper around the libsodium C library.
+
+If multiple are selected it will use the faster implementation by default.
```
Lakelezz: Missing comma after "selected".

build.rs
```diff
@@ -7,5 +7,11 @@ compile_error!("You have the `http` or `gateway` feature enabled, \
     - `native_tls_backend` uses SChannel on Windows, Secure Transport on macOS, \
     and OpenSSL on other platforms.\n\
     If you are unsure, go with `rustls_backend`.");
+#[cfg(all(feature = "voice", not(any(feature = "secretbox", feature = "sodiumoxide"))))]
+compile_error!("You have the `voice` feature enabled, either the `secretbox` or `sodiumoxide` feature must be selected to let Serenity use `voice`.\n\
+               - `secretbox` uses a pure rust secretbox implementation that will work on any system.\n\
+               - `sodiumoxide` uses an optimized C-library\n\
```
Lakelezz: Please change "optimized" to "optimised". Also add missing fullstop.

build.rs
```diff
@@ -7,5 +7,11 @@ compile_error!("You have the `http` or `gateway` feature enabled, \
...
+               If you are cross-compiling (using --target) or compiling for Windows go with `secretbox`.");
```
Lakelezz: No need to explain "--target", if someone is cross-compiling, they will know how in the first place.

Cargo.toml
```diff
@@ -150,7 +154,10 @@ native_tls_backend = ["reqwest/default-tls", "tungstenite/tls"]
 model = ["builder", "http"]
 standard_framework = ["framework", "uwl", "command_attr"]
 utils = ["base64"]
-voice = ["byteorder", "gateway", "audiopus", "rand", "sodiumoxide"]
+voice = ["byteorder", "gateway", "audiopus", "rand"]
+voice-secretbox = ["voice", "secretbox"]
```
Lakelezz: We use underscore for features.
Furthermore, orientate yourself on our naming for the HTTP-backend.
Use "sodiumoxide_backend"/"secretbox_backend".

We will have to see how we are going to rename this to avoid confusion with now four features containing the word "backend". This might be "voice/sodiumoxide" etc., but that is out of the scope here.

Lakelezz (changes requested):

build.rs
```diff
@@ -7,5 +7,11 @@ compile_error!("You have the `http` or `gateway` feature enabled, \
     - `native_tls_backend` uses SChannel on Windows, Secure Transport on macOS, \
     and OpenSSL on other platforms.\n\
     If you are unsure, go with `rustls_backend`.");
+#[cfg(all(feature = "voice", not(any(feature = "secretbox", feature = "sodiumoxide"))))]
```
Lakelezz, line 10: Please add an empty line between both `compile_error`-macros, also make sure to refer to `*_backend` instead of the actual crate names.

arqunis:
I am closing this as there hasn't been any activity for half a year.

PR #661, opened 2019-07-17
Commit some modifications to the command_attr codebase

arqunis:
1) Tweaked attribute parsing to no longer care about the name and separated validation of correct data to its function. Consequently, the `AttributeOption` trait was simplified.
2) `format_args!` instead of `format!` for error message creation. [`Error::new`](https://docs.rs/syn/0.15.39/syn/parse/struct.Error.html#method.new) accepts any type `impl`'ing `Display`. [`Arguments`](https://doc.rust-lang.org/nightly/std/fmt/struct.Arguments.html) is one that the compiler can cheaply produce and need not be allocated, unlike `String`.
3) A `try_r!` macro for early-returning a `TokenStream` out of a syn error when one's present.

### DIFF
--- a/command_attr/src/attributes.rs
+++ b/command_attr/src/attributes.rs
@@ -1,10 +1,12 @@
-use crate::util::*;
-use crate::OnlyIn;
-use proc_macro2::Span;
-use std::fmt;
 use syn::parse::{Error, Result};
 use syn::spanned::Spanned;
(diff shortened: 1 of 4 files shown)


Lakelezz:
Do you really think `try_r` is a good name?

arqunis:
No. It was the first thing I thought of really, with the `_r` suffix because `try` is a reserved keyword in the 2018 edition. Perhaps it could be named `report_if_err` or something like that?

PR #662, opened 2019-07-17
Try to always ignore bots and webhooks when configured by the user

arqunis:
The `WithWhiteSpace` struct is also re-exported from the `configuration` module.


src/framework/standard/mod.rs
```diff
@@ -598,6 +598,14 @@ impl StandardFramework {
 
 impl Framework for StandardFramework {
     fn dispatch(&mut self, mut ctx: Context, msg: Message, threadpool: &ThreadPool) {
+        if let Some(error) = self.should_fail_common(&msg) {
+            if let Some(dispatch) = &self.dispatch {
```
Lakelezz, line 603: Missing newline after `if let`-block start.

Lakelezz (approved):

PR #668, opened 2019-07-26
Rewrite command parsing yet again

arqunis:
But this time, try to be more efficient.

The new code now creates a cache of all valid states we can dispatch commands from groups. This eases the parser in that it only needs to search through this cache a few times to affirm itself that the message is correct and a command may be dispatched.


Lakelezz (approved):

arqunis:
I've just realised that the `parse` module has never been public. Meaning we're free to play with it however we please. 
This allowed me to alter a few things I thought I didn't have a permit to do in the first place. The latest commit, though originally for a single `try` function as opposed to very similar `try_command` and `try_group` functions, contains these new alterations.

I'm pointing this out since it seems @Lakelezz was also [fooled thinking it was public api](https://github.com/serenity-rs/serenity/commit/9cc8816ec56295541230c87992500fee0aaa8696#diff-b1e5a85b1459fbf9883e5445e62fc796).

PR #671, opened 2019-07-27
Add a method to create and iterable of Members in a Guild

alch-emi:
Adds `.members_iter()` to both `Guild` and `GuildId`.  This produces an iterator containing *all* members of that guild, as produced by multiple calls to `.members()`, instead of just the first 1,000.

As I understand it, there is currently no way to produce a list of all members in a guild without either making multiple calls or not having a guarantee that all members will be included, especially for large servers.  That means that if you want a guaranteed way to view all members of a guild, you must implement a pattern much like this.  Since I would expect that this isn't an uncommon need, adding a method to do this easily seems like a good idea.

I chose to have the iterator produce `Result<Member>` instead of just `Member`s directly, as this allows the user to handle errors that occur after the first fetch of members.  Alternative implementations might have `.members_iter()` return a result, and then have that iterator produce only `Member`s.  However, this would again cause the problem that this iterator does not provide a guarantee of providing all members, since it would have to fail silently.

I'd also like to acknowledge that while I didn't make an issue preceding this pull request, I understand completely that this might not be approved simply because it is deemed unnecessary or unfitting, and am not submitting this under any expectation that it will be accepted.

This is also my first PR, not just to this repository, but any Rust-based repository in general.  As such, I'm very likely to have made poor design or style decisions in my code.  I'm very open to critical feedback in this area, and would appreciate any notes, no matter how nit-picky.  Thanks!

Edit: ~~I'm having trouble predicting failures of `macOS_no_gateway` and `macOS_stable_no_cache` with `cargo test --all-features` (hence the abundance of force-pushes).  Is this me failing to use the tools correctly, or is there a different way to preform that check locally?  (Sorry if that's off topic)~~ I was using the tools wrong

### DIFF
--- a/src/model/guild/guild_id.rs
+++ b/src/model/guild/guild_id.rs
@@ -795,6 +795,7 @@ pub struct MembersIter<H: AsRef<Http>> {
     http: H,
     buffer: Vec<Member>,
     after: Option<UserId>,
+    tried_fetch: bool,
 }
 
 #[cfg(all(feature = "http", feature = "cache"))]
@@ -805,6 +806,7 @@ impl<H: AsRef<Http>> MembersIter<H> {
             http: http,
             buffer: Vec::new(),
             after: None,
+            tried_fetch: false,
         }
     }
 
@@ -817,15 +819,20 @@ impl<H: AsRef<Http>> MembersIter<H> {
     /// members to be fetched, then this marks `self.after` as None, indicating
     /// that no more calls ought to be made.
     fn refresh(&mut self) -> Result<()> {
+        // Number of profiles to fetch
+        let grab_size: u64 = 1000;
+
         let next_members = self.guild_id
-            ._members(self.http.as_ref(), Some(1000), self.after)?;
+            ._members(self.http.as_ref(), Some(grab_size), self.after)?;
 
          //Get the last member.  If shorter than 1000, there are no more results anyway
-        self.after = next_members.get(1000)
+        self.after = next_members.get(grab_size as usize - 1usize)
             .map(|member| member.user_id());
 
         // Reverse to optimize pop()
         self.buffer = next_members.into_iter().rev().collect();
+
+        self.tried_fetch = true;
 
         Ok(())
     }
@@ -836,7 +843,7 @@ impl<H: AsRef<Http>> Iterator for MembersIter<H> {
     type Item = Result<Member>;
 
     fn next(&mut self) -> Option<Result<Member>> {
-        if self.buffer.len() == 0 && self.after.is_some() {
+        if self.buffer.len() == 0 && self.after.is_some() || !self.tried_fetch {
             if let Err(e) = self.refresh() {
                 return Some(Err(e))
             }
@@ -847,7 +854,7 @@ impl<H: AsRef<Http>> Iterator for MembersIter<H> {
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         let buffer_size = self.buffer.len();
-        if self.after.is_none() {
+        if self.after.is_none() && self.tried_fetch {
             (buffer_size, Some(buffer_size))
         } else {
             (buffer_size, None)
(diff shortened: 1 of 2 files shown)


src/model/guild/guild_id.rs
```diff
@@ -757,3 +784,83 @@ impl<'a> From<&'a Guild> for GuildId {
...
+
+        let next_members = self.guild_id
+            ._members(self.http.as_ref(), Some(grab_size), self.after)?;
+
+         //Get the last member.  If shorter than 1000, there are no more results anyway
+        self.after = next_members.get(grab_size as usize - 1usize)
+            .map(|member| member.user_id());
+
+        // Reverse to optimize pop()
+        self.buffer = next_members.into_iter().rev().collect();
```
arqunis: This can be optimised to reduce an allocation and be potentially faster.
Make the `next_members` variable `mut` and use the [`reverse`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.reverse) method before assigning to `buffer`:

```rust
let mut next_members = ...;

next_members.reverse();
self.buffer = next_members;
```

src/model/guild/guild_id.rs
```diff
@@ -757,3 +784,83 @@ impl<'a> From<&'a Guild> for GuildId {
...
+        Ok(())
+    }
+}
+
+#[cfg(all(feature = "http", feature = "cache"))]
+impl<H: AsRef<Http>> Iterator for MembersIter<H> {
+    type Item = Result<Member>;
+
+    fn next(&mut self) -> Option<Result<Member>> {
+        if self.buffer.len() == 0 && self.after.is_some() || !self.tried_fetch {
```
```diff
- self.buffer.len() == 0
+ self.buffer.is_empty()
```

src/model/guild/guild_id.rs
```diff
@@ -757,3 +784,83 @@ impl<'a> From<&'a Guild> for GuildId {
...
+    /// that no more calls ought to be made.
+    fn refresh(&mut self) -> Result<()> {
+        // Number of profiles to fetch
+        let grab_size: u64 = 1000;
+
+        let next_members = self.guild_id
+            ._members(self.http.as_ref(), Some(grab_size), self.after)?;
+
+         //Get the last member.  If shorter than 1000, there are no more results anyway
+        self.after = next_members.get(grab_size as usize - 1usize)
```
arqunis: The compiler should be smart enough to know that the `1` is of the `usize` type by the left hand side of the subtraction.
 ```suggestion
         self.after = next_members.get(grab_size as usize - 1)
```

File: src/model/guild/guild_id.rs [hunk 4]
```diff
@@ -757,3 +784,83 @@ impl<'a> From<&'a Guild> for GuildId {
...
+    type Item = Result<Member>;
+
+    fn next(&mut self) -> Option<Result<Member>> {
+        if self.buffer.len() == 0 && self.after.is_some() || !self.tried_fetch {
+            if let Err(e) = self.refresh() {
+                return Some(Err(e))
+            }
+        }
+
+        self.buffer.pop().map(|i| Ok(i))
```
arqunis: The `Ok` variant is a function of its own.

```suggestion
        self.buffer.pop().map(Ok)
```

arqunis (changes requested):
> This is also my first PR, not just to this repository, but any Rust-based repository in general.

Well, I shall give you an applause for that. Though there are some points I'll address, they're minor/not of substantial value, so they don't matter too much. Meaning you did great for your first try. 👍

alch-emi:
Thanks so much for the feedback!  I really appreciate it.  I've implemented all of the improvements you recommended.  What do you think about the architecture over all?  Do the method signatures make sense?  Is the overall flow reasonable?  Is everything implemented that ought to be?

arqunis (approved):

arqunis:
You've nailed the architecture down well, no complaint there. I'd approve the method signatures too, but reviewing again reveals I've missed something.

[The method on `Guild`](https://github.com/serenity-rs/serenity/pull/671/files#diff-e1c86b244465a2bb338d53758aa2c7efR914) should borrow the receiver (i.e `&self`). The iterator expects a http client and the guild's id, which can be copied cheaply, to work, really. There's nothing forbidding further usage of the guild after `members_iter` is run. 

Once you resolve the above, I'll gladly merge this.

alch-emi:
Oh, nice catch!  Thank you!

Issue #669, opened 2019-07-26
Add support for block quoting


anyputer:
Add ways of block quoting to the `MessageBuilder` and maybe `ContentModifier`.

Single line quotes could easily be `\n> {}\n`.
Multi-line quotes are weird in the way that there's no way to terminate them, so you can only add them to the end of a message. They start with `>>> `, then everything after it gets turned into a block. This means that it would be harder to get right than normal content modifiers. My suggestion is to have a `.quote_after()` method or something on the message builder that quotes everything after it. It should only be able to be called once. They can contain all sorts of content modifiers.

PR #672, opened 2019-07-30
Add quoting functionality to MessageBuilder

AregevDev:
Hello,
This PR adds quoting support for `MessageBuilder`.
It addresses #669.

Discord recently added the option to format a message with a `quote`:
image

I added the following method to be compatible with the existing API for pushing to `MessageBuilder`:
`push_quote`
`push_quote_line`
`push_quote_safe`
`push_quote_line_safe`
`start_quote` - this is used for declaring multiline quotes, any pushes after this one will be quoted

All working properly + all tests passing
Have a nice day!

### DIFF
--- a/src/utils/message_builder.rs
+++ b/src/utils/message_builder.rs
@@ -840,7 +840,7 @@ impl MessageBuilder {
     }
 
     /// Starts a multi-line quote, every push after this one will be quoted
-    pub fn start_quote(mut self) -> Self {
+    pub fn quote_rest(mut self) -> Self {
         self.0.push_str("\n>>> ");
 
         self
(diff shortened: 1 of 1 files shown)


AregevDev:
I consider renaming `start_quote` to `quote_rest` as @Erk- suggested, feedback?

arqunis:
I approve of his suggestion. `start_quote` would make sense if there were an `end_quote` counterpart, but since it administers the quoting to the remainder of the message without there being any indicator for it to stop, it's logically better to name it `quote_rest`.

AregevDev:
Done

anyputer:
how about making the new line a requirement and not having seperate new line methods?

AregevDev:
This would remove the API compatibility and some more convenience

anyputer:
i disagree

PR #674, opened 2019-08-01
Added group removal and non-consuming group adding functions to StandardFramework

Jerald:
At the very least, this is nice if you ever want to hard disable a group of commands on the fly. Otherwise, it allows people to do more fun things with `StandardFramework`.

### DIFF
--- a/src/framework/standard/mod.rs
+++ b/src/framework/standard/mod.rs
@@ -383,6 +383,15 @@ impl StandardFramework {
         self
     }
 
+    /// Removes a group from being used in the framework.
+    /// 
+    /// Note: does _not_ return `Self` like many other commands. This is because
+    /// it's not intended to be chained as the other commands are.
+    pub fn remove_group(&mut self, group: &'static CommandGroup) {
+        // Iterates through the vector and if a given group _doesn't_ match, we retain it
+        self.groups.retain(|&(g, _)| g != group)
+    }
+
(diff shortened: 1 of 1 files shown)


src/framework/standard/mod.rs
```diff
@@ -383,6 +383,35 @@ impl StandardFramework {
...
+    /// Adds a group to be used by the framework. Primary use-case is runtime modification
+    /// of groups in the framework; will _not_ mark the framework as initialized. Refer to
+    /// [`group`] for adding groups in initial configuration.
+    /// 
+    /// Note: does _not_ return `Self` like many other commands. This is because
+    /// it's not intended to be chained as the other commands are.
+    /// 
+    /// [`group`]: #method.group
+    pub fn group_add(&mut self, group: &'static CommandGroup) {
+        // Implementation is duplicated from above group function
```
arqunis: Then could you alter `group` to use this method? I'd rather not want to be shooting my, or anyone else's foot when committing modifications to `group` and forgetting to apply those to `group_add` as well or vice versa.

arqunis (changes requested):

src/framework/standard/mod.rs
Jerald: Changed!

Sorry for the delay, apparently I totally missed that you commented this...

arqunis (approved):

PR #680, opened 2019-08-08
Play a YouTube Search's first Video

Sreyas-Sreelal:
The current implementation of ytdl needs a youtube url.This pull request adds new functionality to play audio from the youtube search queries.

### DIFF
--- a/src/voice/streamer.rs
+++ b/src/voice/streamer.rs
@@ -318,6 +318,8 @@ pub fn ytdl(uri: &str) -> Result<Box<dyn AudioSource>> {
     Ok(pcm(true, ChildContainer(ffmpeg)))
 }
 
+/// Creates a streamed audio source from youtube search results with `youtube-dl`,`ffmpeg` and `ytsearch`
+/// Searches for video in youtube with a limit of 1
 pub fn ytdl_search(name: &str) -> Result<Box<dyn AudioSource>> {
     let ytdl_args = [
         "-f",
--- a/src/voice/streamer.rs
+++ b/src/voice/streamer.rs
@@ -318,8 +318,8 @@ pub fn ytdl(uri: &str) -> Result<Box<dyn AudioSource>> {
     Ok(pcm(true, ChildContainer(ffmpeg)))
 }
 
-/// Creates a streamed audio source from youtube search results with `youtube-dl`,`ffmpeg` and `ytsearch`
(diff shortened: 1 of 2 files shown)


Flat:
I feel like this should be up to the user to supply the url, but at the same time we _are_  just using youtube-dl directly.

Lakelezz:
Hello, thank you for your pull request : )

However neither the pull request nor the documented code really explain how the function intents to behave. Does it take the first result? If so, then this should be explained.

Sreyas-Sreelal:
It only takes one result (ytsearch1)

Lakelezz:
This should be mentioned in the documentation of the function.

Sreyas-Sreelal:
added docs

src/voice/streamer.rs
```diff
@@ -318,6 +318,8 @@ pub fn ytdl(uri: &str) -> Result<Box<dyn AudioSource>> {
     Ok(pcm(true, ChildContainer(ffmpeg)))
 }
 
+/// Creates a streamed audio source from youtube search results with `youtube-dl`,`ffmpeg` and `ytsearch`
```
Lakelezz: Missing fullstop at the end of the sentence.
Consider adding a comma before the "and".
Please write YouTube instead of "youtube".

src/voice/streamer.rs
```diff
@@ -318,6 +318,8 @@ pub fn ytdl(uri: &str) -> Result<Box<dyn AudioSource>> {
...
+/// Searches for video in youtube with a limit of 1
```
Lakelezz: Missing fullstop at the end of the sentence.
Also consider changing the sentence:
"Searches for the first listed YouTube video."

Lakelezz (changes requested):
Please fix commented issues.

src/voice/streamer.rs
dlee13: Perhaps I'm splitting hairs here, but I suggest, "Takes the first video listed from a Youtube search."

Lakelezz: Yes, that's even better.
Maybe replace "a YouTube search" with "the YouTube search" since a search is already mentioned in the first doc line.

Sreyas-Sreelal:
Done

Lakelezz (approved):

Lakelezz:
Thanks for your pull request :)

PR #683, opened 2019-08-11
Link to ShardMessenger::chunk_guilds from the Guild.members field

alch-emi:
The [Guild `members` field][1] details that not all users might be available, but doesn't specify that how one might request that the users be made available.  In addition, the method that one would use to accomplish this is kinda hard to find, being on a struct that you can write an entire program without touching once.  Linking to that method from the members field might clarify this a lot.  Sorry for the small PR, but I think this might be an important change (that would certainly have saved me a ton of time and confusion)

[1]: https://docs.rs/serenity/0.6.3/serenity/model/guild/struct.Guild.html#structfield.members

### DIFF
--- a/src/model/guild/mod.rs
+++ b/src/model/guild/mod.rs
@@ -110,8 +110,14 @@ pub struct Guild {
     /// if the [`member_count`] is greater than the `LARGE_THRESHOLD` set by
     /// the library.
     ///
+    /// You can manually request that the remaining members be sent by calling
+    /// [`ShardManager::chunk_guilds`], after which chunks containing guild
+    /// members will be asynchronously received until all members are
+    /// available.
+    ///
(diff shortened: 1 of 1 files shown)


arqunis (approved):

Lakelezz (changes requested):
Hello, thanks for your pull request :)

I personally would advice against explaining the function in the first place. The link should suffice. 

My reasoning: If the function ever changes, the user will always have the updated information when clicking on the link. Furthermore, the function's documentation must be good enough to stand on its own ground. 

Therefore, explaining that this functions allows to manually request remaining guild members, should be sufficient.

alch-emi:
Makes sense!  I tend to get a bit overexplainy.  How does this look?  Should it maybe be reduced to just a `**See Also**`?

Lakelezz (approved):

Lakelezz:
It's fine, thanks : )

PR #686, opened 2019-08-16
Allow delimiters to be set on a per command basis

arqunis:


Lakelezz:
The only issue ascending from this is supporting it in help. 
I know this is not part of this pull request but it surely needs to be added sooner or later, before we release preferably.
Adding this to help would change the number of characters generated and might lead to too long messages; it's a behavioural change after all.

What's your point of view? Should our next release be v0.7 then? Maybe together with new API changes that are incoming, even though we should be able to release them on v0.6 due to our non-exhaustive API-update.

arqunis:
> Should our next release be v0.7 then?

Sure, I guess.
However, as someone had mentioned on Discord, we could publish the framework in its own crate, letting us to release breaking changes to it independently of Serenity (as is the case for `command_attr`, kind of). It's not an integral part of the library, restraining ourselves from releasing a minor increment (0.0.x) because the framework received incompatible API changes to prior releases is really a hindrance. 

The framework, in comparison to the rest of the library, is inadequate for some users. It provides the convenience and hides the complexity that would be expected of a multi-feature framework, but it lacks control that users would much appreciate so they could tailor it precisely to their needs. I've been considering conceiving a modular framework of sorts that could be connected in the way users want. Of course, that's an imperious wish, but it could be done if the framework were disjointed from the rest of Serenity and allowed experiments of its own.

Anyhow, as I've said, we can deal with help via the 0.7 release.

Lakelezz (approved):

PR #691, opened 2019-08-21
Update audit log model to reflect Discord's optional/mixed types

ikkerens:
There is an error in the currently used data structure serenity which causes Json errors on models where serenity does not expect null values, or mixed type values. While this particular change does not properly wrap all the types, it instead focuses on ensuring it works, not that it works well.

Good enough for a temporary solution I'd say.


Erk-:
Should be retargeted at next as it includes breaking changes

arqunis (approved):

Lakelezz (approved):

PR #699, opened 2019-08-25
Utilise Discord's new precision and reset-after headers

### LINKED ISSUE #694: Implement more precise rate limiting
Implement more precise rate limiting cf. https://github.com/discordapp/discord-api-docs/pull/1064

### LINKED ISSUE #698: Make rate limiting work independently of the system clock
https://github.com/discordapp/discord-api-docs/pull/1069

arqunis:
The precision header (`X-RateLimit-Precision`) with the `millisecond` value lets us inform Discord to provide the milliseconds that it otherwise would've rounded up to a second. This finally allows us to implement ratelimiting for the reaction endpoints properly, as they're the only ones that use millisecond precision, compared to other endpoints.

The reset-after header (`X-RateLimit-Reset-After`) gives the reset time directly in seconds (milliseconds via the precision header), unlike its bigger sibling (`X-RateLimit-Reset`) which gives the reset time in a Unix timestamp (seconds/milliseconds since January 1st, 1970 UTC). This can cause varying degrees of inaccuracy depending on how desync the (user's) system's time is with Discord's servers as we used the system time to verify whether or not the reset time has already passed and to extract the value for the current thread to sleep. 
However, reset-after does not pose these problems, and since it's essentially the sleep value, there're no calculations needed — it can just be used as one. Though from the code, I've come to the conclusion that we don't have to use the reset header at all. Its first purpose in our ratelimiting code was to check whether the bucket has reset, and if it did, update the `remaining` value to that of the `limit`. The post-hook code also updates the `remaining` and `limit` values from the http response, which already contains their (potentially) new values, so I removed the code from the pre-hook. I made the post-hook code still parse the reset header because it could be useful for debugging.

And lastly, I removed the time-difference offset as it's been superseded by reset-after.

Closes #694.
Closes #698.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -143,6 +143,7 @@ gateway = [
     "utils",
 ]
 http = []
+absolute_ratelimits = ["http"]
 rustls_backend = ["reqwest/rustls-tls", "tungstenite", "rustls", "webpki", "webpki-roots"]
 native_tls_backend = ["reqwest/default-tls", "tungstenite/tls"]
 model = ["builder", "http"]
--- a/README.md
+++ b/README.md
@@ -137,6 +137,10 @@ the HTTP functions.
 connected to and audio can be sent/received.
 - **default_native_tls**: Default features but using `native_tls_backend`
 instead of `rustls_backend`.
+- **absolute_ratelimits**: Whether the library should use your system clock to avoid
+ratelimits, or use the interval given by Discord that might be less efficient
+due to latency in the network. If you turn this feature on, it is recommended to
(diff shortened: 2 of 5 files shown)


Flat (approved):

Erk- (approved):

Lakelezz:
So, I see a problem with `X-RateLimit-Reset-After`.

As being brought up in https://github.com/discordapp/discord-api-docs/pull/1069:

> with the caveat that network latency will cause you to be less efficient than using X-RateLimit-Reset with a synchronized clock source.

This would be a clear case where we want to have a feature instead of just a changed behaviour.

PR #706, opened 2019-08-28
Use transport compression

Erk-:
This does currently not really work at all as it does not seem to decompress it


Erk-:
Closed as I need to restructure the implementation in a way that would not lead me to have massive buffers in memory at all times.

PR #707, opened 2019-08-29
Add the ability to have multiple examples in commands

mendess:
Adds the ability to use the `#[example(s)]` macro multiple times on commands to add multiple examples.

It looks like this:
image

This does change a some public struct fields, namely
- `command_attr::structures::Options`
-  `framework::standard::structures::ComandOptions`
The latter being the most important since it's on the public API and could cause breakage.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -133,7 +133,7 @@ pub struct Command<'a> {
     availability: &'a str,
     description: Option<&'static str>,
     usage: Option<&'static str>,
-    usage_sample: Option<&'static str>,
+    usage_sample: Vec<&'static str>,
     checks: Vec<String>,
 }
 
@@ -527,7 +527,7 @@ fn nested_group_command_search<'a>(
                     aliases: options.names[1..].to_vec(),
                     availability: available_text,
                     usage: options.usage,
-                    usage_sample: options.example,
+                    usage_sample: options.examples.to_vec(),
                 },
             });
         }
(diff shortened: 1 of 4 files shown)


src/framework/standard/help_commands.rs
```diff
@@ -1080,13 +1080,20 @@ fn send_single_command_embed(
...
 
-            if let Some(ref example) = command.usage_sample {
-                let full_example_text = if let Some(first_prefix) = command.group_prefixes.get(0) {
-                    format!("`{} {} {}`", first_prefix, command.name, example)
-                } else {
-                    format!("`{} {}`", command.name, example)
-                };
-
+            if !command.usage_sample.is_empty() {
+                let format_example: Box<dyn Fn(&str) -> String> =
```
arqunis: I'd prefer to allocate as little as possible.

```rust
let mut full_sample_text = String::new();

let first_prefix = command.group_prefixes.get(0);

for example in &command.usage_sample {
    if let Some(first_prefix) = first_prefix {
        let _ = writeln!(
            &mut full_sample_text, 
            "`{} {} {}`", 
            first_prefix, command.name, example
        );
    } else {
        let _ = writeln!(
            &mut full_sample_text, 
            "`{} {}`", 
            command.name, example
        );
    }
}
```

src/framework/standard/help_commands.rs
```diff
@@ -1312,20 +1319,29 @@ fn single_command_to_plain_string(help_options: &HelpOptions, command: &Command<
...
-            );
-        } else {
-            let _ = writeln!(
-                result,
-                "**{}**: `{} {}`",
-                help_options.usage_label, command.name, example
-            );
-        }
+    if !command.usage_sample.is_empty() {
+        let mut format_example: Box<dyn FnMut(&mut String, &str)> =
```
arqunis:
```rust
let first_prefix = command.group_prefixes.get(0);

for example in &command.usage_sample {
    if let Some(first_prefix) = first_prefix {
        let _ = writeln!(
            &mut full_sample_text, 
            "**{}**: `{} {} {}`", 
            help_options.usage_label, first_prefix, command.name, example
        );
    } else {
        let _ = writeln!(
            &mut full_sample_text, 
            "**{}**: `{} {}`", 
            help_options.usage_label, command.name, example
        );
    }
}
```

arqunis (changes requested):

arqunis (approved):

PR #715, opened 2019-09-11
Add optional inviter field to Invite

ikkerens:
The Discord API optionally sends this when requesting an invite through the invite endpoint. Serenity does currently not add it to the Invite struct, this PR changes that.

### DIFF
--- a/src/model/invite.rs
+++ b/src/model/invite.rs
@@ -218,7 +218,7 @@ impl Deref for InviteUser {
     type Target = UserId;
 
     fn deref(&self) -> &Self::Target {
-        return &self.id
+        &self.id
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/model/invite.rs
```diff
@@ -184,7 +199,30 @@ impl Invite {
...
+
+/// InviteUser implements a Deref to UserId so it gains the convenience methods
+/// for converting it into a [`User`] instance.
+///
+/// [`User`]: ../user/struct.User.html
+impl Deref for InviteUser {
+    type Target = UserId;
+
+    fn deref(&self) -> &Self::Target {
+        return &self.id
```
arqunis:
```suggestion
        &self.id
```

arqunis (changes requested):

arqunis (approved):

PR #718, opened 2019-09-20
Change AttachmentType to use Cows instead

ikkerens:
This PR actually consists of three changes to make AttachmentType significantly more useable and (when done properly) significantly more efficient.

- AttachmentType variants that used a tuple before now use a struct with named fields, making the code more self-documenting.
- AttachmentType variants that needed a filename now use String instead of &str, preventing a `.to_string()` call later on and giving the end library user the responsibility (which in turn also prevents any potential lifetime pitfalls)
- AttachmentType Bytes now uses a `std::borrow::Cow` instead of a lifetimed-borrow. When uploading attachments it is not always desirable to pass a reference due to the way people structure their applications, especially as later on when actually sending the attachment the reference is copied by calling `to_vec`. Using a cow annihilates this completely because using `Cow::Owned::into_owned()` is a no-op, thus preventing a potential copy of up to 50MB.

It's worth noting that this is definitely a breaking change (hence the next branch), but with some potential huge gains.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1456,16 +1456,16 @@ impl Http {
         for file in files {
 
             match file.into() {
-                AttachmentType::Bytes((bytes, filename)) => {
+                AttachmentType::Bytes{data, filename} => {
                     multipart = multipart
-                        .part(file_num.to_string(), Part::bytes(bytes.to_vec())
-                            .file_name(filename.to_string()));
+                        .part(file_num.to_string(), Part::bytes(data.into_owned())
+                            .file_name(filename));
                 },
-                AttachmentType::File((file, filename)) => {
+                AttachmentType::File{file, filename} => {
                     multipart = multipart
                         .part(file_num.to_string(),
                             Part::reader(file.try_clone()?)
-                                .file_name(filename.to_string()));
+                                .file_name(filename));
                 },
                 AttachmentType::Path(path) => {
                     multipart = multipart
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1456,12 +1456,12 @@ impl Http {
         for file in files {
 
             match file.into() {
-                AttachmentType::Bytes{data, filename} => {
+                AttachmentType::Bytes{ data, filename } => {
                     multipart = multipart
                         .part(file_num.to_string(), Part::bytes(data.into_owned())
                             .file_name(filename));
                 },
-                AttachmentType::File{file, filename} => {
+                AttachmentType::File{ file, filename } => {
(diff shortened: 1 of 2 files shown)


src/http/client.rs
```diff
@@ -1456,16 +1456,16 @@ impl Http {
         for file in files {
 
             match file.into() {
-                AttachmentType::Bytes((bytes, filename)) => {
+                AttachmentType::Bytes{data, filename} => {
```
arqunis: Formatting nitpick: Spaces after `{` and before `}`.
```suggestion
                AttachmentType::Bytes{ data, filename } => {
```

src/http/client.rs
```diff
@@ -1456,16 +1456,16 @@ impl Http {
...
                     multipart = multipart
-                        .part(file_num.to_string(), Part::bytes(bytes.to_vec())
-                            .file_name(filename.to_string()));
+                        .part(file_num.to_string(), Part::bytes(data.into_owned())
+                            .file_name(filename));
                 },
-                AttachmentType::File((file, filename)) => {
+                AttachmentType::File{file, filename} => {
```
arqunis: Ditto.
```suggestion
                AttachmentType::File{ file, filename } => {
```

src/http/mod.rs
```diff
@@ -195,17 +196,17 @@ impl LightMethod {
...
     /// Indicates that the `AttachmentType` is a `File`
-    File((&'a File, &'a str)),
+    File{ file: &'a File, filename: String },
     /// Indicates that the `AttachmentType` is a `Path`
     Path(&'a Path),
     #[doc(hidden)]
     __Nonexhaustive,
 }
 
 impl<'a> From<(&'a [u8], &'a str)> for AttachmentType<'a> {
```
arqunis: Since we no longer depend on the filename's livability, its lifetime should be removed:
```suggestion
impl<'a> From<(&'a [u8], &str)> for AttachmentType<'a> {
```
This has the same effect as if we used two distinct lifetimes (`<'a, 'b>`) for the data (`'a`) and the filename (`'b`).

src/http/mod.rs
```diff
@@ -223,7 +224,7 @@ impl<'a> From<&'a PathBuf> for AttachmentType<'a> {
 }
 
 impl<'a> From<(&'a File, &'a str)> for AttachmentType<'a> {
```
arqunis: Ditto.

arqunis (changes requested):

arqunis (approved):

PR #724, opened 2019-09-26
Interpret doc-comments as the description attribute

### LINKED ISSUE #723: Bring #[description] in line with rust's #[doc]
Rust's doc attribute can be used multiple times to add newlines without awkwardly splitting the strings over newlines:
```rust
#[doc = "This is"]
#[doc = " a "]
#[doc = "doc comment"]
```
is equivalent to
```rust
#[doc = "This is\n a \ndoc comment"]
```
[Offical Docs](https://doc.rust-lang.org/rustdoc/the-doc-attribute.html)  
I would like that  the `#[description]` macro would support the same (or at least a similar) behavior.

arqunis:
Fixes #723

### DIFF
--- a/command_attr/src/structures.rs
+++ b/command_attr/src/structures.rs
@@ -88,20 +88,6 @@ fn parse_argument(arg: FnArg) -> Result<Argument> {
     }
 }
 
-#[inline]
-fn parse_doc(doc: Attribute) -> Attribute {
-    Attribute {
-        pound_token: doc.pound_token,
-        style: doc.style,
-        bracket_token: doc.bracket_token,
-        path: Path::from(PathSegment::from(Ident::new(
-            "description",
-            Span::call_site(),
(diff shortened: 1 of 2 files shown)


command_attr/src/structures.rs
```diff
@@ -91,7 +92,7 @@ fn parse_argument(arg: FnArg) -> Result<Argument> {
 pub struct CommandFun {
     /// `#[...]`-style attributes.
     pub attributes: Vec<Attribute>,
-    /// Populated by either `#[cfg(...)]` or `#[doc = "..."]` (the desugared form of doc-comments) type of attributes.
+    /// Populated by `#[cfg(...)]` type of attributes.
```
Flat: type ~~of~~ attributes

Flat (changes requested):
Minor grammar nit

PR #742, opened 2019-11-19
Cover all error cases for audit log deserialization

ikkerens:
Continuation of #741 

Previous PR did not cover all cases of this error, seeing as serde has different ways to do things in some cases.
(The madness that is serde does not help my sanity)

### DIFF
--- a/src/model/guild/audit_log.rs
+++ b/src/model/guild/audit_log.rs
@@ -454,7 +454,7 @@ impl<'de> Deserialize<'de> for AuditLogs {
                             users = Some(map.next_value::<Vec<User>>()?);
                         }
                         Ok(None) => break, // No more keys
-                        Err(e) => if format!("{}", e).contains("unknown field") {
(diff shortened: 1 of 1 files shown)


src/model/guild/audit_log.rs
```diff
@@ -454,7 +454,7 @@ impl<'de> Deserialize<'de> for AuditLogs {
                             users = Some(map.next_value::<Vec<User>>()?);
                         }
                         Ok(None) => break, // No more keys
-                        Err(e) => if e.description().contains("unknown field") {
+                        Err(e) => if format!("{}", e).contains("unknown field") {
```
arqunis: nitpick:
```suggestion
                        Err(e) => if e.to_string().contains("unknown field") {
```
Implementors of `Display` get a free `ToString` implementation that is equivalent to the `format!` macro here.

arqunis (changes requested):

arqunis (approved):

PR #747, opened 2019-12-01
Implement Default for CommandOptions and GroupOptions

### LINKED ISSUE #739: Implement Default for CommandOptions, GroupOptions
These structs have semantic defaults (values set if you don't use the `#[command]` related attributes), but they don't implement `Default`, which is a pain when creating these structures dynamically. Implementing `Default` on those structs should be pretty easy.

(Having all the members be `&'static` references instead of owned values is also annoying...)

LeSeulArtichaut:
Resolves #739.

### DIFF
--- a/src/framework/standard/structures/mod.rs
+++ b/src/framework/standard/structures/mod.rs
@@ -29,6 +29,7 @@ impl Default for OnlyIn {
     fn default() -> Self { Self::None }
 }
 
+#[derive(Debug, Default, PartialEq)]
 pub struct CommandOptions {
     /// A set of checks to be called prior to executing the command. The checks
     /// will short-circuit on the first check that returns `false`.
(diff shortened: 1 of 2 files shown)


arqunis (approved):
I would rather prefer the `Default` impls be consistent with the values from the `#[command]` and `group!` macros. However, since the point of the issue that was raised had been dynamic creation of the option structs, I'll accept this simple solution.

LeSeulArtichaut:
@acdenisSK What have I missed? I thought this was already consistant with the implementation of the macro. Please tell me, I’ll rewrite the code.

arqunis:
These values: https://github.com/serenity-rs/serenity/blob/11ea4fb881b7dac8bf0de61d039aa4245876bd51/command_attr/src/structures.rs#L298-L299

The macros themselves have to build the option structs from scratch. They also turn on certain options. It is only after they've been set that the macros begin applying the user's values to the respective options.

But as I said, I'm fine with the changes as they are right now. Therefore I shall be merging this pr as it is.

Thank you for the effort of making this pull request.

PR #751, opened 2019-12-13
Fix regression

LeSeulArtichaut:


src/framework/standard/structures/mod.rs
```diff
@@ -235,6 +221,21 @@ pub struct HelpOptions {
     pub indention_prefix: &'static str,
 }
 
+#[derive(Debug, PartialEq)]
```
arqunis: This should also derive `Default` to avoid regressing https://github.com/serenity-rs/serenity/commit/918273b8e936796e6424b0c28c0a929f7ce6bf03.

```suggestion
#[derive(Debug, Default, PartialEq)]
```

arqunis (changes requested):

arqunis (approved):

PR #752, opened 2019-12-19
Use #[non_exhaustive]

LeSeulArtichaut:
With the arrival of Rust 1.40, we can now write non exhaustive enums/structs with the built-in `#[non_exhaustive]` attribute.
This PR replaces all occurences of the `__Nonexhaustive` hack previously in place with the new attribute.

I opened this so we can discuss whether this change is appropriate or not for Serenity, as this may force users to use Rust 1.40. However, if merged to `current`, it will only affect Serenity 0.8 and later, so I personally think this is fine.


LeSeulArtichaut:
@Flat Should we try to re-launch the checks? Or is this fine?

Flat:
Seems the checks are still broken for 1.40. Pending review we will also need to reach a consensus if this can be merged to current.

arqunis:
This pull request hasn't been updated since December; I'm closing this.

Flat:
Did it need an update? I believe the checks just needed to be reran after they were updated and determine the correct merge target.

PR #753, opened 2019-12-19
Upgrade to reqwest v0.10.0

nickelc:
The `rustls` feature is back in the `master` branch for the upcoming version 0.10 of reqwest.

I'll update this PR when reqwest v0.10 is released or merge it as it is and i create another PR for the release version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,8 +56,9 @@ version = "1"
 
 [dependencies.reqwest]
 default-features = false
+features = ["blocking", "json"]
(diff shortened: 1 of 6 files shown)


Erk-:
Is the ParseError no longer accessible from request? If not you will have to add url to the http feature in Cargo.toml.

nickelc:
Yes, ParseError is no longer used and re-exported in reqwest.

Erk- (approved):

PR #755, opened 2019-12-23
Fix Guild::edit_role_position example

### LINKED ISSUE #754: Incorrect doc example
https://docs.rs/serenity/0.7.4/serenity/model/guild/struct.Guild.html#method.edit_role_position

The example in the doc is missing the context parameter.

LeSeulArtichaut:
Example now mimics GuildId::edit_role_position
Closes #754

### DIFF
--- a/src/model/guild/mod.rs
+++ b/src/model/guild/mod.rs
@@ -718,8 +718,8 @@ impl Guild {
     /// Change the order of a role:
     ///
     /// ```rust,ignore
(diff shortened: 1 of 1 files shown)


arqunis:
The `v0.7.x` branch shall only get fixes related to broken functionality and the alike. Could you change it back to `current`? It's a very insignificant change for a method that's rarely used, thus I would not like to dedicate a release specifically for it, nor leave it sitting there.

PR #773, opened 2020-01-27
Add support for the Watching activity type.

evieluvsrainbows:
It *is* undocumented in the Discord API documentation, however it has been supported by the API for a  while. It isn't used that often by Discord bots, but it might be good to support it, just in case.

### DIFF
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -163,6 +163,53 @@ impl Activity {
         }
     }
 
+    /// Creates an `Activity` struct that appears as a `Watching <name>`
+    /// status.
+    ///
+    /// **Note**: Maximum `name` length is 128.
+    ///
+    /// # Examples
+    ///
+    /// Create a command that sets the current watching status:
+    ///
+    /// ```rust,no_run
+    /// use serenity::model::gateway::Activity;
+    /// use serenity::model::channel::Message;
+    /// # #[cfg(feature = "framework")]
+    /// use serenity::framework::standard::{Args, CommandResult, macros::command};
+    /// # #[cfg(feature = "client")]
+    /// use serenity::client::Context;
+    ///
+    /// # #[cfg(feature = "framework")]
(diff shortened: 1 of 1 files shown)


evieluvsrainbows:
I have personally tested this with my own Discord bot, and I can confirm that this works. Also, turns out that it is documented, however it's documented as part of the Discord Game SDK instead of the normal Discord API. Kind of confused as to why, but Discord is Discord, I guess 😂

Lakelezz:
As has been discussed before: https://github.com/serenity-rs/serenity/pull/409#issuecomment-427842549

Also see Discord's stance: 
https://github.com/discordapp/discord-api-docs/issues/463
https://github.com/discordapp/discord-api-docs/issues/978

Therefore I advise against adding this, as it is seen as a placeholder.

evieluvsrainbows:
> As has been discussed before: [#409 (comment)](https://github.com/serenity-rs/serenity/pull/409#issuecomment-427842549)
> 
> Also see Discord's stance:
> [discordapp/discord-api-docs#463](https://github.com/discordapp/discord-api-docs/issues/463)
> [discordapp/discord-api-docs#978](https://github.com/discordapp/discord-api-docs/issues/978)
> 
> Therefore I advise against adding this, as it is seen as a placeholder.

It might be seen as a placeholder, but it's still an available Activity type. And it is exposed and documented in the Game SDK which shares some similarities from what I can see with the normal Discord API. And because some bots use said activity type, it means that if you have a user command for example, and parse activities, that activity won't show up or even be parsed by the library. That seems like an issue to me.

Erk-:
We could add it but with the documentation hidden.

I also asked a bit around and it seems others implement it.

arqunis:
I am rejecting this addition for one of Discord's activity types. It is undocumented, and as it had been said by Discord, they may change or remove it at any moment without anyone's approval.

PR #778, opened 2020-02-05
model::user::User now implements Default.

allquixotic:
This is useful for unit testing code that relies on the user model without needing a live client context.

[x] Builds successfully (rustc stable 1.40.0)
[x] Unit tests pass

### DIFF
--- a/src/model/user.rs
+++ b/src/model/user.rs
@@ -458,6 +458,19 @@ impl Hash for User {
     }
 }
 
+impl Default for User {
+    fn default() -> Self {
+        User {
+            id: UserId(210),
+            avatar: Some("abc".to_string()),
+            bot: true,
+            discriminator: 1432,
+            name: "test".to_string(),
+            _nonexhaustive: (),
+        }
+    }
+}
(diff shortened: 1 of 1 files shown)


src/model/user.rs
```diff
@@ -458,6 +458,19 @@ impl Hash for User {
     }
 }
 
+impl Default for User {
+    fn default() -> Self {
```
arqunis, line 462: Documentation should be added to inform users which default values this method assumes for a `User`.

e.g.:
```rust
/// Initializes a `User` with default values. Setting the following:
/// - **id** to `UserId(210)`
/// - **avatar** to `Some("abc")`
/// ....
```

arqunis (changes requested):

arqunis:
Because of the inactivity, I have taken the liberty of [merging this PR with my requested changes](https://github.com/serenity-rs/serenity/commit/f940fc293f2b21afefc51dddc3ecad98ce29fbe3) manually, but with credit where it is due.

PR #781, opened 2020-02-10
Allows `add_file` to support an image URL

Elinvynia:
This is very shitty, tips on improving appreciated

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1432,7 +1432,11 @@ impl Http {
                         .file(file_num.to_string(), path)?;
                 },
                 AttachmentType::Image(url) => {
-                    let parsed_url = match Url::parse(url) {
+                    let filename = Url::parse(url)
+                        .ok()
+                        .and_then(|url| url.path_segments())
+                        .and_then(|segments| segments.last())
+                        .ok_or_else(|| Error::Url(url.to_string())?;
                         Ok(url) => url,
                         Err(_) => return Err(Error::Url(url.to_string())),
                     };
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1434,7 +1434,7 @@ impl Http {
                 AttachmentType::Image(url) => {
                     let mut picture: Vec<u8> = vec![];
                     let req = &mut self.client.get(url).send()?;
-                    let _ = std::io::copy(&mut *req, &mut picture);
+                    std::io::copy(&mut *req, &mut picture)?;
                     multipart = multipart
                         .part(file_num.to_string(), Part::bytes(Cow::Borrowed(&picture[..]).into_owned())
                             .file_name(format!("{}.webp", file_num)));
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1436,18 +1436,7 @@ impl Http {
                         .ok()
                         .and_then(|url| url.path_segments())
                         .and_then(|segments| segments.last())
-                        .ok_or_else(|| Error::Url(url.to_string())?;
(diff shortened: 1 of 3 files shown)


src/http/client.rs
```diff
@@ -1430,6 +1431,14 @@ impl Http {
                     multipart = multipart
                         .file(file_num.to_string(), path)?;
                 },
+                AttachmentType::Image(url) => {
+                    let mut picture: Vec<u8> = vec![];
+                    let req = &mut self.client.get(url).send()?;
+                    let _ = std::io::copy(&mut *req, &mut picture);
+                    multipart = multipart
+                        .part(file_num.to_string(), Part::bytes(Cow::Borrowed(&picture[..]).into_owned())
+                            .file_name(format!("{}.webp", file_num)));
```
Flat: This should use the actual file extension unless you are converting the image to webp format.

src/http/client.rs
```diff
@@ -1430,6 +1431,14 @@ impl Http {
                     multipart = multipart
                         .file(file_num.to_string(), path)?;
                 },
+                AttachmentType::Image(url) => {
+                    let mut picture: Vec<u8> = vec![];
+                    let req = &mut self.client.get(url).send()?;
+                    let _ = std::io::copy(&mut *req, &mut picture);
```
Flat: You should handle the result of `std::io::copy` here instead of ignoring it with a `let _` binding

Flat (changes requested):

src/http/client.rs
```diff
@@ -1430,6 +1431,26 @@ impl Http {
...
+                    let path_segments = match parsed_url.path_segments() {
+                        Some(p) => p,
+                        None => return Err(Error::Url(url.to_string())),
+                    };
+                    let filename = match path_segments.last() {
+                        Some(l) => l,
+                        None => return Err(Error::Url(url.to_string())),
+                    };
+                    let mut picture: Vec<u8> = vec![];
+                    let req = &mut self.client.get(url).send()?;
```
```suggestion
                    let mut req = self.client.get(url).send()?;
```

src/http/client.rs
```diff
@@ -1430,6 +1431,26 @@ impl Http {
...
+                    std::io::copy(&mut *req, &mut picture)?;
```
```suggestion
                    std::io::copy(&mut req, &mut picture)?;
```

src/http/client.rs
```diff
@@ -1430,6 +1431,26 @@ impl Http {
                     multipart = multipart
                         .file(file_num.to_string(), path)?;
                 },
+                AttachmentType::Image(url) => {
+                    let parsed_url = match Url::parse(url) {
```
```suggestion
                    let filename = Url::parse(url)
                        .ok()
                        .and_then(|url| url.path_segments())
                        .and_then(|segments| segments.last())
                        .ok_or_else(|| Error::Url(url.to_string())?;
```

src/http/mod.rs
```diff
@@ -210,7 +212,13 @@ impl<'a> From<(&'a [u8], &str)> for AttachmentType<'a> {
 }
 
 impl<'a> From<&'a str> for AttachmentType<'a> {
-    fn from(s: &'a str) -> AttachmentType<'_> { AttachmentType::Path(Path::new(s)) }
+    fn from(s: &'a str) -> AttachmentType<'_> {
```
arqunis, line 217:
```suggestion
    /// Constructs an `AttachmentType` from a string.
    /// This string may refer to the path of a file on disk, or the http url to an image on the internet.
    fn from(s: &'a str) -> AttachmentType<'_> {
```

Issue #785, opened 2020-02-14
Slow deserialazing in fire function.


noituri:
So after using `ctx.http.get_guild_roles` I experienced very slow deserialization.
I found out that `from_reader` is slower for some reason. Changing this to `from_str` fixed my issue.
https://github.com/Noituri/serenity/commit/e425c2dfd9322c0525b68c8de8f321f0df7b4d47

noituri:
The fix has been merged

PR #787, opened 2020-02-15
Deserializing from `str` in `fire` is faster

noituri:


noituri:
https://github.com/serenity-rs/serenity/issues/785

Erk-:
Would it be faster again to do it the following way?
```rust
response.json::<T>().map_err(From::from)
```

noituri:
So from what I've seen `.json` uses `from_slice` so in theory it should work just fine. I haven't done any tests yet but I think it will do the job.

noituri:
I did some tests and it works perfectly.

Erk- (approved):

evieluvsrainbows (approved):

PR #797, opened 2020-03-03
Implemented various kick_with_reason() methods.

vicky5124:
modified:   src/http/client.rs
	modified:   src/http/routing.rs
	modified:   src/model/guild/guild_id.rs
	modified:   src/model/guild/member.rs
	modified:   src/model/guild/mod.rs
	modified:   src/model/guild/partial_guild.rs
	modified:   src/utils/mod.rs

Implemented a new method `model::guild::Member.kick_with_reason(Http, Option<&str>)` that allows the user to kick a member from a guild with an optional reason. 
Should not be a breaking change. (EDIT: it is a breaking change due to the kick_member() method)

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1366,7 +1366,6 @@ impl Http {
             route: RouteInfo::KickMember {
                 guild_id,
                 user_id,
-                // Supplement an empty string when `reason` is `None`.
                 reason: reason,
             },
         })
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1359,12 +1359,6 @@ impl Http {
 
     /// Kicks a member from a guild.
     pub fn kick_member(&self, guild_id: u64, user_id: u64, reason: Option<&str>) -> Result<()> {
-        let r;
-        if let Some(_) = reason {
-            r = reason
-        } else {
-            r = Some("");
-        }
 
         self.wind(204, Request {
             body: None,
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1358,11 +1358,22 @@ impl Http {
     }
 
     /// Kicks a member from a guild.
-    pub fn kick_member(&self, guild_id: u64, user_id: u64) -> Result<()> {
+    pub fn kick_member(&self, guild_id: u64, user_id: u64, reason: Option<&str>) -> Result<()> {
+        let r;
+        if let Some(_) = reason {
+            r = reason
+        } else {
+            r = Some("");
+        }
+
         self.wind(204, Request {
             body: None,
             headers: None,
-            route: RouteInfo::KickMember { guild_id, user_id },
+            route: RouteInfo::KickMember {
+                guild_id,
+                user_id,
+                reason: r
+            },
         })
     }
 
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1366,7 +1366,7 @@ impl Http {
             route: RouteInfo::KickMember {
                 guild_id,
                 user_id,
-                reason: reason,
+                reason,
             },
         })
     }
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1358,7 +1358,7 @@ impl Http {
     }
 
     /// Kicks a member from a guild.
-    pub fn kick_member(&self, guild_id: u64, user_id: u64, reason: Option<&str>) -> Result<()> {
+    pub fn kick_member(&self, guild_id: u64, user_id: u64, reason: &str) -> Result<()> {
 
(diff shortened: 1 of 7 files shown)


src/http/client.rs
```diff
@@ -1358,11 +1358,22 @@ impl Http {
...
 
     /// Kicks a member from a guild.
-    pub fn kick_member(&self, guild_id: u64, user_id: u64) -> Result<()> {
+    pub fn kick_member(&self, guild_id: u64, user_id: u64, reason: Option<&str>) -> Result<()> {
+        let r;
+        if let Some(_) = reason {
+            r = reason
+        } else {
+            r = Some("");
+        }
```
```suggestion
```

src/http/client.rs
```diff
@@ -1358,11 +1358,22 @@ impl Http {
...
+
         self.wind(204, Request {
             body: None,
             headers: None,
-            route: RouteInfo::KickMember { guild_id, user_id },
+            route: RouteInfo::KickMember {
+                guild_id,
+                user_id,
+                reason: r
```
```suggestion
                // Supplement an empty string when `reason` is `None`.
                reason: reason.or(Some("")),
```

arqunis (changes requested):

src/model/guild/member.rs
```diff
@@ -325,6 +325,37 @@ impl Member {
...
+    /// }
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// Same as [`kick`]
+    ///
+    /// [`kick`]: #method.kick
+    #[cfg(feature = "http")]
+    pub fn kick_with_reason(&self, cache_http: impl CacheHttp, reason: Option<&str>) -> Result<()> {
```
NovusTheory: I don't know if it's just me but it feels weird to have a method named kick_with_reason but making the reason optional. It kind of seems to counteract what the method name is wanting.

vicky5124: Did this so Args.remains() could be used directly as a reason

src/http/client.rs
```diff
@@ -1358,11 +1358,17 @@ impl Http {
     }
 
     /// Kicks a member from a guild.
-    pub fn kick_member(&self, guild_id: u64, user_id: u64) -> Result<()> {
+    pub fn kick_member(&self, guild_id: u64, user_id: u64, reason: Option<&str>) -> Result<()> {
```
Scotsguy: This is a `pub` function, so this is a breaking change, is it not?

vicky5124: It seems so.
Easily fixed by passing None as an additional argument

Scotsguy: Easily fixed, sure, but a breaking change nonetheless

src/model/guild/member.rs
arqunis: I agree with Novus. If the method tells its purpose is the ability to kick someone from a guild with a reason as an explanation why, it should be a given that it will always expect a reason from the user. It should never be `None`.

This should apply to the other `kick`s you have changed. I would like to relieve the library from breaking the interface too much for something as simple as kicking with a reason.

arqunis (changes requested):

src/http/client.rs
```diff
@@ -1367,7 +1367,7 @@ impl Http {
                 guild_id,
                 user_id,
                 // Supplement an empty string when `reason` is `None`.
-                reason: reason.or(Some("")),
+                reason: reason,
```
```suggestion
                reason,
```

src/http/client.rs
```diff
@@ -1367,7 +1367,7 @@ impl Http {
                 guild_id,
                 user_id,
                 // Supplement an empty string when `reason` is `None`.
```
```suggestion

```

src/model/guild/member.rs
```diff
@@ -355,7 +355,7 @@ impl Member {
     ///
     /// [`kick`]: #method.kick
     #[cfg(feature = "http")]
-    pub fn kick_with_reason(&self, cache_http: impl CacheHttp, reason: Option<&str>) -> Result<()> {
+    pub fn kick_with_reason(&self, cache_http: impl CacheHttp, reason: &str) -> Result<()> {
```
arqunis, line 358: Alright, you have almost got my all of wishes, but you seem to have missed one from my last review. 
I wanted for you administer this name change to all of the other `kick` methods as well, so that we wouldn't break the higher-level interface people would most likely use.

What I mean is that, for each `kick` method, there should be an accompanying `kick_with_reason` that the former would use. In the same way you have done it with `Member::kick`.

arqunis (changes requested):

arqunis (approved):

Issue #786, opened 2020-02-15
Future Development of Serenity


Lakelezz:
This issue tries to look into what's highly requested and where the limited development power should go to.
I would like to know what you want to have in Serenity! Please reason why :)
As async Serenity is already in development, I will leave that out of this issue.

Here are some plans in no particular order:

- New Voice
Serenity has bare minimum support for voice activities. A playlist support could be one of the many things we could have. See issue #633 

- New Help System
Improving customisability, improving code, and maybe looking into paged help. Note, paged help is already possible if you make your own help function.

- New Cache 
There are some tricky deadlock cases that could hugely benefit from a redesign. See issue #584 

- New Bucket System
The current bucketing is very simple. Fine-grained buckets, that can be manually invalidated, serialised, could be useful.

- Framework Rework
Maybe a dynamic framework could be interesting. 

This list are thoughts I came up with, if there is anything else, feel free to mention and explain it!
Looking forward for your thoughts and wishes, thanks for your time.

moxian:
The most requested feature I see is "i want to create a poll, and have my bot do something, when people react to this specific message in a specific way" (e.g. "assign role A when reacting with emoji A, and kick from the server when reacting with emoji B").

Maybe in more generic terms - the ability to dynamically create long-lived(?) callbacks tied(??) to specific messages.

evieluvsrainbows:
A feature I'm interested in seeing in the future is a collector system, like Discord.js has. Collectors would allow bot developers to write, for example, commands that listen for user input after an  initial command has been sent, and respond with extra information. One example would maybe be network-based search commands, like a command that uses an API to e.g. search for songs from Spotify, and if there are multiple search results matching their query, allow users to pick which result they want, and then respond with the appropriate result. Could also be useful for e.g. reactions.

Issue #729, opened 2019-10-15
Async and Futures


MOZGIII:
It's time to revisit #298.

MTRNord:
This is still sadly blocking a project of mine :( cant use it with other messaging services that use another "tokio thread system"

TapGhoul:
I am talking with the peeps on Discord, I may look into trying to do this. It's not a small task but it's definitely something I'm open to investigate. Will be looking at the tokio 0.2.0 branch (std futures).

TapGhoul:
Update: Until async fn is available for traits, you can't really expose this to the user in an ergenomic fashion (making all the callbacks async) without using something like async-trait - which in of itself feels like a bit of a hack. We can expose the tokio runtime instance - and even allow you to provide your own - however there is no perfect solution as of yet, sadly.

I can look into updating everything under the hood, but the end result will not be as nice as I would have hoped, due to missing key rust features for this to work properly. I don't know if I'll commit to this as of yet as the payoff I'm hoping for is blocked by async fn in traits.

See: https://areweasyncyet.rs/

mgostIH:
@Silvea12 Since async in traits is still quite a hard blocker, what about using [async-trait](https://crates.io/crates/async-trait) until they get stabilized? I don't think the performance impact of allocating would be a huge deal in this project and it would allow to easily continue the development instead of stalling for what may be 1+ years.
Moreover, switching from the crate to the stable feature would be painless, as you just need to remove a macro.

TapGhoul:
@mgostIH I had looked into that. My 2 worries are lock-in, as well as the fact that the end user is, as a result, required to also use the macro if they are to extend the client, which makes it more difficult for new rust users to reason about at a glance (I struggled with this a bit when I was initially learning) - but it's definitely been a consideration.
My thoughts as an ideal (in my head) solution was mentioned in my above comment.

> We can expose the tokio runtime instance - and even allow you to provide your own - however there is no perfect solution as of yet, sadly.

We can run each call on its own async call (though coming from go this is bad practice - but this is what other libraries acting as runtimes had done anyway which I found quite ergonomic to use) as a standard call and just provide an instance to the tokio runtime to allow for users to spin off tasks from these async-wrapped sync calls. It's slightly higher overhead, but it means less (even if still minimal) boilerplate.
Means that if you don't care about async, your stuff is already run on a thread pool. If you do, however, you can quite easily take full advantage of the tokio runtime.

[...]

PR #803, opened 2020-03-12
Serenity.await

Lakelezz:
This pull request rewrites Serenity to be asynchronous.

As highly demanded (#729), Serenity is finally async.
For a long time, it was nothing but a dream, but now it's reality.

There are still a few things following in this PR, but it's ready to be used, see example 5.

Missing bits:
- [x] `voice`-module;
- [x] `voice`-examples (6, 10);
- [x] scheduling/eventing example (12);
- [x] updating all docs, doc-tests, and code examples in other files;
- [x] `native_tls_backend`;
- [ ] final code refactoring;

To fix:
- [x] Serenity becomes unresponsive, no shard activity, goes offline;

Also, collectors (as mentioned in #786) were added too.

You can test it already!


Scotsguy:
I've updated my bot to serenity.await already: https://github.com/Scotsguy/background-cat/tree/async
(Here's the updating commit, as an example: https://github.com/Scotsguy/background-cat/commit/fd14980c0ddfaea02a1e03dc15a99a61da1ef5bb)

nickelc:
I've successfully updated my bot. :tada: 
Everything seems to work fine with my async coded i use with the current version.

What i don't understand is that `StandardFramework`'s `Args` is partial async.
```rust
let filter = match args.single::<u32>().await {
    Ok(id) => Id::eq(id),
    Err(_) => Fulltext::eq(args.rest()),
};
```

Lakelezz:
> What i don't understand is that `StandardFramework`'s `Args` is partial async.

I'm sorry for this, there is an internal locked state happening inside `Args`, however @acdenisSK will rework `Args` soon to enable even more possibilities, it will help with the API too!


> I've successfully updated my bot. 🎉
> Everything seems to work fine with my async coded i use with the current version.

Very glad to hear this! If there is trouble, let me know!

Lakelezz:
The latest changes address the partial async API of the `Args`-struct, this introduced new breaking changes - if you ported your application already.

---

On another note, example 12 will be replaced by an example about event collectors.
An event collector, e.g. a message collector, can await replies from a user you are interacting with.
It's currently being implemented, however, there is already a working test implementation, it's not merged yet, but here is a sneak peek:

```rust
    let messages = msg.author.id
        .await_reply()
        .collect_limit(3)
        .timeout(Duration::from_secs(5))
        .receive_all(&ctx)
        .await;

    if messages.is_empty() {
        if let Err(why) = msg.reply(&ctx.http, "You did not send any messages!").await {
            println!("Error sending message: {:?}", why);
        }
    }
```

It will await up to 3 messages sent by the author of `msg` for up to 5 seconds.

hiratara:
I am very much looking forward to Serenity.await 😄

I found some bugs in `dispatch.rs`, and I sent some fixes.

* https://github.com/Lakelezz/serenity/pull/2
* https://github.com/Lakelezz/serenity/pull/4

I'm sorry if you've already checked these PRs. I'd appreciate it if these fixes could be available in the `await` branch.

Lakelezz:
Yes, thank you. I merged them!

nickelc:
I've started to fix doc comment examples, starting with `src/client/*`.

arranf:
Is there any way to help with this effort?

src/model/utils.rs
```diff
@@ -67,7 +67,7 @@ pub fn deserialize_members<'de, D: Deserializer<'de>>(
     let mut members = HashMap::new();
 
     for member in vec {
-        let user_id = member.user.read().id;
+        let user_id = futures::executor::block_on(member.user.read()).id;
```
ankoh: My bot is consistently deadlocking here.
I'm fairly new to Rust, but I'm getting here via `async fn recv_event` of the ShardRunner.
Isn't `block_on` in an asynchronous function considered to be very dangerous?

ankoh:
```
tokio::task::block_in_place(|| { futures::executor::block_on(member.user.read()).id })
```
This "fixes" the deadlock but would be a rather ugly hack.
The exact same problem occurs quite a few times whenever you de-/serialize a tokio RWLock and later need the ids for HashMaps.

Lakelezz:
> 
> Is there any way to help with this effort?

Testing is one way to help with the effort! The more testing gets done, the faster I can merge this into Serenity. The last chunk is being added soon: Collectors!

src/model/utils.rs
Lakelezz: Yes, this is not good, indeed. I will look into this.

Lakelezz:
Collectors are now usable! Please test and report bugs or design suggestions : )
In order to get started, check out the reworked example 12!

I will address some bugs next!

hiratara:
Hello,

I've tested the `06-voice` example and noticed the `~play` command hangs the bot. https://github.com/Lakelezz/serenity/pull/7 fixes the problem.

I also suggest the https://github.com/Lakelezz/serenity/pull/6 for minor refactoring. It's not a mandatory change, but I believe that it is a good thing for the project 😄

src/model/utils.rs
Lakelezz: Here is an insight into the issue: https://github.com/tokio-rs/tokio/issues/2376

ankoh: The idiomatic approach would certainly be an async serde-rs/json that doesnt exist (yet).
https://github.com/serde-rs/json/issues/575

The only option that I currently see is to skip the RWLocks during de-/serialization und introduce them afterwards. Either by using a dedicated copy of the types for serde or by using a private field for each rwlock.

ankoh: Of course, `tokio::task::block_in_place` remains an unattractive, yet working solution.
For deserialization, `block_in_place` doesn't sound super evil since the objects haven't been shared with anyone yet.
For serialization, one could just deeply copy the root to make sure the read will never block as well.

Lakelezz:
A small tip, due to https://github.com/tokio-rs/tokio/issues/2376 your bot may start stalling.
You can bypass this for now by pinning tokio to "0.2.13" while I'm working on a solution for this problem.

Celti:
Is it expected that `AttachementType` keeps using `std::fs::File` and not `tokio::fs::File`?

Lakelezz:
> 
> 
> Is it expected that `AttachementType` keeps using `std::fs::File` and not `tokio::fs::File`?

Might make more sense.

---

On another note, collectors have been reworked! They are now streams. Import `use serenity::futures::stream::StreamExt;` or  `use serenity::futures::stream::TryStreamExt;` to put them to use.
This does not affect utility methods on models that await a single item (`await_reply` and `await_reaction`). These will still be futures returning that optional item.
Their plural variant - `await_reactions` and `await_replies` - return a future building the stream, so add your timeout and other things and then simply await the builder to work with the stream.

Lakelezz:
We switched to `tokio::fs::File` now!

Lakelezz:
Just like on the `next` branch, gateway intents are now usable!

Lakelezz:
Tokio blocking/stalling/locking has been fixed! Some models have been adjusted as a result, changes should be straight forward though!

src/model/channel/mod.rs
```diff
@@ -203,8 +199,8 @@ impl Channel {
     #[inline]
     pub async fn is_nsfw(&self) -> bool {
```
arqunis: Redundant `async`
```suggestion
    pub fn is_nsfw(&self) -> bool {
```

src/model/guild/member.rs
```diff
@@ -225,7 +222,7 @@ impl Member {
     pub async fn display_name(&self) -> Cow<'_, String> {
```
arqunis: Ditto
```suggestion
    pub fn display_name(&self) -> Cow<'_, String> {
```

arqunis (changes requested):

examples/02_transparent_guild_sharding/src/main.rs
```diff
@@ -48,7 +48,10 @@ async fn main() {
     // Configure the client with your Discord bot token in the environment.
     let token = env::var("DISCORD_TOKEN")
         .expect("Expected a token in the environment");
-    let mut client = Client::new(&token, Handler).await.expect("Err creating client");
+    let mut client = Client::new(&token, Handler)
```
nickelc:
```suggestion
    let mut client = Client::new(&token)
```

examples/05_command_framework/src/main.rs
```diff
@@ -113,32 +119,86 @@ If you want more information about a specific command, just pass the command as
...
     help_options: &'static HelpOptions,
     groups: &[&'static CommandGroup],
     owners: HashSet<UserId>
 ) -> CommandResult {
-    help_commands::with_embeds(context, msg, args, help_options, groups, owners)
+    help_commands::with_embeds(context, msg, args, help_options, groups, owners).await
+}
+
+#[hook]
+async fn before(ctx: &mut Context, msg: &Message, command_name: &str) -> bool {
```
nickelc:
```suggestion
async fn before(ctx: &Context, msg: &Message, command_name: &str) -> bool {
```

examples/05_command_framework/src/main.rs
```diff
@@ -113,32 +119,86 @@ If you want more information about a specific command, just pass the command as
...
+    let mut data = ctx.data.write().await;
+    let counter = data.get_mut::<CommandCounter>().expect("Expected CommandCounter in TypeMap.");
+    let entry = counter.entry(command_name.to_string()).or_insert(0);
+    *entry += 1;
+
+    true // if `before` returns false, command processing doesn't happen.
+}
+
+#[hook]
+async fn after(_ctx: &mut Context, _msg: &Message, command_name: &str, command_result: CommandResult) {
```
nickelc:
```suggestion
async fn after(_ctx: &Context, _msg: &Message, command_name: &str, command_result: CommandResult) {
```

nickelc: +the rest of the hooks

Lakelezz:
I aim to merge this pull request soon, we are still checking if the OOM bug - that affects only certain users - is fixed with the recent changes.

The final breaking change has been merged, it brings a new cache designed to clone items instead of providing them locked. This includes items' fields, all structs will be without locks now.

About the cache redesign: Deadlocks caused by cache locking are a common and hard to debug error, therefore the cache's fields are now private. The single lock around the cache was broken down into multiple locks around each field.
Furthermore, the known methods will perform a clone, if you wish you to avoid cloning full structs, use the `*_field`-methods that have been added.
They take a closure in which you clone the required field.

Oh and of course: If there is anything missing on the cache, feel free to request it!

kageurufu:
I migrated my toy bot to use async as a test, and got an oom this morning.

```
[2732306.340126] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1000.slice/user@1000.service,task=shitbot,pid=1554827,uid=1000                                            
[2732306.340152] Out of memory: Killed process 1554827 (shitbot) total-vm:10989016kB, anon-rss:10686496kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:21012kB oom_score_adj:0
[2732306.784876] oom_reaper: reaped process 1554827 (shitbot), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

Let me know what I can do to help

arqunis:
Closing this in favour of #905 for continuing the async/await development due to inactivity here.

Issue #633, opened 2019-06-25
Voice Redesign


Erk-:
This was previously a discussion in the voice team but it would be better to have it here in public
(ping @FelixMcFelix @Flat)

FelixMcFelix:
Copied from the internal discussion...

---

A concern I have with voice is that many external operations (play, pause, volume set/check) rely upon locking a shared object. This introduces an extra layer of busy-work for users while making it possible for them to carelessly block the audio thread. This is something we need to be careful about when we introduce callbacks.

Some ideas I have for this:

* Keep a similar design to what we have now. Users have direct access to LockedAudio objects and callbacks would presumably offer direct access to these same objects. We make the risks of performing any operations explicit in our documentation.

* Move to a command-based aproach. Users are given an AudioFacade (or similar object) to send individual commands/requests to the audio thread, which it then executes between requests for audio frames. More complex logic can be sent as a closure, a sole point in the API where we need to make the risks of blocking apparent. This also presents a good opportunity to isolate other sources of blocking (packet sends, mallocs, etc).

In the first case, thread blocking (and potential audio glitches) are still liable to happen. I'm not convinced that an RwLock would help, because the audio thread always requires write access to the audio source.

The second case, while minimising user-caused blocking, makes it harder for users to query an Audio object (as they must now wait for a reply from one of the audio threads). We can cache things like metadata, which should be static, but this could be a pain point.

Flat:
Original discussion start. Any additional feedback welcome.

> **Topics and ideas to rework the voice modules to be more user friendly and feature complete.**
> Suggestions:
> 
> 
> - Replace youtube-dl and ffmpeg command calling with rust native libraries or FFI
> -  Provide metadata from playing tracks that are accessible from commands such as title, artist, filename, bit-rate etc. (via ffprobe, website api or scraping)
> - Callbacks for audio status (Finished, error, started, etc)
> - Optional parameters to disconnect or stop sub-processes automatically if no other users are in the channel, or if the bot in not in a voice channel
> - Add examples for queuing tracks

MOZGIII:
Just so that I understand the current design on serenity overall correctly - when command handler reaches interacts with the system it may performs various blocking operations (i.e. lock, io, etc) and all of them do actually block the thread?

MOZGIII:
[...]

PR #806, opened 2020-03-18
Voice Rework -- Events, Track Queues

FelixMcFelix:
Following on from #633, this implements a proof-of-concept for an improved audio frontend. The largest change is the introduction of events and event handling: both by time elapsed and by track events, such as ending or looping. Following on from this, the library now includes a basic, event-driven track queue system (which people seem to ask for unusually often). A new sample, `examples/13_voice_events`, demonstrates both the `TrackQueue` system and some basic events via the `~queue` and `~play_fade` commands.

Locks are removed from around the control of `Audio` objects, which should allow the backend to be moved to a more granular futures-based backend solution in a cleaner way. There are more issues I would like to tackle before merging:

User-facing tasks:

- [x] Global track/time events.
- [x] Seekable/loopable audio streams.
- [x] In-memory buffering (compressed/uncompressed) of streams.
- [x] Audio input metadata discovery/use.
- [x] Migrate audio control/query functions to async.
- [x] Move to subcrate.
- [x] Twilight support.
- [x] (Provably) survive rebalances.

Backend tasks:

- [x] Rework RTP parsing (since packet decoding appears to have somehow broken down when there are multiple senders). I'm currently writing a library based on [pnet](https://docs.rs/pnet).
- [x] Voice gateway version 4 support.
- [x] Rework input system for audio tracks.
- [x] Rework container/decoding system for audio tracks.
- [x] Separate tasks (audio mixing, collection, event handling, network) out from single thread.
- [x] Async conversion of above tasks.

---

Primarily, I'm looking for feedback on the user-facing API. Audio state queries moving to what is an essentially async API is a large semantic change -- while I'm convinced it's for the better (and will prevent people from hitting audio glitches, and lock contention due to constantly checking state), it's not what people are used to. This could probably be made sync-ish by having the audio threads update a shared `AudioState` which the handles have a reference to, if need be.


vicky5124:
It is possible to combine `voice::streamer::ytdl()` and `voice::streamer::ytdl_search()` using `--default-search auto` on the youtube-dl command parameters.

FelixMcFelix:
@nitsuga5124 Agreed, they have a lot in common. I'll simplify them a bit after I get compressed buffers working.

FelixMcFelix:
I've detailed the data structure I used in 
636ff61 and 2b9c4f6 [here](https://mcfelix.me/blog/shared-buffers/) to make it easier to audit this PR down the line.

examples/14_voice_events_queue/src/main.rs
```diff
@@ -0,0 +1,501 @@
...
+            },
+        };
+
+        // This handler object will allow you to, as needed,
+        // control the audio track via events and further commands.
+        let song = handler.play_source(source.into());
+        let send_http = ctx.http.clone();
+        let chan_id = msg.channel_id;
+
+        // This shows how t0 periodically fire an event, in this case to
```
Flat: s/t0/to/

src/prelude.rs
```diff
@@ -29,4 +29,4 @@ pub use crate::model::ModelError;
 #[cfg(feature = "utils")]
 pub use crate::utils::{TypeMap, TypeMapKey};
 #[cfg(feature = "voice")]
-pub use crate::voice::VoiceError;
+pub use crate::voice::VoiceError;
```
Flat: Line end change?

FelixMcFelix: Accidentally stripped trailing newline.

src/voice/connection.rs
```diff
@@ -143,40 +165,32 @@ impl Connection {
...
 
+        // Follow Discord's IP Discovery procedures, in case NAT tunnelling is needed.
+        let mut bytes = [0; IpDiscoveryPacket::const_packet_size()];
         {
-            let mut bytes = [0; 70];
-
-            (&mut bytes[..]).write_u32::<BigEndian>(ready.ssrc)?;
-            udp.send_to(&bytes, destination)?;
+            let mut view = MutableIpDiscoveryPacket::new(&mut bytes[..])
+                .expect("[Voice] There are enough bytes to encode a discovery packet.");
```
Flat: This will panic stating no issue. Is that intended?

src/voice/connection.rs
```diff
@@ -143,40 +165,32 @@ impl Connection {
...
-            // indicating the "end of the address".
-            let index = bytes
-                .iter()
-                .skip(4)
-                .position(|&x| x == 0)
-                .ok_or(Error::Voice(VoiceError::FindingByte))?;
+        let (len, _addr) = udp.recv_from(&mut bytes)?;
+        {
+            let view = IpDiscoveryPacket::new(&bytes[..len])
+                .expect("[voice] Should be enough space to view a packet we just fully received.");
```
Flat: Again, if this panics there does not seem to be an issue to the user.

src/voice/connection.rs
```diff
@@ -192,15 +206,21 @@ impl Connection {
...
-            "[Voice] WS heartbeat duration given as {}ms, adjusted to {}ms.",
+            "[Voice] WS heartbeat duration {}ms.",
             hello.heartbeat_interval,
-            temp_heartbeat,
         );
 
+        let mut packet = [0u8; VOICE_PACKET_MAX];
+
+        let mut rtp = MutableRtpPacket::new(&mut packet[..])
+            .expect("There is provably enough space to squeeze in an RTP header.");
```
Flat: Similar to previous comments

src/voice/connection.rs
```diff
@@ -374,95 +406,111 @@ impl Connection {
...
+            let now = Instant::now();
+            let temp_len = stream.mix(&mut mix_buffer, vol);
+            let later = Instant::now();
 
-            if !finished {
+            *time_in_call += later - now;
+            *entry_points += 1;
+
+            if *entry_points % 1000 == 0 {
+                println!("Average cost {:?}ns", time_in_call.as_nanos()/(*entry_points as u128));
```
Flat: This should likely be a `debug!()`, `trace!()` or possibly `info!()`

src/voice/connection.rs
```diff
@@ -559,67 +606,70 @@ impl Connection {
...
-            let mut cursor = &mut packet[..HEADER_LEN];
-            cursor.write_all(&[0x80, 0x78])?;
-            cursor.write_u16::<BigEndian>(self.sequence)?;
-            cursor.write_u32::<BigEndian>(self.timestamp)?;
-            cursor.write_u32::<BigEndian>(self.ssrc)?;
-        }
+                   -> Result<()> {
+        let index = {
+            let mut rtp = MutableRtpPacket::new(&mut self.packet[..])
+                .expect("There is provably enough space to squeeze in an RTP header.");
```
Flat: Ditto

src/voice/connection.rs
```diff
@@ -559,67 +606,70 @@ impl Connection {
...
-            secretbox::seal(slice, &nonce, &self.key)
-        };
-        let index = HEADER_LEN + crypted.len();
-        packet[HEADER_LEN..index].clone_from_slice(&crypted);
+        self.udp.send_to(&self.packet[..index], self.destination)?;
 
-        self.sequence = self.sequence.wrapping_add(1);
-        self.timestamp = self.timestamp.wrapping_add(960);
+        let mut rtp = MutableRtpPacket::new(&mut self.packet[..])
+            .expect("There is provably enough space to squeeze in an RTP header.");
```
Flat: Ditto

src/voice/connection.rs
```diff
@@ -291,6 +310,19 @@ impl Connection {
         packet: &[u8],
         nonce: &mut Nonce,
         ) -> Result<()> {
+        match demux::demux(packet) {
+            Demuxed::Rtp(rtp) => {},
+            Demuxed::Rtcp(rtcp) => {
+                println!("{:?}", rtcp);
```
Flat: `debug!()` or `trace!()`

Same for next line.

Flat, line 316: `debug!()` or `trace!()`

Same for next line.

Flat (changes requested):
Some nits about panic strings and printlns.

src/voice/connection.rs
FelixMcFelix: Yes: the `bytes` is sized to have exactly enough space for an IpDiscovery packet, so failure *cannot* happen. I'm using `expect` when we can make assurances ahead of time (the `discortp`   library can only fail if there are too few bytes).

src/voice/connection.rs
FelixMcFelix: See above.

src/voice/connection.rs
FelixMcFelix: This one maybe should be changed to an `Error`, since it could fail if discord completely change the API (unlike the other cases, which will only fail if the compiler is faulty).

src/voice/connection.rs
FelixMcFelix: This (and `entry_points`) will be removed once the patch is finalised.

src/voice/connection.rs
FelixMcFelix: See above.

src/voice/connection.rs
FelixMcFelix: See above.

src/voice/connection.rs
FelixMcFelix: Will be replaced with actual logic to pass through to a `Receiver`, replacing the audio receiving system.

src/voice/connection.rs
Flat: The use of expect is fine, but the message should say something like too few bytes were passed then as it will only show that string if that is the case. If the code is ever changed to cause the expect to trigger.

FelixMcFelix:
I believe this is as done as a synchronous implementation would be. It's been a while, but I hope this makes for a significant improvement.

evieluvsrainbows:
Small question: how will this be ported to async serenity once it's merged? I believe once async serenity is done that might be merged first. I could most def could be wrong about that though.

FelixMcFelix:
> 
> 
> Small question: how will this be ported to async serenity once it's merged? I believe once async serenity is done that might be merged first. I could most def could be wrong about that though.

Depends on how soon we want this integrated. Async-serenity was a moving target when I got started on the rewrite, so I've kept it in mind from a design perspective, but couldn't target it at writing time.

* Easiest approach is to let voice "do its own thing" internally for now, and adapt user-facing code to async. This mainly refers to `Handler` and `TrackHandle`, who send/receive messages into/from the voice tasks.
* Beyond that, `Input` should work on `AsyncRead` + `AsyncSeek`, and internal tasks should be made async functions analogous to @Lakelezz's current port of old-voice.
* Point of contention -- event subsystem could in theory accept `async fn` callbacks from users, but I believe this will stop us from being able to cheaply hand over references to the spawned tasks. Trade-off there.

The removal of locks, shared ownership of `WsClient`, etc. which have been done shouldn't make this too onerous, I'd hope, but I can't look into this for some time. The push to get this reasonably finished has definitely been draining.

examples/10_voice_receive/src/main.rs
```diff
@@ -137,7 +87,106 @@ fn join(ctx: &Context, msg: &Message, mut args: Args) -> CommandResult {
...
     if let Some(handler) = manager.join(guild_id, connect_to) {
-        handler.listen(Some(Box::new(Receiver::new())));
+        // Receiving audio (and similar tasks) are achieved
+        // by creating a global event handler.
+        handler.add_global_event(
+            CoreEvent::SpeakingStateUpdate.into(),
+            |ctx| {
+                if let EventContext::SpeakingStateUpdate(s) = ctx {
+                    // You can implement logic here so that you can differentiate users'
+                    // SSRCs and map the SSRC to the User ID and maintain this state.
```
Flat: May be better to use the full version before using an initialization so everyone knows what an SSRC is.

src/voice/connection.rs
```diff
@@ -474,55 +323,10 @@ impl Connection {
...
-
-        // Send the voice websocket keepalive if it's time
-        self.check_keepalive_timer()?;
-
-        // Send UDP keepalive if it's time
-        self.check_audio_timer()?;
 
         // Reconfigure encoder bitrate.
         // From my testing, it seemed like this needed to be set every cycle.
+        // FIXME: test ths again...
```
Flat: Still need testing?

src/voice/input/cached/mod.rs
```diff
@@ -0,0 +1,1533 @@
...
+        let old_pos = self.cache.pos as u64;
+
+        let (valid, new_pos) = match pos {
+            SeekFrom::Current(adj) => {
+                // overflow expected in many cases.
+                let new_pos = old_pos.wrapping_add(adj as u64);
+                (adj >= 0 || (adj.abs() as u64) <= old_pos, new_pos)
+            }
+            SeekFrom::End(adj) => {
+                // FIXME: make this check for metadata as the basis?
```
Flat: Any further plans for this?

Flat (dismissed):
Aside from a few clarifications on documentation looks good on my first pass.

src/voice/connection.rs
FelixMcFelix: Tested: this behaviour is no longer necessary since we started using `audiopus`. Removed and refactored bitrate setting to be less eager/repetitive.

src/voice/input/cached/mod.rs
FelixMcFelix: I think it might be better to do this in the slow (and correct) way we have right now, where the file is loaded to the end first. It will be tricky to defensively program around possibly mis-reported durations etc. Time-based seeking always uses `SeekFrom::Start(..)`, so the other variants are mainly implemented for the sake of completeness.

examples/10_voice_receive/src/main.rs
FelixMcFelix: I've expanded the event contexts to list every field (and include the relevant `model::events` structs). Is this what you were looking for?

Flat: Actually was hoping to clarify that SSRC are synchronization sources for those not familiar with RTP. But breaking those out in the example is a good change either way.

FelixMcFelix: Okay, wrote an extra paragraph explaining this. Thanks!

FelixMcFelix:
Clarified and fixed where appropriate. Thanks!

tomadimitrie:
Hi, when will this be merged? I am writing the voice part of my bot and I wanted to know whether I should use the old API or this new one

FelixMcFelix:
@tomadimitrie In all likelihood, this will be landing after/with async-await. Parts of this API will change to accommodate the new paradigm. If you absolutely need some of these features, then go ahead, but be aware that things might change.

tomadimitrie:
> @tomadimitrie In all likelihood, this will be landing after/with async-await. Parts of this API will change to accommodate the new paradigm. If you absolutely need some of these features, then go ahead, but be aware that things might change.

Got it, thanks:)

vicky5124:
Feature request: Could this rework re-structure the features so the voice gateway and sending the audio itself go separate?
For example, the "voice-gateway" feature will enable the bot to be able to join a voice channel and get events, along with the handler and other related stuff. And then have the "audio" feature, that will add the libsodium and libopus dependencies, along with all the youtube-dl and ffmpeg methods, to be able to play audio on those connections.

Having this features split will allow people who use Lavalink for sending audio, to compile the bot without having libopus and libsodium installed, making the compilation easier (specially cross compilation for opus), and also saving of dependencies that are not going to be used.

FelixMcFelix:
That should be doable. I was intending to break this into a sub-crate when targeting async so that the gateway component could be swapped between serenity and twilight – so adding another gate to remove the audio driver would be useful.

While I need to confirm what lavalink needs from this, since I've never used/looked at it, I'm currently working on making sure the shared caches behave well with AsyncRead/AsyncSeek, so direct work on the PR will be a little delayed.

vicky5124:
lavalink-rs only depends on the library connecting to voice chat and the handler that gets returned, as that's where the token, endpoint and session_id are, which Lavalink requires to play the audio.

FelixMcFelix:
Given the coming async changes, I've duplicated the current state of the rework into [its own branch](https://github.com/FelixMcFelix/serenity/tree/voice-rework-sync-freeze) for the handful of people currently using it.

FelixMcFelix:
image
From some initial testing, I'll be pursuing synchronous input reading, packet mixing, and encoding. The above chart compares an async (on tokio rt-threaded, 4 core) mixer with the old synchronous implementation of that part of the pipeline.
 * Y-Axis measures milliseconds (release mode). The goal is to remain below 20 at all times (50 frames per second). This is mixing one track from ytdl -> ffmpeg, audio-only (bandcamp.com).
 * Right plots are used to zoom in, since Excel doesn't support logscale boxplots (and I didn't feel like gnuplot today).
 * "Load" refers to 85 other tasks with similar/greater workloads -- 850us blocking window at 50Hz -- spawned as tokio tasks (async) or as full threads (sync). Load is added before spawning the measured voice thread/task.
   * I'm not convinced that the extra busywait sync workload is representative here -- the OS might be acting cleverly to hide what's happening. I can't comment on the sync differences too authoritatively.

Main takeaways are that significant compute-heavy async task loads (i.e., other voice threads) increase the risk of deadline slippage significantly, and that synchronous packet prep is cheaper than asynchronous.
Aside from this, synchronous ergonomics (`Read`, `BufRead`, `Seek` vs their async counterparts) are a factor in this decision too, being easier to handle in general, particularly if we want to make custom user inputs/transforms easier to implement.

vicky5124:
Does this PR fix the issue where if a rebalance happens, the audio stops completely, the client losses track of voice connections and the bot eventually disconnects?

FelixMcFelix:
I don't believe I've fixed that, Nitsuga.

I've been looking at the gateway integration somewhat on the subcrate front, and if I had to guess I'd say that [these lines](https://github.com/serenity-rs/serenity/blob/current/src/client/bridge/gateway/shard_runner.rs#L606-L609) may very well be the culprit? I haven't had any rebalances in quite some time, but this almost certainly drops the inner handle and kills the internal tasks.

lucawen:
So... its likely to almost done ? What is missing to just go to merge and release?

Flat:
Code review and testing, in addition to Felix considering the PR as ready.

FelixMcFelix:
First bash at breaking out into a subcrate, needing major cleanup but under the right conditions it works for now.

Main questions I had stuffed into the commit message:
* What happens to the parts in `model::*` which only appear in the voice gateway API? Do we move these into the subcrate? I don't believe twilight handle these events at all, for instance.
* Do we still need the "voice" feature in the core serenity crate? This now gates access only to the voice client trait, and adds Option<...> fields to clients, runners and other managers.

Rationale and user-facing benefits are listed there, too.

FelixMcFelix:
@Flat @acdenisSK consider this done.

# Open Questions

- [ ] CI Integration for voice-model?
- [ ] CI Integration for songbird?
- [ ] Cargo toml authors for songbird?
- [ ] New "serenity-rs/..." repo?

# Post-merge

- [ ] Publish voice-model before serenity
- [ ] Publish songbird after serenity

songbird/examples/twilight/src/main.rs
```diff
@@ -0,0 +1,378 @@
...
+                let _success = handle.pause();
+                false
+            }
+            _ => {
+                let _success = handle.play();
+                true   
+            }
+        };
+
+        let action = if paused { "Unpaused " } else { "Paused" };
```
Erk-:
```suggestion
        let action = if paused { "Unpaused" } else { "Paused" };
```

songbird/examples/twilight/src/main.rs
```diff
@@ -0,0 +1,378 @@
...
+    
+    let content = if let Some(handle) = store.get(&guild_id) {
+        if handle.is_seekable() {
+            let _success = handle.seek_time(std::time::Duration::from_secs(position));
+            format!("Seeked to {}s", position)
+        } else {
+            format!("Track is not compatible with seeking!")
+        }
+    } else {
+        format!("No track to change volume!")
```
Erk-: Should be seek

songbird/src/driver/tasks/mixer.rs
```diff
@@ -0,0 +1,439 @@
...
+            .conn_active
+            .as_mut()
+            .expect("Shouldn't be mixing packets without access to a cipher + UDP dest.");
+
+        let mut nonce = Nonce::default();
+        let index = {
+            let mut rtp = MutableRtpPacket::new(&mut self.packet[..]).expect(
+                "Too few bytes in self.packet for RTP header.\
+                    (Blame: VOICE_PACKET_MAX?)",
+            );
```
Erk-: Is this a fatal error? Also it is one of those that should not happen or should it return a Error instead?

songbird/src/driver/tasks/mixer.rs
```diff
@@ -0,0 +1,439 @@
...
+                payload[TAG_SIZE..TAG_SIZE + len].clone_from_slice(opus_frame);
+                len
+            };
+
+            let final_payload_size = TAG_SIZE + payload_len;
+
+            let tag = conn
+                .cipher
+                .encrypt_in_place_detached(&nonce, b"", &mut payload[TAG_SIZE..final_payload_size])
+                .expect("Encryption failed?");
```
Erk-: Similar here would it not be better to return a error?

songbird/src/driver/tasks/udp_tx.rs
```diff
@@ -0,0 +1,45 @@
...
+};
+use tracing::{error, info, instrument, trace};
+
+#[instrument(skip(udp_msg_rx))]
+pub(crate) async fn runner(udp_msg_rx: Receiver<UdpTxMessage>, ssrc: u32, mut udp_tx: SendHalf) {
+    info!("UDP transmit handle started.");
+
+    let mut keepalive_bytes = [0u8; MutableKeepalivePacket::minimum_packet_size()];
+    let mut ka = MutableKeepalivePacket::new(&mut keepalive_bytes[..])
+        .expect("Insufficient bytes given to keepalive packet.");
```
Erk-: Similar here, is this something that could happen often.

examples/e16_voice_events_queue/src/main.rs
```diff
@@ -0,0 +1,585 @@
+//! Example demonstrating how to make use of individual track audio events,
+//! and how to use the `TrackQueue` system.
+//!
+//! Requires the "cache", "methods", and "voice" features be enabled in your
```
Flat: "standard_framework" instead of "methods"

songbird/Cargo.toml
```diff
@@ -0,0 +1,146 @@
+[package]
+authors = ["Alex M. M. <acdenissk69@gmail.com>"]
```
Flat: Should this not be yourself?

songbird/src/driver/connection/error.rs
```diff
@@ -0,0 +1,105 @@
...
+    CryptoModeUnavailable,
+    /// An indicator that an endpoint URL was invalid.
+    EndpointUrl,
+    /// Discord hello/ready handshake was violated.
+    ExpectedHandshake,
+    /// Discord failed to correctly respond to IP discovery.
+    IllegalDiscoveryResponse,
+    /// Could not parse Discord's view of our IP.
+    IllegalIp,
+    /// Miscellaneous I/O error..
```
Flat: One too many periods.

songbird/src/input/cached/compressed.rs
```diff
@@ -0,0 +1,303 @@
...
+            if out.is_none() && raw_len > 0 {
+                loop {
+                    // NOTE: we don't index by raw_len because the last frame can be too small
+                    // to occupy a "whole packet". Zero-padding is the correct behaviour.
+                    match self
+                        .encoder
+                        .encode_float(&sample_buf[..samples_in_frame], &mut self.last_frame[..])
+                    {
+                        Ok(pkt_len) => {
+                            debug!("Next packet to write has {:?}", pkt_len);
```
Flat: Perhaps change this to trace?

songbird/src/input/cached/compressed.rs
```diff
@@ -0,0 +1,303 @@
...
+                            self.frame_pos = 0;
+                            self.last_frame.truncate(pkt_len);
+                            break;
+                        },
+                        Err(OpusError::Opus(OpusErrorCode::BufferTooSmall)) => {
+                            // If we need more capacity to encode this frame, then take it.
+                            debug!("Resizing inner buffer (+256).");
```
Flat: Same here.

songbird/src/lib.rs
```diff
@@ -0,0 +1,74 @@
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/FelixMcFelix/serenity/voice-rework/songbird/songbird.png",
+    html_favicon_url = "https://raw.githubusercontent.com/FelixMcFelix/serenity/voice-rework/songbird/songbird-ico.png"
+)]
+#![deny(missing_docs)]
+//! ![project logo][logo]
+//!
+//! Songbird is an async, cross-library compatible voice system for Discord, written in Rust. The library offers:
+//!  * A (standalone) gateway frontend compatible with [serenity] and [twilight] using the `"gateway"` and `"[serenity/twilight]-[rustls/native]"` features.
+//!  * An optionally standalone driver for voice calls, via the `"driver"` feature.
```
Erk-: Does this feature make it possible to run it without serenity or twilight? I think the wording could be a bit different if so.

songbird/src/manager.rs
```diff
@@ -0,0 +1,350 @@
...
+                user_id: user_id.into(),
+            }),
+            calls: Default::default(),
+            sharder: Sharder::Twilight(cluster),
+        })
+    }
+
+    /// Set the bot's user, and the number of shards in use.
+    ///
+    /// This *must not* be called more than once, or after any calls have begun.
```
Erk-: Could this be such that it just does not do anything if you run it more than once?

Erk- (changes requested):
Overall looks good had a few nits when I looked through it

songbird/src/lib.rs
FelixMcFelix: The gateway doesn't support any other backend at the moment. It could be made to do so with a `Box<dyn Trait>` in `Shard` etc., but I had to cull the feature creep somewhere.

The driver is entirely standalone. It is possible to run only the driver if you so wish (i.e., you can create a `ConnectionInfo` in your python bot, and somehow pass that into a rust program using this crate to roll-your-own-lavalink). I'll explain this further.

songbird/src/manager.rs
FelixMcFelix: I think so, yes. Will do.

songbird/src/driver/tasks/mixer.rs
FelixMcFelix: This is fatal, in that if the constant VOICE_PACKET_MAX is too small then no packet can ever be constructed, let alone sent (due to the constant requirements on self.packet length).

songbird/src/driver/tasks/mixer.rs
FelixMcFelix: I think this would be better here, yes. I think I looked into when/where errors could appear from the XSalsa/**/ crate at the time we were migrating to the pure rust impl, and judged that encryption should never fail in this case, but I suppose it'll be better to guard against upstream changes here.

songbird/src/driver/tasks/udp_tx.rs
FelixMcFelix: Keepalive packet size should always exactly equal its minimum size to be constructed at all (again, constant length requirement on array). This should remain fatal, I think.

songbird/src/driver/tasks/mixer.rs
```diff
@@ -0,0 +1,439 @@
...
+                use MixerMessage::*;
+
+                match self.mix_rx.try_recv() {
+                    Ok(AddTrack(mut t)) => {
+                        t.source.prep_with_handle(self.async_handle.clone());
+                        let _ = self.add_track(t, &interconnect);
+                    },
+                    Ok(SetTrack(t)) => {
+                        self.tracks.clear();
+                        let _ = interconnect.events.send(EventMessage::RemoveAllTracks);
```
Flat: There are many `let _` in this module. Are these results safe to ignore?

songbird/src/input/ytdl_src.rs
```diff
@@ -0,0 +1,76 @@
...
+        .args(pre_args)
+        .arg("-i")
+        .arg("-")
+        .args(&ffmpeg_args)
+        .stdin(youtube_dl.stdout.ok_or(Error::Stdout)?)
+        .stderr(Stdio::null())
+        .stdout(Stdio::piped())
+        .spawn()?;
+
+    let metadata = Metadata::from_ytdl_uri(uri).await;
```
Flat: Instead of calling youtube-dl twice, once to pipe to ffmpeg and another to fetch metadata it may be better to use youtube-dl to dump the playlist/url to the stream and metadata at the same time so that only one call is made to the remote site.

examples/e16_voice_events_queue/src/main.rs
```diff
@@ -0,0 +1,585 @@
...
+
+    Ok(())
+}
+
+#[command]
+async fn join(ctx: &Context, msg: &Message) -> CommandResult {
+    let guild = match msg.guild(&ctx.cache).await {
+        Some(guild) => guild,
+        None => {
+            check_msg(msg.channel_id.say(&ctx.http, "Groups and DMs not supported").await);
```
Flat: Instead of checking every time you can rule out that it is not a guild channel by using the `#[only_in("guilds")]` trait. and let serenity handle this for you.

Flat (commented):
This was just a high level review until I have proper time to test and dig deeper. Overall looks good to me, a few nits and questions. Thanks for the immense effort in refactoring audio!

songbird/src/manager.rs
FelixMcFelix: Turns out I already made this NOP after an initialisation, just need to update comment.

songbird/src/driver/tasks/mixer.rs
FelixMcFelix: Reworked this now, this did in fact require more thought (due to redesign of packet handling threads some time after the mixer rework).

songbird/src/input/ytdl_src.rs
FelixMcFelix: Is there a natural way of doing this, built into either ffmpeg or ytdl? I understand that ytdl embeds its metadata within the output stream when called normally, but extracting it is another problem:
* Having ytdl tell us what we need requires the "-j" option or similar, preventing any stream output. This led to the current solution.
* ffprobe seems to be required to get in-stream metadata in JSON or other easy-to-parse format from media. This works well with an on-disk input source, but consumes piped output as far as I can tell -- which still needs to go into ffmpeg.

I'd really like it if there's a way to wrangle ffmpeg (or ytdl) to, e.g., put f32 PCM stream output on Stdout and dump (only!) metadata to Stderr, but I'm not nearly familiar enough with them right now to pull that off.

examples/e16_voice_events_queue/src/main.rs
FelixMcFelix: Did not know about this til now, applied to all voice examples.

songbird/src/input/ytdl_src.rs
Flat: You can use `youtube-dl --dump-json` this will give you a dump of all the metadata of the target media. You can then use the `url` field to pass to ffmpeg to play the actual media. You can see how mpv uses this (sort of, it's a lua binding) here https://github.com/mpv-player/mpv/blob/5dcfe32ff21bf560a27db3a75f773f73afdf6dcf/player/lua/ytdl_hook.lua

FelixMcFelix: I feel we did this before (i.e., years ago), and ran into all sorts of terrifying DASH issues where ytdl sources would just stop at random points which ffmpeg just gave up on. I'll try `--print-json`, which seems to do what I want.

FelixMcFelix: Figured it out now!

arqunis (approved):

Flat (approved):
Approving without local testing as this is being merged to the next branch any bugs can be ironed out before release.

arqunis:
I'll merge this as is. The songbird crate will remain in the repository until the 0.10 release, at which point it will be moved to its own repository and properly tested.

PR #828, opened 2020-04-14
Add Watching Activity constructor

sheldonkwok:
I noticed the Watching Activity was missing so I added it. I'm not sure if it was intentionally left out though since there's a clear gap in the Activity enum code

### DIFF
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -163,7 +163,7 @@ impl Activity {
         }
     }
 
-    /// Creates a `Game` struct that appears as a `Listening to <name>` status.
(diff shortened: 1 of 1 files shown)


Arzte:
https://github.com/serenity-rs/serenity/pull/773 as per previous attempts/discussions to add this, the Watching Activity is undocumented and thus has no guarantees from Discord that it won't change tomorrow.

You can see further discussion here on this https://github.com/serenity-rs/serenity/pull/409#issuecomment-427842549

Discord's Stance:
https://github.com/discord/discord-api-docs/issues/463
https://github.com/discordapp/discord-api-docs/issues/978

sheldonkwok:
Ah missed that the issue. Thanks

Issue #815, opened 2020-04-05
Support Intents


Arzte:
Most events bots see they don't care about (Eg presence updates), Discord has introduced an api feature to select which events you want to receive, allowing you to only receive events you actually care about.

https://github.com/discord/discord-api-docs/pull/1307

Lakelezz:
Gateway Intents were added: #836

PR #836, opened 2020-04-18
Add Gateway Intents

Lakelezz:
This adds support for gateway intents, you can specify these on the client via:

```rust
    let mut client = Client::new_with_extras(&token, |f| {
        f.intents(GatewayIntents::GUILDS).event_handler(Handler)
    })?;
```
By default, Serenity does not specify any intents.
Read more about gateway intents here: https://discordapp.com/developers/docs/topics/gateway#gateway-intents

Additionally, adds example 13 showcasing how to use gateway intents.

Closes: #815

### DIFF
--- a/src/client/bridge/gateway/intents.rs
+++ b/src/client/bridge/gateway/intents.rs
@@ -1,0 +1,133 @@
+use bitflags::__impl_bitflags;
+use serde::{
+    de::{Deserialize, Deserializer},
+    ser::{Serialize, Serializer},
+};
+
+#[derive(Copy, PartialEq, Eq, Clone, PartialOrd, Ord, Hash)]
+/// [Gateway Intents] will limit the events your bot will receive via the gateway.
+/// By default, no intents are specified by Serenity.
+///
+/// [Gateway Intents]: https://discordapp.com/developers/docs/topics/gateway#gateway-intents
+pub struct GatewayIntents {
+    /// The flags composing gateway intents.
+    ///
+    /// # Note
+    /// Do not modify this yourself; use the provided methods.
+    /// Do the same when creating, unless you're absolutely certain that you're giving valid intents flags.
+    pub bits: u64
+}
+
+__impl_bitflags! {
+    GatewayIntents: u64 {
+        /// Enables following gateway events:
+        ///
+        /// - GUILD_CREATE
+        /// - GUILD_DELETE
+        /// - GUILD_ROLE_CREATE
+        /// - GUILD_ROLE_UPDATE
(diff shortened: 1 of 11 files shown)


src/client/bridge/gateway/intents.rs
```diff
@@ -0,0 +1,133 @@
...
+use serde::{
+    de::{Deserialize, Deserializer},
+    ser::{Serialize, Serializer},
+};
+
+#[derive(Copy, PartialEq, Eq, Clone, PartialOrd, Ord, Hash)]
+/// [Gateway Intents] will limit the events your bot will receive via the gateway.
+/// By default, no intents are specified by Serenity.
+///
+/// [Gateway Intents]: https://discordapp.com/developers/docs/topics/gateway#gateway-intents
```
arqunis:
```suggestion
/// [Gateway Intents] will limit the events your bot will receive via the gateway.
/// By default, no intents are specified by Serenity.
///
/// [Gateway Intents]: https://discordapp.com/developers/docs/topics/gateway#gateway-intents
#[derive(Copy, PartialEq, Eq, Clone, PartialOrd, Ord, Hash)]
```

src/gateway/shard.rs
```diff
@@ -2,10 +2,11 @@ use crate::constants::{self, close_codes};
 use crate::internal::prelude::*;
 use crate::model::{
     event::{Event, GatewayEvent},
-    gateway::Activity,
+    gateway::{Activity},
```
arqunis: I'm assuming the addition of these braces was an accident?
```suggestion
    gateway::Activity,
```

src/client/bridge/gateway/intents.rs
```diff
@@ -0,0 +1,133 @@
...
+pub struct GatewayIntents {
+    /// The flags composing gateway intents.
+    ///
+    /// # Note
+    /// Do not modify this yourself; use the provided methods.
```
arqunis, line 16: > use the provided methods

There aren't any, however, so this statement is misleading. You should add them in the same way we do for the [`Permissions` flags](https://github.com/serenity-rs/serenity/blob/current/src/model/permissions.rs#L272).

arqunis (changes requested):

src/client/bridge/gateway/intents.rs
Lakelezz: My bad, I forgot that, they will be added of course!

src/gateway/shard.rs
Lakelezz: Yes, that was an accident.

arqunis (approved):

PR #840, opened 2020-04-19
Fix inconsistency in event_handler argument naming

scott-mitchell:


arqunis:
The `current` branch is in a feature-freeze. It only accepts critical bugfixes right now (and for the conceivable future, as part of LTS). Since your change is tiny, and fixes just the documentation a bit, I have manually merged your commit into the `next` branch.

https://github.com/serenity-rs/serenity/commit/622c962f0636af12c1985eb2891bd41ce53a650c

PR #849, opened 2020-04-26
Replace `&mut Context` with `&Context`

### LINKED ISSUE #846: Turn `&mut Context` to `&Context`
As `Context`'s fields offer internal mutability, there is no point in giving out a mutable reference to `Context`.

Th3-M4jor:
Resolves #846 

I think I replaced every instance and every example should build, please do inform if I missed any.

This is a breaking change (obviously)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,18 @@
 All notable changes to this project will be documented in this file.
 This project mostly adheres to [Semantic Versioning][semver].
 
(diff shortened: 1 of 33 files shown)


arqunis (approved):
This looks good to me, but I suggest [removing these two `impl`s as well][impls], as they would be useless if this PR is merged.

[impls]: https://github.com/Th3-M4jor/serenity/blob/next/src/http/mod.rs#L95

PR #853, opened 2020-05-04
Perform an audit on all feature gates in the repository

arqunis:
Many feature gates in the model module are unnecessary, as they're already implied by the `model` feature (it depends on the `http` and `builder` features). Some were entirely redundant. I've removed some of the feature gates for the http module as well. In what world is the `http` feature not enabled in the http module?

I've fixed some of the function declarations. I also noticed that a few methods on `Role` no longer need to depend on the `cache` feature due to the addition of `Role::guild_id`. They've been fixed too, and a method that was used to find the `GuildId` the role belonged to has been deprecated.

This depends on #852 to be merged first.

### DIFF
--- a/src/builder/edit_message.rs
+++ b/src/builder/edit_message.rs
@@ -17,7 +17,7 @@ use std::collections::HashMap;
 /// # #[cfg(feature = "framework")]
 /// # use serenity::framework::standard::{CommandResult, macros::command};
 /// #
-/// # #[cfg(all(feature = "http", feature = "framework"))]
+/// # #[cfg(all(feature = "model", feature = "utils", feature = "framework"))]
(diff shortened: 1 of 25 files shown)


Flat (commented):
Is there a reason for the general addition of `#[inline]` on many functions?

arqunis:
They could benefit from being inlined in users' code. Their bodies are very small -- most being just a http call to the respective endpoint. Which means that after optimizations are performed, the only thing that's left in the generated machine code of their bots is the http call that is required.

Much of the `model` module had been littered with `#[inline]`s anyway, all used in the same reason I have described.

Flat (approved):

Lakelezz (approved):

PR #858, opened 2020-05-06
Add GuildId::disconnect_member

mendess:
Adds the ability to disconnect a member from a voice channel in the guild.

This implementation was based on [this discussion](https://discordapp.com/channels/381880193251409931/381912587505500160/707567000460853339) in the serenity discord.

This was thrown together without much thought so feedback is welcome

### DIFF
--- a/src/builder/edit_member.rs
+++ b/src/builder/edit_member.rs
@@ -77,4 +77,15 @@ impl EditMember {
         let num = Value::Number(Number::from(channel_id.0));
         self.0.insert("channel_id", num);
     }
+
+    /// Disconnects the user from their voice channel if any
+    ///
+    /// Requires the [Move Members] permission.
+    ///
+    /// [Move Members]: ../model/permissions/struct.Permissions.html#associatedconstant.MOVE_MEMBERS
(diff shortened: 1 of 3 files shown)


arqunis:
New features aren't accepted in the `current` branch at the time, please use `next`.

arqunis:
The method should also be replicated for the `GuildChannel` and `Member` structs by using the implementation of `GuildId`.

vicky5124:
If this is going for next, EditMember.voice_channel() should take an Option

mendess:
Ok, will implement all those changes when I get another break

Th3-M4jor:
To throw my two cents in, we can avoid breaking the API in places where we're changing fn's to take Options if we take advantage of the fact that `T` is trivially convertible to `Option<T>`.

For example
```rust
pub fn voice_channel<C: Into<ChannelId>>(&mut self, channel_id: C)
```
could in theory be changed to
```rust
pub fn voice_channel<C: Into<ChannelId>, T: Into<Option<C>>>(&mut self, channel_id: T)
```

mendess:
Closing in favor of #860

PR #860, opened 2020-05-12
Add voice channel disconnect

mendess:
Follow up on pull request #858. Now pointing to the right branch.

I opted not to implement [this](https://github.com/serenity-rs/serenity/pull/858#issuecomment-624653918) as I think having separate functions for these things is better.

Once again feedback welcome

### DIFF
--- a/src/model/guild/guild_id.rs
+++ b/src/model/guild/guild_id.rs
@@ -577,6 +577,32 @@ impl GuildId {
         Some(guild.name.to_string())
     }
 
+    /// Disconnects a member from a voice channel in a guild.
+    ///
+    /// Requires the [Move Members] permission.
+    ///
+    /// [Move Members]: ../permissions/struct.Permissions.html#associatedconstant.MOVE_MEMBERS
+    #[cfg(feature = "http")]
+    #[inline]
+    pub fn disconnect_member<U>(self, http: impl AsRef<Http>, user_id: U) -> Result<()>
+        where U: Into<UserId> {
+        self._disconnect_member(&http, user_id.into())
+    }
+
+    #[cfg(feature = "http")]
+    fn _disconnect_member(
+        self,
+        http: impl AsRef<Http>,
+        user_id: UserId,
+    ) -> Result<()> {
+        let mut map = Map::new();
+        map.insert(
+            "channel_id".to_string(),
+            Value::Null
+        );
+        http.as_ref().edit_member(self.0, user_id.0, &map)
+    }
+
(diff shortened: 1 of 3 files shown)


src/model/guild/guild_id.rs
```diff
@@ -568,6 +568,32 @@ impl GuildId {
         http.as_ref().edit_member(self.0, user_id.0, &map)
     }
 
+    /// Disconnects a member from a voice channel in a guild.
+    ///
+    /// Requires the [Move Members] permission.
+    ///
+    /// [Move Members]: ../permissions/struct.Permissions.html#associatedconstant.MOVE_MEMBERS
+    #[cfg(feature = "http")]
```
arqunis: This feature gate is pointless. The `model` feature depends on the `http` feature, so there is no need to check the latter.
```suggestion

```

src/model/guild/guild_id.rs
```diff
@@ -568,6 +568,32 @@ impl GuildId {
...
+    #[inline]
+    pub fn disconnect_member<U>(self, http: impl AsRef<Http>, user_id: U) -> Result<()>
+        where U: Into<UserId> {
+        self._disconnect_member(&http, user_id.into())
+    }
+
+    #[cfg(feature = "http")]
```
arqunis: Ditto.
```suggestion
```

arqunis (changes requested):

arqunis (approved):

src/model/guild/guild_id.rs
```diff
@@ -568,6 +568,30 @@ impl GuildId {
         http.as_ref().edit_member(self.0, user_id.0, &map)
     }
 
+    /// Disconnects a member from a voice channel in a guild.
```
```suggestion
    /// Disconnects a member from a voice channel in the guild.
```
The a/an article would be appropriate here if the user were to pass a guild id to the method. However, since we're already working on a guild id, it implies which guild the member will be disconnected in.

src/model/guild/member.rs
```diff
@@ -327,6 +327,35 @@ impl Member {
...
+    }
+
+    fn _move_to_voice_channel<H>(&self, http: H, c: ChannelId) -> Result<()>
+    where
+        H: AsRef<Http>
+    {
+        self.guild_id.move_member(http, self.user.read().id, c)
+    }
+
+    /// Disconnects the member their voice channel if any.
```
arqunis:
```suggestion
    /// Disconnects the member from their voice channel if any.
```

arqunis (changes requested):

arqunis (approved):

Issue #804, opened 2020-03-18
Framework Redesign


Lakelezz:
Once #803 lands, it will be interesting to consider a rework of the framework.

Right now, the framework is represented as a trait, allowing everyone to implement their own framework.
However this raises a maintenance problem, there might be portions of the framework a new framework would like to keep. Hereby mentioned portions could be divided into front API, dispatch evaluation, and command dispatch.
The first offers the user an API to use, the second will evaluate whether an attempt to run a command is valid, the last will interpret the final type of command data.

We personally want to develop a new simple framework as an experiment to discover a replacement for the current single framework-trait we have right now. Unlike the standard framework, this simple framework won't require procedural macros at all.
Furthermore, a second framework shall be developed that expands on the implemented portions of the simple framework by allowing to dynamically serialise/deserialise/delete commands, but limits the data format of commands.

We must be aware that serialising fn-pointers is not fully feasible. Serialising and deserialising Rust-code is not planned either. The idea of dynamic commands may be more about the idea of interpretable code.

This issue will serve as a tracker for updates and discussions on this matter.

arqunis:
An example of a simple framework would be:
```rust
use serenity::framework::{SimpleFramework, CommandResult, Args};
// `CommandResult` being an alias to `Result<(), Box<std::error::Error>>`

async fn ping(ctx: Arc<Context>, msg: Arc<Message>, _args: Args) -> CommandResult {
    msg.channel_id.say("pong!").await?;
    Ok(())
}

let mut framework = SimpleFramework::new(|config| config.prefix("!").delimiter(" "));
// SimpleFramework::add(name: &str, f: impl Fn(Arc<Context>, Arc<Message>, Args) -> F) 
//    where F: Future<Output=CommandResult> 
framework.add("ping", ping);
// There will also be an help command implicitly provided, as if people did
// framework.add(help);
// themselves

let mut client = Client::new_with_framework(
     /* DISCORD_TOKEN */, 
     /* EventHandler */, 
    framework
);
```

The purpose of this framework would be, as told by this issue, simple and to the point, without the use of procedural macros. It would have a limited set of configurable options. This is great if people need a command framework up-and-running in a matter of seconds without fiddling with macros or other antics as a result of weird interaction with some options. And, if people require a greater, more powerful framework, they will have an easy choice to switch between simpler frameworks and sophisticated frameworks (cf. the dynamic framework), as the latter would be based off the former. Thus, they can expand the bot as it grows with the least effort.

[...]

PR #861, opened 2020-05-14
[WIP] Simple framework

Th3-M4jor:
Adds a simple framework option to Serenity as mentioned in #804 
Allowing users to use a framework without using procedural macros.

The intent is to have this ready to merge with `next` after `Serenity.await` has been merged as this PR is based on #803 

- [x] initial prototyping and proof of concept
- [x] create default help command
- [x] add feature gates and have shared code with `StandardFramework` where appropriate
- [x] add documentation and examples


src/framework/simple/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+                self.run_cmd(&ctx, &msg, &cmd_name, &tokens[1..]).await;
+
+            }
+
+        } else if let Some(normal_msg_fn) = &self.normal_message_fn {
+            normal_msg_fn(&ctx, &msg).await;
+        }
+        
+    }
+}
```
arqunis: The arguments could be handled better, but just to make a quick and efficient improvement to your implementation:
```suggestion
    if !msg.content.starts_with(&self.prefix) {        
        if let Some(normal_message) = &self.normal_message_fn {
            normal_message(&ctx, &msg).await;
        }
    
        return;
    }
    
    let cmd_and_args = &msg.content[self.prefix.len()..];
    
    if cmd_and_args.is_empty() {
        if let Some(prefix_only) = &self.prefix_only_cmd {
            prefix_only(&ctx, &msg).await;
        }
        
        return;
    }
    
    let tokens = cmd_and_args.split(&self.delimiter).collect::<Vec<&str>>();
    
    let cmd_name = if self.case_insensitive {
        Cow::Owned(tokens[0].to_lowercase())
    } else {
        Cow::Borrowed(&tokens[0])
    };
    
    self.run_cmd(&ctx, &msg, &cmd_name, &tokens[1..]).await;
```

This requires to import `Cow` as to avoid allocating the command name on every invocation if the user did not enable case insensitivity.

arqunis (changes requested):

src/framework/simple/mod.rs
Th3-M4jor: I went ahead and rewrote to use the Args struct within the standard framework, and moved it to a shared module

src/framework/simple/mod.rs
```diff
@@ -149,35 +156,52 @@ impl SimpleFramework {
...
 
-                if let Some(prefix_only) = &self.prefix_only_cmd {
-                    prefix_only(&ctx, &msg).await;
-                    return;
-                }
+        let text = msg.content.trim_start();
 
+        if !text.starts_with(&self.prefix) {
+            self.run_normal_message(&ctx, &msg).await;
+        } else {
```
arqunis: I prefer flat code when inside of a function. What I mean by "flat" code is that there's minimal indentation, so it doesn't vigorously stretch to the right; otherwise we get ["hadouken" code](https://external-preview.redd.it/6TTm4tS5pMCiq2KlgviubpGKt1QzvetOi2Z6Z3lB1Nc.jpg?auto=webp&s=c1df961d84c8d76048a93c1d261e425d3737d779) if left unchecked. Early returns can help avoid the latter. They also split the erroneous cases from the successful case, which will happen at the end of the function.
```suggestion
        if !text.starts_with(&self.prefix) {
            self.run_normal_message(&ctx, &msg).await;
            return;
        }
```

src/framework/simple/mod.rs
```diff
@@ -149,35 +156,52 @@ impl SimpleFramework {
...
+            let cmd_and_args = &text[self.prefix.len()..];
+            if cmd_and_args.is_empty() {
+                self.run_prefix_only_cmd(&ctx, &msg).await;
```
arqunis:
```suggestion
        let cmd_and_args = &text[self.prefix.len()..];
        if cmd_and_args.is_empty() {
            self.run_prefix_only_cmd(&ctx, &msg).await;
            return;
        }
```

arqunis (changes requested):

Th3-M4jor:
I believe this is more or less done.
Further, the method I used to `Box::pin` futures, could probably be used to remove the need for the `#[hook]` procedural macro within the `StandardFramework`

tdonovan4:
Would you consider making the traits defined in [simple/traits.rs](https://github.com/Th3-M4jor/serenity/blob/16ff58568981e9cb75ede0f72bd748d49f9e7f17/src/framework/simple/traits.rs) public? 

I wanted to try this framework by implementing another on top of it but I ran into a problem. Basically, I'm trying to make a function that takes a command's function and passes it to the inner SimpleFramework using the `add()` method.  

When trying to store the command's function, I had the same issue with lifetimes as you and so I implemented my own version of AsyncFn3. However, since both traits are different, I can't pass the function to `add()`. 

I made [an example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=94fea6669402f4e5cdaff348dbd0a645) on the Rust playground to illustrate this. 

I tried various methods to try to fix this issue, but I couldn't find a solution (lifetimes using HRTBs are hard). It might be possible, but at this point, I think making the traits public is the best solution to my problem.

Overall, I think the simple framework could be a great way to remove boilerplate code when trying to create a custom framework.

Th3-M4jor:
Done 👍

Th3-M4jor:
Closing this so as to rebase from new `await_next` branch. Determined it was simpler to copy changes over than resolve all the merge conflicts

PR #867, opened 2020-05-18
Fixed unrecognised_command() not working if the last group added has prefixes.

vicky5124:
```
Fixed unrecognised_command() not triggering when the last group in the standard framework has a prefix.

	modified:   src/framework/standard/parse/mod.rs
```

### DIFF
--- a/src/framework/standard/parse/mod.rs
+++ b/src/framework/standard/parse/mod.rs
@@ -373,8 +373,8 @@ pub fn command(
                 if res.is_ok() {
                     check_discrepancy(ctx, msg, config, &group.options)?;
 
-                    return res;
                     is_prefixless = true;
+                    return res;
                 }
 
                 last = res;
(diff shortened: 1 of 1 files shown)


src/framework/standard/parse/mod.rs
```diff
@@ -371,6 +374,7 @@ pub fn command(
                     check_discrepancy(ctx, msg, config, &group.options)?;
 
                     return res;
+                    is_prefixless = true;
```
arqunis: The assignment ought to appear before the return statement. It's unreachable code otherwise.
```md
warning: unreachable statement
--> src/framework/standard/parse/mod.rs:377:21
|
|                     return res;
|                     ---------- any code following this expression is unreachable
|                     is_prefixless = true;
|                     ^^^^^^^^^^^^^^^^^^^^^ unreachable statement
|
= note: `#[warn(unreachable_code)]` on by default
```

```suggestion
                    is_prefixless = true;
                    return res;
```

arqunis (changes requested):

arqunis (approved):

PR #869, opened 2020-05-19
Mention projects that extend the functionality of the library instead

arqunis:
The only relation we have with the "related projects" is that they also wrap the Discord API in their target languages. This section really only serves one purpose: to acknowledge the existence of these wrappers for other languages. Sure, we raise awareness of them but it doesn't help the consumers of Serenity. Hence, we have decided to replace with another section that might be of use to them.

For now, we're only mentioning the `serenity-lavalink` project, which may be particularly useful to users playing audio/music with Serenity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ Voice + youtube-dl:
 
 # Projects extending Serenity
 
-- [serenity-lavalink][project:serenity-lavalink] - A wrapper around [Lavalink][repo:lavalink], an audio sending node based on [Lavaplayer][repo:lavaplayer]
+- [serenity-lavalink][project:serenity-lavalink] - An interface to [Lavalink][repo:lavalink], an audio sending node based on [Lavaplayer][repo:lavaplayer]
 
 [`Cache`]: https://docs.rs/serenity/*/serenity/cache/struct.Cache.html
 [`Client::new`]: https://docs.rs/serenity/*/serenity/client/struct.Client.html#method.new
--- a/README.md
+++ b/README.md
@@ -192,14 +192,9 @@ Voice + youtube-dl:
 
 - youtube-dl (Arch: `community/youtube-dl`)
 
-# Related Projects
+# Projects extending Serenity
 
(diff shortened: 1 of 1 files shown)


Lakelezz:
I think adding a brief description per listed item would be a great addition!

README.md
```diff
@@ -192,14 +192,9 @@ Voice + youtube-dl:
...
-# Related Projects
+# Projects extending Serenity
 
-- [Discord.net][library:Discord.net]
-- [JDA][library:JDA]
-- [disco][library:disco]
-- [discordrb][library:discordrb]
-- [discord.js][library:discord.js]
-- [discord.py][library:discord.py]
+- [serenity-lavalink][project:serenity-lavalink] - A wrapper around [Lavalink][repo:lavalink], an audio sending node based on [Lavaplayer][repo:lavaplayer]
```
Flat: I believe it would be better to describe it as an interface to Lavalink. Lavalink is a stand alone server so the only way to interface with it it through its network API.

Flat (approved):

README.md
```diff
@@ -192,14 +192,9 @@ Voice + youtube-dl:
...
-# Related Projects
+# Projects extending Serenity
 
-- [Discord.net][library:Discord.net]
-- [JDA][library:JDA]
-- [disco][library:disco]
-- [discordrb][library:discordrb]
-- [discord.js][library:discord.js]
-- [discord.py][library:discord.py]
+- [serenity-lavalink][project:serenity-lavalink] - An interface to [Lavalink][repo:lavalink], an audio sending node based on [Lavaplayer][repo:lavaplayer]
```
Lakelezz: The readme uses `:` mostly, so I would prefer if we change this.
```suggestion
- [serenity-lavalink][project:serenity-lavalink]: An interface to [Lavalink][repo:lavalink], an audio sending node based on [Lavaplayer][repo:lavaplayer]
```

Lakelezz (approved):

PR #872, opened 2020-05-24
Add with-nickname mention filtering out

### LINKED ISSUE #850: Message::content_safe not replacing <@!...> user mentions
https://github.com/serenity-rs/serenity/blob/620faa2c5dc371ece8d38a4cd571f96084e89b5a/src/model/channel/message.rs#L303

`Message::content_safe` currently replaces whatever `UserId::mention` returns in
https://github.com/serenity-rs/serenity/blob/620faa2c5dc371ece8d38a4cd571f96084e89b5a/src/model/misc.rs#L81
However, `<@!...>` user mentions will not be replaced.

ks129:
Nicknames will result that mention have `!` after `@` and now
`content_safe` filters that out too.

Closes #850

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -300,7 +300,14 @@ impl Message {
             at_distinct.push_str(&u.name);
             at_distinct.push('#');
             let _ = write!(at_distinct, "{:04}", u.discriminator);
-            result = result.replace(&u.mention(), &at_distinct);
+            let mut m = u.mention();
+            // Find out is there need to add `!` to mention.
+            if result.contains(&m) {
+                result = result.replace(&m, &at_distinct);
+            } else {
+                m.insert(2, '!');
+                result = result.replace(&m, &at_distinct);
+            }
         }
 
         // Then replace all role mentions.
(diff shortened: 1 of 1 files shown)


src/model/channel/message.rs
```diff
@@ -300,7 +300,14 @@ impl Message {
...
             let _ = write!(at_distinct, "{:04}", u.discriminator);
-            result = result.replace(&u.mention(), &at_distinct);
+            let mut m = u.mention();
+            // Find out is there need to add `!` to mention.
+            if result.contains(&m) {
+                result = result.replace(&m, &at_distinct);
+            } else {
+                m.insert(2, '!');
+                result = result.replace(&m, &at_distinct);
+            }
```
```suggestion
            // Check whether we're replacing a nickname mention or a normal mention.
            // `UserId::mention` returns a normal mention. If it isn't present in the message, it's a nickname mention.
            if !result.contains(&m) {
                m.insert(2, '!');
            }

            result = result.replace(&m, &at_distinct);
```

arqunis (changes requested):

ks129:
@acdenisSK Applied your suggestion now.

PR #877, opened 2020-06-03
Add pagination for CurrentUser::guilds

NilsIrl:


src/model/user.rs
```diff
@@ -166,7 +166,19 @@ impl CurrentUser {
     /// ```
     #[inline]
```
arqunis:
```suggestion

```
This should no longer be marked as `#[inline]`. I fear it would cause more harm than good.
It's a beneficial attribute for small functions, as it permits them to be inlined across crate boundaries. But if they're longer than three statements, then I believe it shouldn't apply to them anymore.

arqunis (changes requested):

src/model/user.rs
NilsIrl: Fixed

arqunis (approved):

PR #878, opened 2020-06-03
Get guilds limit optional

NilsIrl:
Depends on #877 

Reflects the optional nature of the parameter in the discord API: https://discord.com/developers/docs/resources/user#get-current-user-guilds

This is more of an "RFC" than anything. Maybe it would be preferable to have a different API. e.g., Have `get_guilds` and `get_guilds_with_limit` etc...

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1175,7 +1175,7 @@ impl Http {
     ///
     /// let guild_id = GuildId(81384788765712384);
     ///
-    /// let guilds = http.as_ref().get_guilds(&GuildPagination::After(guild_id), 10).unwrap();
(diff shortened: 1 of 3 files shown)


Th3-M4jor:
It might be better to have two different functions simply for a cleaner API, although you could also leverage the fact that T is trivially convertible to Option<T> to the same effect, but with a more complex function signature

arqunis:
I agree with @Th3-M4jor's first suggestion more. The current `get_guilds` method should be renamed to `get_guilds_with_limit` and a new `get_guilds` method should be put in its place without the limit.

NilsIrl:
In terms of pagination, should we have a different function? Or user `Option` or add another member to `GuildPagination` that doesn't specify pagination?

Issue #865, opened 2020-05-16
Examples cannot be built with either beta or nightly cargo


arqunis:
Cargo in both beta and nightly versions now emits an error claiming that package names cannot start with a digit.
```markdown
error: failed to parse manifest at `/home/alex/Projects/serenity/examples/01_basic_ping_bot/Cargo.toml`

Caused by:
  the name `01_basic_ping_bot` cannot be used as a package name, the name cannot start with a digit
``` 

This does not, currently, affect stable cargo.

arqunis:
Fixed by https://github.com/serenity-rs/serenity/pull/882 and https://github.com/serenity-rs/serenity/pull/883.

PR #882, opened 2020-06-11
Prepend 'e' to all example names

arqunis:
Stable cargo now disallows package names starting with a digit, thus preventing our CI from compiling the examples. Fixes https://github.com/serenity-rs/serenity/issues/865.


examples/e03_struct_utilities/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "04_message_builder"
```
arqunis, line 2: Ignore the `Cargo.toml` diffs for these two examples. I do not know why git swapped the files and considers them as one "renamed" to the other.

Erk- (approved):

PR #884, opened 2020-06-11
voice: use `xsalsa20poly1305` crate for encryption

tarcieri:
This commit replaces the `sodiumoxide` crate, a wrapper for the libsodium C library, with the pure Rust `xsalsa20poly1305` crate from the RustCrypto project:

https://github.com/RustCrypto/AEADs/tree/master/xsalsa20poly1305

This should reduce compile times as `libsodium` is a somewhat large C library and you are only using one algorithm from it. The `xsalsa20poly1305` crate, on the other hand, is a lightweight pure Rust dependency that implements only the algorithm you are using.

Note that the advertised MSRV of this crate is 1.41+, however earlier today new release of the dependency causing that, `generic-array`, reduced the MSRV to 1.36.0:

https://github.com/fizyk20/generic-array/pull/102

You can read more about RustCrypto and our recent releases in this post:

https://www.reddit.com/r/rust/comments/h0em5n/ann_new_rustcrypto_releases_aead_blockcipher/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,12 +68,6 @@ version = "0.7"
 optional = true
 version = "0.16"
 
-[dependencies.sodiumoxide]
-default-features = false
-features = ["std"]
-optional = true
-version = "0.2"
-
 [dependencies.threadpool]
 optional = true
 version = "1"
@@ -98,6 +92,10 @@ version = "0.21"
 [dependencies.webpki-roots]
 optional = true
 version = "0.18"
+
+[dependencies.xsalsa20poly1305]
+optional = true
+version = "0.4.2"
 
 [dev-dependencies.http_crate]
 version = "0.2"
@@ -154,7 +152,7 @@ native_tls_backend = ["reqwest/native-tls", "tungstenite/tls"]
 model = ["builder", "http"]
 standard_framework = ["framework", "uwl", "command_attr", "static_assertions"]
 utils = ["base64"]
-voice = ["byteorder", "gateway", "audiopus", "rand", "sodiumoxide"]
+voice = ["byteorder", "gateway", "audiopus", "rand", "xsalsa20poly1305"]
 
 [package.metadata.docs.rs]
 all-features = true
--- a/src/voice/connection.rs
+++ b/src/voice/connection.rs
@@ -26,7 +26,6 @@ use audiopus::{
 use parking_lot::Mutex;
 use rand::random;
 use serde::Deserialize;
-use sodiumoxide::crypto::secretbox::{self, Key, Nonce};
 use std::{
     collections::HashMap,
     io::Write,
@@ -45,6 +44,10 @@ use std::{
         JoinHandle
     },
     time::Duration
+};
+use xsalsa20poly1305::{
+    aead::{Aead, NewAead},
+    Key, Nonce, XSalsa20Poly1305, KEY_SIZE,
 };
 
 use super::audio::{AudioReceiver, AudioType, HEADER_LEN, SAMPLE_RATE, DEFAULT_BITRATE, LockedAudio};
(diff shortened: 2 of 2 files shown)


src/voice/connection.rs
```diff
@@ -676,8 +678,12 @@ fn encryption_key(client: &mut WsClient) -> Result<Key> {
...
                 }
 
-                return Key::from_slice(&desc.secret_key)
-                    .ok_or(Error::Voice(VoiceError::KeyGen));
+                if desc.secret_key.len() == KEY_SIZE {
+                    let key = Key::from_slice(&desc.secret_key);
+                    return Ok(XSalsa20Poly1305::new(key));
+                } else {
+                    return Err(Error::Voice(VoiceError::KeyGen));
+                }
```
tarcieri, line 689: This pattern is unfortunate. Many of our other RustCrypto crates have a `new_var` method for this purpose. I'll look into getting that added to the upstream trait.

tarcieri: Added a slice-based cipher init API here, although it might be awhile before we ship it:

https://github.com/RustCrypto/traits/pull/191

If you'd like I can add a TODO

Cargo.toml
```diff
@@ -99,6 +93,10 @@ version = "0.21"
 optional = true
 version = "0.18"
 
+[dependencies.xsalsa20poly1305]
+optional = true
+version = "0.4.2"
```
FelixMcFelix, line 98:
```suggestion
version = "0.4"
```
This may be better, unless the new crate is expected to introduce breaking changes in patch versions.

src/voice/connection.rs
```diff
@@ -602,7 +604,7 @@ impl Connection {
 
         let crypted = {
             let slice = &packet[HEADER_LEN..HEADER_LEN + len];
-            secretbox::seal(slice, &nonce, &self.key)
+            self.cipher.encrypt(&nonce, slice).unwrap()
```
FelixMcFelix: Could you please change this to an `expect` if encryption is never actually able to fail?
```suggestion
            self.cipher.encrypt(&nonce, slice)
                .expect("[Voice] Couldn't encrypt packet.")
```
Looking at `aead`'s `Error` type doesn't reveal much about this, so I expect you know more (i.e., if this is only "fallible" to fit the `Aead` interface). Adding a variant to "voice/error.rs" and using `?` would be a good idea if otherwise.

src/voice/connection.rs
FelixMcFelix: This seems fine for now.

FelixMcFelix (changes requested):
Thanks for the PR, looks good (and should remove occasional external library issues to boot).

Other than a few nits, it's a welcome change; API mostly matches libsodium (including in-place/detached modes, which we're using in the rewrite, so we can migrate that too). I don't believe this is a breaking change since the crate is entirely internal, and the minimum compiler version is now below `current`'s 1.39. Will test manually in the morning since CI for examples is broken on `current`.

Cargo.toml
tarcieri: Normally I agree, but in writing this PR I actually added a few features it's using. See the [CHANGELOG.md](https://github.com/RustCrypto/AEADs/blob/master/xsalsa20poly1305/CHANGELOG.md).

Cargo will grab the latest version, but note if it were ever to be yanked, the earlier versions are incompatible.

src/voice/connection.rs
FelixMcFelix: Sorry, this went through while I was reviewing. A TODO/FIXME would be welcome then, thanks.

src/voice/connection.rs
tarcieri: There's presently only one possible runtime error in the current implementation:

https://github.com/RustCrypto/AEADs/blob/master/xsalsa20poly1305/src/lib.rs#L286-L289

The current implementation can't trigger it, and in short, you'd really have to go out of your way to trigger it. If you ever did, encryption would always fail 100% of the time.

That said, I can still change it to `expect` if you like for futureproofing.

src/voice/connection.rs
tarcieri: Added

FelixMcFelix (approved):

FelixMcFelix:
Tested locally, runs fine for me. Thanks!

PR #885, opened 2020-06-12
Message::react: return the created Reaction.

meithecatte:
`Reaction::user` has grown a bit. Should it still be `#[inline]`?


arqunis:
It no longer should be. Could you also fix the example to reflect your changes?

meithecatte:
Sorry for that, the CI passes now.

Issue #886, opened 2020-06-17
A `get_link` method for Message


bakcxoj:
The `Message` struct contains enough information to generate a message link string.

The code could look something like this:

```rust
let guild_id = self.guild_id
    .map(|id| id.0.to_string())
    .unwrap_or(String::from("@me"));

format!("https://discord.com/channels/{}/{}/{}", guild_id, self.channel_id.0, self.id.0)
```

_I don't know if this code works. It's just a proof of concept. That's why this is an issue and not a PR_

bakcxoj:
Closing as resolved

PR #887, opened 2020-06-18
Add a method to retrieve a link to a message

arqunis:
Closes https://github.com/serenity-rs/serenity/issues/886.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -661,6 +661,16 @@ impl Message {
         self.guild_id.as_ref().and_then(|guild_id| self.author.nick_in(cache_http, *guild_id))
     }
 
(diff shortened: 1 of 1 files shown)


bakcxoj (approved):
Not a fan of `https://discord.com/channels` being duplicated, but otherwise LGTM!

Flat (approved):

Flat:
The alternative would be having a variable just for the domain/channels which IMO makes the code less clear and readable just to avoid a small duplication of a static string.

arqunis:
Aside from that, it would also mean a separate allocation just for the guild id or "@me" string, which is the reason I went for the duplication. If it's possible, reduce allocations to a minimum.

PR #889, opened 2020-06-20
Added url() method to Webhook.

vicky5124:
```	
modified:   src/model/webhook.rs
```
non-breaking change.
Adds a method to Webhook to return a valid url.

### DIFF
--- a/src/model/webhook.rs
+++ b/src/model/webhook.rs
@@ -237,7 +237,7 @@ impl Webhook {
 
     /// Returns the url of the webhook.
     ///
-    /// ```rust,no_run
+    /// ```rust,ignore
     /// assert_eq!(hook.url(), "https://discord.com/api/webhooks/245037420704169985/ig5AO-wdVWpCBtUUMxmgsWryqgsW3DChbKYOINftJ4DCrUbnkedoYZD0VOH1QLr-S3sV")
     /// ```
     pub fn url(&self) -> String {
--- a/src/model/webhook.rs
+++ b/src/model/webhook.rs
@@ -238,7 +238,7 @@ impl Webhook {
     /// Returns the url of the webhook.
     ///
     /// ```rust,no_run
-    /// assert_eq!(hook.url(), "https://discord.com/api/webhooks/245037420704169985/ig5AO-wdVWpCBtUUMxmgsWryqgsW3DChbKYOINftJ4DCrUbnkedoYZD0VOH1QLr-S3sV".to_string())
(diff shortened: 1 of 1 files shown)


src/model/webhook.rs
```diff
@@ -234,6 +234,15 @@ impl Webhook {
             Err(why) => Err(why),
         }
     }
+
+    /// Returns the url of the webhook.
+    ///
+    /// ```rust,no_run
+    /// assert_eq!(hook.url(), "https://discord.com/api/webhooks/245037420704169985/ig5AO-wdVWpCBtUUMxmgsWryqgsW3DChbKYOINftJ4DCrUbnkedoYZD0VOH1QLr-S3sV".to_string())
```
arqunis:
```suggestion
    /// assert_eq!(hook.url(), "https://discord.com/api/webhooks/245037420704169985/ig5AO-wdVWpCBtUUMxmgsWryqgsW3DChbKYOINftJ4DCrUbnkedoYZD0VOH1QLr-S3sV")
```
[`String` implements equality comparisons to `&str`](https://doc.rust-lang.org/nightly/std/string/struct.String.html#impl-PartialEq%3C%26%27a%20str%3E), you don't need to elevate the literal to a `String` before being able to use it in the `assert_eq!` macro.

arqunis (changes requested):

src/model/webhook.rs
```diff
@@ -234,6 +234,15 @@ impl Webhook {
             Err(why) => Err(why),
         }
     }
+
+    /// Returns the url of the webhook.
+    ///
+    /// ```rust,no_run
```
arqunis:
```suggestion
    /// ```rust,ignore
```
`no_run` tells rustdoc to disallow running the doctest, but still compile it. Obviously from CI, your example fails because `hook` is a nonexistent variable. The `ignore` directive tells rustdoc to not compile the doctest at all, so you wouldn't have to worry about errors.

If you don't like this suggestion, then there's another method that is already employed in this file. Create a fake `Webhook` instance using `Http`:
```suggestion
    /// ```rust,no_run
    /// # use serenity::http::Http;
    /// # use std::sync::Arc;
    /// #
    /// # let http = Arc::new(Http::default());
    /// let id = 245037420704169985;
    /// let token = "ig5AO-wdVWpCBtUUMxmgsWryqgsW3DChbKYOINftJ4DCrUbnkedoYZD0VOH1QLr-S3sV";
    /// 
    /// let hook = http.as_ref().get_webhook_with_token(id, token)
    ///     .expect("valid webhook");
```

arqunis (changes requested):

arqunis (approved):

PR #891, opened 2020-06-22
Support animated emojis in Emoji#fmt

zack37:


src/model/guild/emoji.rs
```diff
@@ -223,7 +223,11 @@ impl Display for Emoji {
     ///
     /// This is in the format of: `<:NAME:EMOJI_ID>`.
```
arqunis: This should mention the format for animated emojis.
```suggestion
    /// This is in the format of either `<:NAME:EMOJI_ID>` for normal emojis, 
    /// or `<a:NAME:EMOJI_ID>` for animated emojis.
```

arqunis (changes requested):
Thanks for your contribution! It's fine, but I'd like just one small change before merging this.

arqunis (approved):

PR #897, opened 2020-06-27
Update dependencies rustls, tungstenite, webpki-roots, parking_lot, and base64 

johnchildren:
## Motivation

Prevents depdenency on two different versions of rustls and http. Previously there were two versions of these crates when compiling this project when compiled/resolved today. Have attached some graphs generated with `cargo deps --all-deps | dot -Tpng > graph.png` before and after this change.

deps_before

deps_after

cargo build --all-features has 206 compilation units now vs 210 before this change.

## Notes

[`tungstenite::client::connect`](https://docs.rs/tungstenite/0.11.0/tungstenite/client/fn.connect.html) now uses a [`IntoClientRequest`](https://docs.rs/tungstenite/0.11.0/tungstenite/client/trait.IntoClientRequest.html), so the import of `tungstenite::handshake::client::Request` is no longer necessary.

The `url::Url` import in `src/internal/ws_impl.rs` was giving a warning, hence the addition of the extra configuration macro.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,7 +66,7 @@ version = "0.7"
 
 [dependencies.rustls]
 optional = true
-version = "0.16"
+version = "0.17"
 
 [dependencies.threadpool]
 optional = true
@@ -75,7 +75,7 @@ version = "1"
 [dependencies.tungstenite]
 default-features = false
 optional = true
-version = "0.9"
+version = "0.11"
 
 [dependencies.typemap]
 optional = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 [dependencies]
 bitflags = "1"
 log = "0.4"
-parking_lot = "0.9"
+parking_lot = "0.11"
 serde_json = "1"
 
 [dependencies.static_assertions]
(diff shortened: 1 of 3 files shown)


Erk-:
You should probably also update webpki-roots in this pr as well that may remove another few dependencies

johnchildren:
> 
> 
> You should probably also update webpki-roots in this pr as well that may remove another few dependencies

Good spot, here's the updated graph after this change:

graph

meithecatte:
`parking_lot`, `audiopus` and `base64` are also out of date. Is there a reason not to bump them?

johnchildren:
> 
> 
> `parking_lot`, `audiopus` and `base64` are also out of date. Is there a reason not to bump them?

Looks like `parking_lot` and `base64` are easy enough to bump without breaking anything but `audiopus` seems to be being updated in this PR: https://github.com/serenity-rs/serenity/pull/803

Updating it seems to introduce non-trivial breaking changes related to the new API which I don't want to duplicate work.

But otherwise, good idea, have updated the two crates mentioned.

arqunis (approved):

Flat (approved):

PR #901, opened 2020-07-05
Add nonce field to guild members chunk

mjsir911:
As per [the docs](https://discord.com/developers/docs/topics/gateway#guild-members-chunk) there is a nonce field in the `GuildMembersChunkEvent` object.

This PR is adding the ability to use it.

Pulled from [terminal-discord/serenity](https://github.com/terminal-discord/serenity), these changes are made to use serenity in [terminal-discord/weechat-discord](https://github.com/terminal-discord/weechat-discord)

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -495,7 +495,7 @@ fn handle_event(
             let event_handler = Arc::clone(event_handler);
 
             threadpool.execute(move || {
-                event_handler.guild_members_chunk(context, event.guild_id, event.members);
+                event_handler.guild_members_chunk(context, event.guild_id, event.members, event.nonce);
             });
         },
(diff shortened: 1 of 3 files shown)


arqunis:
This needs to be redirected to the `next` branch. At its present state, `current` does not accept feature additions. Additionally, this is also a breaking change to the `EventHandler` interface.

mjsir911:
> Additionally, this is also a breaking change to the `EventHandler` interface.

I'm not sure what you mean here, could you elaborate? Is there anything actionable I need to do?

arqunis:
I was expanding on my reasoning to switch the target branch to `next`. If anyone has implemented a listener to the `guild_members_chunk` event, your change will break their code because they are going to have incompatible function declarations. 

Just switching the branch is enough, you don't have to do anything else.

arqunis (approved):

PR #903, opened 2020-07-05
Fix check behaviour in the help command

jmgao:
Four vaguely related commits in one pull request:

* Fix reversed strikethrough_commands_tip_in_{dm,guild}.
* Respect owner_privilege in help.
* ~Expose CustomisedHelpData fields.~
* ~Allow deletion of help command strikethrough text.~

I think the first three should be mostly uncontroversial, but the last one changes behavior, so maybe it'd need to go into `next`? Let me know if you want me to move commits around.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -395,15 +395,17 @@ fn check_command_behaviour(
     let b = check_common_behaviour(&ctx, msg, &options, owners, help_options);
 
     if b == HelpBehaviour::Nothing {
-       for check in options.checks {
-           if !check.check_in_help {
-               break;
-           }
+       if !options.owner_privilege || !owners.contains(&msg.author.id) {
+           for check in options.checks {
+               if !check.check_in_help {
+                   continue;
+               }
 
-           let mut args = Args::new("", &[]);
+               let mut args = Args::new("", &[]);
 
-           if let CheckResult::Failure(_) = (check.function)(ctx, msg, &mut args, options) {
-               return help_options.lacking_conditions;
+               if let CheckResult::Failure(_) = (check.function)(ctx, msg, &mut args, options) {
+                   return help_options.lacking_conditions;
+               }
            }
        }
     }
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -897,9 +897,9 @@ pub fn create_customised_help_data<'a>(
     }
 
     let strikethrough_command_tip = if msg.is_private() {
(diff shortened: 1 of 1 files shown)


arqunis:
> but the last one changes behavior, so maybe it'd need to go into `next`? 

Absolutely. Furthermore, you're exposing a private api, thereby constituting the addition of a feature. `current` is in a feature-freeze, it only accepts bug fixes at the moment. So `next` is more appropriate.

Addressing the last commit, you have made `#[strikethrough_commands_tip_in_guild(false)]`/`#[strikethrough_commands_tip_in_dm(false)]` worthless. I made `false` as a method of preventing the generation of the default tip text, but it would mean nothing once I merge this. I believe it's a good change, though. Out of all the attributes defined in `command_attr`, the strikethrough attributes have the most complicated way of parsing and interpreting their input.

You should [alter this impl](https://github.com/serenity-rs/serenity/blob/current/command_attr/src/attributes.rs#L238-L259) to no longer accept a boolean and *just* optionally parse a string. Here's what I mean:
```rust
impl AttributeOption for Option<String> {
    fn parse(values: Values) -> Result<Self> {
        validate(&values, &[ValueKind::Name, ValueKind::Equals, ValueKind::SingleList])?;

        Ok(values.literals.get(0).map(|l| l.to_str()))
    }
}
```

This way, if the user doesn't pass anything to the attributes, they will be provided with the default generated tip text. Otherwise, they will override it with their own text. `#[strikethrough_commands_tip_in_guild("")]`/`#[strikethrough_commands_tip_in_dm("")]` would mean no tip text at all.

src/framework/standard/help_commands.rs
```diff
@@ -395,15 +395,17 @@ fn check_command_behaviour(
...
 
     if b == HelpBehaviour::Nothing {
-       for check in options.checks {
-           if !check.check_in_help {
-               break;
-           }
+       if !options.owner_privilege || !owners.contains(&msg.author.id) {
+           for check in options.checks {
+               if !check.check_in_help {
+                   break;
```
arqunis: This should also be fixed.
```suggestion
                   continue;
```
It's a bug to stop executing more checks if one check does not desire to be applied to the help command.

arqunis (changes requested):

src/framework/standard/help_commands.rs
jmgao: Done.

jmgao:
Got it, I'll open a separate PR to next for those commits.

arqunis (approved):

PR #904, opened 2020-07-05
Expose CustomisedHelpData fields, improve help strikethrough message attribute.

jmgao:


src/framework/standard/help_commands.rs
```diff
@@ -110,38 +110,38 @@ macro_rules! warn_about_failed_send {
...
-    checks: Vec<String>,
+    pub name: &'static str,
+    pub group_name: &'static str,
+    pub group_prefixes: &'a [&'static str],
+    pub aliases: Vec<&'static str>,
+    pub availability: &'a str,
+    pub description: Option<&'static str>,
+    pub usage: Option<&'static str>,
+    pub usage_sample: Vec<&'static str>,
+    pub checks: Vec<String>,
```
arqunis, line 138:
```suggestion
    pub checks: Vec<String>,
    pub(crate) _nonexhaustive: (),
```
This permits future fields to be added without breaking users' code when they depend on `Command`.

You don't have to do this for the other two. I doubt they're going to change frequently.

arqunis (changes requested):

src/framework/standard/help_commands.rs
jmgao: Done.

arqunis (approved):

PR #905, opened 2020-07-06
Serenity.await (continued)

arqunis:
The author of the [Serenity.await](https://github.com/serenity-rs/serenity/pull/803) PR has been absent for months, with no words from them at all. It is also in a bad shape in terms of its mergebility. It contains several merge conflicts and hasn't been updated to the latest state of the `next` branch.

This PR eliminates these issues, and until a response is received from the author, the development of `async`ifying Serenity will go forward here.

In order to update the previous PR, I had to manually recommit all of the changes present in it. This was a painful process, but the changes should be basically the same between the PRs. I introduced improvements myself in a few places while I was rummaging through the previous PR. Also to note, while the commits are under my name, it is actually @Lakelezz and her efforts who made the port to async. This PR is to continue the efforts and prepare it for being merged into the repository. 

The problem of OOMs has not been fixed yet. It is still misunderstood where the root of the problem is, and occurs (and doesn't occur) in people's bots differently.

Use this PR by putting the url of my repository and branch in your `Cargo.toml`:
```toml
[dependencies.serenity]
git = "https://github.com/acdenisSK/serenity"
branch = "await_next"
```


kageurufu:
I updated my bot yesterday morning, and have had one OoM since then.

```
[3247937.402851] VM Thread invoked oom-killer: gfp_mask=0x100cca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=0
[3247937.402876]  oom_kill_process.cold+0xb/0x10
[3247937.403005] [  pid  ]   uid  tgid total_vm      rss pgtables_bytes swapents oom_score_adj name
[3247937.403410] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=7b9aab197d186968a86bd46530b172488fc2d95fc210b2dccba7d7905426898f,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1000.slice/user@1000.service,task=shitbot,pid=1845665,uid=1000
[3247937.403435] Out of memory: Killed process 1845665 (shitbot) total-vm:12744028kB, anon-rss:12440616kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:24472kB oom_score_adj:1000
[3247937.849762] oom_reaper: reaped process 1845665 (shitbot), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

I went ahead and set `MemoryHigh` and `MemoryMax` in my service file (0.9G and 1.0G respectively) to prevent it causing problems for other services on this server. Is there anything I can do to help debug this?

vicky5124:
> I updated my bot yesterday morning, and have had one OoM since then.
> 
> ```
> [3247937.402851] VM Thread invoked oom-killer: gfp_mask=0x100cca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=0
> [3247937.402876]  oom_kill_process.cold+0xb/0x10
> [3247937.403005] [  pid  ]   uid  tgid total_vm      rss pgtables_bytes swapents oom_score_adj name
> [3247937.403410] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=7b9aab197d186968a86bd46530b172488fc2d95fc210b2dccba7d7905426898f,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1000.slice/user@1000.service,task=shitbot,pid=1845665,uid=1000
> [3247937.403435] Out of memory: Killed process 1845665 (shitbot) total-vm:12744028kB, anon-rss:12440616kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:24472kB oom_score_adj:1000
> [3247937.849762] oom_reaper: reaped process 1845665 (shitbot), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
> ```
> 
> I went ahead and set `MemoryHigh` and `MemoryMax` in my service file (0.9G and 1.0G respectively) to prevent it causing problems for other services on this server. Is there anything I can do to help debug this?

trace level logs (preferably with tokio tracing) is the best you can do to help, as we'll be able to see at what point the bot spikes in memory usage to get an OOM.

kageurufu:
I've added the following to the start of my `main()`, anything else I need to do?

```rust
    tracing::subscriber::set_global_default(
        tracing_subscriber::fmt()
            .with_max_level(Level::TRACE)
            .with_writer(io::stderr)
            .finish(),
    )
    .expect("Failed to set global tracing subscriber");
```

EDIT: I'm not seeing a bunch of trace-level logging, I must be missing something

vicky5124:
> I've added the following to the start of my `main()`, anything else I need to do?
> 
> ```rust
>     tracing::subscriber::set_global_default(
>         tracing_subscriber::fmt()
>             .with_max_level(Level::TRACE)
>             .with_writer(io::stderr)
>             .finish(),
>     )
>     .expect("Failed to set global tracing subscriber");
> ```
> 
> EDIT: I'm not seeing a bunch of trace-level logging, I must be missing something

Try this, it's what i use.
```rust
let subscriber = tracing_subscriber::FmtSubscriber::builder()
    .with_max_level(Level::TRACE)
    .with_writer(io::stderr)
    .finish();
tracing::subscriber::set_global_default(subscriber)?;
```

trungfinity:
This is nice. I was looking for it and came across this PR literally 1 minute after your last push. I checked against our Discord bot, refactored to using `async`/`await`, and it works nicely!

LeSeulArtichaut:
Small nit with the `#[hook]` macro:
I'm unable to use any lint attribute (e.g. `#[allow]`) for the function.

In this code:
```rust
#[hook]
#[allow(warnings)]
async fn some_hook() {}
```
The attribute gets ignored entirely.
I tried to work around this by setting the lint level inside the hook:

```rust
#[hook]
async fn some_hook() {
    #![allow(warnings)]
}
```
But now I get this much weirder error (with no other context or diagnostics):
```
error: expected square brackets
```
And of course using the hook doesn't work (E0425 cannot find value in this scope ).

I don't think this should block this PR being merged, but since it isn't merged yet I thought opening an issue didn't make sense. Thanks in advance!

arqunis:
The latest commit should address your problem.

vova616:
working great for me too, just a few questions
is there a way for it to work with async-std?
does 'after' has to accept Box::pin? 

Thanks!

```
let mut client = Client::new(key.into_string().unwrap())
    .event_handler(Handler)
    .framework(StandardFramework::new()
    .configure(|c| c.prefix("~").delimiters(vec![", ", ","])) // set the bot's prefix to "~"
    .group(&GENERAL_GROUP)
    .after(|client, msg, command_name, error| {
        Box::pin(async move {
            match error {
                Ok(()) => {},
                Err(why) => {
                    println!("Command '{}' returned error {:?}", command_name, why);
                    let content = format!("Command failed: {:?}", why);
                    let _ = msg.reply(&client, content).await;
                },
            };
        })
    })).await?;
```

vicky5124:
@vova616 
It is not possible not use async-std with serenity.await, as one of the main dependencies (reqwests) only works with tokio.

The way you define client with a framework an other additions has also changed some. You will now have to replace all the closures with their own functions, using the #[hook] macro.
https://5124.mywire.org/tmp/serenity-await/serenity/framework/standard/macros/attr.hook.html
You can see how to use this on the example 5 and in the documentation.

SystemNeoSpl:
Maybe it would be interesting to see if an async-std compatible version could be made, perhaps with `surf` instead of `reqwest`.

Of course, I think the best course of action for now is just to finish async/.await implementation as a whole before ya get into the whole mess of supporting multiple runtimes.

dlee13:
You can still use libs that depend on async-std together with serenity. You'll just end up having a lot of extra dependencies. Async-std 1.6 uses smol for its executor, so you can add smol as a dependency and enable the tokio support feature.

arqunis:
If you're willing to use the nightly toolchain of Rust, you can now use the `#[hook]` attribute on closures. [These are the flags you're going to need][flags], and [this is how it looks like][look]. This should ease migration to or use of the async standard framework.

[flags]: https://github.com/acdenisSK/kitty/blob/master/src/main.rs#L1-L3
[look]: https://github.com/acdenisSK/kitty/blob/master/src/main.rs#L95-L104

Flat:
If you would like to help diagnose the OOM bug and run your bot on Linux please see https://github.com/KDE/heaptrack to collect statistics about the heap of your bot while it is running. You can then share the file or a screenshot of the top memoryleaks from heaptrack GUI. This will only help diagnosis if you actually experience the OOM bug.

kageurufu:
I am currently running under heaptrack, and will update when I have an oom event to share.

kageurufu:
Well, it took a few days, but I forced an oom by allocating memory in another process until the killer ran

Heres my bottom-up allocations:
image

heres the full [heaptrack stats file](https://kageurufu.net/heaptrack.shitbot.7301.zst)

Flat:
Unfortunately we still have not been able to pin down the OOM bug. Other than it seems to occur in the shard runner and possibly due to a panic. I am of the opinion that this PR be merged to get more exposure and hopefully information that can help us track down this bug. It only seems to affect a small portion of users and I am unaware of the commonality that causes this.

As a side note, if you run into this bug and your bot is open source please let me know so I can take a closer look between bots where this is an issue and those that it is not. (I'm hoping it is software and not hardware)

kageurufu:
@Flat that plan of action sounds good to me.

My bot is not currently open source, but I'd be happy to give any developers here access to the relevant repos (I've got a private vendored submodule as well), just let me know who to give read access to

vicky5124:
I agree with Flat. This PR has been pretty much ready for way too long, with only that 1 OOM issue that has been there since the start. The only real way to have it fixed is by exposing it to more people, to get more data and information about the cause.

Not having this PR released is also making other PRs have to be paused, like the voice rework, and it also prevents the lavalink-rs library to be on crates.io, making it's exposure way lower.

arqunis:
I am merging this and making a release candidate for v0.9.0. If any issues pop up after the candidate has been published onto https://crates.io, please let me know!

recapitalverb:
absolutely beautiful. will there be some kind of guide to port code for v0.8.x to v0.9?

arqunis:
Firstly, you now need a runtime to execute futures. Serenity (or more precisely, its dependencies) uses `tokio`, thus you should (and have to) use as well.

`Cargo.toml`:
```toml
[dependencies.tokio]
version = "0.2"
features = ["macros"]
```
`main.rs`:
```rust
#[tokio::main]
async fn main() {
    // your code here
}
```

Secondly, `async fn` inside traits is currently unstable. However, to not lose out on the benefits, we have added a dependency to allow `async fn` in traits on stable. It provides a proc-macro that can be applied on any trait and its `impl`s. Unlike `tokio`, you need not include it as a dependency in your `Cargo.toml`. It is re-exported from Serenity.
```rust
use serenity::client::EventHandler;
use serenity::async_trait;

struct Handler;
 
#[async_trait]
impl EventHandler for Handler {
     // your code here
}
```

Thirdly, the creation of the client is different. It has adopted a future-style builder to accommodate configuring the client.
```rust
let token = "...";
let mut client = Client::new(token)
     .event_handler(Handler)
     .await
     .expect("client");
 ```
You can find more options at https://docs.rs/serenity/0.9.0-rc.0/serenity/client/struct.ClientBuilder.html.

Fourthly, the cache doesn't give locks anymore. You can only ever clone data out of it. In previous versions, Deadlocks used to be a frequent occurrence, and to avoid them, users had to restructure their code. 

You can refer to the new cache interface at https://docs.rs/serenity/0.9.0-rc.0/serenity/cache/struct.Cache.html.

Fifthly, standard framework callbacks (e.g., `before`, `after`, `on_dispatch_error`) should now be extracted into functions with the `#[hook]` macro.
```rust
use serenity::framework::standard::macros::hook;

#[hook]
async fn before(ctx: &Context, msg: &Message, command_name: &str) -> bool {
    // your code here
}

StandardFramework::new()
     .before(before)
```
That is if you want your bot to compile on the stable toolchain of Rust. If you're on nightly, you can enable these features:
```rust
#![feature(stmt_expr_attributes)]
#![feature(proc_macro_hygiene)]
#![feature(async_closure)]
```
And instead apply the `#[hook]` macro on a closure:
```rust
StandardFramework::new()
     .before(#[hook] |ctx, msg, command_name| async {

     }) 
```
Sixthly, the default help commands provided by the standard framework now return `Option<Message>`. This permits you to manipulate the generated message, such as to delete or edit it.
```rust
fn my_help(
    context: &Context,
    msg: &Message,
    args: Args,
    help_options: &'static HelpOptions,
    owners: HashSet<UserId>,
) -> CommandResult {
    help_commands::with_embeds(context, msg, args, help_options, groups, owners)
}
```
Should now be:
```rust
async fn my_help(
    context: &Context,
    msg: &Message,
    args: Args,
    help_options: &'static HelpOptions,
    owners: HashSet<UserId>,
) -> CommandResult {
    let _ = help_commands::with_embeds(context, msg, args, help_options, groups, owners).await;
    Ok(())
}
```

Seventhly...well, this will be obvious. Add `async`s and `.await`s wherever they are required.

That should be about it. I advise seeing the examples if you're still confused how code should look now. 
https://github.com/serenity-rs/serenity/tree/current/examples

hammypants:
Is there anyway to hook into the `tokio` runtime used? or provide our own?

Meshing serenity with another `tokio`-using library or set of code is messy, or impossible[or close enough] if the versions do not match _exactly_.

dmarcuse:
Serenity specifies version 0.2 of tokio, so any version of tokio that's compatible (i.e. >= 0.2.0, < 0.3) should be automatically unified by cargo in the dependency tree. If you need to use another version of tokio, you can use `tokio-compat` or `tokio-compat-02`, which provide compatibility layers that allow you to use tokio 0.1 or tokio 0.3 as well.

hammypants:
@dmarcuse first off, thank you, I did not know of `tokio-compat-02`. Unfortunately, I'm running into a myriad of type-length limit errors while using it. I'm not sure it's worth pursuing any further at this point; will probably wait until serenity moves to 0.3+.

dmarcuse:
The type length limit can be adjusted with `#![type_length_limit="..."]` at the top of your crate root if I remember correctly, does that not work here?

PR #911, opened 2020-07-10
Add mentions_me helper

mysteriouspants:
I found that I wanted this particular API in a recent project of mine. Review and improvement welcome!

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -618,6 +618,19 @@ impl Message {
         self.mentions_user_id(user.id)
     }
 
+    /// Checks whether the message mentions the current user.
+    pub fn mentions_me(&self, cache_http: impl CacheHttp) -> Result<bool> {
+        #[cfg(feature = "cache")]
+        {
+            if let Some(cache) = cache_http.cache() {
+                return Ok(self.mentions_user_id(cache.read().user.id));
+            }
+        }
+
+        let current_user = cache_http.http().get_current_user()?;
+        Ok(self.mentions_user_id(current_user.id))
+    }
+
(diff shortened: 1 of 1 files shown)


src/model/channel/message.rs
```diff
@@ -618,6 +618,14 @@ impl Message {
...
     }
 
+    /// Checks whether the message mentions the current user.
+    pub fn mentions_me(&self, http: impl AsRef<Http>) -> Result<bool> {
+        match http.as_ref().get_current_user() {
+            Ok(current_user) => Ok(self.mentions_user_id(current_user.id)),
+            Err(e) => Err(e)
+        }
+    }
+
```
arqunis: Despite the HTTP client being all that is needed to fulfill the task of this function, it's inefficient to invoke a REST call each time. The cache is already aware of the user information about the bot. I recommend retrieving the bot's id using the cache if the user passes it.
```suggestion
    pub fn mentions_me(&self, cache_http: impl CacheHttp) -> Result<bool> {
        #[cfg(feature = "cache")]
        {
            if let Some(cache) = cache_http.cache() {
                return Ok(self.mentions_user_id(cache.read().user.id));
            }
        }
        
        let current_user = http.as_ref().get_current_user()?;
        Ok(self.mentions_user_id(current_user.id))
    }

```

arqunis (changes requested):

src/model/channel/message.rs
mysteriouspants: Oh that makes quite a lot of sense, thanks for the tip! I've updated the PR with your suggestion.

arqunis (approved):

PR #912, opened 2020-07-11
Replace the in-house typemap with typemap_rev

kingbri1:
I migrated the serenity typemap over to a standalone crate. Doing this
removes an extra file (typemap.rs) which is now unnecessary. I am planning
on publishing the crate on crates.io in the not-too-distant future, but for now,
the git is the same as the current typemap utility.

I have invited acdenis to the typemap_rev repo so he can modify it as needed
since it is mostly his code.

Signed-off-by: kingbri <bdashore3@gmail.com>


arqunis:
Cargo prevents publishing crates with unpublished dependency crates. That accounts for git dependencies. I won't merge this PR until `typemap_rev` gets published on https://crates.io.

kingbri1:
I'll get it published as soon as I run a few tests

kingbri1:
This commit fixes the issue. Let me know if the dependency should be optional or not.

arqunis:
You should add the crate under the `client` feature.
https://github.com/serenity-rs/serenity/blob/d355d344b6ebbcbb60de0fc8e0331b9c3e4583c9/Cargo.toml#L130-L134
Subsequently, the feature gate should be updated to the `client` feature.
https://github.com/serenity-rs/serenity/blob/d355d344b6ebbcbb60de0fc8e0331b9c3e4583c9/src/prelude.rs#L29-L30

bdashore3:
Like this for the cargo.toml?

```rust
client = [
    "gateway",
    "http",
    "threadpool",
]

[dependencies.typemap_rev]
version = "0.1.2"
```

kingbri1:
Ignore that last comment, I figured it out

src/prelude.rs
```diff
@@ -20,13 +20,12 @@ pub use parking_lot::{Mutex, RwLock};
 
 #[cfg(feature = "client")]
 pub use crate::client::{Client, ClientError, Context, EventHandler, RawEventHandler};
+pub use typemap_rev::{TypeMap, TypeMapKey};
```
arqunis, line 24:
```suggestion
#[cfg(feature = "client")]
pub use typemap_rev::{TypeMap, TypeMapKey};
```
Feature gates apply on only one item succeeding them.

arqunis (changes requested):

arqunis (approved):

PR #913, opened 2020-07-13
Added teams support.

vicky5124:
`TeamMember.membership_state` could be an enum, but i couldn't figure out how to make it deserializable.
```
	modified:   src/model/application.rs
```

### DIFF
--- a/src/model/application.rs
+++ b/src/model/application.rs
@@ -114,10 +114,9 @@ pub struct Team {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct TeamMember {
     /// The member's membership state.
+    pub membership_state: MembershipState,
+    /// The list of permissions of the member on the team.
     ///
-    /// 1: Invited
-    /// 2: Accepted
-    pub membership_state: u8,
     /// NOTE: Will always be ["*"] for now.
     pub permissions: Vec<String>,
     /// The ID of the team they are a member of.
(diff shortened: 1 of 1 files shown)


src/model/application.rs
```diff
@@ -89,6 +91,38 @@ pub struct CurrentApplicationInfo {
...
     pub bot_require_code_grant: bool,
+    pub team: Option<Team>,
     #[serde(skip)]
     pub(crate) _nonexhaustive: (),
 }
+
+/// Information about the Team group of the application.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct Team {
+    /// The icon of the TEam.
```
arqunis:
```suggestion
    /// The icon of the team.
```

src/model/application.rs
```diff
@@ -89,6 +91,38 @@ pub struct CurrentApplicationInfo {
...
+}
+
+/// Infromation about a Member on a Team.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct TeamMember {
+    /// The member's membership state.
+    ///
+    /// 1: Invited
+    /// 2: Accepted
+    pub membership_state: u8,
```
arqunis: > `TeamMember::membership_state` could be an enum, but i couldn't figure out how to make it deserializable.

We have a macro built for this purpose, `enum_number!`. Here's an example of its usage:
https://github.com/serenity-rs/serenity/blob/current/src/model/gateway.rs#L357-L396

arqunis (changes requested):

src/model/application.rs
vicky5124: Resolved on ec0286b.

arqunis (approved):

PR #920, opened 2020-07-24
Add nick to PartialMember

asherkin:
Discord includes this in the partial member info included with message events.

https://discord.com/developers/docs/resources/guild#guild-member-object

### DIFF
--- a/src/model/guild/member.rs
+++ b/src/model/guild/member.rs
@@ -520,6 +520,10 @@ pub struct PartialMember {
     pub joined_at: Option<DateTime<FixedOffset>>,
     /// Indicator of whether the member can speak in voice channels.
     pub mute: bool,
+    /// The member's nickname, if present.
+    ///
+    /// Can't be longer than 32 characters.
+    pub nick: Option<String>,
     /// Vector of Ids of [`Role`]s given to the member.
     pub roles: Vec<RoleId>,
     #[serde(skip)]


arqunis:
`current` is in a feature-freeze. It doesn't accept new features right now. And addition of a field is considered a new feature. Would you mind switching the target branch to `next`?

arqunis (approved):

asherkin:
> And addition of a field is considered a new feature.

Ah, I thought it was safe here because of the `_nonexhaustive`.

> Would you mind switching the target branch to `next`?

Thanks for the review - that's done now.

arqunis:
> Ah, I thought it was safe here because of the `_nonexhaustive.`

We employ the use of a private field called `_nonexhaustive` to prevent making breaking changes when we add new fields. Rust treats new fields as a breaking change because it breaks initialization and pattern matching of the structure.

Imagine a struct `Foo` with fields `bar` and `baz`:
```rust
struct Foo {
    bar: i32,
    baz: f64,
}
```
And it is used like so:
```rust
let foo = Foo {
    bar: 4,
    baz: 3.14,
};

do_something(&foo);

let Foo { bar, baz } = foo;

let result = add(bar, baz);
```
This code will fail to compile if we add another field.

1. Initialization 
Rust dictates that all values (including instances of structs) must be defined and initialized completely. If we left out the new field, its value will be undefined and will be up to the machine to decide the value of the field. The language wants to absolutely prevent that, as undefined values have caused bugs in the past (though, nowadays you have [`MaybeUninit`](https://doc.rust-lang.org/nightly/std/mem/union.MaybeUninit.html) to sidestep this in a safe manner).

2. Pattern matching
In order for pattern matching to work, you have to be accurate and precise. You have to explicitly declare which values you want to match on, and which want to bind to a variable, and which to ignore. In our case we're matching the whole struct. This means that the declaration of the struct has to match our pattern exactly. Adding a new field breaks this condition.

The private fields forbids initializing the struct, and forces users to ignore any other potential fields when pattern matching it. Á la:
```rust
let foo = Some(Foo { bar: 2, baz: 2.71 });

if let Some(Foo { bar, baz, ... }) = foo {
    // ....
}
```

So, no, there isn't any relation between `_nonexhaustive` and new features. All it does it makes sure we don't have to release a major version on the addition of fields or variants (we do this on enums as well).

Anyway, thank you for the PR!

PR #933, opened 2020-08-18
Remove some chrono

qm3ster:
This PR is only about changes that don't affect public API

### DIFF
--- a/src/internal/timer.rs
+++ b/src/internal/timer.rs
@@ -1,51 +1,45 @@
-use chrono::{DateTime, Duration, Utc};
-use std::time::Duration as StdDuration;
-use tokio::time::delay_for;
+use tokio::time::{delay_until, Duration, Instant};
 
 #[derive(Debug)]
 pub struct Timer {
-    due: DateTime<Utc>,
+    due: Instant,
     duration: Duration,
 }
 
 impl Timer {
+    /// construct timer, initially set and due `duration_in_ms` in the future
     pub fn new(duration_in_ms: u64) -> Timer {
-        let duration = Duration::milliseconds(duration_in_ms as i64);
+        let duration = Duration::from_millis(duration_in_ms);
 
         Timer {
-            due: Utc::now() + duration,
+            due: Instant::now() + duration,
             duration,
         }
     }
 
+    /// block until next due time resetting afterwards
     pub async fn hold(&mut self) {
-        let due_time = (self.due.timestamp() * 1000) + i64::from(self.due.timestamp_subsec_millis());
-        let now_time = {
-            let now = Utc::now();
-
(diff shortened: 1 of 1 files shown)


src/internal/timer.rs
```diff
@@ -1,51 +1,45 @@
...
 
 #[derive(Debug)]
 pub struct Timer {
-    due: DateTime<Utc>,
+    due: Instant,
     duration: Duration,
 }
 
 impl Timer {
+    /// construct timer, initially set and due `duration_in_ms` in the future
```
arqunis: Doc-strings should be written as proper sentences. They should start with an uppercase letter, and end with a full stop.

```suggestion
    /// Construct a `Timer`, which is initially set to expire in due `duration_in_ms` time from the current instant.
```

src/internal/timer.rs
```diff
@@ -1,51 +1,45 @@
...
+        let duration = Duration::from_millis(duration_in_ms);
 
         Timer {
-            due: Utc::now() + duration,
+            due: Instant::now() + duration,
             duration,
         }
     }
 
+    /// block until next due time resetting afterwards
```
arqunis:
```suggestion
    /// Block until the due time. The timer will be reset afterwards.
```

src/internal/timer.rs
```diff
@@ -1,51 +1,45 @@
...
-                delay_for(StdDuration::from_millis(sleep_time as u64)).await;
-            }
-        }
-
-        self.due = self.due + self.duration;
+        delay_until(self.due).await;
+        self.increment();
     }
 
+    /// returns true and resets the timer if due
```
arqunis:
```suggestion
    /// Returns a boolean indicating whether the timer has expired. the timer will also be reset if it has.
```

src/internal/timer.rs
```diff
@@ -1,51 +1,45 @@
...
+        if Instant::now() >= self.due {
+            self.increment();
             true
         } else {
             false
         }
     }
 
-    pub fn reset(&mut self) { self.due = Utc::now() + self.duration; }
+    /// reset timer to be 1 duration after previous due time
```
arqunis:
```suggestion
    /// Resets the timer by incrementing the due time using the duration that was passed upon construction of the timer. 
```

src/internal/timer.rs
```diff
@@ -1,51 +1,45 @@
...
+    fn increment(&mut self) {
+        self.due += self.duration
+    }
+
+    /// reset timer to be 1 duration from **now**
```
arqunis:
```suggestion
    /// Resets the timer by restarting from the current instant and setting its due time to the duration that was passed upon construction. This has the same effect as constructing a new `Timer`.
```

arqunis (changes requested):

arqunis (approved):

PR #935, opened 2020-08-20
Goodbye FixedOffset

qm3ster:
Be very afraid, users will be broken by:
1. replacing `DateTime<FixedOffset>` with `DateTime<Utc>` everywhere
2. BucketBuilder methods taking u64 not i64

Look through the commits though, there might be something good 🤷🤷‍♂️🤷‍♀️


arqunis:
Could you remove the last commit? I would prefer to apply `cargo fmt` on the repository in a separate PR.

qm3ster:
Done.

arqunis (approved):

Issue #932, opened 2020-08-18
Add a Display impl for Permissions


arqunis:
The purpose of this `Display` implementation will be to print the name of the permission as it is shown in the Discord client:
`Permissions::MANAGE_CHANNELS` -> "Manage Channels"
`Permissions::VIEW_AUDIT_LOG` -> "View Audit Log".

If many permissions are contained within the `Permissions` struct, the implementation should return a string concatenating all of the names into a listing:
`Permissions::MANAGE_MESSAGES & Permissions::MANAGE_ROLES` -> "Manage Messages and Manage Roles"
`Permissions::MANAGE_MESSAGES & Permissions::MANAGE_CHANNELS & Permissions::MANAGE_EMOJIS` -> "Manage Messages, Manage Channels, and Manage Roles"

PR #939, opened 2020-08-23
Implement `Display` for `Permissions`

AriusX7:
Implements #932.

I wrote the `get_permission_names` method by hand first but decided to use a macro later to make it easier to add new permissions.

For the `USE_VAD` permission, I used "Use Voice Activity" as the name as it's used by Discord on the Developer Portal. Similarly, I used the name used on the portal for the `MENTION_EVERYONE` permission. It is called "Mention @everyone, @here, and All Roles" in the app.

### DIFF
--- a/src/model/permissions.rs
+++ b/src/model/permissions.rs
@@ -587,14 +587,14 @@ impl Display for Permissions {
         let total = names.len();
         for (i, &name) in names.iter().enumerate() {
             if i > 0 && i != total - 1 {
-                let _ = write!(f, ",");
+                let _ = write!(f, ", ");
             }
 
             if total > 1 && i == total - 1 {
-                let _ = write!(f, " and");
+                let _ = write!(f, " and ");
             }
 
-            let _ = write!(f, " {}", name);
+            let _ = write!(f, "{}", name);
         }
 
         Ok(())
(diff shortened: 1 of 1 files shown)


src/model/permissions.rs
```diff
@@ -477,3 +567,24 @@ impl Serialize for Permissions {
...
 }
+
+impl Display for Permissions {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        let names = self.get_permission_names();
+
+        let total = names.len();
+        for (i, &name) in names.iter().enumerate() {
+            if i > 0 && i != total - 1 {
+                let _ = write!(f, ", ");
```
arqunis:
```suggestion
                write!(f, ", ")?;
```

src/model/permissions.rs
```diff
@@ -477,3 +567,24 @@ impl Serialize for Permissions {
...
+            }
+
+            if total > 1 && i == total - 1 {
+                let _ = write!(f, " and ");
```
arqunis:
```suggestion
                write!(f, " and ")?;
```

src/model/permissions.rs
```diff
@@ -477,3 +567,24 @@ impl Serialize for Permissions {
...
+            }
+
+            let _ = write!(f, "{}", name);
```
arqunis:
```suggestion
            write!(f, "{}", name)?;
```

arqunis (changes requested):

arqunis (approved):

PR #944, opened 2020-08-25
Add support for invite events

arqunis:
This adds handlers for [the `INVITE_CREATE` and `INVITE_DELETE` events](https://discord.com/developers/docs/topics/gateway#invite-create).

Note, however, that this is untested. My bot that I have been trying to test these events has not been able to receive them, thus I ask others if their bots can and that they can get dispatched from Serenity.

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -594,6 +594,20 @@ async fn handle_event(
                 }}
             });
         },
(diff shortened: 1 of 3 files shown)


arqunis:
[The person who requested for the support of these events](https://github.com/Adatan) has confirmed that the implementation works, thus I am merging this.

PR #952, opened 2020-09-05
Add animated emoji support for parse_emoji()

kingbri1:
Allows for users to check if an emoji mention contains an animated emoji and returns a bool with that value.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -247,7 +247,7 @@ pub fn parse_mention(mention: impl AsRef<str>) -> Option<u64> {
     }
 }
 
-/// Retrieves the name and Id from an emoji mention, in the form of an
+/// Retrieves the animated state, name and Id from an emoji mention, in the form of an
 /// `EmojiIdentifier`.
 ///
 /// If the emoji usage is invalid, then `None` is returned.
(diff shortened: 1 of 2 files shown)


kingbri1:
Since creating a reaction requires an animated bool in a ReactionType, rather than guessing if the emoji is animated or not, use the emoji mention.

src/utils/mod.rs
```diff
@@ -287,13 +287,20 @@ pub fn parse_emoji(mention: impl AsRef<str>) -> Option<EmojiIdentifier> {
...
         let mut name = String::default();
         let mut id = String::default();
+        let mut animated = false;
 
-        for (i, x) in mention[2..].chars().enumerate() {
+        if &mention[1..3] == "a:" {
+            animated = true;
+        }
+
+        let start = animated as usize + 2;
```
arqunis: This is clever, but I prefer readability more.
```suggestion
        let start = if animated { 3 } else { 2 }; 
```

src/utils/mod.rs
```diff
@@ -287,13 +287,20 @@ pub fn parse_emoji(mention: impl AsRef<str>) -> Option<EmojiIdentifier> {
         return None;
     }
 
-    if mention.starts_with("<:") && mention.ends_with('>') {
+    if (mention.starts_with("<:") || mention.starts_with("<a:")) && mention.ends_with('>') {
         let mut name = String::default();
         let mut id = String::default();
+        let mut animated = false;
```
arqunis:
```suggestion
        let animated = &mention[1..3] == "a:";
```

arqunis (changes requested):

kingbri1:
Changes added and force pushed. Also fixed errors regarding tests.

arqunis (approved):

PR #954, opened 2020-09-08
Switched from log to tracing.

vicky5124:
This may need a check for places that may be spammy and have instrument, but from what i tested, nothing that spams has instrument with command arguments.

### DIFF
--- a/examples/e08_env_logging/src/main.rs
+++ b/examples/e08_env_logging/src/main.rs
@@ -88,7 +88,7 @@ async fn main() {
 }
 
 // Currently, the instrument macro doesn't work with commands.
-// if you wish to intrument commands, use it on the before function.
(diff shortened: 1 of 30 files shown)


examples/e08_env_logging/src/main.rs
```diff
@@ -26,25 +36,64 @@ impl EventHandler for Handler {
...
         .await
         .expect("Err creating client");
 
     if let Err(why) = client.start().await {
         error!("Client error: {:?}", why);
     }
 }
+
+// Currently, the instrument macro doesn't work with commands.
+// if you wish to intrument commands, use it on the before function.
```
BlackHoleFox: Minor typo here:
```suggestion
// if you wish to instrument commands, use it on the before function.
```

arqunis (approved):

PR #959, opened 2020-09-13
Use #[non_exhaustive] where applicable

LikeLakers2:
Hi there! This changes all of those odd `__Nonexhaustive` enum cases, so that they instead use [the `#[non_exhaustive]` attribute](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute).

~~However, I opted not to add the `#[non_exhaustive]` attribute to any structs where it could apply (that is, those who had a `__nonexhaustive: ()` field), since I'm unsure if the effects would be acceptable:~~

> ~~Non-exhaustive structs could have additional fields added in future. Therefore, non-exhaustive structs cannot be constructed in external crates using the traditional `Struct { .. }` syntax; cannot be matched against without a wildcard `..`; and struct update syntax will not work.~~

~~That said, if it's desired that I add it to any applicable structs, I can do so.~~

Side note: I did notice #752 while searching to see if there was an existing PR that made this change -- but being that the PR is from December 2019, I imagine there would be lots of merge conflicts and other misc issues that would make it hard to merge that. So I decided to make my own PR.

### DIFF
--- a/src/builder/create_channel.rs
+++ b/src/builder/create_channel.rs
@@ -127,7 +127,6 @@ impl CreateChannel {
             let (id, kind) = match perm.kind {
                 PermissionOverwriteType::Member(id) => (id.0, "member"),
                 PermissionOverwriteType::Role(id) => (id.0, "role"),
-                PermissionOverwriteType::__Nonexhaustive => unreachable!(),
             };
 
             json!({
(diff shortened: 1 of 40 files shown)


arqunis:
> since I'm unsure if the effects would be acceptable

The private field enforces the effects already. The attribute just makes it more apparent to the user and the compiler that a struct/enum is non-exhaustive, thus tooling can support the fact (e.g., rustdoc showing it in documentation). You should switch the structs to use the attribute as well.

LikeLakers2:
@acdenisSK I've added `#[non_exhaustive]` to any applicable structs, as you requested.

arqunis (approved):

PR #961, opened 2020-09-14
Fix parsing of commands that have requirements

arqunis:
This fixes parsing behaviour of commands that have requirements. 

Under normal circumstances, if the requirements are not met, the framework should return a dispatch error. Of course, It does return a dispatch error, but only when there is a single group. Once there are several (at least two) groups, the dispatch error that is supposed to be returned gets replaced with the "UnrecognisedCommand" error. This is because we do not check that we had a dispatch error at all, so the framework does not have a problem replacing it with a useless error.

This also removes redundant discrepancy checks. Each valid group or command is already checked for discrepancies before they are returned from their parsing functions.

### DIFF
--- a/src/framework/standard/parse/mod.rs
+++ b/src/framework/standard/parse/mod.rs
@@ -316,6 +316,10 @@ impl From<DispatchError> for ParseError {
     }
 }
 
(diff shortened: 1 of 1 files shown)


kingbri1:
Works on all my bots that have had this issue

PR #985, opened 2020-10-01
Fixed all clippy warnings

ikkerens:
All of them. No more clippy warnings.
Had to #[allow] a few things, most notably things that would result in breaking API or clippy-compiler disagreements.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -583,6 +583,7 @@ impl Client {
     ///
     /// [`Client`]: #struct.Client.html
     /// [`Future`]: https://doc.rust-lang.org/std/future/trait.Future.html
+    #[allow(clippy::new_ret_no_self)] // Remnant of the async transition, but changing it would be breaking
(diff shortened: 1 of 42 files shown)


src/client/mod.rs
```diff
@@ -583,6 +583,7 @@ impl Client {
     ///
     /// [`Client`]: #struct.Client.html
     /// [`Future`]: https://doc.rust-lang.org/std/future/trait.Future.html
+    #[allow(clippy::new_ret_no_self)] // Remnant of the async transition, but changing it would be breaking
     pub fn new<'a>(token: impl AsRef<str>) -> ClientBuilder<'a> {
         ClientBuilder::new(token)
     }
```
Erk-: Maybe make a function named `builder` that does this and mark new as deprecated

Erk- (commented):
lgtm, just a small nit

arqunis (approved):

PR #991, opened 2020-10-03
Initial implementation for inline replies

ikkerens:
Initial implementation for https://github.com/discord/discord-api-docs/pull/2118

I also added in the support for v8, which will need to be uncommented when that gets implemented. Do you guys keep track of a v8 TODO list somewhere?

Full disclaimer though, I have no idea if this works as it is untested vs the Discord API (~~they haven't implemented this yet, so don't merge until they have~~).

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -112,6 +112,8 @@ pub struct Message {
     pub message_reference: Option<MessageReference>,
     /// Bit flags describing extra features of the message.
     pub flags: Option<MessageFlags>,
+    /// The message that was replied to using this message.
+    pub referenced_message: Option<Box<Message>>, // Boxed to avoid recusion
     #[serde(skip)]
     pub(crate) _nonexhaustive: (),
 }
@@ -863,6 +865,8 @@ pub enum MessageType {
     NitroTier2 = 10,
     /// An indicator that the guild has reached nitro tier 3
     NitroTier3 = 11,
+    // /// An message reply.
+    // InlineReply = 19, // TODO Enable this when v8 hits.
 }
 
 enum_number!(
@@ -879,6 +883,7 @@ enum_number!(
         NitroTier1,
         NitroTier2,
         NitroTier3,
+        // InlineReply, // TODO Enable this when v8 hits
     }
 );
 
@@ -975,6 +980,28 @@ pub struct MessageReference {
     pub guild_id: Option<GuildId>,
     #[serde(skip)]
     pub(crate) _nonexhaustive: (),
+}
+
+impl From<&Message> for MessageReference {
+    fn from(m: &Message) -> Self {
+        Self {
+            message_id: Some(m.id),
+            channel_id: m.channel_id,
+            guild_id: m.guild_id,
+            _nonexhaustive: ()
+        }
+    }
+}
+
+impl From<(ChannelId, MessageId)> for MessageReference {
+    fn from(pair: (ChannelId, MessageId)) -> Self {
+        Self {
+            message_id: Some(pair.1),
+            channel_id: pair.0,
+            guild_id: None,
+            _nonexhaustive: ()
+        }
+    }
 }
 
 /// Channel Mention Object
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -904,6 +904,7 @@ impl MessageType {
             NitroTier1 => 9,
             NitroTier2 => 10,
             NitroTier3 => 11,
+            InlineReply => 19,
         }
     }
 }
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -866,7 +866,7 @@ pub enum MessageType {
     /// An indicator that the guild has reached nitro tier 3
     NitroTier3 = 11,
     // /// An message reply.
-    // InlineReply = 19, // TODO Enable this when v8 hits.
+    InlineReply = 19,
 }
 
 enum_number!(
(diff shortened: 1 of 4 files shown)


src/model/channel/message.rs
```diff
@@ -879,6 +883,7 @@ enum_number!(
         NitroTier1,
         NitroTier2,
         NitroTier3,
+        // InlineReply, // TODO Enable this when v8 hits
```
arqunis: This can be enabled now.

arqunis (changes requested):

ikkerens:
I've enabled the v8 enums, but I can't personally test this inline replies feature with the lib myself atm.

arqunis (approved):
Besides the [absence of the variant](https://github.com/serenity-rs/serenity/pull/991/files#diff-a95a446fbee2360e72976dd53c83ca8ff33a6319978fe62252ae0bdbb385faf5R906) causing failure in CI, this pull request is fine. I have tested it myself in the Serenity Discord server and works as it should.

vicky5124:
shouldn't Message.reply be updated to use this instead?
i wanna be able to use this extensively  thought out my bot, and having it limited to the message builder doesn't make this convenient.

This is also missing the option to disable the mention to the original author of the message, as possible to be done with the client.

Flat:
I agree reply should be changed to use this, it avoids confusion for new users but may be unexpected behavior for current users. Possibly a notice stating the change prominently in the next release may suffice.

Lakelezz:
My suggestions is to update `reply` to perform the new inline replies, but this is a breaking change as it changes behaviour. Therefore we would have to rebase to the breaking channel.
Additionally, implement `reply_ping`, which sets the toggle for whether the inline reply mentions to `true` and `reply_mention`, which would perform the old school reply we had for a long time (if that's even still needed).

kingbri1:
Agreed with this. These 3 methods give the end user the ability to choose instead of having one firm option.

HTG-YT:
I suggest add another method called `respond` to just send a message normally...?

HTG-YT:
I am not really sure about that 🤣

Lakelezz (changes requested):
As I mentioned above, it would be great to add these please before we can merge it: 
`reply`, performs the inline reply without pinging;
`reply_ping`, performs the inline reply with a ping;
`reply_mention`, performs the original `reply`, such as `@name:`.

HTGAzureX1212:
> As I mentioned above, it would be great to add these please before we can merge it:
> `reply`, performs the inline reply without pinging;
> `reply_ping`, performs the inline reply with a ping;
> `reply_mention`, performs the original `reply`, such as `@name:`.

I don't know if this approach is acceptable: `reply` takes a `ping` `bool` parameter, if `true` is passed in, performs the inline reply with a ping and vice versa.

iddm:
> > As I mentioned above, it would be great to add these please before we can merge it:
> > `reply`, performs the inline reply without pinging;
> > `reply_ping`, performs the inline reply with a ping;
> > `reply_mention`, performs the original `reply`, such as `@name:`.
> 
> I don't know if this approach is acceptable: `reply` takes a `ping` `bool` parameter, if `true` is passed in, performs the inline reply with a ping and vice versa.

This would break the backwards compatibility, and even if we aren't following the semver, this is just not convenient.

HTGAzureX1212:
I meant that to merge `reply` and `reply_ping`, and leave `reply_mention` as is

src/model/channel/message.rs
```diff
@@ -977,6 +983,28 @@ pub struct MessageReference {
     pub(crate) _nonexhaustive: (),
 }
 
+impl From<&Message> for MessageReference {
```
iddm, line 986: Just curious here, isn't `impl From<AsRef<Message>>` better?

iddm:
> I meant that to merge `reply` and `reply_ping`, and leave `reply_mention` as is

I might have been confused a bit, please correct me if I indeed have wrongly understood you.
The [reply method](https://docs.rs/serenity/0.9.1/serenity/model/channel/struct.Message.html#method.reply) has the following interface now in 0.9.1:

```rust
pub async fn reply<'_>(
    &'_ self,
    cache_http: impl CacheHttp,
    content: impl Display
) -> Result<Message>
```
If we are to add a bool parameter there like this
```rust
pub async fn reply<'_>(
    &'_ self,
    ping: bool,
    cache_http: impl CacheHttp,
    content: impl Display
) -> Result<Message>
```
then the current users will have to also add it to all of their invocations of it, what isn't convenient.

HTGAzureX1212:
> > I meant that to merge `reply` and `reply_ping`, and leave `reply_mention` as is
> 
> I might have been confused a bit, please correct me if I indeed have wrongly understood you.
> The [reply method](https://docs.rs/serenity/0.9.1/serenity/model/channel/struct.Message.html#method.reply) has the following interface now in 0.9.1:
> 
> ```rust
> pub async fn reply<'_>(
>     &'_ self,
>     cache_http: impl CacheHttp,
>     content: impl Display
> ) -> Result<Message>
> ```
> 
> If we are to add a bool parameter there like this
> 
> ```rust
> pub async fn reply<'_>(
>     &'_ self,
>     ping: bool,
>     cache_http: impl CacheHttp,
>     content: impl Display
> ) -> Result<Message>
> ```
> 
> then the current users will have to also add it to all of their invocations of it, what isn't convenient.

Was it the original `reply` method in 0.9.1 the new **inline** reply?

What I meant was, the **inline** reply method should have a `ping` `bool` parameter to determine whether to ping the user that was **inline** replied.

arqunis:
To avoid stalling this pull request further, it has been decided to merge it now. The adaptation of the `Message::reply` method will come in a separate commit/pull request.

PR #996, opened 2020-10-05
Added an example for client data.

vicky5124:
This very useful feature of serenity was kinda being buried down by the huge amount of information obtained on the example 5, so i made it on it's own example, showing how to use it safely in various ways.

	modified:   Makefile.toml
	modified:   azure-build-examples.yml
	modified:   examples/Cargo.toml
	modified:   examples/README.md
	new file:   examples/e14_global_data/Cargo.toml
	new file:   examples/e14_global_data/Makefile.toml
	new file:   examples/e14_global_data/src/main.rs

### DIFF
--- a/examples/e14_global_data/src/main.rs
+++ b/examples/e14_global_data/src/main.rs
@@ -32,7 +32,7 @@ impl TypeMapKey for CommandCounter {
 struct MessageCount;
 
 impl TypeMapKey for MessageCount {
-    // While you will be using RwLock or Mutex most of the times you want to modify data,
+    // While you will be using RwLock or Mutex most of the time you want to modify data,
     // sometimes it's not required; like for example, with static data, or if you are using other
     // kinds sof atomic operators.
     //
--- a/examples/e14_global_data/src/main.rs
+++ b/examples/e14_global_data/src/main.rs
@@ -50,7 +50,7 @@ async fn before(ctx: &Context, msg: &Message, command_name: &str) -> bool {
 
     let counter_lock = {
         // While data is a RwLock, it's recommended that you always open the lock as read.
-        // this is mainly done to avoid Deadlocks for having a possible writer waiting for multiple
+        // This is mainly done to avoid Deadlocks for having a possible writer waiting for multiple
         // readers to close.
         let data_read = ctx.data.read().await;
 
--- a/examples/e14_global_data/src/main.rs
+++ b/examples/e14_global_data/src/main.rs
@@ -56,7 +56,7 @@ async fn before(ctx: &Context, msg: &Message, command_name: &str) -> bool {
 
         // Since the CommandCounter Value is wrapped in an Arc, cloning will not duplicate the
         // data, instead the reference is cloned.
-        // We wap every value on an arc, as to keep the data lock open for the least time possible,
+        // We wap every value on in an Arc, as to keep the data lock open for the least time possible,
         // to again, avoid Deadlocking it.
         data_read.get::<CommandCounter>().expect("Expected CommandCounter in TypeMap.").clone()
     };
--- a/examples/e14_global_data/src/main.rs
+++ b/examples/e14_global_data/src/main.rs
@@ -34,7 +34,7 @@ struct MessageCount;
 impl TypeMapKey for MessageCount {
     // While you will be using RwLock or Mutex most of the time you want to modify data,
     // sometimes it's not required; like for example, with static data, or if you are using other
-    // kinds sof atomic operators.
+    // kinds of atomic operators.
     //
     // Arc should stay, to allow for the data lock to be closed early.
     type Value = Arc<AtomicUsize>;
--- a/examples/e14_global_data/src/main.rs
+++ b/examples/e14_global_data/src/main.rs
@@ -57,7 +57,7 @@ async fn before(ctx: &Context, msg: &Message, command_name: &str) -> bool {
         // Since the CommandCounter Value is wrapped in an Arc, cloning will not duplicate the
         // data, instead the reference is cloned.
         // We wap every value on in an Arc, as to keep the data lock open for the least time possible,
-        // to again, avoid Deadlocking it.
+        // to again, avoid deadlocking it.
         data_read.get::<CommandCounter>().expect("Expected CommandCounter in TypeMap.").clone()
     };
 
--- a/examples/e14_global_data/src/main.rs
+++ b/examples/e14_global_data/src/main.rs
@@ -123,7 +123,7 @@ async fn main() {
     // write mode, but there's a small thing catch:
     // There can only be a single writer to a given lock open in the entire application, this means
     // you can't open a new write lock until the previous write lock has closed.
-    // This is not the case with Read locks, read locks can be open indefinitely, BUT as soon as
+    // This is not the case with read locks, read locks can be open indefinitely, BUT as soon as
     // you need to open the lock in write mode, all the read locks must be closed.
     //
     // You can find more information about deadlocks in the Rust Book, ch16-03:
--- a/examples/e14_global_data/src/main.rs
+++ b/examples/e14_global_data/src/main.rs
@@ -89,7 +89,7 @@ impl EventHandler for Handler {
             };
 
             // Atomic operations with ordering do not require mut to be modified.
-            // in this case, we want to increase the message count by 1.
+            // In this case, we want to increase the message count by 1.
             // https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.fetch_add
             count.fetch_add(1, Ordering::SeqCst);
         }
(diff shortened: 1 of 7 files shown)


examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+struct MessageCount;
+
+impl TypeMapKey for MessageCount {
+    // While you will be using RwLock or Mutex most of the times you want to modify data,
+    // sometimes it's not required; Like for example, with static data, or if you are using other
+    // kinds sof atomic operators.
+    //
+    // Arc shoud stay, to allow for the data lock to be closed early.
```
arqunis: Typo.
```suggestion
    // Arc should stay, to allow for the data lock to be closed early.
```

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+
+impl TypeMapKey for CommandCounter {
+    type Value = Arc<RwLock<HashMap<String, u64>>>;
+}
+
+struct MessageCount;
+
+impl TypeMapKey for MessageCount {
+    // While you will be using RwLock or Mutex most of the times you want to modify data,
+    // sometimes it's not required; Like for example, with static data, or if you are using other
```
arqunis: The proceeding word after a semicolon cannot be capitalised (unless, for instance, said word is a proper noun).
```suggestion
    // sometimes it's not required; like for example, with static data, or if you are using other
```

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+struct CommandCounter;
+
+impl TypeMapKey for CommandCounter {
+    type Value = Arc<RwLock<HashMap<String, u64>>>;
+}
+
+struct MessageCount;
+
+impl TypeMapKey for MessageCount {
+    // While you will be using RwLock or Mutex most of the times you want to modify data,
```
arqunis: "Most of the time" is synonymous with *usually* and is the more correct version here. "Most of the times" would be used in cases where you are specifically talking about occasions when an action/activity happens (or does not).
```suggestion
    // While you will be using RwLock or Mutex most of the time you want to modify data,
```

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+#[commands(ping, command_usage, owo_count)]
+struct General;
+
+#[hook]
+async fn before(ctx: &Context, msg: &Message, command_name: &str) -> bool {
+    println!("Running command '{}' invoked by '{}'", command_name, msg.author.tag());
+
+    let counter_lock = {
+        // While data is a RwLock, it's recommended that you always open the lock as read.
+        // this is mainly done to avoid Deadlocks for having a possible writer waiting for multiple
```
arqunis:
```suggestion
        // This is mainly done to avoid Deadlocks for having a possible writer waiting for multiple
```

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+        // data, instead the reference is cloned.
+        // We wap every value on an arc, as to keep the data lock open for the least time possible,
+        // to again, avoid Deadlocking it.
+        data_read.get::<CommandCounter>().expect("Expected CommandCounter in TypeMap.").clone()
+    };
+
+    // Just like with client.data in main, we want to keep write locks open the least time
+    // possible, so we wrap them on a block so they get automatically closed at the end.
+    {
+        // The HashMap of CommandCounter is wrapped on RwLock; Since we want to write to it, we will
```
arqunis: Ditto. And additionally, the hashmap is contained *in* an `RwLock`, not placed on its top or somewhere on its head.
```suggestion
        // The HashMap of CommandCounter is wrapped in an RwLock; since we want to write to it, we will
```

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    async fn message(&self, ctx: Context, msg: Message) {
+        if msg.content.to_lowercase().contains("owo") {
+            // Since data is located in Context, this means you are also able to use it within events!
+            let count = {
+                let data_read = ctx.data.read().await;
+                data_read.get::<MessageCount>().expect("Expected MessageCount in TypeMap.").clone()
+            };
+
+            // Atomic operations with ordering do not require mut to be modified.
+            // in this case, we want to increase the message count by 1.
```
arqunis:
```suggestion
            // In this case, we want to increase the message count by 1.
```

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    let counter_lock = {
+        // While data is a RwLock, it's recommended that you always open the lock as read.
+        // this is mainly done to avoid Deadlocks for having a possible writer waiting for multiple
+        // readers to close.
+        let data_read = ctx.data.read().await;
+
+        // Since the CommandCounter Value is wrapped in an Arc, cloning will not duplicate the
+        // data, instead the reference is cloned.
+        // We wap every value on an arc, as to keep the data lock open for the least time possible,
```
arqunis: Ditto.
```suggestion
        // We wap every value on in an Arc, as to keep the data lock open for the least time possible,
```

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+        // to again, avoid Deadlocking it.
```
arqunis:
```suggestion
        // to again, avoid deadlocking it.
```

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+        .framework(framework)
+        .await
+        .expect("Err creating client");
+
+    // This is wherere we can initially insert the data we desire into the "global" data TypeMap.
+    // client.data is wrapped on a RwLock, and since we want to insert to it, we have to open it in
+    // write mode, but there's a small thing catch:
+    // There can only be a single writer to a given lock open in the entire application, this means
+    // you can't open a new write lock until the previous write lock has closed.
+    // This is not the case with Read locks, read locks can be open indefinitely, BUT as soon as
```
arqunis:
```suggestion
    // This is not the case with read locks, read locks can be open indefinitely, BUT as soon as
```

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+        Err(_) => {
+            msg.reply(ctx, "I require an argument to run this command.").await?;
+            return Ok(());
+        }
+    };
+
+    // Yet again, we want to keep the locks open for the least time possible.
+    let amount = {
+        // Since we only want to read the data and not write to it, we open it in read mode;
+        // and since this is open in read mode, it means that there can be multiple locks open at
```
arqunis: Here I suggest two ways:
1) Remove the conjunction *and*.
```suggestion
        // Since we only want to read the data and not write to it, we open it in read mode;
        // since this is open in read mode, it means that there can be multiple locks open at
```
The semicolon already does the job of joining two independent clauses together, removing the need for conjunctions such as *and*, *but*, *for*, etc..

2) Change the semicolon to a comma.
```suggestion
        // Since we only want to read the data and not write to it, we open it in read mode,
        // and since this is open in read mode, it means that there can be multiple locks open at
```

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+        // the data lock in read mode, as it will avoid a lot of possible deadlocks.
+        let data_read = ctx.data.read().await;
+
+        // Then we obtain the value we need from data, in this case, we want the command counter.
+        // The returned value from get() is an Arc, so the reference will be cloned, rather than
+        // the data.
+        let command_counter_lock = data_read.get::<CommandCounter>().expect("Expected CommandCounter in TypeMap.").clone();
+
+        let command_counter = command_counter_lock.read().await;
+        // and we return a usable value from it.
```
arqunis:
```suggestion
        // And we return a usable value from it.
```

arqunis (changes requested):

examples/e14_global_data/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+impl TypeMapKey for CommandCounter {
+    type Value = Arc<RwLock<HashMap<String, u64>>>;
+}
+
+struct MessageCount;
+
+impl TypeMapKey for MessageCount {
+    // While you will be using RwLock or Mutex most of the time you want to modify data,
+    // sometimes it's not required; like for example, with static data, or if you are using other
+    // kinds sof atomic operators.
```
arqunis:
```suggestion
    // kinds of atomic operators.
```

arqunis (changes requested):

arqunis (approved):

PR #1015, opened 2020-10-12
Update & reduce dependencies

nickelc:
- Switch to async-tungstenite/tokio-rustls
- futures only with the `std` feature drops the futures-executor dep

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ tracing = "0.1"
 tracing-futures = "0.2"
 serde_json = "1"
 async-trait = "0.1"
-futures = "0.3"
 
 [dependencies.static_assertions]
 optional = true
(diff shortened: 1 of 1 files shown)


arqunis:
Seems you will have to tweak CI as well. Enabling both `tokio-native-tls` and `tokio-rustls` features of `async_tungstenite` [results in a module conflict](https://github.com/serenity-rs/serenity/pull/1015/checks?check_run_id=1243705059#step:6:278). Unless a CI job is set up to build with only certain features, by default, we build jobs with all features. Subsequently, the aforementioned features clash because of this.

nickelc:
PR is ready. The feature clash is fixed in async-tungstenite.

arqunis (approved):

PR #1035, opened 2020-10-24
Changed `utils::shard_id` to accept `Into<u64>`

JellyWX:
This is beneficial as it allows you to pass GuildId objects into the shard_id function


JellyWX:
Need to squash those commits...

arqunis (approved):

Issue #888, opened 2020-06-18
Builtin std_framework checks only check for Member objects on cache


vicky5124:
Builtin checks like has_permissions only checks on Cache, this makes checks like has_permission trigger to false if a user is not in the cache, situation which can happen due to custom intents that disable PRESENCE.

Solution: make the checks fallback to Http if a user's Member is not in Cache.

Lakelezz:
This should be fixed as of #1036 and #1087!
Thanks for the issue.

Issue #1009, opened 2020-10-07
TODO Update list for API V8


vicky5124:
- [x] (???) Update intents to default to all but privileged; this is necessary as intents are required on V8
- [x] Do not cache Members by default, it can severely slow down the startup speed of the bot if the members and presence intents are not present.
- [x] (#1036) Depend less on cache internally (specially members due to privileged intents being mandatory), make most of the internal methods fall back to http if not present in cache.
- [x] (#1063) Permissions field is now a String.
- [x] (#1043) `game` fields have been replaced with the first element in activities.
- [x] (#1063) Channel Permission Overwirtes `type` field is now serialized as an integer rather than a String.
- [x] (#1043) `embed_enabled` and `embed_channel_id` have been replaced with this: https://discord.com/developers/docs/resources/guild#guild-widget-object
- [x] (Doesn't need fixing) Form error responses now have a format that is better aligned with "regular errors".
- [x] (#1043) Rate limit headers are sending seconds instead of milliseconds. 
- [x] (#1043) The X-RateLimit-Precision header is no longer respected.
- [x] (#1043) Bots no longer get CHANNEL_CREATE for private channels. Remove the private_channel_create event.
- [x] (#1043) Bans `delete-message-days` is replaced with `delete_message_days`
- [x] (#1043) Removed roles, premium_since, and nick from Presence Update gateway event.
- [x] (#991) Inline Replies.
- [x] (#1043) Remove `guild_subscriptions` as it isn't supported anymore due to v6 requiring intents now.
- [x] (#1042) Change the `Vec<GuildId>` to a singular GuildId in `chunk_guilds` related APIs as it only support one guild per request now.
- [x] (#1073) Update `VoiceState` to include `guild_id`, `member`, and `self_video` fields.
- [x] (Fixed Itself) Fix Sequence off warnings on PresenceUpdateEvent (#1082)

NOTE: This list may be non-exhaustive.

ikkerens:
To add, there are two things to uncomment in my pending PR: https://github.com/serenity-rs/serenity/pull/991

iirelu:
I don't know if it's related to this issue but sometime in the last few weeks, the Message::member method no longer works. Just returns None. Do I have to wait until this issue is fixed?

ikkerens:
> I don't know if it's related to this issue but sometime in the last few weeks, the Message::member method no longer works. Just returns None. Do I have to wait until this issue is fixed?

[...]

PR #1036, opened 2020-10-24
member_permissions without cache

JellyWX:
Still a bit of a work in progress. Changes here are:

* Member function `member_permissions_cache` which retains old behaviour
* `member_permissions` is now async with a Result<Permissions> return type
* `member_permissions` still only checks cache if cache is enabled; otherwise it checks on HTTP
* Functions such as `Guild::has_perms` and `Member::permissions` use the `member_permissions_cached` to retain old behaviour. 

It seems like it'll be a pain to blend this all under one since it requires changing a lot of arguments from `Cache` to `CacheHttp`, but this at least seems like a start.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -62,6 +62,7 @@ use super::{
 };
 #[cfg(all(feature = "cache", feature = "http"))]
 use crate::{
+    http::CacheHttp,
     cache::Cache,
     client::Context,
     framework::standard::CommonOptions,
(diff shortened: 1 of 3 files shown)


vicky5124:
Member::permissions and Guild::has_perms should probably be changed to also use Http
this would very likely fix #888, and it would be better for v8, as we need to depend less on the members cache, due to the privileged intents.

JellyWX:
@nitsuga5124 When you say v8, do you mean serenity 0.8 or webhook v8? I will have a look at changing those too, I just put this together fairly quickly to be able to open a PR concerning it

arqunis:
v8 refers to [version 8 of Discord's gateway](https://discord.com/developers/docs/topics/gateway#gateways-gateway-versions).

vicky5124:
You can see #1009 for the v8 switch to-do list, which includes depending less on members cache.

JellyWX:
kicked that together, still need to test it properly

JellyWX:
Sorry, been busy with other stuff. This seems to be working fine when the cache is still enabled but I haven't tested it on other configurations of features

arqunis (approved):

PR #1041, opened 2020-10-27
Add usable prefix to default configuration

Miezhiko:
Not sure but for me it looks more consistent since there is default `Delimiter` and I was thinking that default configuration should be usable without changes, now it's not seems like because on_mention is disabled and prefix is not set.
I've picked ! because I think that it's widely used.

### DIFF
--- a/src/framework/standard/configuration.rs
+++ b/src/framework/standard/configuration.rs
@@ -419,12 +419,11 @@ impl Configuration {
     ///     .prefix("!"));
     /// ```
     pub fn prefix(&mut self, prefix: &str) -> &mut Self {
-        self.prefixes =
-            if prefix.is_empty() {
-                vec![]
-            } else {
-                vec![prefix.to_string()]
-            };
+        self.prefixes = if prefix.is_empty() {
+            vec![]
+        } else {
+            vec![prefix.to_string()]
+        };
 
         self
     }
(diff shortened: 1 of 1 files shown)


Flat (changes requested):
This line needs updated to state the new default https://github.com/serenity-rs/serenity/blob/d565058380249e88bc98a583c17fe8a18efef908/src/framework/standard/configuration.rs#L407

Additionally most of the framework examples use `~` as a prefix. I think this either needs to be the default or the examples updated to `!` it should also be mentioned that you can set no prefix by passing an empty vec i.e. `.prefix(vec![])`

Miezhiko:
didn't knew about `~` will change on it tomorrow, some later

Flat (approved):

src/framework/standard/configuration.rs
```diff
@@ -417,7 +419,12 @@ impl Configuration {
     ///     .prefix("!"));
     /// ```
     pub fn prefix(&mut self, prefix: &str) -> &mut Self {
-        self.prefixes = vec![prefix.to_string()];
+        self.prefixes =
+            if prefix.is_empty() {
+                vec![]
+            } else {
+                vec![prefix.to_string()]
+            };
```
arqunis: Formatting nitpick.
```suggestion
        self.prefixes = if prefix.is_empty() {
            vec![]
        } else {
            vec![prefix.to_string()]
        };
```

arqunis (changes requested):

arqunis (approved):

PR #1043, opened 2020-10-29
Gateway v8: Initial Draft

james7132:
Initial changes for supporting v8, mostly low hanging fruit:

 * Updated base API URI to use v8 instead of v6.
 * Removed roles, premium_since, and nick from PresenceUpdateEvent.
 * Removed Presence.activity since it is no longer being populated (Use Presence.activities instead).
 * Remove `private_channel_create` event as these will no longer be issued by Discord anymore.
 * Changed delete-message-days query param to delete_message_days for bans.
 * Update embed_channel_id and embed_enabled to widget_channel_id and widget_enabled. 
 * Remove guild_subscriptions as it is now a noop in v8.
 * Updated rate limit handling to read the updated rate limit headers properly. Removed X-RateLimit-Precision header on outgoing HTTP requests.
 * Update deserialization of PermissionOverwrites to use the integer based values instead of "member" and "role". This may break Audit Log serialization as they will be string serialized there.
 * Changed how gateway intents are specified:
   * Default ClientBuilder's intents to all but privileged intents.
   * Removed ClientBuilder.add_intent (either need to remove it and rely entirely on ClientBuilder.intents or add a  remove_intent function)
   * Made all references to GatewayIntents in Shard* types and ClientBuilder non-optional.


src/client/dispatch.rs
```diff
@@ -337,13 +337,6 @@ async fn handle_event(
             // This could potentially be annoying to handle when otherwise wanting to normally take care of a new channel.
```
arqunis: This comment should be removed as it doesn't apply anymore.

src/model/gateway.rs
```diff
@@ -567,10 +547,8 @@ impl Serialize for Presence {
         }
 
         let mut state = serializer.serialize_struct("Presence", 5)?;
```
arqunis: The number should be updated to mirror the number of fields that are going to be serialized.
```suggestion
        let mut state = serializer.serialize_struct("Presence", 3)?;
```

arqunis (changes requested):

src/client/dispatch.rs
james7132: Done.

src/model/gateway.rs
james7132: Done

james7132:
Tests should be fixed and comments addressed. PTAL.

arqunis (approved):

PR #1053, opened 2020-11-01
Event type

casey:
This adds a method to get the `EventType` of an `Event`. I needed this for logging and statistics. Hopefully it's useful to others!

For some reason I was getting compilation failures, but adding the lifetime fixed them.

I also added `EventType::name`, which returns the screaming-snake-case name as originally returned by the API. This doesn't apply cleanly on its own, so I included it in this PR, since it's related.


casey:
The error on beta and nightly is very similar to the compilation error I was getting. Has that been seen before?

arqunis:
No. The lifetime you added is also wrong. The `prefix` function is meant to return a `Cow` that references the message's content (stored inside `Stream`). Now the `Cow` also references `Context`, which it doesn't have to. Could you remove the lifetime?

casey:
Will do, I'll push a diff in a second.

One question, would you prefer that event names, since they appear in multiple places, be factored into constants? I.e. `GUILD_ROLE_DELETE` instead of `"GUILD_ROLE_DELETE"`.

arqunis:
Sure.

casey:
Done, I removed the lifetime and refactored event name strings into constants. I'm not getting the compilation error anymore, so it must have been something wrong on my end.

src/model/event.rs
```diff
@@ -1979,6 +2025,101 @@ pub enum EventType {
...
 }
 
+impl From<&Event> for EventType {
+    fn from(event: &Event) -> EventType {
+        event.event_type()
+    }
+}
+
+impl EventType {
+    const CHANNEL_CREATE: &'static str = "CHANNEL_CREATE";
```
arqunis, line 2035: The compiler can deduce the lifetime of references in `const`s and `static`s by itself. You should remove the `'static` lifetimes to make the code more aesthetically appealing.
```suggestion
    const CHANNEL_CREATE: &str = "CHANNEL_CREATE";
```

arqunis (changes requested):

casey:
I believe that currently, it is only possible to omit the static lifetime for normal constants, not associated constants. (See [this issue](https://github.com/rust-lang/rust/issues/38831).)

As an alternative, I could make them not associate constants, i.e. move them out of `impl EventType` and into the containing module.

arqunis:
That's a pity. I thought that works for associated constansts as well, but it's apparently intentional that it doesn't. You can ignore my change, your pull request is fine as is. I'll merge it later. And once I also investigate the error, it is happening in CI again, this time just for the framework example...

arqunis:
I think we may have a case of a [heisenbug](https://en.wikipedia.org/wiki/Heisenbug) tormenting us. I've tried testing from my copy of serenity, your copy of serenity, switching between the stable, beta, and nightly toolchains, yet I am unable to consistently reproduce the error.  I also just tried restarting all of the CI jobs and the error has disappeared.

Well, I guess I have the liberty to merge your pull request, but the error is strange nonetheless.

arqunis (approved):

casey:
Very weird! I know I saw it locally a few times, so it must not be CI-specific.

Thank you for reviewing!

PR #1063, opened 2020-11-07
Fix serialization of `PermissionOverwrite`s

arqunis:
~~Despite [Discord claiming that the type of `PermissionOverwrites` is an integer](https://discord.com/developers/docs/resources/channel#overwrite-object-overwrite-structure) in v8, it is still a string like before. This reverses the change made in https://github.com/serenity-rs/serenity/pull/1043 to deserialize strings again. It also adds a visitor to future-proof the code once Discord decides to actually update the data type.~~ #1064 actually fixed the issue. This pull request has been repurposed to fix serialization of `PermissionOverwrite`s instead.

### DIFF
--- a/src/model/channel/mod.rs
+++ b/src/model/channel/mod.rs
@@ -377,8 +377,8 @@ impl<'de> Deserialize<'de> for PermissionOverwrite {
         let data = PermissionOverwriteData::deserialize(deserializer)?;
 
         let kind = match &data.kind {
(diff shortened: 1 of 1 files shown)


FelixMcFelix:
Not yet sure this is the cause. Adding `?v=8` to the gateway URL and testing from there at the moment.

FelixMcFelix:
#1064 should fix the root cause.

Issue #1066, opened 2020-11-08
`add_file` associated fns for `ExecuteWebhook`


JellyWX:
Since files can be sent over webhooks as attachments, it would make sense to have an associated function to attach a file to `ExecuteWebhook` similar to the method/s in `CreateMessage`

PR #1071, opened 2020-11-09
Sending files with webhooks

JellyWX:
To address issue #1066 .

I only set this up for one file since I didn't realize initially that webhooks support multiple files

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -34,6 +34,7 @@ use tokio::{
     io::AsyncReadExt,
     fs::File,
 };
+use crate::http::routing::Route;
 
 pub struct Http {
     pub(crate) client: Arc<Client>,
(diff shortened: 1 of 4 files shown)


src/http/client.rs
```diff
@@ -883,6 +884,93 @@ impl Http {
...
+    ///
+    /// [`HttpError::UnsuccessfulRequest`]: enum.HttpError.html#variant.UnsuccessfulRequest
+    pub async fn execute_webhook_with_files<'a, T, It: IntoIterator<Item=T>>(
+        &self,
+        webhook_id: u64,
+        token: &str,
+        wait: bool,
+        files: It,
+        map: JsonMap) -> Result<Option<Message>>
+        where T: Into<AttachmentType<'a>> {
```
arqunis: Formatting nitpick.

```suggestion
        map: JsonMap
    ) -> Result<Option<Message>>
        where 
            T: Into<AttachmentType<'a>>,
    {
```

src/http/client.rs
```diff
@@ -883,6 +884,93 @@ impl Http {
...
+        let mut multipart = reqwest::multipart::Form::new();
+
+        for (file_num, file) in files.into_iter().enumerate() {
+
```
```suggestion
```

arqunis (changes requested):

JellyWX:
that should resolve changes. there's a couple of notes id like to add though (since I could resolve them/change what ive done off what you suggest):

* the code is adapted from the `send_files` method. However, to avoid using a new `unsafe` block, I used `files.into_iter().enumerate()`. it would be trivial to modify `send_files` to remove the unsafe block from there, but i presume it is there for a good reason? see:


        let mut file_num = "0".to_string();

        for file in files {
            ...
            
            unsafe {
                let vec = file_num.as_mut_vec();
                vec[0] += 1;
            }
        }


* also, `send_files` has this loop towards the end (which is also accompanied by some code in the calling method that modifies the map before it is passed in). I have replaced both of those with a single line in the webhook one. see original:


        for (k, v) in map {
            match v {
                Value::Bool(false) => multipart = multipart.text(k.clone(), "false"),
                Value::Bool(true) => multipart = multipart.text(k.clone(), "true"),
                Value::Number(inner) => multipart = multipart.text(k.clone(), inner.to_string()),
                Value::String(inner) => multipart = multipart.text(k.clone(), inner),
                Value::Object(inner) => multipart = multipart.text(k.clone(), serde_json::to_string(&inner)?),
                _ => continue,
            };
        }


see modified:
 

        multipart = multipart.text("payload_json", serde_json::to_string(&map)?);

arqunis (approved):
Tested locally, function works as expected. You can apply those changes you did to the normal `send_files` function if you wish.

arqunis (approved):

PR #1072, opened 2020-11-10
Only add roles that the member does not have

### LINKED ISSUE #1046: member#add_roles do not check if the member already have the given roles
The method `member#add_role` (single) starts by checking if the member already have the role and exit if it's the case. However, this is not the case with the `member#add_roles` (multiple) method : 

This method starts by adding the roles given as parameters to the member's role list without checking for duplicates. This will cause the following call to `edit_member` to fail if there is a duplicate (for exemple, if a role already granted to the user is passed to this method). 

As a result, the method will then remove all roles given in parameters from the member's role list. This is bad because the duplicated role will now be absent from the list (while the user on discord will still have it).

I would suggest to check for duplicates when adding the roles in the `role_ids` parameter to the member's role list.
Also, it would better to add them to the list once the function succeed instead of adding them at the begining no matter what and then attempting to remove them if the function fails.

This bug was experienced on the version 0.9.0-rc4 of Serenity.

arqunis:
Should fix #1046.

This is applied in 0.9. The commit will be cherry-picked onto `current` and `next` when it is merged.

### DIFF
--- a/src/model/guild/member.rs
+++ b/src/model/guild/member.rs
@@ -82,7 +82,11 @@ impl Member {
     /// [`Role`]: struct.Role.html
     /// [Manage Roles]: ../permissions/struct.Permissions.html#associatedconstant.MANAGE_ROLES
     pub async fn add_roles(&mut self, http: impl AsRef<Http>, role_ids: &[RoleId]) -> Result<()> {
-        self.roles.extend_from_slice(role_ids);
+        for id in role_ids {
+            if !self.roles.contains(id) {
+                self.roles.push(*id);
+            }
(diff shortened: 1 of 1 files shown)


SadiinsoSnowfall:
I think it would be better to add the roles only if the call to edit_member is successful (as it is done in the add_role method).

SadiinsoSnowfall:
You could also create a Vec capable of holding all the current roles as well as the new ones.
The current code will cause a re-allocation of the `roles` vector nearly every time (if one of the new roles is not already contained in `self.roles`)

PR #1073, opened 2020-11-10
Update VoiceState model

sam-kirby:
The VoiceState model now contains the following additional fields:
* guild_id (optional)
* member (optional)
* self_video

Due to the member object as serialized by Discord lacking a guild_id
field, a custom deserializer was implemented.

When a member object is received, it is used to update the cache. This
is particularly useful when considering bots lacking priveleged intents,
as caching this object can prevent the need for future API calls.

This also allows for the correct functioning of the
GuildChannel::members function in the absence of the Members intent.

A similar approach would bring similar advantages if employed regarding
MessageCreate and related events, though this is likely to be a breaking
change. Consider adding to #1009.


src/model/voice.rs
```diff
@@ -49,13 +58,229 @@ impl fmt::Debug for VoiceState {
...
+            self_stream,
+            self_video,
+            session_id,
+            suppress,
+            token,
+            user_id,
+            _nonexhaustive: ()
+        })
+    }
+}
```
arqunis: These should be placed inside the `deserialize` function below. Outside of deserialization of `VoiceState`, these types are useless and pollute the module's namespace.

arqunis (changes requested):

arqunis (approved):

PR #1074, opened 2020-11-11
Songbird driver configuration implemented

FelixMcFelix:
This adds user control over the encryption scheme negotiated between the driver and discord (and implements the remaining variants, `xsalsa20_poly1305_suffix` and `xsalsa20_poly1305_lite`), the number of tracks to preallocate, and whether any decoding/decryption work is performed on received packets. These can be set for all `Call`s, or on a per-`Call`/`Driver` basis.

This causes breaking changes on RTP and RTCP packet events, as there may be no decoded packet (-> `Option<...>`) and that received packets may have a trailing encryption nonce (`packet_end_pad`).

### DIFF
--- a/songbird/src/driver/config.rs
+++ b/songbird/src/driver/config.rs
@@ -44,38 +44,38 @@ pub struct Config {
 }
 
 impl Default for Config {
-	fn default() -> Self {
-		Self {
-			crypto_mode: CryptoMode::Normal,
-			decode_mode: DecodeMode::Decrypt,
-			preallocated_tracks: 1,
-		}
-	}
+    fn default() -> Self {
+        Self {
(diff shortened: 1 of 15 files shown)


arqunis (approved):

songbird/src/driver/config.rs
```diff
@@ -1,10 +1,82 @@
-use super::CryptoMode;
+use super::{CryptoMode, DecodeMode};
 
 /// Configuration for the inner Driver.
 ///
 /// At present, this cannot be changed.
```
arqunis: Now it can be.
```suggestion
```

arqunis (changes requested):

arqunis (approved):

PR #1076, opened 2020-11-13
Provide youtube-dlc as an alternative to youtube-dl 

peppizza:
youtube-dl's progress has slowed down and will very often error with `ERROR: Unable to extract JS player URL`.  youtube-dlc is a drop-in replacement for youtube-dl that is being updated and does not have that error https://github.com/blackjack4494/yt-dlc

### DIFF
--- a/songbird/Cargo.toml
+++ b/songbird/Cargo.toml
@@ -107,6 +107,7 @@ default = [
     "serenity-rustls",
     "driver",
     "gateway",
+    "youtube-dlc"
 ]
 gateway = [
     "flume",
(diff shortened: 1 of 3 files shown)


vicky5124:
this could probably be made into a feature, to let users choose between either one, and not make this a breaking change.
youtube-dl is more accessible as it's available on most linux package managers, so it's support should stay.

peppizza:
maybe, however the error that i posted above is very frequent with youtube-dl and it's not clear to a user of the library what the issue is

vicky5124:
it could be mentioned in a documentation line, as a note

peppizza:
youtube-dl and youtube-dlc are now both features with youtube-dlc being the default

Flat:
As the tried and tested program `youtube-dl` should probably be the default. Claims that `youtube-dlc` is more maintained would also need some further information.

Lakelezz:
I do agree that `youtube-dl` should stay our default for now, suddenly changing the default to a new fork sounds adventurous.

peppizza:
youtube-dl is now the default feature and I updated the compile_error message to reflect it

peppizza:
moving to https://github.com/serenity-rs/songbird

PR #1085, opened 2020-11-23
fix: strikethrough tip macro typo in examples

thebongy:
Found a small documentation issue in the frameworks example
https://docs.rs/serenity/0.8.6/serenity/framework/standard/struct.HelpOptions.html#structfield.strikethrough_commands_tip_in_dm

There are also several other macros defined in https://docs.rs/serenity/0.8.6/serenity/framework/standard/macros/attr.help.html , I can also update the example to include those, let me know if I should include that too :)


arqunis (approved):
>  let me know if I should include that too :)

That won't be necessary, but thanks.

PR #1086, opened 2020-11-26
Add functions to use inlined replies on `Message`

Lakelezz:
This pull request adds `reply`, `reply_ping`, and `reply_mention` to the `Message`-struct.
- `reply` is an inlined reply that won't ping.
- `reply_ping` is a pinging inlined reply.
- `reply_mention` performs a reply based on the old Serenity format: `@user {text}`.

### DIFF
--- a/src/builder/create_allowed_mentions.rs
+++ b/src/builder/create_allowed_mentions.rs
@@ -15,7 +15,7 @@ pub enum ParseValue {
     Roles,
 }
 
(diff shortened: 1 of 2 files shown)


iddm:
Will you please release `0.9.3` after this?

evieluvsrainbows (approved):
Tested this last night; everything is confirmed to be working well.

Issue #1088, opened 2020-11-27
Expose option<guildId> in message_delete event handler


SadiinsoSnowfall:
The current handler for the MESSAGE_DELETE event is the following : 
`message_delete(&self, ctx: Context, channel_id: ChannelId, deleted_message_id: MessageId)`

But the discord API does send a `Option<GuildId>` in this event payload: (https://discord.com/developers/docs/topics/gateway#message-delete)

Furthermore this field is already parsed and stored in the event definition (see [Events.rs/MessageDeleteEvent](https://github.com/serenity-rs/serenity/blob/current/src/model/event.rs#L931)).

I think this field should be exposed in the event handler (by modifying [event_handler.rs](https://github.com/serenity-rs/serenity/blob/current/src/client/event_handler.rs#L239) and [dispatch.rs](https://github.com/serenity-rs/serenity/blob/current/src/client/dispatch.rs#L614) accordingly).

There is also a similar design issue with the MESSAGE_DELETE_BULK event handler.

Issue found on Serenity 0.9.1

SadiinsoSnowfall:
I can open a pull request if needed. The problem is that this would be a breaking change for all users implementing the message_delete and message_delete_bulk functions of the EventHandler trait (though somewhat easy to fix).

arqunis:
You are free to open a pull request. Make sure you target the `current` branch, `v0.9.x` cannot accept breaking changes.

PR #1090, opened 2020-11-28
Fix for #1088 - GuildId in message_delete handler

SadiinsoSnowfall:
The changes pass all the tests when using `cargo test --all-features`.
I also have tested this version on a real discord bot to ensure that it works (both in DMs and in a guild).

### DIFF
--- a/src/client/event_handler.rs
+++ b/src/client/event_handler.rs
@@ -235,12 +235,12 @@ pub trait EventHandler: Send + Sync {
 
     /// Dispatched when a message is deleted.
     ///
-    /// Provides the channel's id and the message's id.
+    /// Provides the guild's id, the channel's id and the message's id.
     async fn message_delete(&self, _ctx: Context, _channel_id: ChannelId, _deleted_message_id: MessageId, _guild_id: Option<GuildId>) {}
 
     /// Dispatched when multiple messages were deleted at once.
     ///
-    /// Provides the channel's id and the deleted messages' ids.
+    /// Provides the guild's id, channel's id and the deleted message's ids.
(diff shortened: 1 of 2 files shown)


src/client/event_handler.rs
```diff
@@ -235,13 +235,13 @@ pub trait EventHandler: Send + Sync {
...
-    /// Provides the channel's id and the message's id.
-    async fn message_delete(&self, _ctx: Context, _channel_id: ChannelId, _deleted_message_id: MessageId) {}
+    /// Provides the guild's id, the channel's id and the message's id.
+    async fn message_delete(&self, _ctx: Context, _channel_id: ChannelId, _deleted_message_id: MessageId, _guild_id: Option<GuildId>) {}
 
     /// Dispatched when multiple messages were deleted at once.
     ///
-    /// Provides the channel's id and the deleted messages' ids.
-    async fn message_delete_bulk(&self, _ctx: Context, _channel_id: ChannelId, _multiple_deleted_messages_ids: Vec<MessageId>) {}
+    /// Provides the guild's id, channel's id and the deleted message's ids.
```
arqunis: The apostrophe's position is intentional for `messages` to signify possessiveness of a plural noun, since a single message does not have many identifiers.
```suggestion
    /// Provides the guild's id, channel's id and the deleted messages' ids.
```

arqunis (changes requested):

SadiinsoSnowfall:
Ok, it should be good now

arqunis (approved):

evieluvsrainbows (approved):

PR #1091, opened 2020-11-28
Add Default Limit And Validate When Fetching Members

### LINKED ISSUE #1050: Guild#members default limit & validation
There is no verification done on the `limit` parameter of the `Guild#members` method. This will cause the call to `Client#get_guild_members` to fail if its value is invalid (not in the 1-1000 range) without any explanation (because Discord will not return a valid json value in this case).

Also, the default value for the `limit` parameter in the discord API (see https://discord.com/developers/docs/resources/guild#list-guild-members) is 1. This means that if `None` is passed to `Guild#members`, the function will only query one member. I think this should be changed to 1000 (the maximum value, because `None` means "no limit" in this case if I understood correctly) by default or at least be documented.

This issue was experienced on Serenity 0.9.0

SadiinsoSnowfall:
Fix #1050 by using the maximum value (1000) as the default value for the member fetch limit and by validating the limit value before sending the request.
This version passed all the tests and ran successfully on a real discord bot.

This may be considered a breaking change as it will change the behaviour of the `guild_id::members`, `Guild::members` and `PartialGuild::members` functions when passing an invalid value or None as the limit.


src/error.rs
```diff
@@ -71,6 +71,11 @@ pub enum Error {
     /// *This only exists for the `GuildId::ban` and `Member::ban` functions. For their cases,
     /// it's the "reason".*
     ExceededLimit(String, u32),
+    /// Input is not in the specified range
```
Lakelezz:
```suggestion
    /// The input is not in the specified range.
```
Just a small nit.

Lakelezz (changes requested):
Thanks for your pull request. 
I just put in some small improvements to the wording of the original doc, but since we are already there, let's improve them as well!

Lakelezz:
Thanks for merging my suggestions, make sure to resolve the conflict!

SadiinsoSnowfall:
My apologies for the delay, it should be good now.

evieluvsrainbows (approved):

Lakelezz:
Thank you for your pull request!

PR #1098, opened 2020-11-30
Replace `CheckResult` with a simple `Result<(), Reason>`

Lakelezz:
The goal is to simplify the API by using the standard library `Result`.
There is no reason to provide a different type for the check-system.

### DIFF
--- a/command_attr/src/lib.rs
+++ b/command_attr/src/lib.rs
@@ -750,7 +750,7 @@ pub fn check(_attr: TokenStream, input: TokenStream) -> TokenStream {
 
     propagate_err!(create_declaration_validations(&mut fun, DeclarFor::Check));
 
-    let res = parse_quote!(serenity::framework::standard::CheckResult);
+    let res = parse_quote!(Result<(), serenity::framework::standard::Reason>);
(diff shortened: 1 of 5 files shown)


command_attr/src/lib.rs
```diff
@@ -750,7 +750,7 @@ pub fn check(_attr: TokenStream, input: TokenStream) -> TokenStream {
 
     propagate_err!(create_declaration_validations(&mut fun, DeclarFor::Check));
 
-    let res = parse_quote!(serenity::framework::standard::CheckResult);
+    let res = parse_quote!(Result<(), serenity::framework::standard::Reason>);
```
arqunis: The absolute path for `Result` should be used in case a type alias is used with the same name.
```suggestion
    let res = parse_quote!(std::result::Result<(), serenity::framework::standard::Reason>);
```

arqunis (changes requested):

peppizza:
I am also for this change as I think using the ? operator inside of checks would be incredibly useful to easily pass down a check failure

arqunis (approved):

PR #1099, opened 2020-12-01
Fix #1092 - MessageTooLong documentation

### LINKED ISSUE #1092: Invalid Documentation About `MessageTooLong`
The documentation for the MessageTooLong  states the following : 
>Indicates that a [`Message`]s content was too long and will not successfully send, as the length is over 2000 codepoints, or 4000 bytes. 
The number of bytes larger than the limit is provided.

But this is not what is computed/returned by the [Message::overflow_length](https://github.com/serenity-rs/serenity/blob/current/src/model/channel/message.rs#L466) function (the function whose return value is used to create said error).

The function is in fact checking that the message length __in characters__ is not greater than 2000, but isn't checking the length in bytes in any way. Furthermore, the return value represent the overflow in characters, not bytes (as it is explained in the function documentation).

Also, I think the `overflow_length` function can be refactored to return an `usize` or an `Option<usize>` instead of an `Option<u64>` and the `MESSAGE_CODE_LIMIT` constant type can be changed to usize. This is because the size of a String is usually measured in usize and the said constant is always casted to an i64 or an u64 when used.

The part of the documentation about the 4000 bytes limit should be either removed, or this limit need to be enforced in the  `overflow_length` function (note that a character is not always 2 bytes long, see [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d1cfebc533cb8e93562bacece669a6e1)).

SadiinsoSnowfall:


src/model/error.rs
```diff
@@ -115,9 +115,9 @@ pub enum Error {
     /// [`Cache`]: ../../cache/struct.Cache.html
     ItemMissing,
     /// Indicates that a [`Message`]s content was too long and will not
-    /// successfully send, as the length is over 2000 codepoints, or 4000 bytes.
+    /// successfully send, as the length is over 2000 codepoints.
     ///
-    /// The number of bytes larger than the limit is provided.
+    /// The number of characters larger than the limit is provided.
```
Lakelezz: Use the term codepoints instead:
```suggestion
    /// The number of code points larger than the limit is provided.
```

Lakelezz (approved):
Thank you for the issue and even following up with yet another pull request. I appreciate it :)

SadiinsoSnowfall:
You're welcome

PR #1100, opened 2020-12-01
Switched from u64 to usize to represent overflow (#1092)

SadiinsoSnowfall:
This PR is a breaking change as it changes the inner type of two Errors (EmbedTooLarge and MessageTooLong) from u64 to usize.

I think a test target should be added to ensure that the `overflow_length` function remains correct but there is nothing about adding unit tests in `contribute.md`.

### DIFF
--- a/src/model/error.rs
+++ b/src/model/error.rs
@@ -74,7 +74,7 @@ pub enum Error {
     DeleteMessageDaysAmount(u8),
     /// Indicates that the textual content of an embed exceeds the maximum
     /// length.
-    EmbedTooLarge(u64),
+    EmbedTooLarge(usize),
     /// An indication that a [guild][`Guild`] could not be found by
     /// [Id][`GuildId`] in the [`Cache`].
     ///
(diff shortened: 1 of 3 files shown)


src/model/error.rs
```diff
@@ -115,12 +115,12 @@ pub enum Error {
     /// [`Cache`]: ../../cache/struct.Cache.html
     ItemMissing,
     /// Indicates that a [`Message`]s content was too long and will not
-    /// successfully send, as the length is over 2000 codepoints, or 4000 bytes.
+    /// successfully send, as the length is over 2000 codepoints.
     ///
-    /// The number of bytes larger than the limit is provided.
+    /// The number of characters larger than the limit is provided.
```
Lakelezz:
```suggestion
    /// The number of code points larger than the limit is provided.
```

Lakelezz (approved):
Thank you for the issue and even following up with a pull request, I appreciate it!

arqunis (approved):

PR #1102, opened 2020-12-02
Make reason an error type

peppizza:
Now that CheckResult is gone, Reason should implement the Error trait to allow for result chaining

### DIFF
--- a/src/framework/standard/structures/check.rs
+++ b/src/framework/standard/structures/check.rs
@@ -3,8 +3,7 @@ use crate::framework::standard::{Args, CommandOptions};
 use crate::model::channel::Message;
 use futures::future::BoxFuture;
 use std::error::Error;
-use std::fmt;
-use std::fmt::{Debug, Display};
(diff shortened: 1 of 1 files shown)


src/framework/standard/structures/check.rs
```diff
@@ -1,9 +1,10 @@
-use std::fmt::Debug;
-use std::fmt;
-use crate::model::channel::Message;
 use crate::client::Context;
 use crate::framework::standard::{Args, CommandOptions};
+use crate::model::channel::Message;
 use futures::future::BoxFuture;
+use std::error::Error;
+use std::fmt;
+use std::fmt::{Debug, Display};
```
arqunis: Import format nitpick.
```suggestion
use std::fmt::{self, Debug, Display};
```

arqunis (changes requested):

arqunis (approved):

Flat (approved):

PR #1109, opened 2020-12-04
Fix clippy warning

nickelc:
clippy::write_with_newline
```
help: use `writeln!()` instead
     |
1042 |         let _ = writeln!(group_text, );
     |                 ^^^^^^^             --
```

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -1039,7 +1039,7 @@ fn flatten_group_to_string(
     };
 
     if summary_or_prefixes {
-        let _ = write!(group_text, "\n");
+        let _ = writeln!(group_text,);
     }
 
     let mut joined_commands = group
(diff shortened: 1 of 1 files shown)


src/framework/standard/help_commands.rs
```diff
@@ -1039,7 +1039,7 @@ fn flatten_group_to_string(
     };
 
     if summary_or_prefixes {
-        let _ = write!(group_text, "\n");
+        let _ = writeln!(group_text,);
```
arqunis:
```suggestion
        let _ = writeln!(group_text);
```

arqunis (changes requested):

src/framework/standard/help_commands.rs
nickelc: i thought the comma is required but yeah it's not.

arqunis (approved):

PR #1123, opened 2020-12-11
Rework buckets to offer fine-grained control

### LINKED ISSUE #1117: Rate Limit buckets rework
Not having them be guild specific will lead to commands being blocked in one server, because it was run in another

Lakelezz:
This pull requests closes #1117. 

Buckets can now apply to channels, users, guilds, categories, or globally.
The check type for buckets has been simplified to `&Message`. The signature no longer provides the optional guild ID, channel ID, and user ID.

This pull requests adds a utility method on `Message` and the `Cache` to retrieve the channel category ID. This addition is being used inside the new bucket code.

Using a bucket applying to categories requires the cache.
Global buckets will count every invocation of a command by every source.

By default, a bucket will work as before, limit per user.

Additionally, example 5 has been updated showcasing the new API addition for buckets.

Please help me testing this! :)

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -750,6 +750,12 @@ impl Message {
         ReactionCollectorBuilder::new(shard_messenger).message_id(self.id.0)
     }
 
+    /// Retrieves the message channel's category ID if the channel has such.
+    #[cfg(feature = "cache")]
+    pub async fn category_id(&self, cache: impl AsRef<Cache>) -> Option<ChannelId> {
+        cache.as_ref().channel_category_id(self.channel_id).await
+    }
+
     pub(crate) fn check_content_length(map: &JsonMap) -> Result<()> {
         if let Some(content) = map.get("content") {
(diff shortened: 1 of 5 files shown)


peppizza:
I tested each bucket in numerous scenarios and they seem to be working as intended so unless someone else finds a bug I think this can be merged.

Thanks for the awesome work @Lakelezz!

Lakelezz:
Thanks for your review :)!

src/model/channel/message.rs
```diff
@@ -750,6 +750,12 @@ impl Message {
         ReactionCollectorBuilder::new(shard_messenger).message_id(self.id.0)
     }
 
+    /// Retrieves the message channel's category ID if the channel has such.
```
Flat:
```suggestion
    /// Retrieves the message channel's category ID if the channel has one.
```

Flat (approved):
Lgtm

arqunis (approved):
Other than Flat's documentation change, this looks good.

PR #1124, opened 2020-12-14
docs: Clarify framework configuration dynamic_prefix docs

### LINKED ISSUE #898: Confusing dynamic_prefix documentation
The description for [`dynamic_prefix`](https://docs.rs/serenity/0.8.6/serenity/framework/standard/struct.Configuration.html#method.dynamic_prefix) says:

> Return None to not have a special prefix for the dispatch, and to instead use the inherited prefix.

Because of the way this is worded, I took this to mean that if the function returns Some it will not use the inherited prefix. (And I am taking "inherited prefix" to mean the prefix that is set by calling `.prefix()` on `Configuration`).

However this is not the case, as the function uses the inherited prefix regardless of what this function returns. If `dynamic_prefix` returns Some, both the inherited prefix and the dynamic prefix are acceptable as prefixes.

What makes this misleading is the usage of the word "instead". By saying "Return None to _instead_ use an inherited prefix", that implies that using the inherited prefix wasn't happening before. In other words, it implies that using the inherited prefix is contingent on `dynamic_prefix` returning None.

Proposed change:

    Sets the prefix to respond to dynamically, in addition to the one configured with [`prefix`].

    Return None to not have a special prefix for the dispatch.

    Note: Defaults to no dynamic prefix check.

(Note the omission of the language "inherited prefix", as I think this is unclear as to the implications of what that means- instead being explicit and mentioning the `prefix` method.)

I think it would also be pertinent to add a note to the description of `prefix`, something like:

    This prefix will always be usable, even if there is a [`dynamic_prefix`] configured.

drklee3:
Clarifies use of `dynamic_prefix` and configured prefix via `prefix()` and `prefixes()`.  Also adds information relating to the default prefix of `~`.

Closes #898

### DIFF
--- a/src/framework/standard/configuration.rs
+++ b/src/framework/standard/configuration.rs
@@ -283,19 +283,26 @@ impl Configuration {
         self
     }
 
-    /// Sets the prefix to respond to dynamically based on conditions.
+    /// Sets the prefix to respond to dynamically, in addition to the one
+    /// configured with [`prefix`] or [`prefixes`].
     ///
-    /// Return `None` to not have a special prefix for the dispatch, and to
-    /// instead use the inherited prefix.
+    /// Return `None` to not have a special prefix for the dispatch and to
+    /// instead use the configured prefix from [`prefix`] or [`prefixes`].
     ///
     /// This method can be called many times to add more dynamic prefix hooks.
     ///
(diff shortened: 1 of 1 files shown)


Lakelezz (approved):

Lakelezz:
Thank you a lot for this contribution! :)

PR #1127, opened 2020-12-16
Add Reverting and Awaiting Bucket Rate Limits

Lakelezz:
There are two new features introduced in this commit:
The first one allows to await a rate limit instead of cancelling the command invocation.
The second one allows to return a ticket to the rate limit bucket.

**Awaiting Rate Limits**
Buckets can now be configured to return an action. The action indicates how a framework should react.
This means, buckets can indicate to cancel or delay command invocation.
In order to set this feature, call the chain-method `await_ratelimits` on the bucket builder.

**Returning Tickets**
If a command never reaches the critical zone that requires a rate limit, we can now undo and return the ticket.
This can be down by returning `RevertBucket` from a command.
The concept is simple, a command may perform an expensive computation or issue a request via web API.
However the command may never reach that part, imagine incorrect user input.
The command would still go on cooldown, which can be a pain for the user.
By returning `RevertBucket`, the cooldown can be undone.

**Bug Fixes**
Following bugs have been discovered and fixed:
- `set_time` being `Option<Instant>` allowed for it being `None` at creation of a new bucket, which allowed twice the ticket count. It's just `Instant` *without* `Option` now.
- If no `limit` is set on the `BucketBuilder`, buckets will always return a rate limit for command invocation. While technically not wrong, logically and ergonomically useless. They now default to the limit of 1.

Please help testing this pull request :)

### DIFF
--- a/src/framework/standard/mod.rs
+++ b/src/framework/standard/mod.rs
@@ -12,11 +12,13 @@ pub use args::{Args, Delimiter, Error as ArgError, Iter, RawArguments};
 pub use configuration::{Configuration, WithWhiteSpace};
 pub use structures::*;
 
-use structures::buckets::Bucket;
+use structures::buckets::{Bucket, BucketAction};
 pub use structures::buckets::BucketBuilder;
 
 use parse::{ParseError, Invoke};
 use parse::map::{CommandMap, GroupMap, Map};
+
+use self::buckets::RevertBucket;
 
 use super::Framework;
 use crate::client::Context;
@@ -273,14 +275,31 @@ impl StandardFramework {
             return Some(DispatchError::BlockedChannel);
         }
 
-        {
-            let mut buckets = self.buckets.lock().await;
+        // Try passing the command's bucket.
+        // We exit the loop if no command ratelimit has been hit or we
+        // early-return when ratelimits cancel the framework invocation.
+        // Otherwise, we delay and loop again to check if we passed the bucket.
+        loop {
+            let mut duration = None;
 
-            if let Some(ref mut bucket) = command.bucket.as_ref().and_then(|b| buckets.get_mut(*b)) {
+            {
+                let mut buckets = self.buckets.lock().await;
 
-                if let Some(rate_limit) = bucket.take(ctx, msg).await {
-                    return Some(DispatchError::Ratelimited(rate_limit))
+                if let Some(ref mut bucket) = command.bucket.as_ref().and_then(|b| buckets.get_mut(*b)) {
+
+                    if let Some(bucket_action) = bucket.take(ctx, msg).await {
+
+                        duration = match bucket_action {
+                            BucketAction::CancelWith(duration) => return Some(DispatchError::Ratelimited(duration)),
+                            BucketAction::DelayFor(duration) => Some(duration),
+                        };
+                    }
                 }
+            }
+
+            match duration {
+                Some(duration) => tokio::time::delay_for(duration).await,
+                None => break,
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/framework/standard/structures/buckets.rs
```diff
@@ -84,34 +140,98 @@ impl TicketCounter {
...
+        if let Some(ref check) = self.check {
+
+            if !(check)(ctx, msg).await {
+                return
+            }
+        }
+
+        if let Some(ticket_owner) = self.tickets_for.get_mut(&id) {
+
+            // Let's remove a ticket if there is one.
```
Flat:
```suggestion
            // Remove a ticket if one is available.
```

src/framework/standard/mod.rs
```diff
@@ -273,15 +275,32 @@ impl StandardFramework {
             return Some(DispatchError::BlockedChannel);
         }
 
-        {
-            let mut buckets = self.buckets.lock().await;
+        // Try passing the command's bucket.
+        // We exit the loop if no command ratelimit has been hit or we
```
Flat:
```suggestion
        // exiting the loop if no command ratelimit has been hit or
```

src/framework/standard/mod.rs
```diff
@@ -273,15 +275,32 @@ impl StandardFramework {
...
+        // early-return when ratelimits cancel the framework invocation.
+        // Otherwise, we delay and loop again to check if we passed the bucket.
```
Flat:
```suggestion
        // Otherwise, delay and loop again to check if we passed the bucket.
```

src/framework/standard/structures/buckets.rs
```diff
@@ -61,16 +68,65 @@ impl Bucket {
...
+/// and then try taking a ticket again.
+pub enum BucketAction {
+    CancelWith(Duration),
+    DelayFor(Duration),
 }
 
 impl TicketCounter {
-    pub async fn take(&mut self, ctx: &Context, msg: &Message, id: u64) -> Option<Duration> {
+    /// Tries to check whether the invocation is permitted by the ticket counter
+    /// and figuratively speaking if a ticket can be taken; it does not return a
```
Flat:
```suggestion
    /// and if a ticket can be taken; it does not return a
```

src/framework/standard/structures/buckets.rs
```diff
@@ -61,16 +68,65 @@ impl Bucket {
...
+    /// a ticket but a duration until a ticket can be taken.
+    ///
+    /// The duration will be wrapped in an action for the caller to perform
+    /// if wanted. This may inform them to directly cancel trying to take
```
Flat:
```suggestion
    /// if wanted. This may inform them to directly cancel trying to take a ticket
```

src/framework/standard/structures/buckets.rs
Flat:
```suggestion
            // Remove a ticket if one is available.
```

src/framework/standard/structures/buckets.rs
```diff
@@ -84,34 +140,98 @@ impl TicketCounter {
...
+            if ticket_owner.tickets > 0 {
+                ticket_owner.tickets -= 1;
+            }
+
+            let delay = self.ratelimit.delay;
+            // Substract one step of time that would have to pass.
+            // This tries to bypass a problem of keeping track of when tickets
+            // were taken.
+            // When a ticket is taken, the bucket sets `last_time`, by
+            // substracting the delay once, we allow for one more ticket to be
```
Flat:
```suggestion
            // substracting the delay, once a ticket is allowed to be
```

src/framework/standard/structures/buckets.rs
```diff
@@ -84,34 +140,98 @@ impl TicketCounter {
...
+            // taken.
+            // If we would reset the value to `None`, we risk resetting the
```
Flat:
```suggestion
            // If the value is set to `None` this could possibly reset the
```

src/framework/standard/structures/buckets.rs
```diff
@@ -240,6 +361,18 @@ impl BucketBuilder {
         self
     }
 
+    /// If this is set to `true`, the invocation of the command will be delayed
+    /// and won't return a duration to wait to dispatch erros, but actually
```
Flat:
```suggestion
    /// and won't return a duration to wait to dispatch errors, but actually
```

arqunis (approved):

PR #1135, opened 2020-12-19
Add Watching activity type under unstable feature.

evieluvsrainbows:
I previously submitted a PR of this nature back in February 2020, which was rejected. However, this PR adds the Watching activity type under the `unstable` feature flag, meaning that this is exempt from any stability guarantee, similar to the `sync_id` and `session_id` fields that were added to `Activity` yesterday. With this addition, the Activity structure is now 100% complete. Grammar for the other activity types' documentation was also fixed, and the `unstable` feature flag has been added to the Features section of the README, and it has been cleaned up a small bit as well.


evieluvsrainbows:
Nevermind. There is no easy way to put this behind unstable, due to the `enum_number!()` macro.

PR #1139, opened 2020-12-21
Add Partial Application object to the Ready Event.

evieluvsrainbows:
Minor PR that adds support for the partial application object that was added to the Ready event in API v8.

### DIFF
--- a/src/model/application.rs
+++ b/src/model/application.rs
@@ -108,6 +108,15 @@ impl fmt::Debug for BotApplication {
     }
 }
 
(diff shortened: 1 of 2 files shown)


Lakelezz:
We probably should add a remark about the flags being unknown yet as they are not documented?

evieluvsrainbows:
Done.

Lakelezz (approved):

Flat (approved):

arqunis (approved):

PR #1142, opened 2020-12-23
Write Mentionable as a Display instead of String

Wolvereness:
This PR lets `.mention()` avoid the intermediate allocation of `String`, as well as provides an API that implements `Copy` for dealing with mentionable strings. This may be breaking for some uses, but the most common may be insertion into format-macros, where this PR should have a painless/unnoticeable transition. Uses that need a `String` can be resolved using `.to_string()` and this is mentioned in the documentation.

Notably, `Emoji` behaves strangely. The name of an emoji is irrelevant when it comes to its use; the id and animation status are what matters. Incidentally, animated emoji are now functional, but that could also be solved as its own issue in the absence of this PR.

Rust does not support deprecating trait implementations. `From<EmojiId> for Mention` would otherwise be deprecated, because it's insufficient information and lacks animation status.

Documentation examples have been added for `Mentionable::mention()` and `Mention`.

### DIFF
--- a/src/model/misc.rs
+++ b/src/model/misc.rs
@@ -8,81 +8,182 @@ use std::error::Error as StdError;
 use std::result::Result as StdResult;
 #[cfg(all(feature = "model", feature = "utils"))]
 use std::str::FromStr;
-#[cfg(all(feature = "model", feature = "utils"))]
 use std::fmt;
 #[cfg(all(feature = "model", any(feature = "cache", feature = "utils")))]
 use crate::utils;
 
 /// Allows something - such as a channel or role - to be mentioned in a message.
 pub trait Mentionable {
-    /// Creates a mentionable string, that will be able to notify and/or create
-    /// a link to the item.
-    fn mention(&self) -> String;
+    /// Creates a [`Mention`] that will be able to notify or create a link to the
+    /// item.
+    ///
+    /// [`Mention`] implements [`Display`], so [`ToString::to_string()`] can
+    /// be called on it, or inserted directly into a [`format_args!`] type of
+    /// macro.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # #[cfg(feature = "client")] {
+    /// # use serenity::model::guild::Member;
+    /// # use serenity::model::channel::GuildChannel;
+    /// # use serenity::model::id::ChannelId;
+    /// # use serenity::prelude::Context;
+    /// # use serenity::Error;
+    /// use serenity::model::misc::Mentionable;
+    /// async fn greet(
+    ///     ctx: Context,
+    ///     member: Member,
+    ///     to_channel: GuildChannel,
+    ///     rules_channel: ChannelId,
+    /// ) -> Result<(), Error> {
+    ///     to_channel.id.send_message(
+    ///         ctx,
+    ///         |m| m.content(format_args!(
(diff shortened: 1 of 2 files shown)


src/model/misc.rs
```diff
@@ -8,81 +8,178 @@ use std::error::Error as StdError;
...
 use std::fmt;
 #[cfg(all(feature = "model", any(feature = "cache", feature = "utils")))]
 use crate::utils;
 
 /// Allows something - such as a channel or role - to be mentioned in a message.
 pub trait Mentionable {
-    /// Creates a mentionable string, that will be able to notify and/or create
-    /// a link to the item.
-    fn mention(&self) -> String;
+    /// Creates a [`Mention`] that will able to notify or create a link to the
```
Flat:
```suggestion
    /// Creates a [`Mention`] that will be able to notify or create a link to the
```

src/model/misc.rs
```diff
@@ -8,81 +8,178 @@ use std::error::Error as StdError;
...
+    /// item.
+    ///
+    /// [`Mention`] implements [`Display`], so [`ToString::to_string()`] can
+    /// be called on it, or inserted directly into a [`format_args!`].
```
Flat:
```suggestion
    /// be called on it, or inserted directly into [`format_args!`].
```

src/model/misc.rs
```diff
@@ -8,81 +8,178 @@ use std::error::Error as StdError;
...
+    fn mention(&self) -> Mention;
 }
 
-impl Mentionable for ChannelId {
-    fn mention(&self) -> String { format!("<#{}>", self.0) }
+#[derive(Debug, Clone, Copy)]
+/// A struct that represents some way to insert a notification, link, or emoji
+/// into a message.
+///
+/// [`Display`] is the primary way of utilizing a [`Mention`], either in a
```
Flat, line 77:
```suggestion
/// [`Display`] is the primary way of utilizing a [`Mention`], either in
```

Flat (commented):
Documentation nits.

Indentation in the model/misc.rs example needs to be fixed as well.

src/model/misc.rs
Wolvereness: This one in particular is referring to multiple macros of that family; `write!` `format!` `writeln!`, but I have no issue changing it.

(same as the other "a `format_args!`")

Flat: in that case instead of removing `a` it may be more appropriate  to say "[`format_args!`] type macro."

Wolvereness:
> Indentation in the model/misc.rs example needs to be fixed as well.

My apologies; I'm not seeing where the issue is. Condensing the lines pushes past the 80-character limit, and I can't spot where it's not 4-spaces-per-level.

On the nits, I'll take care of that after work.

Flat:
No that was my fault, disregard the indentation comment. The GitHub app doesn't show it correctly apparently. Just checked it on the web and it looks fine.

Lakelezz:
Please ensure to pass all CI runs!

Wolvereness:
@Lakelezz I fixed the run that was failing, but something else is going wrong and I'm very certain it's not relating to anything I did.

/edit: Rebased, which caused the CI to run again, and all-passing.

arqunis (approved):

PR #1143, opened 2020-12-23
Improve permission checking

### LINKED ISSUE #1134: Improve permission checking
I think serenity's permission checking could be made more robust and flexible. As an example, [`GuildChannel::permissions_for_user`](https://docs.rs/serenity/0.9.2/serenity/model/channel/struct.GuildChannel.html#method.permissions_for_user) is both asynchronous and fallible, but it neither fetches uncached members nor fails if they are uncached, because it relies only on [`Guild::user_permissions_in`](https://docs.rs/serenity/0.9.2/serenity/model/guild/struct.Guild.html#method.user_permissions_in) which is neither asynchronous nor fallible.

The design I would like to see is changing `Guild::user_permissions_in` to use arguments of `&GuildChannel` and `&Member`. There are failure cases in this function that are almost entirely covered by an invalid `ChannelId` or `UserId` being passed, and changing the types in this way means these failures are ruled out by construction. With this change, functions like `GuildChannel::permissions_for_user` can try to get the `Member` object from the `Guild`'s cache themselves, and if it is nonexistent, can either fetch the `Member` or return an error, as appropriate.

The less appealing alternative to me is simply making `Guild::user_permissions_in` fallible. There are two different failure cases, so an `Option` can't practically be used to encode the result, which means `Result<Permission>` is the next best option. This forces users to match on the error they get to see if it was a missing channel or missing member, and `Error` has a lot more surface area than is necessary to encode the actual result of this function. Once they've determined what caused the error, there is no clear way to recover from the error, since there is no method that doesn't require already having a cached `Member` and `GuildChannel`. One could be added, but then this method becomes somewhat redundant and, in my opinion, a strictly less robust method than the first solution I suggested.

I think this is almost a necessary change. Bots very frequently rely on discord permissions to determine what actions are appropriate for the bot to carry out on behalf of a user. With incorrect permission checking, bots based on serenity can have disastrous bugs, like leaking private messages from channels that users cannot read themselves but that the bot incorrectly calculated they could.

I'm open to implementing these changes myself, but I would like to see what direction should be taken to get my changes approved.

ThatsNoMoon:
These are currently the minimum changes I think would workably solve the problems with permissions in #1134. However, I think another breaking change *could* make the API nicer to use. If `GuildChannel::user_permissions_in` used `CacheHttp` instead of `AsRef<Cache>` it could fetch any uncached users. I'm guessing `Context` is being used most of the time and it implements both so I don't think this would be a roadblock for migrating for many and it would make that method correct more often.

Closes #1134.

### DIFF
--- a/src/framework/standard/mod.rs
+++ b/src/framework/standard/mod.rs
@@ -822,7 +822,27 @@ pub(crate) async fn has_correct_permissions(
     if options.required_permissions().is_empty() {
         true
     } else if let Some(guild) = message.guild(&cache).await {
-        let perms = guild.user_permissions_in(message.channel_id, message.author.id);
+        let channel = match guild.channels.get(&message.channel_id) {
+            Some(channel) => channel,
+            None => return false,
+        };
+        let member = match guild.members.get(&message.author.id) {
+            Some(member) => member,
+            None => return false,
+        };
+
+        let perms = match guild.user_permissions_in(channel, member) {
+            Ok(perms) => perms,
+            Err(e) => {
+                tracing::error!(
+                    "(╯°□°）╯︵ ┻━┻ error getting permissions for user {} in channel {}: {}",
+                    member.user.id,
+                    channel.id,
+                    e
+                );
+                return false;
+            }
+        };
 
         perms.contains(*options.required_permissions())
     } else {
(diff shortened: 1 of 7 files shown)


arqunis (approved):

Lakelezz (approved):

Lakelezz:
Thank you for this pull request! :)

PR #1144, opened 2020-12-24
Add Max Delay Limits, Delay Hook, and RateLimitInfo

Lakelezz:
This commit limits the maximum amount of rate limit delayed invocations.
By setting this limit to 0, all invocations will error instantly. Otherwise, they will only error once the limit is hit.
Maximum delay limits prevent a command user to spam the bot with delayed tasks, by default we allow 1 delay task.

Once a delay will be suggested, the bucket will hook an optional function.
This function may inform the user about the delay as delays are not forwarded to dispatch errors.

Errors will now contain a rate limit error, providing a comprehensive note about the rate limit.
They allow to avoid spamming "Try again in X seconds"-messages by checking `is_first_try`.
Additionally the info contains the duration, the active delays, maximum delays, and the action taken by the invocation handler.

Example 5 has been adjusted accordingly. The text is now clearer about what builder method affects the bucket in what way.

Please help testing this! :)

### DIFF
--- a/examples/e05_command_framework/src/main.rs
+++ b/examples/e05_command_framework/src/main.rs
@@ -135,6 +135,7 @@ async fn my_help(
     let _ = help_commands::with_embeds(context, msg, args, help_options, groups, owners).await;
     Ok(())
 }
+
 #[hook]
 async fn before(ctx: &Context, msg: &Message, command_name: &str) -> bool {
     println!("Got command '{}' by user '{}'", command_name, msg.author.name);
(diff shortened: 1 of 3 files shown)


arqunis (approved):

arqunis (approved):
Works as intended. Also, there is a warning for unused `use std::time::Duration` in `standard/mod.rs`, since in the error type it got replaced with `RateLimitInfo`.

Lakelezz:
Fixed it!

PR #1174, opened 2021-01-07
Format the repository and add a workflow for formatting and linting

arqunis:
Now that we've exhausted all of our pull requests (except for #878, not sure what to do with it), we can advance towards syntactical consistency. This entails formatting code with options defined in `rustfmt.toml` before merging. To ensure this, a new workflow called *lint* is added that will check for warnings and suggestions from clippy and formatting so code adheres to the output of `rustfmt`.


acdenisSK:
Don't mind the failure for format -- it is pulling in code from this repository and not my branch that has the formatting.

Lakelezz (approved):

PR #1181, opened 2021-01-09
Update to async-tungstenite 0.12

nickelc:
The newest version no longer pulls in  `pin-project` and `bytes` 0.5.


Lakelezz:
This is a breaking change, as we expose `TungsteniteError`, please target `next`.

arqunis (approved):

PR #1175, opened 2021-01-07
Builders and Implementations For Interaction Response

SinsofSloth:
Feel free to rename the methods and struct names if they're too wordy or redudant.

PR #1186, opened 2021-01-12
Add builders for creating interactions

Th3-M4jor:
While this is grossly incomplete, and still needs a lot more documentation,
I would like to get feedback on what's there before continuing.

### DIFF
--- a/src/builder/create_interaction.rs
+++ b/src/builder/create_interaction.rs
@@ -7,9 +7,9 @@ use crate::{
 };
 
 #[derive(Clone, Debug, Default)]
-pub struct InteractionOptionChoices(pub Vec<Value>);
+pub struct CreateInteractionOptionChoices(pub Vec<Value>);
 
-impl InteractionOptionChoices {
+impl CreateInteractionOptionChoices {
     #[inline]
     pub fn add_int<D: ToString>(&mut self, name: D, value: i32) -> &mut Self {
         self.add_value(name.to_string(), Value::Number(serde_json::Number::from(value)))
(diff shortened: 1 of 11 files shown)


Th3-M4jor:
I believe this is ready for review.
Could add functions for creating and sending these to `model::interactions` but I think that should wait until after #1175 is merged

src/builder/create_interaction.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub fn create_sub_option<F>(&mut self, f: F) -> &mut Self
+    where F: FnOnce(&mut CreateInteractionOption) -> &mut CreateInteractionOption {
+        let mut data = CreateInteractionOption::default();
+        f(&mut data);
+        let new_option = utils::hashmap_to_json_map(data.0);
+        let options = self.0.entry("options").or_insert_with(|| Value::Array(Vec::new()));
+        if let Some(opt_arr) = options.as_array_mut() {
+            opt_arr.push(Value::Object(new_option));
+        };
+        self
```
arqunis:
```suggestion
        self.add_sub_option(data)
```

src/builder/create_interaction.rs
```diff
@@ -0,0 +1,188 @@
...
+    where F: FnOnce(&mut CreateInteractionOption) -> &mut CreateInteractionOption {
+        let mut data = CreateInteractionOption::default();
+        f(&mut data);
+        let new_option = utils::hashmap_to_json_map(data.0);
+        let options = self.0.entry("options").or_insert_with(|| Value::Array(Vec::new()));
+        if let Some(opt_arr) = options.as_array_mut() {
+            opt_arr.push(Value::Object(new_option));
+        };
+        // should this be changed to have None be a panic?
+        self
```
arqunis:
```suggestion
        self.add_interaction_option(data)
```

arqunis (changes requested):
This is good, but is there a reason the target branch is `next`? I don't see why this addition should have to be released in 0.11.

Th3-M4jor:
You're right, this probably can be merged into `next` and `current`.

arqunis (approved):

arqunis:
Manually rebased onto `current`: https://github.com/serenity-rs/serenity/commit/0d743fdd3b7cc0ea28500547ff0a9caff4a83c3f

PR #1188, opened 2021-01-14
Add permissions to EditChannel builder

pcbzzz:


arqunis (approved):

src/builder/edit_channel.rs
```diff
@@ -115,4 +119,59 @@ impl EditChannel {
...
+    /// use serenity::model::permissions::Permissions;
+    ///
+    /// // Assuming a guild have already been bound.
+    /// let permissions = Some(PermissionOverwrite {
+    ///     allow: Permissions::READ_MESSAGES,
+    ///     deny: Permissions::SEND_TTS_MESSAGES,
+    ///     kind: PermissionOverwriteType::Member(UserId(1234)),
+    /// });
+    ///
+    /// guild.create_channel(http, |c| {
```
arqunis: This should show an example of editing a channel, not creating a new one.

src/builder/edit_channel.rs
```diff
@@ -115,4 +119,59 @@ impl EditChannel {
...
+    /// # use std::sync::Arc;
+    /// #
+    /// # async fn run() -> Result<(), Box<dyn std::error::Error>> {
+    /// #     let http = Arc::new(Http::default());
+    /// #     let mut guild = GuildId(0).to_partial_guild(&http).await?;
+    /// use serenity::model::channel::{PermissionOverwrite, PermissionOverwriteType};
+    /// use serenity::model::id::UserId;
+    /// use serenity::model::permissions::Permissions;
+    ///
+    /// // Assuming a guild have already been bound.
```
arqunis: "guild" is the subject of the sentence, which is in third person.
```suggestion
    /// // Assuming a guild has already been bound.
```

This typo is also present in the example for [`CreateChannel::permissions`][perms], could you fix it there too?

[perms]: https://github.com/serenity-rs/serenity/blob/current/src/builder/create_channel.rs#L108

arqunis (changes requested):

src/builder/edit_channel.rs
pcbzzz: Fixed both. Considering it was a direct copy of that function I'm not surprised, I noticed the `create_channel` after I submitted and just hoped it was overlooked 😜

src/builder/edit_channel.rs
```diff
@@ -115,4 +119,59 @@ impl EditChannel {
...
+    /// use serenity::model::permissions::Permissions;
+    ///
+    /// // Assuming a guild has already been bound.
+    /// let permissions = Some(PermissionOverwrite {
+    ///     allow: Permissions::READ_MESSAGES,
+    ///     deny: Permissions::SEND_TTS_MESSAGES,
+    ///     kind: PermissionOverwriteType::Member(UserId(1234)),
+    /// });
+    ///
+    /// guild.edit_channel(http, |c| {
```
arqunis: Guild models don't have an `edit_channel` method (though they should). Use [`GuildChannel::edit`](https://serenity-rs.github.io/serenity/current/serenity/model/channel/struct.GuildChannel.html#method.edit).

arqunis (changes requested):

arqunis (approved):

PR #1194, opened 2021-01-18
Update the contributing guide to reflect the state of the project

arqunis:


Lakelezz:
I like this.

We should add a note about comments. As their nature is to always appear in the line before the relevant section instead of the same.
E.g., if you comment on the interesting line 333 of code, the comment goes to line 332.

Comments are intended to be in British English and should form a proper English sentence; end with a full stop (`.`).
Categories within an API comment block should be prepended with `#`, allowing to reference it in links. 
An example: 
```
# Panic
This function will panic if `level` is below `-1`. 
```

Additionally, we should explain that doc-links are supposed to use `crate::module::sub_module::struct`-style instead of hard links. External hyperlinks are obviously exempted from this guideline.

Lakelezz:
We need to add section about `unsafe`.
If code containing the Rust keyword `unsafe` is committed, the particular sections must be reasoned with a comment.
The comment must entail the necessity of the usage and explain why the code is sound.
Comprehensive comments are required to ensure a well-defined audit can be performed by everyone interested.

Lakelezz (approved):
Thank you for doing all these changes! :)

PR #1201, opened 2021-01-26
Enable log feature in tracing

SystemNeoSpl:
This allows tracing to emit `log::Record`s. That way, users of `serenity` won't be forced into using `tracing`, and instead can pick their own logging implementation (e.g. `slog`, `env_logger`, `fern`, etc.)


Flat (approved):
Makes sense to add this feature. Further details here https://docs.rs/tracing/0.1.22/tracing/#crate-feature-flags

arqunis (approved):

Issue #1198, opened 2021-01-21
Adding stricter Clippy lints to Serenity


Th3-M4jor:
I'm opening this issue to discuss if we should be a little more granular with the Clippy lints we deny globally, and if so which ones should we deny globally?

The full list can be found [here](https://rust-lang.github.io/rust-clippy/master/index.html).
I'd like to deny these globally to start with:
- [x] ~~[expect_used](https://rust-lang.github.io/rust-clippy/master/index.html#expect_used)~~
- [x] [unwrap_used](https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_used)
- [x] [non_ascii_literal](https://rust-lang.github.io/rust-clippy/master/index.html#non_ascii_literal)

And maybe this one:
- [x] [missing_errors_doc](https://rust-lang.github.io/rust-clippy/master/index.html#missing_errors_doc)

As discussed in the Discord:
- [x] [let_underscore_must_use](https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_must_use)

I don't mind writing the PR to add and fix them once we settle on a list.

Th3-M4jor:
Decided against adding a lint for usage of `expect()`, because currently all instances of it should panic.

vicky5124:
As mentioned on the discord, implement [this](https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_must_use) clippy lint too.

PR #1203, opened 2021-01-26
Add clippy lint for denying usage of `unwrap()`

Th3-M4jor:
This PR removes most usages of `unwrap()` and adds `#![deny(clippy::unwrap_used)]` to `lib.rs`, as mentioned in #1198.

Added `#[allow(clippy::unwrap_used)]` to functions where it should always be `Some` or `Ok` and it'd be a breaking change to instead return an Error/None.

### DIFF
--- a/src/builder/create_allowed_mentions.rs
+++ b/src/builder/create_allowed_mentions.rs
@@ -75,10 +75,8 @@ impl CreateAllowedMentions {
     pub fn parse(&mut self, value: ParseValue) -> &mut Self {
         let val = self.0.entry("parse").or_insert_with(|| Value::Array(Vec::new()));
 
-        if let Some(arr) = val.as_array_mut() {
-            arr.push(json![value]);
-        }
-        // Should there be an "else" block that panics?
+        let arr = val.as_array_mut().expect("Must be an array");
+        arr.push(json![value]);
 
         self
     }
@@ -94,10 +92,8 @@ impl CreateAllowedMentions {
     pub fn empty_parse(&mut self) -> &mut Self {
         let val = self.0.entry("parse").or_insert_with(|| Value::Array(Vec::new()));
 
-        if let Some(arr) = val.as_array_mut() {
-            arr.clear();
-        }
-        // Should there be an "else" block that panics?
+        let arr = val.as_array_mut().expect("Must be an array");
+        arr.clear();
 
         self
     }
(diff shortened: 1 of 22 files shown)


Th3-M4jor:
Update: added `#![deny(clippy::non_ascii_literals)]`, allowing them in the list of discord join messages in `constants.rs`.
Removed all the "tableflip" text-faces from tracing macros since error/warning messages probably shouldn't include that anyway. They might be logged to or read by something that's not unicode aware.

src/builder/create_allowed_mentions.rs
```diff
@@ -73,11 +73,13 @@ impl CreateAllowedMentions {
...
-        if let Some(val) = self.0.get_mut("parse") {
-            val.as_array_mut().unwrap().push(json![value]);
-        } else {
-            self.0.insert("parse", Value::Array(vec![json!(value)]));
+        let val = self.0.entry("parse").or_insert_with(|| Value::Array(Vec::new()));
+
+        if let Some(arr) = val.as_array_mut() {
+            arr.push(json![value]);
         }
+        // Should there be an "else" block that panics?
```
arqunis: It's a programmatic error for the `parse` field to not be an array, though it's very unlikely to happen. The field is managed by these builder methods, which ensure the field is an array. The only way it could be another type is if the user inserted the field themselves, since the `HashMap` is public. For the sake of being pedantic, I say there should be a panic.

```rust
let arr = val.as_array_mut().expect("must be an array");
arr.push(json![value]);
```

src/builder/create_allowed_mentions.rs
```diff
@@ -90,11 +92,13 @@ impl CreateAllowedMentions {
...
-        if let Some(val) = self.0.get_mut("parse") {
-            val.as_array_mut().unwrap().clear();
-        } else {
-            self.0.insert("parse", Value::Array(vec![]));
+        let val = self.0.entry("parse").or_insert_with(|| Value::Array(Vec::new()));
+
+        if let Some(arr) = val.as_array_mut() {
+            arr.clear();
         }
+        // Should there be an "else" block that panics?
```
arqunis: Ditto.

arqunis (changes requested):

Th3-M4jor:
There are a few other builders that should also panic like this then. I'll push a change that does this to all of them.

arqunis (approved):

PR #1205, opened 2021-01-28
Handle nested sub-commands in help

Lakelezz:
## Problem
First, a group with a prefix caused sub-commands to fail in help.
Additionally, nesting sub-commands inside sub-commands did not work either.

## Solution
This fix allows sub-commands to be nested multiple levels.
Also, groups holding the command may have any amount of prefixes or aliases.

On a side-note, this pull request changes and adds some documentation to the help system.

Please help testing this by running help on previously failed and working commands to catch regressions.


vicky5124:
Tested and it works fine :+1:

PR #1218, opened 2021-02-07
Don't require `cache` feature for `MembersIter` and `GuildId::members_iter`

AriusX7:
The `MembersIter` struct and `GuildId::members_iter` method don't need the `cache` feature to work.

Edit: I assumed this is a breaking change because someone might've used a trait to add a method named `members_iter` to `GuildId` with `cache` disabled. Please correct me if I'm wrong.

### DIFF
--- a/src/model/guild/guild_id.rs
+++ b/src/model/guild/guild_id.rs
@@ -907,7 +907,7 @@ impl<'a> From<&'a Guild> for GuildId {
 
 /// A helper class returned by [`GuildId::members_iter`]
 #[derive(Clone, Debug)]
-#[cfg(all(feature = "model"))]
+#[cfg(feature = "model")]
 pub struct MembersIter<H: AsRef<Http>> {
     guild_id: GuildId,
     http: H,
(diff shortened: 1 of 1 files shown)


src/model/guild/guild_id.rs
```diff
@@ -909,7 +907,7 @@ impl<'a> From<&'a Guild> for GuildId {
 
 /// A helper class returned by [`GuildId::members_iter`]
 #[derive(Clone, Debug)]
-#[cfg(all(feature = "model", feature = "cache"))]
+#[cfg(all(feature = "model"))]
```
arqunis: As we're checking for just one feature now, the `all` predicate can be dropped.
```suggestion
#[cfg(feature = "model")]
```

src/model/guild/guild_id.rs
```diff
@@ -918,7 +916,7 @@ pub struct MembersIter<H: AsRef<Http>> {
     tried_fetch: bool,
 }
 
-#[cfg(all(feature = "model", feature = "cache"))]
+#[cfg(all(feature = "model"))]
```
```suggestion
#[cfg(feature = "model")]
```

arqunis (changes requested):

arqunis (approved):

arqunis:
> I assumed this is a breaking change because someone might've used a trait to add a method named `members_iter` to `GuildId` with cache disabled. Please correct me if I'm wrong.

This is possible, but very unlikely. In case someone did, they should've expected they're going to run into conflicts once they have the `cache` feature enabled (or once the feature restriction has been removed). I've merged this into `next`, but I'll backport it to `current`, since I believe no one has done this.

PR #1223, opened 2021-02-10
Properly fix tracing version to force a version over or equal to .23

vicky5124:
Just added ^ to the versions so a version over or equal to the supported is guaranteed to be used.

	modified:   Cargo.toml
	modified:   examples/e06_sample_bot_structure/Cargo.toml
	modified:   examples/e07_env_logging/Cargo.toml

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ serde_json = "1"
 async-trait = "0.1"
 
 [dependencies.tracing]
-version = "0.1.23"
+version = "^0.1.23"
 features = ["log"]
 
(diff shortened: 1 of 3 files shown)


nickelc:
that's the same.
see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio
```
The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.
```

PR #1231, opened 2021-02-17
Capture sub-commands and aliases in `similar_command`

### LINKED ISSUE #1212: Levenshtein approximation for command aliases too
Say you have a command named `info` with the alias `about`, `help inf` will ask if the user meant `info` but `help abou` will say the command wasn't found, I think it should ask if they meant `info` in both cases

Lakelezz:
# Description
This pull request reconsiders whether the current help-item is a command-name, command-alias, or sub-command.
If the item is not an exact match but within the user-defined levenshtein distance, the item will be added to the `similar_command`-list.
If a group is not qualifying for the help-system due to its prefix, it will be skipped, implying performance gains.

# Type of Change
This pull request qualifies as an *enhancement*, relates to code residing in the *framework*, and closes #1212.

# How Has This Been Tested?
This pull request used the command framework example (example 5).

I ran `~help emoji nek`, got commands exclusively from the `emoji`-group. This includes `cat`, `neko`, and `dog`, previously this included the main name `cat` only.

Additionally I ran `~help uppe`, the system suggested `upper` and `~help upper sec` suggested `sub` and `secret`. 

Last but not least, I successfully executed `~help em`, `~help emoji neko`, `~help emoji`, and `~help upper sub` to see if alias-groups, alias-commands, main command-names, and sub-commands still work.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -404,7 +404,7 @@ fn nested_commands_search<'rec, 'a: 'rec>(
 
                 for command_name in command.options.names {
                     if name == *command_name {
-                        command_found = Some(&(*command_name).clone());
+                        command_found = Some((*command_name).clone());
 
                         break;
                     }
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -400,7 +400,15 @@ fn nested_commands_search<'rec, 'a: 'rec>(
             let mut command = *command;
 
             let search_command_name_matched = {
-                let command_found = command.options.names.iter().find(|n| **n == name).cloned();
+                let mut command_found = None;
+
+                for command_name in command.options.names {
+                    if name == *command_name {
+                        command_found = Some(command_name.clone());
+
+                        break;
+                    }
+                }
 
                 if command_found.is_some() {
                     command_found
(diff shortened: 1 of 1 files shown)


src/framework/standard/help_commands.rs
```diff
@@ -545,7 +550,11 @@ fn nested_group_command_search<'rec, 'a: 'rec>(
                 },
             }
 
-            group.options.prefixes.iter().any(|prefix| trim_prefixless_group(prefix, name));
+            if !group.options.prefixes.is_empty()
+                && !group.options.prefixes.iter().any(|prefix| trim_prefixless_group(prefix, name))
+            {
```
arqunis, line 555: This can be simplified to calling just `any`. If the iterator is empty, `any` will return `false`.

```suggestion
            if !group.options.prefixes.iter().any(|prefix| trim_prefixless_group(prefix, name)) {
```

Lakelezz: Yes, but that's the problem. This will cause an empty `prefixes` to evaluate the condition to `true` and trigger `continue`, skipping any further recursion on it.

A help execution such as `~help upper su` or `~help upper sub` will break fully then. They have no prefix and thus are not considered and skipped.

`any` on an empty `prefixes` evaluates to `false`, negated by `!` this turns into `true`, the condition is met, `continue` reached.

`group.options.prefixes.is_empty()` will return `true` and negated by `!` it will turn into `false`, the condition requires the conjunction `&&` and short-circuits as there is a `false` already, `continue` is unreached.

arqunis: Ah, I see. I went a little too far with pursuit of optimisation that I didn't consider the change in behaviour if you applied my suggestion. You can disregard my comment.

Lakelezz: Thanks for paying attention to the code though, I think it's a good way to ensure code is valid :)

PR #1233, opened 2021-02-18
Ignore Missing Cache Items on Permission Checks

### LINKED ISSUE #1166: `utils::user_has_perms` Fails When Cache Lacks Items
`Message::delete` fails on the bot's own message during the permissions check in `utils::user_has_perms` with `Err(ChannelNotFound)` as private channels are not cached anymore (Bots no longer receive channel create events for private channels on v8).

Basic example:

```rust
let dm_chan = msg.author.create_dm_channel(&ctx).await.unwrap().id;

let m = dm_chan
  .say(&ctx, "hello")
  .await
  .unwrap();

dbg!(m.delete(&ctx).await);
// Err(ChannelNotFound)
```

Permission check for `Message::delete` with the cache should be skipped(?) if the message is both the bot's own and in a private channel.  It may also fail in other cases where it uses the cache.

Also, if private channels are no longer cached, it could probably be removed from the cache struct entirely.

Lakelezz:
# Description
When the cache is enabled but the cache is not populated with all the data needed for a permission check, the check failed with an error.
This pull request changes model methods to run the HTTP-request still, as attempting to run the request is cheaper than requesting data for a check.

# Type of Change
This pull request qualifies as a *fix*, relates to code residing in the *utils* and *model*, and closes #1166.
It fixes the behaviour, as the methods were never truly designed to fail, nonetheless, the new gateway intents causes the cache to receive fewer events by default.

# How Has This Been Tested?
Attempted to execute methods that would run a permission check while the gateway intents are disabled.
I would be happy if people can try to run their bots with this pull request, especially if you encountered #1166,

### DIFF
--- a/src/model/utils.rs
+++ b/src/model/utils.rs
@@ -218,7 +218,7 @@ pub fn serialize_gen_map<K: Eq + Hash, S: Serializer, V: Serialize>(
 }
 
 /// Tries to find a user's permissions using the cache.
-/// Unline [`user_has_perms`], this function will return `true` even when
+/// Unlike [`user_has_perms`], this function will return `true` even when
 /// the permissions are not in the cache.
 #[cfg(all(feature = "cache", feature = "model"))]
 #[inline]
--- a/src/model/utils.rs
+++ b/src/model/utils.rs
@@ -217,6 +217,27 @@ pub fn serialize_gen_map<K: Eq + Hash, S: Serializer, V: Serialize>(
     seq.end()
 }
 
+/// Tries to find a user's permissions using the cache.
+/// Unline [`user_has_perms`], this function will return `true` even when
+/// the permissions are not in the cache.
+#[cfg(all(feature = "cache", feature = "model"))]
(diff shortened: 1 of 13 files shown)


src/model/utils.rs
```diff
@@ -217,6 +217,27 @@ pub fn serialize_gen_map<K: Eq + Hash, S: Serializer, V: Serialize>(
     seq.end()
 }
 
+/// Tries to find a user's permissions using the cache.
+/// Unline [`user_has_perms`], this function will return `true` even when
```
arqunis: Typo.
```suggestion
/// Unlike [`user_has_perms`], this function will return `true` even when
```

arqunis (changes requested):

src/model/utils.rs
Lakelezz: That's a funny one!

arqunis (approved):

Flat (approved):

PR #1234, opened 2021-02-20
Add user public flags field

### LINKED ISSUE #1220: Add `public_flags` to `User`
The representation of the [User object](https://discord.com/developers/docs/resources/user#user-object-user-structure) is lacking the `public_flags`-field.

miqbalrr:
closes #1220 

from [`here`](https://discord.com/developers/docs/resources/user#user-object-user-structure) `public_flags` field has an integer type, but i don't know exactly in real execution.

all default value for this new `public_flags` field is 0

### DIFF
--- a/src/model/user.rs
+++ b/src/model/user.rs
@@ -4,7 +4,7 @@ use std::fmt;
 #[cfg(feature = "model")]
 use std::fmt::Write;
 
-use bitflags::bitflags;
+use bitflags::__impl_bitflags;
 use futures::future::{BoxFuture, FutureExt};
 #[cfg(feature = "model")]
 use serde_json::json;
@@ -47,7 +47,7 @@ pub struct CurrentUser {
     #[serde(rename = "username")]
     pub name: String,
     pub verified: Option<bool>,
-    pub public_flags: UserPublicFlags,
+    pub public_flags: Option<UserPublicFlags>,
 }
 
 #[cfg(feature = "model")]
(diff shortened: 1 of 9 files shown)


src/model/user.rs
```diff
@@ -46,6 +46,7 @@ pub struct CurrentUser {
     #[serde(rename = "username")]
     pub name: String,
     pub verified: Option<bool>,
+    pub public_flags: u8,
```
drklee3: I believe this should be at least a `u32` since there are [17 flags](https://discord.com/developers/docs/resources/user#user-object-user-flags).

The last flag `1 << 17` would be represented as `100000000000000000` in binary, so at minimum it requires 18 bits. The `<<` arithmetic left shift moves `1` to the left 17 positions.

drklee3 (commented):
Good start, small change with the data type and an additional use of the `bitflags` crate. Some other modules in serenity that also use bitflags might be good reference like the [gateway ActivityFlags](https://github.com/serenity-rs/serenity/blob/9bbb25aac4d651804286f333eb503a72d41e473b/src/model/gateway.rs#L385-L402) or [permissions](https://github.com/serenity-rs/serenity/blob/9bbb25aac4d651804286f333eb503a72d41e473b/src/model/permissions.rs).

src/model/user.rs
miqbalrr: okay, gimme a couple time for it,

src/model/user.rs
```diff
@@ -446,6 +448,43 @@ pub struct User {
...
+        /// User's flag as system
+        const SYSTEM = 0b00000000_00000000_00010000_00000000;
+        /// User's flag as bug hunter level 2
+        const BUG_HUNTER_LEVEL_2 = 0b00000000_00000000_01000000_00000000;
+        /// User's flag as verified bot
+        const VERIFIED_BOT = 0b00000000_00000001_00000000_00000000;
+        /// User's flag as early verified bot developer
+        const EARLY_VERIFIED_BOT_DEVELOPER = 0b00000000_00000010_00000000_00000000;
+    }
 }
```
arqunis: This should be implemented like it is for [`Permissions`](https://github.com/serenity-rs/serenity/blob/current/src/model/permissions.rs#L207-L319) and [other bitflags structs](https://github.com/serenity-rs/serenity/blob/1ab52a056f94760a7436927c98c08ea27da3a0ad/src/model/channel/message.rs#L1121-L1138).

arqunis (changes requested):

src/model/user.rs
miqbalrr: teach me how to use bitflags in the right way

src/model/user.rs
```diff
@@ -449,41 +449,43 @@ pub struct User {
...
 
-bitflags! {
-    /// User public flags
-    #[derive(Default, Deserialize, Serialize)]
-    pub struct UserPublicFlags: u32 {
-        /// User's flag as none
-        const NONE = 0b00000000_00000000_00000000_00000000;
+/// User's public flags
+#[derive(Clone, Deserialize, Serialize)]
```
arqunis: This can implement `Copy` as four bytes are very cheap for the processor to copy.

```suggestion
#[derive(Clone, Copy, Deserialize, Serialize)]
```

arqunis (changes requested):

src/model/user.rs
arqunis: I apologise that I hadn't gotten to you in a while, but from the latest commit, you got it right. I do have a small nitpick, but otherwise good job!

src/model/user.rs
zacck-zz: @miqbalrr I can finish this one off is you don't have the time :)

miqbalrr: Sorry, i wasn't check the notification. i Will finish later

arqunis (approved):

arqunis:
Thanks!

PR #1235, opened 2021-02-20
Initial Parse trait PR

kangalio:
This PR adds a new `serenity::utils::Parse` trait, mainly for https://github.com/serenity-rs/framework/issues/24.

As described in the docstring below, the `Parse` trait works like `std::str::FromStr`, but it also supports Discord model types like `Member` or `Message`.
```rust
/// Parse a value from a string in context of a received message.
///
/// This trait is a superset of [`std::str::FromStr`] (the trait used in `str::parse`). The
/// difference is that this trait supports serenity-specific Discord types like [`Member`], [`Role`],
/// or [`Emoji`].
///
/// Trait implementations may do network requests as part of their parsing procedure.
///
/// Useful for implementing command argument parsing frameworks.
#[async_trait::async_trait]
pub trait Parse: Sized {
    /// The associated error which can be returned from parsing.
    type Err;

    /// Parses a string `s` as a command parameter of this type.
    async fn parse(ctx: &Context, msg: &Message, s: &str) -> Result<Self, Self::Err>;
}
```

In this PR, support for `Member` and `Message` is implemented.

In a later PR, I hope to add support for more types:
- `User`/`UserId`
- `Channel`/`ChannelId`
- `GuildId`
- `ChannelCategory`
- `Role`/`RoleId`
- `Invite`
- `Emoji`/`EmojiId`/`EmojiIdentifier`
- `Colour`
- did I miss one?

### DIFF
--- a/src/utils/parse.rs
+++ b/src/utils/parse.rs
@@ -30,7 +30,10 @@ impl<T: std::str::FromStr> Parse for T {
 /// Error that can be returned from [`Member::parse`].
 #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
 pub enum MemberParseError {
+    /// The guild in which the parser was invoked is not in cache.
     GuildNotInCache,
+    /// The provided member string failed to parse, or the parsed result cannot be found in the
+    /// guild cache data.
     NotFoundOrMalformed,
 }
 
@@ -104,7 +107,9 @@ impl Parse for Member {
 /// Error that can be returned from [`Message::parse`].
 #[derive(Debug)]
 pub enum MessageParseError {
+    /// When the provided string does not adhere to any known guild message format
     Malformed,
+    /// When message data retrieval via HTTP failed
     Http(SerenityError),
     /// When the `gateway` feature is disabled and the required information was not in cache.
     HttpNotAvailable,
--- a/src/utils/parse.rs
+++ b/src/utils/parse.rs
@@ -7,8 +7,8 @@ use crate::{model::prelude::*, prelude::*};
 /// Parse a value from a string in context of a received message.
 ///
 /// This trait is a superset of [`std::str::FromStr`] (the trait used in `str::parse`). The
-/// difference is that this trait supports serenity-specific Discord types like [`Member`], `Role`],
-/// or [`Emoji`].
+/// difference is that this trait supports serenity-specific Discord types like [`Member`],
+/// [`Role`], or [`Emoji`].
 ///
 /// Trait implementations may do network requests as part of their parsing procedure.
 ///
@@ -61,6 +61,12 @@ impl std::fmt::Display for MemberParseError {
 impl Parse for Member {
     type Err = MemberParseError;
 
+    #[cfg(not(feature = "cache"))]
+    async fn parse(_: &sconteContext, _: &Message, _: &str) -> Result<Self, Self::Err> {
+        Err(MemberParseError::GuildNotInCache)
+    }
+
+    #[cfg(feature = "cache")]
     async fn parse(ctx: &Context, msg: &Message, s: &str) -> Result<Self, Self::Err> {
         let guild = msg.guild(&ctx.cache).await.ok_or(MemberParseError::GuildNotInCache)?;
 
@@ -109,6 +115,8 @@ impl Parse for Member {
 pub enum MessageParseError {
     Malformed,
     Http(SerenityError),
+    /// When the `gateway` feature is disabled and the required information was not in cache
+    HttpNotAvailable,
 }
 
 impl std::error::Error for MessageParseError {
@@ -116,6 +124,7 @@ impl std::error::Error for MessageParseError {
         match self {
             Self::Malformed => None,
             Self::Http(e) => Some(e),
+            Self::HttpNotAvailable => None,
         }
     }
 }
@@ -125,8 +134,12 @@ impl std::fmt::Display for MessageParseError {
         match self {
             Self::Malformed => {
                 write!(f, "provided string did not adhere to any known guild message format")
-            }
+            },
             Self::Http(e) => write!(f, "failed to request message data via HTTP: {}", e),
+            Self::HttpNotAvailable => write!(
+                f,
+                "gateway feature is disabled and the required information was not in cache"
+            ),
         }
     }
 }
@@ -164,10 +177,15 @@ impl Parse for Message {
             .or_else(extract_from_message_url)
             .ok_or(MessageParseError::Malformed)?;
 
+        #[cfg(feature = "cache")]
         if let Some(msg) = ctx.cache.message(channel_id, message_id).await {
-            Ok(msg)
-        } else {
+            return Ok(msg);
+        }
+
+        if cfg!(feature = "gateway") {
             ctx.http.get_message(channel_id.0, message_id.0).await.map_err(MessageParseError::Http)
+        } else {
+            Err(MessageParseError::HttpNotAvailable)
         }
     }
 }
--- a/src/utils/parse.rs
+++ b/src/utils/parse.rs
@@ -177,7 +177,7 @@ impl Parse for Message {
             return Ok(msg);
         }
 
-        if cfg!(feature = "gateway") {
+        if cfg!(feature = "http") {
             ctx.http.get_message(channel_id.0, message_id.0).await.map_err(MessageParseError::Http)
         } else {
             Err(MessageParseError::HttpNotAvailable)
--- a/src/utils/parse.rs
+++ b/src/utils/parse.rs
@@ -29,6 +29,7 @@ impl<T: std::str::FromStr> Parse for T {
 }
 
 /// Error that can be returned from [`Member::parse`].
+#[non_exhaustive]
 #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
 pub enum MemberParseError {
     /// The guild in which the parser was invoked is not in cache.
@@ -104,6 +105,7 @@ impl Parse for Member {
 }
 
 /// Error that can be returned from [`Message::parse`].
+#[non_exhaustive]
 #[derive(Debug)]
 pub enum MessageParseError {
     /// When the provided string does not adhere to any known guild message format
--- a/src/utils/parse.rs
+++ b/src/utils/parse.rs
@@ -1,0 +1,173 @@
+use crate::{model::prelude::*, prelude::*};
+
+// Questions:
+// - Should the error enums be #[non_exhaustive]?
+// - Should the MessageParseError implement source()? It may be too much boilerplate (maybe add thiserror dependency?)
+
+/// Parse a value from a string in context of a received message.
+///
+/// This trait is a superset of [`std::str::FromStr`] (the trait used in `str::parse`). The
+/// difference is that this trait supports serenity-specific Discord types like [`Member`], `Role`],
+/// or [`Emoji`].
+///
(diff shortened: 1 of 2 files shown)


src/utils/parse.rs
```diff
@@ -0,0 +1,186 @@
...
+            .or_else(extract_from_message_id)
+            .or_else(extract_from_message_url)
+            .ok_or(MessageParseError::Malformed)?;
+
+        #[cfg(feature = "cache")]
+        if let Some(msg) = ctx.cache.message(channel_id, message_id).await {
+            return Ok(msg);
+        }
+
+        if cfg!(feature = "gateway") {
```
arqunis: The `gateway` feature implies the `http` feature, but I don't see why we couldn't depend on `http` directly. 
```suggestion
        if cfg!(feature = "http") {
```

arqunis (changes requested):

src/utils/parse.rs
kangalio: That's just my unfamiliarity with serenity's features (I just use default features for my own bots :sweat_smile:) Thanks for the correction

kangalio:
Do you know how to stop GitHub from showing the "Changes requested" warning/info? The requested changes have been resolved, what else does it want

arqunis (approved):

arqunis:
It has to be approved by the reviewer for it to go away.

src/utils/parse.rs
```diff
@@ -0,0 +1,186 @@
...
+/// 2. Lookup by mention.
+/// 3. Lookup by name#discrim
+/// 4. Lookup by name
+/// 5. Lookup by nickname
+#[cfg(feature = "cache")]
+#[async_trait::async_trait]
+impl Parse for Member {
+    type Err = MemberParseError;
+
+    #[doc = ""] // to get impl documentation to show
```
Lakelezz: Comments go before the related code item.
```suggestion
    // to get impl documentation to show
    #[doc = ""] 
```

src/utils/parse.rs
```diff
@@ -0,0 +1,186 @@
...
+            .ok_or(MemberParseError::NotFoundOrMalformed)
+    }
+}
+
+/// Error that can be returned from [`Message::parse`].
+#[derive(Debug)]
+pub enum MessageParseError {
+    Malformed,
+    Http(SerenityError),
+    /// When the `gateway` feature is disabled and the required information was not in cache
```
Lakelezz:
```suggestion
    /// When the `gateway` feature is disabled and the required information was not in cache.
```

src/utils/parse.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+impl std::fmt::Display for MessageParseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Malformed => {
+                write!(f, "provided string did not adhere to any known guild message format")
+            },
+            Self::Http(e) => write!(f, "failed to request message data via HTTP: {}", e),
```
Lakelezz:
```suggestion
            Self::Http(e) => write!(f, "Failed to request message data via HTTP: {}.", e),
```

src/utils/parse.rs
```diff
@@ -0,0 +1,186 @@
...
+            Self::HttpNotAvailable => None,
+        }
+    }
+}
+
+impl std::fmt::Display for MessageParseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Malformed => {
+                write!(f, "provided string did not adhere to any known guild message format")
```
Lakelezz:
```suggestion
                write!(f, "Provided string did not adhere to any known guild message format.")
```

src/utils/parse.rs
```diff
@@ -0,0 +1,186 @@
...
+///
+/// The lookup strategy is as follows (in order):
+/// 1. Lookup by "{channel ID}-{message ID}" (retrieved by shift-clicking on "Copy ID")
+/// 2. Lookup by message ID (the message must be in the context channel)
+/// 3. Lookup by message URL
+#[async_trait::async_trait]
+impl Parse for Message {
+    type Err = MessageParseError;
+
+    #[doc = ""] // to get impl documentation to show
```
Lakelezz:
```suggestion
    // to get impl documentation to show
    #[doc = ""] 
```

src/utils/parse.rs
```diff
@@ -0,0 +1,186 @@
+use crate::{model::prelude::*, prelude::*};
+
+// Questions:
+// - Should the error enums be #[non_exhaustive]?
+// - Should the MessageParseError implement source()? It may be too much boilerplate (maybe add thiserror dependency?)
```
Lakelezz: This does not belong in final code, but should be discussed before merging this.

src/utils/parse.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// Parse a value from a string in context of a received message.
+///
+/// This trait is a superset of [`std::str::FromStr`] (the trait used in [`str::parse`]). The
+/// difference is that this trait supports serenity-specific Discord types like [`Member`],
+/// [`Role`] (soon), or [`Emoji`] (soon).
```
Lakelezz: If it does not support the mentioned items, they should not be listed here (see the *soon*), but items that it supports. Alternatively, generalise the sentence.

src/utils/parse.rs
```diff
@@ -0,0 +1,186 @@
...
+impl std::fmt::Display for MessageParseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Malformed => {
+                write!(f, "provided string did not adhere to any known guild message format")
+            },
+            Self::Http(e) => write!(f, "failed to request message data via HTTP: {}", e),
+            Self::HttpNotAvailable => write!(
+                f,
+                "gateway feature is disabled and the required information was not in cache"
```
Lakelezz:
```suggestion
                "Gateway feature is disabled and the required information was not in cache."
```

Lakelezz (changes requested):
Thanks for your pull request.
I changed the punctuation and I do think the comments have a few questions that must be solved *before* merging.

src/utils/parse.rs
kangalio: I see, I wasn't sure which casing should be preferred for error messages inside serenity.

It may be of interest that even existing code isn't fully consistent about this: some code uses uppercase errors, some lowercase:
Screenshot_20210223_201001

kangalio: Another thing: with the convention being to add a trailing full stop to error messages, wouldn't `{}.` expand to two full stops? Think `Failed to request message data via HTTP: Some HTTP specific error..` (note the `..`)

kangalio:
Thank you for the review Lakelezz. As suggested, here are remaining questions to discuss (extracted from comment inside code):

- Should the error enums be #[non_exhaustive]?
- Should the MessageParseError implement source()? It may be too much boilerplate (maybe add thiserror dependency?)

src/utils/parse.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// Parse a value from a string in context of a received message.
+///
+/// This trait is a superset of [`std::str::FromStr`] (the trait used in [`str::parse`]). The
+/// difference is that this trait aims to support serenity-specific Discord types like [`Member`]
+/// or [`Message`].
+///
+/// Trait implementations may do network requests as part of their parsing procedure.
+///
+/// Useful for implementing command argument parsing frameworks.
```
arqunis:
```suggestion
/// Useful for implementing argument parsing in command frameworks.
```

src/utils/parse.rs
```diff
@@ -0,0 +1,184 @@
+use crate::{model::prelude::*, prelude::*};
```
arqunis:
```suggestion
use crate::model::prelude::*;
use crate::prelude::*;
```

src/utils/parse.rs
```diff
@@ -0,0 +1,184 @@
...
+    async fn parse(_: &Context, _: &Message, s: &str) -> Result<Self, Self::Err> {
+        T::from_str(s)
+    }
+}
+
+/// Error that can be returned from [`Member::parse`].
+#[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
+pub enum MemberParseError {
+    GuildNotInCache,
+    NotFoundOrMalformed,
```
arqunis: These variants should also be documented.

src/utils/parse.rs
```diff
@@ -0,0 +1,184 @@
...
+/// 3. Lookup by name#discrim
+/// 4. Lookup by name
+/// 5. Lookup by nickname
+#[cfg(feature = "cache")]
+#[async_trait::async_trait]
+impl Parse for Member {
+    type Err = MemberParseError;
+
+    // To get impl documentation to show.
+    #[doc = ""]
```
arqunis: I don't find much reason to have to do this. I've tried building the documentation without this and by expanding the impl, I was able to read the above documentation just fine. 

```suggestion
```

src/utils/parse.rs
```diff
@@ -0,0 +1,184 @@
...
+            .cloned()
+            .ok_or(MemberParseError::NotFoundOrMalformed)
+    }
+}
+
+/// Error that can be returned from [`Message::parse`].
+#[derive(Debug)]
+pub enum MessageParseError {
+    Malformed,
+    Http(SerenityError),
```
arqunis: These variants should be documented.

src/utils/parse.rs
```diff
@@ -0,0 +1,184 @@
...
+/// The lookup strategy is as follows (in order):
+/// 1. Lookup by "{channel ID}-{message ID}" (retrieved by shift-clicking on "Copy ID")
+/// 2. Lookup by message ID (the message must be in the context channel)
+/// 3. Lookup by message URL
+#[async_trait::async_trait]
+impl Parse for Message {
+    type Err = MessageParseError;
+
+    // To get impl documentation to show.
+    #[doc = ""]
```
arqunis: Ditto.
```suggestion
```

src/utils/parse.rs
```diff
@@ -0,0 +1,184 @@
+use crate::{model::prelude::*, prelude::*};
+
+/// Parse a value from a string in context of a received message.
+///
+/// This trait is a superset of [`std::str::FromStr`] (the trait used in [`str::parse`]). The
```
arqunis: Users can refer to `FromStr`'s documentation to discover where it is used, no need to explain it here.
```suggestion
/// This trait is a superset of [`std::str::FromStr`]. The
```

arqunis (changes requested):

arqunis:
> Should the error enums be `#[non_exhaustive]`?

All public structs and enums should ideally be marked `#[non_exhaustive]`. It allows us to add new fields/variants without breaking the API.

> Should the `MessageParseError` implement `source()`? It may be too much boilerplate (maybe add `thiserror` dependency?)

If `source` can be implemented for error structs/enums, it should be. Also, while it could ease writing error structs/enums, I wouldn't add `thiserror` just for that. If we were writing them often, it could be beneficial, but I'd like to avoid increasing compilation times for the library.

kangalio:
Okay, I made the newly added error enums non_exhaustive. They also have a `source` implementation where appropriate (it already existed, not sure in which commit I added that).

Are there more things left to tweak?

arqunis (approved):

Issue #1224, opened 2021-02-11
simd-json feature flag


Licenser:
Hi,
we recently started using serenity as in our project. We usually are quite cautious of computational cost and rely on simd accelerated json parsing/rendering. I would love to open a PR to add a feature flag to alow choosing between serde and simd-json but since that'll be quite a bit of work wanted to make sure if that would be a welcome contribution :)?

arqunis:
> if that would be a welcome contribution :)

Yes, it would be. I remember considering to add `simd-json` support to the library a while ago to enhance serialisation/deserialisation performance for those that have SIMD available, but then I forgot about it. I'll gladly merge the pull request once it is open and reviewed.

Zaitara:
This was added in #1237 as mentioned above.

PR #1237, opened 2021-02-21
simd-json feature flag

Licenser:
As suggested in #1224 the first stab as a `simd-json` feature flag. Any thoughts about this so far?

I got a few questions on the code, in a few places data is translated to a `Value` first then into a `struct`, I was wondering what the history of that is?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -157,6 +157,9 @@ unstable_discord_api = []
 utils = ["base64"]
 voice = ["client", "model"]
 
+# Enables simd accelerated parsing
+simdjson = ["simd-json"]
+
 # Backends to pick from:
 # - Rustls Backends
 rustls_backend = ["reqwest/rustls-tls", "async-tungstenite/tokio-rustls", "tokio", "rustls_backend_marker", "bytes"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ serde_json = "1"
 async-trait = "0.1.9"
 
 [dependencies.simd-json]
-version = "0.4.1"
+version = "0.4"
 optional = true
 
 [dependencies.tracing]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,11 @@ bitflags = "1.1"
 serde_json = "1"
 async-trait = "0.1.9"
 
+[dependencies.simd-json]
+version = "0.3.25"
+optional = true
+
+
 [dependencies.tracing]
 version = "0.1.23"
 features = ["log"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,9 +18,8 @@ serde_json = "1"
 async-trait = "0.1.9"
 
 [dependencies.simd-json]
-version = "0.3.25"
+version = "0.4.1"
 optional = true
-
 
 [dependencies.tracing]
 version = "0.1.23"
@@ -136,10 +135,23 @@ package = "http"
 default = ["default_no_backend", "rustls_backend"]
 default_native_tls = ["default_no_backend", "native_tls_backend"]
 default_tokio_0_2 = ["default_no_backend", "rustls_tokio_0_2_backend"]
-default_native_tls_tokio_0_2 = ["default_no_backend", "native_tls_tokio_0_2_backend"]
+default_native_tls_tokio_0_2 = [
+    "default_no_backend",
+    "native_tls_tokio_0_2_backend",
+]
 
 # Serenity requires a backend, this picks all default features without a backend.
-default_no_backend = ["builder", "cache", "client", "framework", "gateway", "model", "http", "standard_framework", "utils"]
+default_no_backend = [
+    "builder",
+    "cache",
+    "client",
+    "framework",
+    "gateway",
+    "model",
+    "http",
+    "standard_framework",
+    "utils",
+]
 
 builder = ["utils"]
 cache = []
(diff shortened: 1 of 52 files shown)


arqunis:
>  in a few places data is translated to a Value first then into a struct, I was wondering what the history of that is?

Do you mean places such as [`Http::get_guild_members`](https://github.com/serenity-rs/serenity/blob/current/src/http/client.rs#L1718-L1742)? The reason that's done is for injecting the `guild_id` field into the objects. The Discord API doesn't send that field in a lot of guild-specific objects. We perform that ourselves for the convenience of users, so that they can directly refer to the guild that the object belongs to without having to search for it.

Licenser:
Ja for example there are a few other places that use `from_value` / `to_value` that  makes sense. It might be worth flipping that around from a performance point of view but that belongs in its own issue / PR :). I was just curious about the history.

Licenser:
The main part is in a good place but I'm unsure what the preference for `voice-model` is. It'll use a lot of the same functionality as the `json.rs` but it feels kind of small to extract into a 3rd crate, would there be objections to just copy & pasting the code into both crates?

arqunis:
Some parts of the `voice-model` crate are already copy-pastes/adaptions from the main crate. Feel free to copy the changes over there as well.

Licenser:
Looked at the voice-model and I don't think that's worth translating just yet. The RawValue of serde_value is quite different than the one from simd-json so making that 'fail over' is near impossible it'd require two entirely different code path which would also pull in the custom struct decoding simd-json does which would also pull in re-visiting the parse to Value, which would also pull in updates to simd-json-derive :joy: (read: it would be a rabbit hole and probably too big for a single PR)

So I would suggest: 
- treat this as a unit of work, it uses simd-json for all en and de-coding other than the voice part. 
- re-visit the parse to value (probably the biggest perf gain to be had here with both simd and serde)
- add simd-json-derive to fast-parse structs
- revisit parsing of the voice-model and switch on the entire struct parsing to alternate between simd and serde

How does that sound?

arqunis:
Sure. I don't consider the `voice-model` crate too important, as not everyone will concern themselves with audio. If it simplifies the process of adding `simd-json`-based parsing for you, we can go ahead with your suggestion.

By the way, while I haven't merged the pull request yet, thanks for the hard work thus far! Those that have SIMD available and are looking for a speed boost in their bots will certainly appreciate your work.

Cargo.toml
```diff
@@ -17,6 +17,11 @@ bitflags = "1.1"
 serde_json = "1"
 async-trait = "0.1.9"
 
+[dependencies.simd-json]
+version = "0.3.25"
+optional = true
+
+
```
arqunis:
```suggestion
```

Cargo.toml
```diff
@@ -167,4 +175,4 @@ native_tls_backend_marker = []
 
 [package.metadata.docs.rs]
 all-features = true
-rustdoc-args = ["--cfg", "docsrs"]
+rustdoc-args = ["--cfg", "docsrs"]
```
arqunis: Missing newline.

src/builder/edit_role.rs
```diff
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 
-use crate::internal::prelude::*;
 use crate::model::{guild::Role, Permissions};
+use crate::{internal::prelude::*, json::from_number};
```
arqunis: These should be separate `use`s.
```suggestion
use crate::internal::prelude::*;
use crate::json::from_number;
```

src/cache/cache_update.rs
```diff
@@ -16,7 +16,7 @@ use super::Cache;
 /// Creating a custom struct implementation to update the cache with:
 ///
 /// ```rust,ignore
-/// use serde_json::json;
+/// use crate::json::json;
 /// use serenity::{
 ///     cache::{Cache, CacheUpdate},
 ///     model::{
```
arqunis: `crate` should be `serenity`, and thus be a part of the import below. Usage of `serde_json` in the example further down should also be updated to be json-impl agnostic.  
```suggestion
/// use serenity::{
///     json::json,
///     cache::{Cache, CacheUpdate},
///     model::{
```

src/json.rs
```diff
@@ -0,0 +1,143 @@
+use serde::de::{Deserialize, DeserializeOwned};
```
arqunis, line 5: There should be module-level documentation (`//!` doc comments) explaining the purpose of this module.

src/model/channel/channel_category.rs
```diff
@@ -2,9 +2,9 @@
 use crate::builder::EditChannel;
 #[cfg(feature = "model")]
 use crate::http::{CacheHttp, Http};
-use crate::model::prelude::*;
 #[cfg(all(feature = "model", feature = "utils"))]
 use crate::utils as serenity_utils;
+use crate::{json::from_number, model::prelude::*};
```
arqunis:
```suggestion
use crate::json::from_number;
use crate::model::prelude::*;
```

src/model/guild/guild_id.rs
```diff
@@ -21,14 +19,19 @@ use crate::collector::{
...
 use crate::utils;
 #[cfg(all(feature = "model", feature = "unstable_discord_api"))]
 use crate::{
     builder::CreateInteraction,
     model::interactions::{ApplicationCommand, Interaction},
 };
+use crate::{
+    json::{from_number, NULL},
+    model::prelude::*,
+};
```
```suggestion
use crate::{
    json::{from_number, NULL},
    builder::CreateInteraction,
    model::{
        prelude::*,
        interactions::{ApplicationCommand, Interaction},
    },
};
```

src/model/guild/partial_guild.rs
```diff
@@ -15,13 +17,16 @@ use crate::collector::{
...
 #[cfg(all(feature = "model", feature = "unstable_discord_api"))]
 use crate::{
     builder::CreateInteraction,
     model::interactions::{ApplicationCommand, Interaction},
 };
+use crate::{
+    json::from_number,
+    model::utils::{deserialize_emojis, deserialize_roles},
+};
+use crate::{json::from_value, model::prelude::*};
```
arqunis:
```suggestion
use crate::{
    json::{from_number, from_value},
    model::{
        prelude::*,
        utils::{deserialize_emojis, deserialize_roles},
    },
};
```

arqunis (changes requested):

src/model/guild/guild_id.rs
Licenser: Those two can not be combined since the first one depends on the `unstable_discord_api` feature the second one doesn't

src/model/guild/partial_guild.rs
Licenser: see above

src/model/guild/guild_id.rs
arqunis: Ah, didn't pay attention to that. This is fine then.

Licenser:
Sorry for the late reply, I didn't see the GH notification on the comments :) should be all good now.

(edit) Note: I had to bump the minimum rust version for the MSRV test to 1.49 to be able to use the simd-json traits

arqunis:
> I had to bump the minimum rust version for the MSRV test to 1.49 to be able to use the simd-json traits

That is unfortunate. We consider increasing the MSRV as a breaking change, which are not allowed on the `current` branch (unless there is an exception, such as fixing a critical bug). You will have to retarget your pull request to the `next` branch.

Anyhow, I don't have any issues with the rest of the pull request. After the target branch is changed to `next`, it should be fine to merge this. Thanks for the effort!

Licenser:
That's perfectly sensible :) I rebased to next and updated the PR to point to it

arqunis (approved):

PR #1239, opened 2021-02-23
docs: Add note on Guild presences field for requiring guild presences intent

drklee3:
A small addition to make it more clear that the guild presence intent is required.

### DIFF
--- a/src/model/guild/mod.rs
+++ b/src/model/guild/mod.rs
@@ -131,6 +131,9 @@ pub struct Guild {
     /// The Id of the [`User`] who owns the guild.
     pub owner_id: UserId,
     /// A mapping of [`User`]s' Ids to their current presences.
+    ///
+    /// This will be empty unless the "guild presences" privileged intent is
+    /// enabled.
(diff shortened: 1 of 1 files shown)


src/model/guild/mod.rs
```diff
@@ -131,6 +131,9 @@ pub struct Guild {
     /// The Id of the [`User`] who owns the guild.
     pub owner_id: UserId,
     /// A mapping of [`User`]s' Ids to their current presences.
+    ///
+    /// This will be empty unless the "guild presences" privileged intent is
```
Lakelezz: This is quite the good clue! That's why I would add the **Info** label.
```suggestion
    /// **Info**:
    /// This will be empty unless the "guild presences" privileged intent is
```

drklee3: 🕵️ Should it be info or note? I see note used quite more often.

Lakelezz: 🕵️ That is a good idea, we can use `**Note**` instead!

PR #1240, opened 2021-02-23
Remove `Box` indirection on `framework`.

Lakelezz:
# Description
This pull request removes the `Box` around the framework, allowing to `as`-cast for the `ClientBuilder::framework_arc`-method.
The `as`-cast is now performed in the method itself, accepting a generic framework instead, easing the use of the API.

# Type of Change
This pull request *breaks the API* in a number of places and relates to code residing in the *framework*-module.
It enhances Serenity by removing an indirection (`Box`), which improves speed and size. Additionally, it takes the burden of casting the framework to a trait object.

# How Has This Been Tested?
The code compiles still.

The removal of the `as`-cast was tested by performing this:

```rust
let framework = Arc::new(StandardFramework::new());
client_builder.framework_arc(Arc::clone(&framework));
```

### DIFF
--- a/src/client/bridge/gateway/shard_manager.rs
+++ b/src/client/bridge/gateway/shard_manager.rs
@@ -346,7 +346,7 @@ pub struct ShardManagerOptions<'a> {
     pub event_handler: &'a Option<Arc<dyn EventHandler>>,
     pub raw_event_handler: &'a Option<Arc<dyn RawEventHandler>>,
     #[cfg(feature = "framework")]
-    pub framework: &'a Arc<Box<dyn Framework + Send + Sync>>,
+    pub framework: &'a Arc<dyn Framework + Send + Sync>,
     pub shard_index: u64,
     pub shard_init: u64,
     pub shard_total: u64,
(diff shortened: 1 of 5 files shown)


src/client/bridge/gateway/shard_manager.rs
```diff
@@ -38,7 +38,7 @@ use crate::CacheAndHttp;
 /// Initialize a shard manager with a framework responsible for shards 0 through
 /// 2, of 5 total shards:
 ///
-/// ```rust,no_run
+/// ```rust
```
arqunis: Are we sure we want to run this? It would call `Http::get_gateway` for each check in CI; I'd like to not spam Discord needlessly.
```suggestion
/// ```rust,no_run
```

arqunis (changes requested):

src/client/bridge/gateway/shard_manager.rs
Lakelezz: That was an unintended regression.

arqunis (approved):

arqunis:
Thanks.

PR #1248, opened 2021-03-02
Fix the type of `token` and add field `kind` to `Webhook`

### LINKED ISSUE #1246: Webhook's token field is optional
The current [Webhook struct](https://docs.rs/serenity/0.10.2/serenity/model/webhook/struct.Webhook.html) does not account for different webhook types and webhook's `token` field being optional. [Discord's documentation on webhooks](https://discord.com/developers/docs/resources/webhook#webhook-object-webhook-structure) marks field `token` as optional and so, it does not get included with type 2 webhooks. Currently (with "0.10.2"), retrieving a type 2 webhook, for example, when fetching guild's webhook list, will fail with the following message:
```
Error while sending HTTP request.

Caused by:
    0: Error while sending HTTP request.
    1: error decoding response body: missing field `token` at line 1 column 8716
    2: missing field `token` at line 1 column 8716
```

arqunis:
## Description

This changes the type for `Webhook::token` to `Option<String>` and adds a new field to the struct called `kind` with the type `WebhookType`. This updates the structure to [conform to Discord's schema](https://discord.com/developers/docs/resources/webhook#webhook-object-webhook-structure) and to resolve https://github.com/serenity-rs/serenity/issues/1246.

## Type of Change

This is a multiclass change in the `model` module. It improves the correspondence with Discord's API for webhooks, and fixes an issue with the `token` definition, but has to break the library's APi to fix it. This targets `current` despite the breaking change, as otherwise webhooks are broken.

## How Has This Been Tested?

This has been tested by calling `Guild::webhooks` in a guild that had both an incoming webhook and a channel follower webhook. The result was a success, returning a list with the webhooks without errors.

### DIFF
--- a/src/model/webhook.rs
+++ b/src/model/webhook.rs
@@ -17,9 +17,46 @@ use crate::http::Http;
 #[cfg(feature = "model")]
 use crate::internal::prelude::*;
 #[cfg(feature = "model")]
+use crate::model::prelude::*;
+#[cfg(feature = "model")]
 use crate::model::ModelError;
 #[cfg(feature = "model")]
 use crate::utils;
+
+/// A representation of a type of webhook.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
+#[non_exhaustive]
+pub enum WebhookType {
+    /// An indicator that the webhook can post messages to channels with
+    /// a token.
+    Incoming = 1,
+    /// An indicator that the webhook is managed by Discord for posting new
+    /// messages to channels without a token.
+    ChannelFollower = 2,
+}
+
+enum_number!(WebhookType {
+    Incoming,
+    ChannelFollower
+});
+
+impl WebhookType {
+    #[inline]
+    pub fn name(&self) -> &str {
+        match self {
+            WebhookType::Incoming => "incoming",
+            WebhookType::ChannelFollower => "channel follower",
+        }
+    }
+
+    #[inline]
+    pub fn num(self) -> u64 {
+        match self {
+            WebhookType::Incoming => 1,
+            WebhookType::ChannelFollower => 2,
+        }
+    }
+}
 
 /// A representation of a webhook, which is a low-effort way to post messages to
 /// channels. They do not necessarily require a bot user or authentication to
@@ -31,6 +68,9 @@ pub struct Webhook {
     ///
     /// Can be used to calculate the creation date of the webhook.
     pub id: WebhookId,
+    /// The type of the webhook.
+    #[serde(rename = "type")]
+    pub kind: WebhookType,
     /// The default avatar.
     ///
     /// This can be modified via [`ExecuteWebhook::avatar_url`].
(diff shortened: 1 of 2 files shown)


src/model/webhook.rs
```diff
@@ -132,9 +180,14 @@ impl Webhook {
     ///
     /// # Errors
     ///
-    /// May return an [`Error::Http`] if the content is illformed, or if the token is invalid.
-    /// May also return an [`Error::Json`] if there is an error in deserializing Discord's response.
+    /// Returns an [`Error::Model`] if the [`token`] is `None`.
+    ///
+    /// May also return an [`Error::Http`] if the content is illformed, or if the token is invalid.
```
Lakelezz:
```suggestion
    /// May also return an [`Error::Http`] if the content is malformed, or if the token is invalid.
```
It's more common to say *malformed*.

src/model/webhook.rs
```diff
@@ -132,9 +180,14 @@ impl Webhook {
...
+    ///
+    /// Or may return an [`Error::Json`] if there is an error in deserializing Discord's response.
```
Lakelezz:
```suggestion
    /// Or may return an [`Error::Json`] if there is an error in deserialising Discord's response.
```

src/model/webhook.rs
```diff
@@ -230,9 +285,14 @@ impl Webhook {
     ///
     /// # Errors
     ///
-    /// May return [`Error::Http`] if the content is illformed, or if the webhook's token is invalid.
-    /// Also, may return an [`Error::Json`] if there is an error deserializing Discord's response.
+    /// Returns an [`Error::Model`] if the [`token`] is `None`.
+    ///
+    /// May also return an [`Error::Http`] if the content is illformed, or if the webhook's token is invalid.
```
Lakelezz:
```suggestion
    /// May also return an [`Error::Http`] if the content is malformed, or if the webhook's token is invalid.
```
As said, *malformed* is the common way to express this.

src/model/webhook.rs
```diff
@@ -230,9 +285,14 @@ impl Webhook {
...
+    ///
+    /// Or may return an [`Error::Json`] if there is an error deserializing Discord's response.
```
Lakelezz:
```suggestion
    /// Or may return an [`Error::Json`] if there is an error deserialising Discord's response.
```

src/model/webhook.rs
```diff
@@ -273,15 +328,20 @@ impl Webhook {
...
     /// # Errors
     ///
-    /// Returns an [`Error::Http`] if the http client errors or if Discord returns an error.
+    /// Returns an [`Error::Model`] if the [`token`] is `None`.
+    ///
+    /// May also return an [`Error::Http`] if the http client errors or if Discord returns an error.
     /// Such as if the `Webhook` was deleted.
     ///
-    /// May also return an [`Error::Json`] if there is an error in deserializing the response.
+    /// Or may return an [`Error::Json`] if there is an error deserializing Discord's response.
```
Lakelezz:
```suggestion
    /// Or may return an [`Error::Json`] if there is an error deserialising Discord's response.
```

Lakelezz (changes requested):
Thanks for fixing this, I just raised some textual concerns. The American wording is not even authored by you, but while we are at it, I hope it's okay for you to change that :)

arqunis:
Done, performed a replace operation with regex for illformed to malformed and deserialization to deserialisation.

Lakelezz (approved):
Perfect, thanks a lot for taking your time to fix this! :)

PR #1258, opened 2021-03-13
Improve conformity and fix updating certain fields for `Member`

arqunis:
## Description

This adds a new field called `premium_since` to the `Member` struct, increasing the conformity to [the `Member` object schema](https://discord.com/developers/docs/resources/guild#guild-member-object). It also fixes handling the update for the `joined_at` and `pending` fields.

## Type of Change

This is an improvement and a fix concerning the `Member` struct, a part of the `model` module.

## How Has This Been Tested?

~~This hasn't been tested yet, but hypothetically it should work.~~ This has been tested by a member from the Serenity guild, works as expected.

### DIFF
--- a/src/model/event.rs
+++ b/src/model/event.rs
@@ -496,13 +496,13 @@ impl CacheUpdate for GuildMemberUpdateEvent {
                 guild.members.insert(self.user.id, Member {
                     deaf: false,
                     guild_id: self.guild_id,
-                    joined_at: None,
+                    joined_at: Some(self.joined_at),
                     mute: false,
                     nick: self.nick.clone(),
                     roles: self.roles.clone(),
                     user: self.user.clone(),
-                    pending: false,
-                    premium_since: None,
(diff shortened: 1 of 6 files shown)


src/model/event.rs
```diff
@@ -473,23 +477,22 @@ impl CacheUpdate for GuildMemberUpdateEvent {
...
+                member.premium_since.clone_from(&self.premium_since);
 
                 item
             } else {
                 None
             };
 
-            if !found {
+            if item.is_none() {
                 guild.members.insert(self.user.id, Member {
```
drklee3, line 496: Should this be inserting `Some(self.joined_at)` in addition to the other `pending` and `premium_since` fields on `self` instead of None/false values?

arqunis: It should, good catch! I made this change without much thought, but taking a second look, the `Member` instance should contain data from the update event when it's missing.

PR #1262, opened 2021-03-16
Add documentation for the `#[hook]` macro

arqunis:
## Description

This simply adds documentation to the `#[hook]` macro, which prior to this had no documentation.

## Type of Change

This is an improvement to the documentation of the `command_attr` crate for one of its macros.

## How Has This Been Tested?

This has been tested by personally proofreading the documentation text and verifying that building documentation for the `command_attr` is successful, which it is.


drklee3 (commented):
I read it over and it looks good :+1: (learned some new things)

However I am not sure if it is very clear to users what `#[hook]` is used for in the context of actually using it with serenity.  I think it would be helpful if it references where and how it is used, like `StandardFramework`'s before / after / normal_message, etc.

arqunis:
> However I am not sure if it is very clear to users what `#[hook]` is used for in the context of actually using it with serenity.  I think it would be helpful if it references where and how it is used, like `StandardFramework`'s before / after / normal_message, etc.

Isn't this segment sufficient enough?
> The macro exists to retain the normal syntax of `async` functions (and closures), while
granting the user the ability to pass those functions to the framework, like command functions
and hooks (`before`, `after`, `on_dispatch_error`, etc.).

There isn't a dedicated section for its usage in Serenity, but I believe that by at least mentioning command functions and hooks, it should be easily inferred what the intended purpose for it is.

drklee3:
> Isn't this segment sufficient enough?
> 
> > The macro exists to retain the normal syntax of `async` functions (and closures), while
> > granting the user the ability to pass those functions to the framework, like command functions
> > and hooks (`before`, `after`, `on_dispatch_error`, etc.).

Ah yeah seems like I missed that part, that's just about what I had in mind!

drklee3 (approved):

PR #1266, opened 2021-03-24
feat: Add Discord API proxy support

drklee3:
### Description

Adds a new `HttpBuilder` to allow configuration of the http client to disable the proxy and/or add an API proxy.  This should be a non-breaking change as it preserves the existing `Http::new` and `Http::new_with_token` fns.

### Tested

Added basic http builder unit tests for default and proxy/disabled ratelimiter.  Also tested with twilight-http-proxy and API requests correctly get sent to the proxy instead of Discord API directly.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2331,10 +2331,14 @@ impl Http {
         T: Into<AttachmentType<'a>>,
     {
         let uri = api!("/channels/{}/messages", channel_id);
-        let url = match Url::parse(&uri) {
+        let mut url = match Url::parse(&uri) {
             Ok(url) => url,
             Err(_) => return Err(Error::Url(uri)),
         };
+
+        if let Some(proxy) = &self.proxy {
+            url.set_host(Some(proxy)).map_err(|e| Error::Http(Box::new(e.into())))?;
+        }
 
         let mut multipart = reqwest::multipart::Form::new();
 
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -4,6 +4,7 @@ use std::{
     fmt,
     future::Future,
     pin::Pin,
+    str::FromStr,
     sync::Arc,
     task::{Context as FutContext, Poll},
 };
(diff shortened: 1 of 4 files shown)


src/http/client.rs
```diff
@@ -2168,11 +2334,17 @@ impl Http {
...
+        let mut url = match Url::parse(&uri) {
             Ok(url) => url,
             Err(_) => return Err(Error::Url(uri)),
         };
 
+        if let Some(proxy) = &self.proxy {
+            url.set_host(proxy.host_str()).map_err(|e| HttpError::Url(e))?;
+            url.set_scheme(proxy.scheme()).map_err(|_| HttpError::InvalidScheme)?;
+            url.set_port(proxy.port()).map_err(|_| HttpError::InvalidPort)?;
+        }
```
drklee3, line 2346: Open to suggestions on refactoring this bit and the `proxy` field in general, feels a bit messy using `Url`.  Alternative way would be the previous version, just using `String` for the proxy field type and using a `String::replace` to create a new `Url`.

src/http/client.rs
```diff
@@ -28,6 +37,157 @@ use crate::http::routing::Route;
...
+            client: None,
+            ratelimiter: None,
+            ratelimiter_disabled: Some(false),
+            token: None,
+            proxy: None,
+            fut: None,
+        }
+    }
+
+    /// Construct a new builder to call methods on for the Http construction.
```
Lakelezz:
```suggestion
    /// Construct a new builder to call methods on for the HTTP construction.
```
Just consistency.

Lakelezz (changes requested):
Just a small nit.

src/http/client.rs
arqunis: Usage of `Url` is fine. It ensures the proxy is given with a valid url. Besides, this seems to be the only place where the url has to be changed to the proxy if one exists.

arqunis (approved):

PR #1267, opened 2021-03-25
Allow editing and deleting messages from webhooks

arqunis:
## Description

This adds support for editing and deleting messages originating from webhooks. Current `Message::{edit, delete}` methods cannot be used as the endpoints for webhook messages are different, with slightly different data that can be submitted when editing a message.

## Type of Change

This is an enhancement to the HTTP and model interfaces by covering more surface area of the Discord API.

## How Has This Been Tested?

This has been tested by sending a webhook message, editing, and then deleting it. All operations have been successful.

### DIFF
--- a/src/http/routing.rs
+++ b/src/http/routing.rs
@@ -655,6 +655,13 @@ impl Route {
         format!(api!("/webhooks/{}/{}?wait={}"), webhook_id, token, wait)
     }
 
+    pub fn webhook_message<D>(webhook_id: u64, token: D, message_id: u64) -> String
+    where
+        D: Display,
+    {
+        format!(api!("/webhooks/{}/{}/messages/{}"), webhook_id, token, message_id)
+    }
+
     #[cfg(feature = "unstable_discord_api")]
     #[cfg_attr(docsrs, doc(cfg(feature = "unstable_discord_api")))]
     pub fn webhook_original_interaction_response<D: Display>(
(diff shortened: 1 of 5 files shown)


src/http/routing.rs
```diff
@@ -655,6 +655,13 @@ impl Route {
         format!(api!("/webhooks/{}/{}?wait={}"), webhook_id, token, wait)
     }
 
+    pub fn webhook_message<D>(webhook_id: u64, token: D, message_id: u64) -> String
```
Lakelezz, line 664: We should add documentation to this function.

Lakelezz (commented):
You added 
> // FIXME: This should probably be another route.

What do you think should be the other route?

src/http/routing.rs
arqunis: I didn't add documentation to the function since the rest of the routing functions themselves didn't have any documentation, and I didn't want to be inconsistent by documenting just one function. Would it be fine to make an exception just for this, or document all of the routing functions as well?

Lakelezz: Oh, you are right. Let's reconsider this later :)

arqunis:
> What do you think should be the other route?

`WebhooksIdMessagesId(u64)` for the `/webhooks/:webhook_id/:token/messages/:message_id` path, based on the pattern of other routes. I didn't incorporate it to the pull request as I wasn't certain if I should add a new route. Thinking about it again, however, I most likely should.

PR #1276, opened 2021-04-01
feat(model): Add `rtc_region` and `video_quality_mode` guildchannel fields

drklee3:
#### Description

Adds new fields `rtc_region` and `video_quality_mode` to `GuildChannel`.

Documentation:

- rtc_region: https://github.com/discord/discord-api-docs/pull/2692
- video_quality_mode: https://github.com/discord/discord-api-docs/pull/2754

#### Tested

`cargo test --all-features` passes.  

Basic test using `ChannelId(channel_id)to_channel(&ctx).await` shows deserialization works with a `rtc_region` override set and `video_quality_mode` set to 720p in channel settings:

### DIFF
--- a/src/model/channel/guild_channel.rs
+++ b/src/model/channel/guild_channel.rs
@@ -104,7 +104,8 @@ pub struct GuildChannel {
     pub slow_mode_rate: Option<u64>,
     /// The region override.
     ///
-    /// **Note**: This is only available for voice and stage channels.
+    /// **Note**: This is only available for voice and stage channels. `None`
+    /// for voice and stage channels means automatic region selection.
     pub rtc_region: Option<String>,
     /// The video quality mode for a voice channel.
     pub video_quality_mode: Option<VideoQualityMode>,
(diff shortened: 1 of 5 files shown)


src/model/channel/guild_channel.rs
```diff
@@ -102,6 +102,12 @@ pub struct GuildChannel {
     /// channels.
     #[serde(default, rename = "rate_limit_per_user")]
     pub slow_mode_rate: Option<u64>,
+    /// The region override.
+    ///
```
FelixMcFelix, line 106: Would you be able to clarify here that a `None` value on this field for a compatible channel (voice/stage) means automatic region selection?

FelixMcFelix (changes requested):
Thanks for this. I think there's just a point on semantics to clear up in the docs, but beyond that LGTM!

FelixMcFelix (approved):

arqunis (approved):

PR #1286, opened 2021-04-09
Added the invite reminder system message and flag

vicky5124:
https://github.com/discord/discord-api-docs/commit/6b816cb6f542ea813c8f4dab797fc0e7ce203305
Also added the possibility to configure the messages that can be sent to the system channel of the guild.

	modified:   src/builder/edit_guild.rs
	modified:   src/framework/standard/structures/check.rs
	modified:   src/model/channel/message.rs
	modified:   src/model/channel/mod.rs
	modified:   src/model/channel/reaction.rs
	modified:   src/model/guild/mod.rs
	new file:   src/model/guild/system_channel.rs

### DIFF
--- a/src/builder/edit_guild.rs
+++ b/src/builder/edit_guild.rs
@@ -265,7 +265,31 @@ impl EditGuild {
         self.0.insert("verification_level", num);
     }
 
-    pub fn system_channel_flags(&mut self, system_channel_flags: SystemChannelFlags) {
+    /// Modifies the notifications that are sent by discord to the configured system channel.
+    ///
+    /// ```rust,no_run
+    /// # use serenity::{http::Http, model::id::GuildId};
+    /// #
+    /// # async fn run() -> Result<(), Box<dyn std::error::Error>> {
+    /// #     let http = Http::default();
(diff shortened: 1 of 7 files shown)


src/builder/edit_guild.rs
```diff
@@ -264,4 +264,8 @@ impl EditGuild {
         let num = Value::Number(Number::from(verification_level.num()));
         self.0.insert("verification_level", num);
     }
+
+    pub fn system_channel_flags(&mut self, system_channel_flags: SystemChannelFlags) {
+        self.0.insert("system_channel_flags", system_channel_flags.bits().into());
+    }
```
arqunis: This should return `&mut Self` to allow chaining the builder with other methods. Documentation should also be added.
```suggestion
    pub fn system_channel_flags(&mut self, system_channel_flags: SystemChannelFlags) -> &mut Self {
        self.0.insert("system_channel_flags", system_channel_flags.bits().into());
        self
    }
```

arqunis (changes requested):

src/builder/edit_guild.rs
vicky5124: Fixed on 1e3f689

arqunis (approved):

PR #1288, opened 2021-04-09
Add voice status to the MemberUpdate event

vicky5124:
https://github.com/discord/discord-api-docs/commit/fa9a0739eeedf3ff380b77b9fdcf1302aa1b0c26

	modified:   src/model/event.rs

### DIFF
--- a/src/model/event.rs
+++ b/src/model/event.rs
@@ -466,6 +466,10 @@ pub struct GuildMemberUpdateEvent {
     pub premium_since: Option<DateTime<Utc>>,
     #[serde(default)]
     pub pending: bool,
+    #[serde(default)]
+    pub deaf: bool,
+    #[serde(default)]
+    pub mute: bool,
 }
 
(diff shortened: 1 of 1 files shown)


arqunis:
These fields should also update the member object if one is present, like it is done for other fields already: https://github.com/serenity-rs/serenity/blob/aaf32708ce4a409f2ae72192ae5e77921f87c6e0/src/model/event.rs#L487-L492.

drklee3:
Just to note the mute and deaf fields don't seem to actually change yet.  Mute / deafen doesn't trigger a member update from what I tested a few days ago.

vicky5124:
> Just to note the mute and deaf fields don't seem to actually change yet. Mute / deafen doesn't trigger a member update from what I tested a few days ago.

They do not trigger on mute/deafening, but if the member happens to be updated while being muted or deaf, the value does seem to be sent.

arqunis (approved):

PR #1291, opened 2021-04-12
Add timeouts to Http Client.

merlleu:
Hello,
I'm having an issue with discord API not responding sometimes, I found out timeouts weren't configured on reqwest client. This Pull request add timeouts for discord (30s timeout for response and 3s for connection).


Lakelezz (changes requested):
Hello, please do not hardcode these, expose them via the `ClientBuilder` and the `Http` structure.

Lakelezz:
Hello, thanks, can you please add a build-method to the [`ClientBuilder`](https://docs.rs/serenity/0.10.5/serenity/client/struct.ClientBuilder.html) too?
By that I mean a method mutating the builder and not constructing it.

merlleu:
Hello, Thanks for your feedback, is this implementation ok for you ?

src/client/mod.rs
```diff
@@ -77,6 +76,8 @@ pub struct ClientBuilder<'a> {
     intents: GatewayIntents,
     #[cfg(feature = "cache")]
     timeout: Option<Duration>,
```
arqunis: Since you have removed the above feature gate of the import, you can also remove the one here. It's not necessary anymore.
```suggestion
    timeout: Option<Duration>,
```

arqunis (changes requested):

src/client/mod.rs
```diff
@@ -287,6 +292,58 @@ impl<'a> ClientBuilder<'a> {
 
         self
     }
+
+    /// Sets how long the http client should wait for the entire response from discord API.
```
arqunis: Discord is a proper noun in this context, thus it should be capitalised.
```suggestion
    /// Sets how long the http client should wait for the entire response from the Discord API.
```

src/client/mod.rs
```diff
@@ -287,6 +292,58 @@ impl<'a> ClientBuilder<'a> {
...
+    /// By default, an http request will never timeout, it could create a deadlock sometimes.
```
arqunis: We prefer British English spelling and pronunciation, where the 'h' in words such as "http" is pronounced rather than being a silent letter. As a result, this affects the choice of the a/an determiner.
```suggestion
    /// By default, a http request will never timeout, it could create a deadlock sometimes.
```

src/client/mod.rs
```diff
@@ -287,6 +292,58 @@ impl<'a> ClientBuilder<'a> {
...
+    ///
+    /// *Info*:
```
arqunis: We use the word *note* in **bold** in the documentation when we want to state additional information that the user should be aware of.
```suggestion
    /// **Note**:
```

src/client/mod.rs
```diff
@@ -287,6 +292,58 @@ impl<'a> ClientBuilder<'a> {
...
+                ));
+            },
+            None => {
+                self.http = Some(Http::new_with_timeouts(
+                    self.http_request_timeout,
+                    self.http_connect_timeout,
+                    "",
+                ));
+            },
+        };
```
arqunis:
```suggestion
        let old_token = match self.http {
            Some(old_http) => &old_http.token,
            None => "",
        };

        self.http = Some(Http::new_with_timeouts(
            self.http_request_timeout,
            self.http_connect_timeout,
            old_token,
        ));
```

src/client/mod.rs
```diff
@@ -287,6 +292,58 @@ impl<'a> ClientBuilder<'a> {
...
```
arqunis:
```suggestion
        let old_token = match self.http {
            Some(old_http) => &old_http.token,
            None => "",
        };

        self.http = Some(Http::new_with_timeouts(
            self.http_request_timeout,
            self.http_connect_timeout,
            old_token,
        ));
```

arqunis (changes requested):
I have some pedantic requests, but otherwise this should be fine. Furthermore, would you mind fixing the conflicts?

src/http/client.rs
```diff
@@ -238,6 +240,33 @@ impl Http {
...
+    /// The timeout is applied from when the request starts connecting until the response body has finished.
+    /// The connect_timeout is applied for only the connect phase.
+    pub fn new_with_timeouts(timeout: std::time::Duration, connect_timeout: std::time::Duration, token: &str) -> Self {
+        let builder = configure_client_backend(Client::builder())
+            .timeout(timeout)
+            .connect_timeout(connect_timeout);
+        
+        let built = builder.build().expect("Cannot build reqwest::Client");
+        let client = Arc::new(built);
+        let client2 = Arc::clone(&client);
```
arqunis: This shouldn't be necessary.
```suggestion
```

arqunis (changes requested):

Milo123459:
Any major updates on this?

PR #1295, opened 2021-04-17
Introduce intradoc-links in a lot of places

kangalio:
I went through the serenity source code using VSCode [regex](https://pastebin.com/raw/28HZKR15) find-and-replace and tried to find and convert as many item references into intradoc links as possible.

There were a few things I am uncertain about:
- should items from std be turned into intradoc-links as well (None, Future, HashSet, File, Path, Result, Vec, String, Arc)?
- should references to `Self` methods be written as
  ```
  Lorem ipsum you can call [`method_name`] to do foo bar.
  ...
  [`method_name`]: Self::method_name
  ```
  or just
  ```
  Lorem ipsum you can call [`Self::method_name`] to do foo bar.
  ```

Regexes used (can't enclose in backquotes because the regexes contain backquotes themselves):
- Initial regex to find all items within code-blocks that are not intradoc-links: (///.*?)([^\\[])(\`[\w:_]+\`)([^\\]]|^) => $1$2[$3]$4
- To find remaining std items, this regex matches camel-case names inside code blocks that are not intradoc links: (///.\*[^\\[])\`([A-Z][a-z][A-Za-z]\*)\`([^\\]]|$) => $1[\`$2\`]$3


arqunis:
> There were a few things I am uncertain about: 
>
> - should items from std be turned into intradoc-links as well (None, Future, HashSet, File, Path, Result, Vec, String, Arc)?
> - should references to Self methods be written as 

For the first point, yes. 
For the second point, it should be the latter. As far I am aware, links to methods of `Self` are on their own lines in a few places already. Thus, for consistency, the same should be done.

Lakelezz:
I agree with acdenisSK!

src/error.rs
```diff
@@ -36,7 +36,7 @@ use crate::model::ModelError;
 pub type Result<T> = StdResult<T, Error>;
 
 /// A common error enum returned by most of the library's functionality within a
-/// custom [`Result`].
+/// custom [[`Result`]].
```
nickelc, line 39: this looks wrong

nickelc:
```suggestion
/// custom [`Result`].
```

kangalio: Thanks for finding those two cases

src/model/channel/channel_id.rs
```diff
@@ -710,6 +710,7 @@ impl ChannelId {
     /// [`CreateMessage::content`]: crate::builder::CreateMessage::content
     /// [Attach Files]: Permissions::ATTACH_FILES
     /// [Send Messages]: Permissions::SEND_MESSAGES
+    /// [`File`]: std::fs::File
```
nickelc: Tokio's `File` is used for `Into<AttachmentType>`.  See example.
```suggestion
    /// [`File`]: tokio::fs::File
```

arqunis (approved):

PR #1305, opened 2021-04-25
Support all guild fields

Zaitara:
This PR adds support to the following guild fields:

- Welcome screens, see discord/discord-api-docs#2398
- rules channel id, public updates channel id
- approximate and maximum member/presence counts, max video channel users
- nsfw
- discovery splash

It also corrects the `EditGuild` builder documentation which was incorrect because system channel is not restricted to community guilds.

This has been tested.

Continued in #1306


Zaitara:
Continued in #1306

PR #1310, opened 2021-04-29
Better errors display

Zaitara:
This PR improves the error system which was introduced in v8: https://discord.com/developers/docs/reference#error-messages

It gives an additionnal `errors` field which lists each problematic field with an error code, message and a path to this field in the request.

It was tested.

### DIFF
--- a/src/http/error.rs
+++ b/src/http/error.rs
@@ -17,7 +17,7 @@ pub struct DiscordJsonError {
     pub code: isize,
     /// The error message.
     pub message: String,
-    /// The full explaned errors with their path in the request
+    /// The full explained errors with their path in the request
     /// body.
     pub errors: Vec<DiscordJsonSingleError>,
 }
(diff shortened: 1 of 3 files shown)


kangalio:
This is great! The somewhat obscure error messages were an annoyance to display coherently to the user, which is hopefully much easier and better now!

Zaitara:
> This is great! The somewhat obscure error messages were an annoyance to display coherently to the user, which is hopefully much easier and better now!

Indeed, it is less readable but much more easier to debug, as it also support several errors and does not show to only one of it, and with a path the developer can immediatly know where is his error coming from.
image

src/http/error.rs
```diff
@@ -4,22 +4,64 @@ use std::{
...
+            errors,
+        })
     }
 }
 
+// impl std::fmt::Debug for DiscordJsonError {
+//     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+//         write!(f, "\"{}\"", self.message)
+//     }
+// }
```
arqunis:
```suggestion
```

arqunis (changes requested):

arqunis (approved):

PR #1332, opened 2021-05-12
Fix message link when discord does not provide `guild_id`

### LINKED ISSUE #1331: Message::link returns broken links
Sometimes (but apparently not always), using `Message::link` doesn't give me the expected `<serverid>/<channelid>/<messageid>` link, but instead generates a link of the form `@me/<channelid>/<messageid>`. That broken link then still links to the right channel and message, but makes discord open it in the DM window (which is also weird behaviour from discords side, but still also a broken link from serenity.

# Expected behaviour
`Message::link()` consistently gives me a valid link to the message

# Current behaviour
`Message::link()` in some circumstances returns a link that contains `@me` where the server-id should be (even when the message i'm linking to _is_ in a server)

Zaitara:
This PR creates a method `Message::link_ensured` that tries to search for the guild Id of a message if Discord does not provide it, as it is the case for messages received through the gateway. `Message::link` fails to generate a correct link when the message lacks the guild Id, as it will always assume the message comes from a DM.

This has been tested.

Closes #1331

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -856,38 +856,16 @@ impl Message {
     /// The link will be valid for messages in either private channels or guilds.
     #[inline]
     pub fn link(&self) -> String {
-        match self.guild_id {
-            Some(guild_id) => format!(
-                "https://discord.com/channels/{}/{}/{}",
-                guild_id.0, self.channel_id.0, self.id.0
-            ),
-            None => format!("https://discord.com/channels/@me/{}/{}", self.channel_id.0, self.id.0),
-        }
+        self.id.link(self.channel_id, self.guild_id)
     }
 
-    /// Same as [`Self::link`] but tries to find the channel
+    /// Same as [`Self::link`] but tries to find the [`GuildId`]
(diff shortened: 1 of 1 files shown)


src/model/channel/message.rs
```diff
@@ -856,13 +856,16 @@ impl Message {
...
-                "https://discord.com/channels/{}/{}/{}",
-                guild_id.0, self.channel_id.0, self.id.0
-            ),
-            None => format!("https://discord.com/channels/@me/{}/{}", self.channel_id.0, self.id.0),
-        }
+        self.id.link(self.channel_id, self.guild_id)
+    }
+
+    /// Same as [`Self::link`] but tries to find the [`GuildId`]
+    /// if discord does not provide it.
```
arqunis: "Discord" is a proper noun (it refers to a specific entity), thus it should be capitalised. 
```suggestion
    /// if Discord does not provide it.
```

arqunis (changes requested):

arqunis (approved):

Issue #1032, opened 2020-10-22
Stricter validate_token checks


leumasme:
The validate_token check can be made more strict, as the parts of the token actually have a meaning.
Split by ".", the token consists of 3 base64-encoded parts:
The UserID as a string, encoded in base64;
A Timestamp of Discords Custom Token Epoch, timestamp of the token creation. (Add +1293840000 => UNIX Timestamp);
A cryptographic base64 HMAC used for the actual authorization.

We notice that just with this info we can already do some extra checks:
Do all of the dot-separated parts only contain valid base64 characters?
Is the first part a Numeric String when base64decoded?
Is the Timestamp in a realistic time range?

The UserID then has further meaning:
Converting the UserID String into a Number and then to Binary:
From Right to Left, Bits 0-12; 13-17; 17-22 are rather uninteresting.
Bits 23 to end converted to decimal are another timestamp of the Discord ID Epoch, which indicates the accounts creation date. (Add +1420070400000 => UNIX Timestamp).
This could also be checked to be in a realistic time range.

These extra checks could more easily catch accidentally adding another character in the middle of ones' token.

kangalio:
> A Timestamp of Discords Custom Token Epoch, timestamp of the token creation. (Add +1293840000 => UNIX Timestamp);

It seems that this is not true anymore; the timestamp now uses the normal UNIX epoch.

While implementing this I noticed that the code yielded the year 2061 for my bot. After removing the Discord epoch correction and using the timestamp as a UNIX timestamp directly, the datetime matched.

Flat:
> 
> 
> > A Timestamp of Discords Custom Token Epoch, timestamp of the token creation. (Add +1293840000 => UNIX Timestamp);
> 
> It seems that this is not true anymore; the timestamp now uses the normal UNIX epoch.
> 
> While implementing this I noticed that the code yielded the year 2061 for my bot. After removing the Discord epoch correction and using the timestamp as a UNIX timestamp directly, the datetime matched.

This will still be true for older tokens. For example mine still needs that constant added, so a check will need to be done. Should be as simple as `if < 1293840000 : add 1293840000`

PR #1342, opened 2021-05-14
Stricter validate_token checks

kangalio:
Implements #1032. In essence, `serenity::client::validate_token` has been changed to fully parse the token's components and check if they are meaningful, to catch more errors.

Tested on one of my bot tokens; the extracted user ID and token creation timestamp are correct.

Question: should the function that extracts the bot user ID and token creation timestamp out of the token be made public?

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -952,22 +952,28 @@ impl Client {
 /// Returns a [`ClientError::InvalidToken`] when one of the above checks fail.
 /// The type of failure is not specified.
 pub fn validate_token(token: impl AsRef<str>) -> Result<()> {
-    let (user_id, generation_time) =
-        parse_token(token.as_ref()).ok_or(Error::Client(ClientError::InvalidToken))?;
+    let token = parse_token(token.as_ref()).ok_or(Error::Client(ClientError::InvalidToken))?;
 
     // Check if timestamps are in a sensible range
-    if user_id.created_at().year() >= 2100 || generation_time.year() >= 2100 {
+    if token.bot_user_id.created_at().year() >= 2100 || token.creation_time.year() >= 2100 {
         return Err(Error::Client(ClientError::InvalidToken));
     }
 
     Ok(())
 }
 
+/// Part of the data contained within a Discord bot token. Returned by [`parse_token`].
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct TokenComponents {
+    pub bot_user_id: UserId,
+    pub creation_time: chrono::NaiveDateTime,
+}
+
 /// Verifies that the token adheres to the Discord token format and extracts the bot user ID and the
 /// token generation timestamp
-fn parse_token(token: &str) -> Option<(UserId, chrono::NaiveDateTime)> {
+pub fn parse_token(token: impl AsRef<str>) -> Option<TokenComponents> {
     // The token consists of three base64-encoded parts
-    let parts: Vec<&str> = token.split('.').collect();
+    let parts: Vec<&str> = token.as_ref().split('.').collect();
     let base64_config = base64::Config::new(base64::CharacterSet::UrlSafe, true);
 
     // First part must be a base64-encoded stringified user ID
@@ -995,5 +1001,8 @@ fn parse_token(token: &str) -> Option<(UserId, chrono::NaiveDateTime)> {
     // Third part is a base64-encoded HMAC that's not interesting on its own
     let _ = base64::decode(parts.get(2)?).ok()?;
 
-    Some((user_id, timestamp))
+    Some(TokenComponents {
+        bot_user_id: user_id,
+        creation_time: timestamp,
+    })
 }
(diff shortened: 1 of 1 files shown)


src/client/mod.rs
```diff
@@ -941,45 +943,59 @@ impl Client {
...
 ///
 /// Returns a [`ClientError::InvalidToken`] when one of the above checks fail.
 /// The type of failure is not specified.
 pub fn validate_token(token: impl AsRef<str>) -> Result<()> {
-    let token = token.as_ref();
+    let (user_id, generation_time) =
+        parse_token(token.as_ref()).ok_or(Error::Client(ClientError::InvalidToken))?;
 
-    if token.is_empty() {
+    dbg!(user_id, generation_time);
```
Zaitara:
```suggestion
```

Flat: I don't think there is value in leaving this debug line in.

src/client/mod.rs
```diff
@@ -941,45 +943,59 @@ impl Client {
...
+
+    // Check if timestamps are in a sensible range
+    if user_id.created_at().year() >= 2100 || generation_time.year() >= 2100 {
```
Flat: This seems arbitrary and creates a Serenity Y2.1K problem.

src/client/mod.rs
```diff
@@ -941,45 +943,59 @@ impl Client {
...
         return Err(Error::Client(ClientError::InvalidToken));
     }
 
+    Ok(())
+}
+
+/// Verifies that the token adheres to the Discord token format and extracts the bot user ID and the
+/// token generation timestamp
+fn parse_token(token: &str) -> Option<(UserId, chrono::NaiveDateTime)> {
```
Zaitara:
```suggestion
pub fn parse_token(token: impl AsRef<str>) -> Option<(UserId, chrono::NaiveDateTime)> {
let token = token.as_ref();
```

src/client/mod.rs
```diff
@@ -941,45 +943,59 @@ impl Client {
...
-///
-/// // tokens must have three parts, separated by periods (this is still
-/// // actually an invalid token):
-/// assert!(validate_token("aaa.abcdefgh.bbb").is_ok());
-///
-/// // the second part must be _at least_ 6 characters long:
-/// assert!(validate_token("a.abcdef.b").is_ok());
-/// assert!(validate_token("a.abcde.b").is_err());
+/// // helpful to prevent typos
+/// assert!(validate_token("Njg4NzYwMjQxMzYzODc3ODg4.C_ikow.j3VupLBuE1QWZng3TMGH0z_UAwg").is_err());
```
Zaitara, line 947: It is not a real token right?

Zaitara (changes requested):

src/client/mod.rs
acdenisSK, line 958:
```suggestion
```

acdenisSK (changes requested):

src/client/mod.rs
```diff
@@ -37,6 +37,7 @@ use std::{
     task::{Context as FutContext, Poll},
 };
 
+use chrono::Datelike;
```
Flat, line 40: @acdenisSK was there a push to remove chrono or are we fine with using it over time still?

src/client/mod.rs
kangalio: Oh crap how did I forget to remove those :facepalm:

src/client/mod.rs
arqunis: I do not recall migrating to the `time` crate over `chrono`. It should be fine to continue using it.

src/client/mod.rs
kangalio: It is fairly arbitrary. Do you have a better suggestion? I think it's not a great idea to query the current time in that function and compare to that; that feels outside the scope of what a "validate_token" function should do; plus it could break depending on OS time configuration.

src/client/mod.rs
kangalio: So you're suggesting to make that method public. In that case would it be good to make a designated struct?
```rust
pub struct TokenComponents {
    bot_user_id: UserId,
    token_creation_time: chrono::NaiveDateTime,
}
```
That way, callers of the function can make it explicit what the "UserId" and the "NaiveDateTime" actually represent.

src/client/mod.rs
kangalio: It's not, it's an adaption of the pseudo token which was there before (see the line above).

src/client/mod.rs
kangalio: Yes, it was an oversight on my part, sorry

src/client/mod.rs
Flat: I do not think this is a necessary step for validation, at the end of the day it gets sent to discord to be accepted or rejected, this would be a more cursory review to ensure the formatting is proper I think. I don't think we are concerned with someone crafting a token with a huge or small timestamp which can't be used anyway.

Flat (approved):

arqunis (approved):

Issue #876, opened 2020-06-02
Add command_name to the dispatch error hook


kingbri1:
`Serenity version: await@a2e8cde`

On a dispatch error, I would like to let the user know what command errored and why. However, this would require parsing the message and filtering the first word without the command prefix. Other included hooks such as `after` and `unrecognized_command` both use command_name to help me know exactly what command is being used.

Example: When someone provides not enough args on my mock command, the dispatch error should fire and be handled by letting the user know that the mock command errored because 0 args were provided and the minimum threshold is 1.

Doing this can also be useful for sending out different error messages based on command name, therefore reducing extra code.

kangalio:
Can this issue be closed now?

kingbri1:
Has someone tested if it works? If so, I'll close the issue.

kangalio:
Yeah, quoting the PR text:

> I tested this by creating a little test bot that triggers almost all possible dispatch error conditions: https://pastebin.com/KNNRM1K1

kingbri1:
Awesome! Thanks for implementing this. Closing issue.

PR #1344, opened 2021-05-15
Dispatch error command name

kangalio:
Implements #876. The dispatch error hook now receives a fourth parameter with the command name.

This is a breaking change so the PR is based on the "next" branch.

I tested this by creating a little test bot that triggers almost all possible dispatch error conditions: https://pastebin.com/KNNRM1K1

Question: is that way of getting the command name correct? I found that part of the code a bit confusing https://github.com/kangalioo/serenity/blob/c82c3031f09a2207c1ffd5e42d95bc3543015edf/src/framework/standard/parse/mod.rs#L514-L549

### DIFF
--- a/src/framework/standard/mod.rs
+++ b/src/framework/standard/mod.rs
@@ -41,42 +41,17 @@ use crate::model::{channel::Message, permissions::Permissions};
 #[cfg(all(feature = "cache", feature = "http", feature = "model"))]
 use crate::model::{guild::Role, id::RoleId};
 
-/// Represents a fail condition under which a command won't be executed.
-#[derive(Debug)]
-pub struct DispatchError {
-    kind: DispatchErrorKind,
-    command_name: String,
-}
-
-impl DispatchError {
-    /// Create a new [`DispatchError`] of certain type with the command name it got triggered on.
-    pub fn new(kind: DispatchErrorKind, command_name: String) -> Self {
-        Self {
-            kind,
-            command_name,
-        }
-    }
-
-    pub fn kind(&self) -> &DispatchErrorKind {
-        &self.kind
-    }
-
-    pub fn command_name(&self) -> &str {
-        &self.command_name
-    }
-}
-
-/// An enum representing all possible fail conditions of a [`DispatchError`].
+/// An enum representing all possible fail conditions under which a command won't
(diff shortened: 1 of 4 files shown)


src/framework/standard/mod.rs
```diff
@@ -722,7 +730,9 @@ impl Framework for StandardFramework {
                     self.should_fail(&ctx, &msg, &mut args, &command.options, &group.options).await
                 {
                     if let Some(dispatch) = &self.dispatch {
-                        dispatch(&mut ctx, &msg, error).await;
+                        let command_name =
+                            command.options.names.get(0).copied().unwrap_or("<unnamed>");
```
arqunis: This case should be impossible to be `None`, but in the case it is, an error should be shown since it would mean a bug in the `#[command]` macro generation. I suggest using index syntax to both simplify the code and to assert that `command.options.names` has to have at least one element -- the command's main name.
```suggestion
                        let command_name = command.options.names[0];
```

src/framework/standard/parse/mod.rs
```diff
@@ -505,17 +511,40 @@ pub async fn command(
             Map::Prefixless(subgroups, commands) => {
                 is_prefixless = true;
 
+                fn command_name_if_recognised(res: &Result<Invoke, ParseError>) -> Option<&str> {
+                    match res {
+                        Ok(Invoke::Command {
+                            command, ..
+                        }) => Some(command.options.names.get(0).copied().unwrap_or("<unknown>")),
```
arqunis: Ditto.
```suggestion
                        }) => Some(command.options.names[0]),
```

src/framework/standard/parse/mod.rs
```diff
@@ -505,17 +511,40 @@ pub async fn command(
...
+                        Ok(Invoke::Help(name)) => Some(name), // is this correct?
```
arqunis: Help commands are handled above, from line 480. This is a useless match arm since `Invoke::Help` isn't conceived by either `handle_group` or `handle_command`, but I don't see harm in it. It can be kept, I suppose.

arqunis (changes requested):

arqunis (approved):

PR #1349, opened 2021-05-18
Make Activity::streaming take a Url

hybras:
Title is self explanatory. I think this is a breaking change, as url parsing is fallible and explicit

### DIFF
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -146,17 +146,18 @@ impl Activity {
     /// # #[cfg(feature = "framework")]
     /// #[command]
     /// async fn stream(ctx: &Context, _msg: &Message, args: Args) -> CommandResult {
-    ///     let STREAM_URL= Url::parse("https://www.youtube.com/watch?v=dQw4w9WgXcQ").unwrap();
-    ///
+    ///     let STREAM_URL = "https://www.youtube.com/watch?v=dQw4w9WgXcQ";
     ///     let name = args.message();
+    ///
     ///     ctx.set_activity(Activity::streaming(&name, STREAM_URL)).await;
     ///
     ///     Ok(())
     /// }
     /// ```
-    pub fn streaming<N>(name: N, url: Url) -> Activity
+    pub fn streaming<'a, N, U>(name: N, url: U) -> Activity
     where
         N: ToString,
+        U: ToString,
     {
         Activity {
             application_id: None,
@@ -175,7 +176,7 @@ impl Activity {
             sync_id: None,
             #[cfg(feature = "unstable_discord_api")]
             session_id: None,
-            url: Some(url.to_string()),
+            url: Some(Url::parse(&url.to_string()).expect("Failed to parse url").into()),
         }
     }
 
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -141,10 +141,12 @@ impl Activity {
     /// # #[cfg(feature = "client")]
     /// use serenity::client::Context;
     ///
+    /// use url::Url;
+    ///
     /// # #[cfg(feature = "framework")]
     /// #[command]
     /// async fn stream(ctx: &Context, _msg: &Message, args: Args) -> CommandResult {
-    ///     const STREAM_URL: &str = "...";
+    ///     let STREAM_URL= Url::parse("https://www.youtube.com/watch?v=dQw4w9WgXcQ").unwrap();
     ///
     ///     let name = args.message();
     ///     ctx.set_activity(Activity::streaming(&name, STREAM_URL)).await;
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -6,6 +6,7 @@ use serde::ser::{Serialize, SerializeStruct, Serializer};
 
 use super::prelude::*;
 use super::utils::*;
+use crate::json::from_value;
 
 /// A representation of the data retrieved from the bot gateway endpoint.
 ///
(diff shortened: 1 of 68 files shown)


hybras:
CI failed due to lints and fmting. Clippy didn't flag my changes, and lint just wanted me to move an import

src/model/gateway.rs
```diff
@@ -151,10 +152,9 @@ impl Activity {
     ///     Ok(())
     /// }
     /// ```
-    pub fn streaming<N, U>(name: N, url: U) -> Activity
+    pub fn streaming<N>(name: N, url: Url) -> Activity
```
arqunis: I find it wrong to force a dependency of Serenity upon users just so they can use this function (the `url` crate is not re-exported from Serenity, thus users will have to define `url` as a dependency in their bots). Could it be changed to a generic that tries to convert to a `Url`? The `Url` type has a `TryFrom<&str>` implementation, which we could use to abstract the construction of the `Url` from users and prevent breaking code as much as possible.

```rust
pub fn streaming<N, U>(name: N, url: U) -> Activity 
where
    N: ToString,
    U: TryInto<Url>,
{
    let url = url.try_into().expect("failed to parse url");
    // ....
}

**Review by arqunis — CHANGES_REQUESTED** [review]

File: src/model/gateway.rs (hunk 1 above)
Comment (hybras) [reply src/model/gateway.rs]: I don't see the harm in users needing to `use url;`. But I'm not opposed to ur suggestions either.

1. Should url be re-exported. If so, where?
2. `TryFrom` it is

Comment (arqunis) [reply src/model/gateway.rs]: There is no harm per se, it just makes the library harder to set up, since you'll need to define `url` as a dependency besides `serenity` (see: [tokio](https://github.com/serenity-rs/serenity/blob/current/examples/e01_basic_ping_bot/Cargo.toml#L8-L9)). It also makes it likely for the `url` of Serenity and the `url` of the user to get out-of-date (for example, we update `url` to 3.0, but the user still depends on 2.2.2, which rustc will classify as separate crates, resulting in types that are incompatible with one another and thus compilation errors).

> Should url be re-exported. If so, where?

At the crate root, [just as it is done for other crates](https://github.com/serenity-rs/serenity/blob/current/src/lib.rs#L131-L137).

Comment (hybras) [reply src/model/gateway.rs]: I couldn't figure out what bound to use that would allow `Url::try_from` generically (in case they added impls for more than `&str`). I also don't understand Rust's Stringy conversions and types, but that's beyond me and this PR. I switched it back to `U: ToString` and internally use `Url::parse`. The signature is now unchanged. 

Because the Url usage is now internal, I did not re-export it

Comment (arqunis) [reply src/model/gateway.rs]: > I couldn't figure out what bound to use that would allow `Url::try_from` generically

That would be the `TryInto` trait I demonstrated in my example snippet. [`TryInto` has a generic implementation for every type that implements `TryFrom`](https://doc.rust-lang.org/nightly/src/core/convert/mod.rs.html#567-576), just like [the regular `Into` and `From` traits](https://doc.rust-lang.org/nightly/src/core/convert/mod.rs.html#533-540). If a type `Foo`, for example, implements `TryFrom<i32>`, then we can depend on this implementation indirectly with `TryInto<Foo>`. This will not only pick up the `TryFrom<i32>` implementation, but also any future `TryFrom` implementations on `Foo`.

Anyway, since we no longer use `Url` in the public API, and the API has not changed, you can retarget your PR to the `current` branch.

Comment (hybras) [reply src/model/gateway.rs]: I *think* then this is what I'm going for. ~~This is back to a breaking change~~. 

Also, thank you for staying with me and fleshing this out

```rust
use url::Url;

use std::convert::TryFrom;
use std::fmt::Debug;

    pub fn streaming<'a, N, U>(name: N, url: U) -> Activity
    where
        N: ToString,
        Url: TryFrom<U>,
        <Url as TryFrom<U>>::Error: Debug,
        U: 'a,
```

hybras: > you can retarget your PR to the current branch

Yes

hybras:
I have a few questions.

1. Is changing the url field of Activity to a url acceptable? Url is serializable and `AsRef<str> , but do the consumers of library care?
2. Is it on the caller or callee to see if its valid (ie, are we taking a string-ish or a url)?
3. streaming currently takes a `ToString`. However, I assume that mostly String and &str are being passed in. If we take some `T where Url : TryFrom<T>`, how much breakage would result (how many people are passing things other than String and &str?)?

arqunis:
Sorry for the late reply.

> 1\. Is changing the url field of `Activity` to a url acceptable? `Url` is serializable and [implements] `AsRef<str>`, but do the consumers of [the] library care?

I personally consider this inconsequential. At most, the `Activity::url` field would be proven at a typesystem-level to be a valid URL, but I doubt the consumers would care about accuracy that much. Besides, Discord should send valid URLs anyway, so converting them into instances of `Url` would be pointless conversions. In the case consumers want to dissect the URL into its individual components, they can depend on the `url` crate and parse it into a `Url` themselves, and only when it's necessary.

> 2\. Is it on the caller or callee to see if its valid (i.e., are we taking a string-ish or a url)?

It should be on the caller to verify that they're sending a valid URL, but it wouldn't be wrong to perform such check ourselves. We already perform a lot of checks on behalf of the consumers (checking for permissions, for instance).

> 3\. `Activity::streaming` currently takes a `ToString`. However, I assume that mostly `String` and `&str` are being passed in. If we take some `T where Url : TryFrom<T>`, how much breakage would result (how many people are passing things other than `String` and `&str`)?

Practically speaking, I guess none. The `ToString` trait is used in order accept as many types as possible that can be converted into a `String`, which are types implementing `ToString` directly, or indirectly via `Display` (`ToString` has a blanket implementation for `T: Display`). But, regarding this function specifically, people will be just passing `&str`s or `String`s. If the bound was changed to use `Url`'s `TryFrom` implementation(s), I don't believe there would be any breakage that would prevent compilation of a consumer's code.

Still, as it is a public API change, we would have merge it into the `next` branch, otherwise we would be in violation of SemVer to release the change in a minor version.

hybras:
I did not mean to imply this *wouldn't* be a breaking change. I'm trying to figure out what this PR should actually cover. I think I'm clear now

Issue #296, opened 2018-03-21
Add an optional X-Audit-Log-Reason header to auditable actions


maddin77:

xSke:
I can tackle this. Been meaning to contribute, and it seems like low-hanging fruit.

My plan is to add a `Option<&str>` parameter to every method in `http/raw.rs` executing an auditable action. This will create major breaking changes to users of that module. The same goes for the `models` module. Alternatively, I can make two variants of every method, one named `x_with_reason`, the other staying the same and defaulting to no reason. Which variant would be preferred?

Erk-:
If you did the latter version it could be added to the current branch without breaking changes, then we can look at how we should structure it on the v0.6.x branch afterwards.

LeSeulArtichaut:
Because it has been left for long, I'd like to work on this.
I'll be adding `with_reason` methods like suggested by @xSke to be compatible with the current version of Serenity. However, to support UTF-8 reasons, I'd need to do percent-encoding. Is it ok to have an additional dependency or should I write my own implementation for this?

Erk-:
I would say that it is better to add a single dependency, https://crates.io/crates/percent-encoding, (which does not depend on any other crate) than to roll it ourselves.

LeSeulArtichaut:
I think I'll give up the idea of having two variants of the same method. There already are a lot of them, and I think having two methods for 30 routes or so, just in the raw Http struct, is a lot, but there are much more methods from all the model structs.

kangalio:
Does this issue still need to be open?

PR #1354, opened 2021-05-21
Add audit log reason support to Http

kangalio:
A second attempt at #296.

This PR only adds audit log reason support to the Http client. Model methods like `Member::unban` or `GuildChannel::edit` still cannot be provided with an audit log reason. This was done because:
- it's unclear how the API for audit log reasons in the model type methods should be designed as to not be a hassle to 90% of use cases where no audit log reason is provided
  - add a `Option<&str>` parameter to all the methods? That's a big breaking change and requires users to litter `None` function arguments all over the place
- adding support to the Http client is the most important thing to open the door for audit log reason support; model types are just extra convenience that can be added later

I tested this by making a [small test bot](https://pastebin.com/E6UiwF11) that invokes roughly half of all modified Http functions. The audit log in the server settings shows the correct reason strings. I couldn't test the other half of Http functions because I couldn't really figure out how to invoke them easily.

Unfortunately, Discord apparently doesn't properly document which HTTP routes support an audit log reason and which don't. I was using [this table](https://discord.com/developers/docs/resources/audit-log#audit-log-entry-object-audit-log-events) as a guideline, but probably there are some API routes where I falsely added or falsely omitted audit log support.

### DIFF
--- a/src/model/guild/member.rs
+++ b/src/model/guild/member.rs
@@ -83,7 +83,8 @@ impl Member {
             return Ok(());
         }
 
-        match http.as_ref().add_member_role(self.guild_id.0, self.user.id.0, role_id.0).await {
+        match http.as_ref().add_member_role(self.guild_id.0, self.user.id.0, role_id.0, None).await
+        {
             Ok(()) => {
                 self.roles.push(role_id);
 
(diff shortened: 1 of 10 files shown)


src/model/guild/member.rs
```diff
@@ -508,7 +513,7 @@ impl Member {
     /// [Ban Members]: Permissions::BAN_MEMBERS
     #[inline]
     pub async fn unban(&self, http: impl AsRef<Http>) -> Result<()> {
```
Zaitara, line 520: Why don't you add support to the reason message here (and in others structures above)? Is it in WIP?

Zaitara (changes requested):

src/model/guild/member.rs
kangalio: This PR only adds audit log support to the Http client for simplicity. I will expand the explanation for this in the PR description, hopefully that will clear it up

Zaitara (approved):

Zaitara:
Can you resolve conflicts? Otherwise lgtm

arqunis (approved):

PR #1355, opened 2021-05-22
Fix embed in interaction follow-up messages

TehPers:
Currently, interaction follow-up messages don't work at all. Discord's API for follow-up messages expects a field called `embeds` which can be "array of up to 10 embed objects", however the current version of the library sends a field called `embed` which is a single, optional embed object.

This PR changes the follow-up message builder to support multiple embeds, also fixing how the embeds are sent to Discord in the process. Additionally, `CreateInteractionResponseFollowup::set_embed` is being renamed to `add_embed` to reflect the change.

~~Since this is a breaking change, I'm targeting `next`. Let me know if I should change `add_embed` back to `set_embed` and target `current` instead to fix it on the current version of the library.~~ Targeting 'current' now since it's locked behind the `unstable_discord_api` feature anyway.

References:
 - [Create Followup Message](https://discord.com/developers/docs/interactions/slash-commands#create-followup-message)
 - [Execute Webhook](https://discord.com/developers/docs/resources/webhook#execute-webhook) ('Create Followup Message' follows this format)

### DIFF
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -84,8 +84,8 @@ impl CreateInteractionResponseData {
         self.set_embed(embed)
     }
 
-    /// Set an embed for the message.
-    pub fn set_embed(&mut self, embed: CreateEmbed) -> &mut Self {
+    /// Adds an embed to the message.
+    pub fn add_embed(&mut self, embed: CreateEmbed) -> &mut Self {
         let map = utils::hashmap_to_json_map(embed.0);
         let embed = Value::Object(map);
 
@@ -95,6 +95,20 @@ impl CreateInteractionResponseData {
             embeds.push(embed);
         }
 
+        self
+    }
+
+    /// Sets a list of embeds to include in the message.
+    ///
+    /// Calling this multiple times will overwrite the embed list.
+    /// To append embeds, call [`Self::add_embed`] instead.
+    pub fn embeds(&mut self, embeds: impl IntoIterator<Item = CreateEmbed>) -> &mut Self {
+        let embeds = embeds
+            .into_iter()
+            .map(|embed| utils::hashmap_to_json_map(embed.0).into())
+            .collect();
+        
+        self.0.insert("embeds", Value::Array(embeds));
         self
     }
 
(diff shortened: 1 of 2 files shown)


src/builder/create_interaction_response_followup.rs
```diff
@@ -90,15 +90,21 @@ impl<'a> CreateInteractionResponseFollowup<'a> {
...
         let mut embed = CreateEmbed::default();
         f(&mut embed);
-        self.set_embed(embed)
+        self.add_embed(embed)
     }
 
-    /// Set an embed for the message.
-    pub fn set_embed(&mut self, embed: CreateEmbed) -> &mut Self {
+    /// Adds an embed to the message.
+    pub fn add_embed(&mut self, embed: CreateEmbed) -> &mut Self {
```
Zaitara, line 98: You should also create a separate `set_embeds` method which would take an array of embeds.

Zaitara (changes requested):

Zaitara:
You can target `current` as it is under the `unstable_discord_api` feature.

src/builder/create_interaction_response_followup.rs
TehPers: Created an `embeds` method to go with the `files` method.

Zaitara:
Could you also add these methods to `CreateInteractionResponseData`?

TehPers:
> 
> 
> Could you also add these methods to `CreateInteractionResponseData`?

Should I also rename `CreateInteractionResponseData::set_embed` -> `add_embed` while I'm at it? `set_embed` seems like an odd name in the first place when it's adding a new embed without affecting the other ones.

src/builder/create_interaction_response.rs
```diff
@@ -81,11 +81,11 @@ impl CreateInteractionResponseData {
     {
```
Zaitara, line 81: Rename this method to `create_embed` (instead of `embed`)

Zaitara (changes requested):

src/builder/create_interaction_response.rs
TehPers: Should I rename it in `CreateInteractionResponseFollowup` as well? That way the names will match each other.

Zaitara: Yes you should

arqunis (approved):

PR #1360, opened 2021-05-24
Fix allowed_mentions use in Message::_reply

### LINKED ISSUE #1236: reply_ping doesn't allow mentioning additional users
reply_ping does not allow mentioning users (other than the user being replied to) in the message body.

`
let msg_content = format!("Pinging other users: {} {} {}", user1.mention(), user2.mention(), user3.mention());
message.reply_ping(&ctx, msg_content).await;
`  

Expected result: 'message' is replied to with an inline reply with the content 'Pinging other users: @user1 @user2 @user3'
Actual behavior: 'message' is replied to with an inline reply but the extra mentions are broken (does not notify users, and on mobile the mentions appear as '@invalid_user'). if one of the extra mentions is of the user being replied to, that mention will actually work.

the user who is being replied to will be properly mentioned, but the additional mentions in the message body are blocked because reply_ping builds the message by setting allowed_mentions to only the replied user. 
see https://docs.rs/serenity/0.10.2/src/serenity/model/channel/message.rs.html#624

reply_ping is at
https://docs.rs/serenity/0.10.2/src/serenity/model/channel/message.rs.html#579

serenity version 0.10.2
rustc version 1.49.0

kangalio:
Fixes #1236.

To control whether an inline reply pings the original message's author, the `Message::_reply` method (which `reply` and `reply_ping` are based on) uses the allowed_mentions method:

```rust
.allowed_mentions(|f| f.replied_user(ping_user));
```

However, as soon as you add the allowed_mentions attribute to a message, Discord disables _all_ pings by default and you have to re-enable them each manually to return to normal ping behavior. Therefore, this PR changes the above code snippet to:

```rust
.allowed_mentions(|f| {
    f.replied_user(ping_user)
        // By providing allowed_mentions, Discord disabled _all_ pings by
        // default so we need to re-enable them
        .parse(crate::builder::ParseValue::Everyone)
        .parse(crate::builder::ParseValue::Users)
        .parse(crate::builder::ParseValue::Roles)
});
```

I confirmed with a small test bot that `Message::reply_ping` now exhibits the expected behavior (i.e. pings within the message work).

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -726,11 +726,16 @@ impl Message {
         }
 
         self.channel_id
-            .send_message(cache_http.http(), |mut builder| {
+            .send_message(cache_http.http(), |builder| {
                 if let Some(ping_user) = inlined {
-                    builder = builder
-                        .reference_message(self)
-                        .allowed_mentions(|f| f.replied_user(ping_user));
+                    builder.reference_message(self).allowed_mentions(|f| {
+                        f.replied_user(ping_user)
+                            // By providing allowed_mentions, Discord disabled _all_ pings by
+                            // default so we need to re-enable them
+                            .parse(crate::builder::ParseValue::Everyone)
+                            .parse(crate::builder::ParseValue::Users)
+                            .parse(crate::builder::ParseValue::Roles)
+                    });
                 }
 
                 builder.content(content)


arqunis (approved):

whopiyush:
When will this fix be pushed to cargo? I'm using v0.10.7 which is the latest according to cargo, still facing this issue. For the time being, I've changed my `cargo.toml` to use serenity from the repository instead.

PR #1361, opened 2021-05-24
Change Presence::user field to handle optional fields

### LINKED ISSUE #1184: PresenceUpdateEvent::Presence::user is empty
Why is the PresenceUpdateEvent::Presence::user field almost always empty in the `presence_update` EventHandler function? The user_id is correct, but the user object is usually empty. It seems, to me, to be returning the presence object before the user has been populated fuly. How can I guarantee that it gets populated? Or do I need to do a separate user lookup call after getting the id?

kangalio:
Fixes #1184.

Quoting AcdenisSK from the issue:

> The deserialization of the user object is wrong. It either expects full data for a user object, or just the user id. It doesn't account for some fields missing, which it should.

This problem was fixed by introducing a new struct `PresenceUser` that uses `Option` for all optional fields. I used `CurrentUser` as a base for the `PresenceUser` because the `CurrentUser` struct seems to have all fields that the presence user field carries.

Discord documentation for the presence object: https://discord.com/developers/docs/topics/gateway#presence-update

I'm a bit unsure about the changes to cache, so there may be stupid mistakes in there :sweat_smile:

I tested the presence deserialization and it seems to work (https://pastebin.com/6QKK6kJ1). The cache changes I did not test because I don't know how to test changes to cache like that.

### DIFF
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -2,7 +2,6 @@
 
 use bitflags::bitflags;
 use serde::de::Error as DeError;
-use serde::ser::{Serialize, SerializeStruct, Serializer};
 
 use super::prelude::*;
 use super::utils::*;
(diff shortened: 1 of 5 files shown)


src/model/gateway.rs
```diff
@@ -549,107 +548,86 @@ pub struct ClientStatus {
...
-impl Serialize for Presence {
-    fn serialize<S>(&self, serializer: S) -> StdResult<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        #[derive(Serialize)]
-        struct UserId {
-            id: u64,
-        }
+    // TODO: add `to_current_user` and `into_current_user`? (converts to Option<CurrentUser>)
```
arqunis: While I don't find the significance of these (how often would they be used?), I also don't see harm in adding the methods. `PresenceUser` does structurally match `CurrentUser`, with the only difference between them being that the former has all but one optional fields. If Discord serves all fields of `PresenceUser`, it can be losslessly converted into `CurrentUser`.

arqunis (changes requested):
Looks good. I've added a comment addressing your TODO. Furthermore, would you mind fixing the formatting so that it passes the CI job?

src/model/gateway.rs
kangalio: If you don't see the significance of the methods, I'll just not add them. Less complexity that way :)

arqunis (approved):

PR #1365, opened 2021-05-26
Make Activity::streaming take a Url::TryFrom

hybras:
This is a continuation of #1349 . That PR was initially a breaking change, but Url supports TryFrom, so breakage was not needed. I have changed to bounds based on that.

### DIFF
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -1,8 +1,12 @@
 //! Models pertaining to the gateway.
 
+use std::convert::TryFrom;
+use std::fmt::Debug;
+
 use bitflags::bitflags;
 use serde::de::Error as DeError;
 use serde::ser::{Serialize, SerializeStruct, Serializer};
+use url::Url;
 
 use super::prelude::*;
 use super::utils::*;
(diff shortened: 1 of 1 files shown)


arqunis:
> so breakage was not needed

Unfortunately... changing bounds is a breaking change, especially if the range of inputs has been constricted from any type that implements `Display` (or `ToString`), to just `&str` because of the lone `TryFrom<&str>` implementation on `Url` right now. When I said that you can retarget your PR to the `current` branch, I meant that just for the internal changes to `Activity::streaming`, as that did not change the public API at all, but this does. You will have to retarget again to the `next` branch, or revert these changes back to the internal ones.

src/model/gateway.rs
```diff
@@ -150,10 +154,12 @@ impl Activity {
...
     /// }
     /// ```
-    pub fn streaming<N, U>(name: N, url: U) -> Activity
+    pub fn streaming<'a, N, U>(name: N, url: U) -> Activity
     where
         N: ToString,
-        U: ToString,
+        Url: TryFrom<U>,
+        <Url as TryFrom<U>>::Error: Debug,
+        U: 'a,
```
arqunis, line 162: I cannot comprehend the need for the lifetime here. We do not express a relationship between types or references, therefore this lifetime is a no-op and unnecessary.
```suggestion
```

arqunis (changes requested):

hybras:
Its clear that I'm not sure what I actually want... I'm just gonna close this

PR #1356, opened 2021-05-22
Clarify slash commands implementation

Zaitara:
This PR clarifies the slash commands and interaction implementation by improving the documentation and renames confusing `create(_global)_application_commands(_permissions)` methods to `set(_global)_application_commands(_permissions)`.

This has been tested.

PR #1368, opened 2021-05-28
Fix return type of closure in CreateApplicationCommandPermissionsData::create_permission

Seppel3210:
The closure returned `&mut Self` (i.e. `CreateApplicationCommandPermission*s*Data`) instead of `&mut CreateApplicationCommandPermissionData` which made it impossible to use in the intended way.

(P.S. I think the name "Create…" for builders instead of "…Builder" is a bit confusing but that's a different thing 🙂)


HarmoGlace:
It is already covered by #1356

PR #1373, opened 2021-06-02
Use Url in Activity streaming

hybras:
This PR is *another* redo of #1349 (wow am I bad at this). I've changed Activity.url to a Url. Url serializes to and from a string, so serialization is mostly unchanged. If u attempt to deserialize an invalid url, it will error. This behavior may not be desired.

I changed U from ToString to AsRef<str>. This covers String, &str, and Url. I'm so happy rust allowed me to say "I want all of these". This should minimize breakage. I chose AsRef<str> over ToString or a TryFrom bound because it seems Url will not be adding any other TryFrom implementations. This intent might be better reflected by using Url::parse instead of Url::try_from internally.

### DIFF
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -1,6 +1,7 @@
 //! Models pertaining to the gateway.
 
 use bitflags::bitflags;
+use reqwest::Url;
 use serde::de::Error as DeError;
 
 use super::prelude::*;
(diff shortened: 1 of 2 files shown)


arqunis (approved):
Seems all right, but would you mind fixing the formatting to be according to the project's formatting rules? Ensure you use the latest nightly version of the `rustfmt` tool.

src/model/gateway.rs
```diff
@@ -1,6 +1,7 @@
 //! Models pertaining to the gateway.
 
 use bitflags::bitflags;
+use reqwest::Url;
```
arqunis: This appears to be a fragile import. While `reqwest` conveniently re-exports the type for us, it's also likely for its `url` and our `url` to diverge. I am quite certain the reason this code works in its current state is due to cargo re-using the same `url` for both `reqwest` and us (that is, only the recent version of `url` is compiled and used across `reqwest` and Serenity). This should be changed to import from the `url` crate directly.
```suggestion
use url::Url;
```

arqunis (changes requested):

arqunis (approved):

PR #1377, opened 2021-06-02
Add the token to webhooks fetched via HTTP

JellyWX:
I'm not sure if this is a good solution to this problem.

# Issue this resolves:
The following flow doesn't work (previously it did):

- Fetch a webhook via `Http::get_webhook_with_token()`
- Execute the webhook via `Webhook::execute()`

The reason this had stopped working is that the webhook, when fetched, doesn't have a token attached to it. So attempting to execute the webhook results in it reporting a `Model(NoTokenSet)` error. I suspect a Discord API change sometime last night has disabled the sending of tokens in the body of the GET webhook route.

This solution in this PR is a bit jank imo

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2681,6 +2681,11 @@ impl Http {
             },
         })
         .await
+        .map(|mut w: Webhook| {
+            w.token = Some(token.to_string());
+
+            w
+        })
     }
 
     /// Kicks a member from a guild.


arqunis (approved):

Zaitara:
It was indeed changed a few hours ago, see https://github.com/discord/discord-api-docs/issues/3056

PR #1378, opened 2021-06-02
Add support for adding attachments in message edits

kangalio:
This PR makes it possible to add new attachments to a message when editing it.

This PR:
- creates a new method `Http::edit_message_and_attachments` (that method is to `edit_message` what `send_files` is to `send_message`)
  - there's a lot of duplication between that new method and `Http::edit_message_and_attachments` and `Http::send_files`, but I don't know enough about serenity's http design to improve the situation
- changes calls of `Http::edit_message` to `Http::edit_message_and_attachments` where appropriate
- creates a new method `EditMessage::attachment` to add a new attachment
- adds a new field to `EditMessage` which stores new attachments (breaking change) 
- adds a lifetime parameter to `EditMessage` (breaking change)
- moves the `AttachmentType` serialization into an HTTP form part into a function to avoid code duplication

Here are [the Discord docs](https://discord.com/developers/docs/resources/channel#edit-message) for the message edit request.

The code was tested with a [small test bot](https://pastebin.com/fHeffUu2).

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1329,7 +1329,8 @@ impl Http {
         map: &Value,
         new_attachments: impl IntoIterator<Item = AttachmentType<'_>>,
     ) -> Result<Message> {
-        // TODO: Avoid duplication with send_files()?
+        // Note: if you need to copy this code for a new method, extract this code into a function
+        // instead and call it in here and in send_files(), to avoid duplication (see Rule Of Three)
 
         let uri = api!("/channels/{}/messages/{}", channel_id, message_id);
         let mut url = Url::parse(&uri).map_err(|_| Error::Url(uri))?;
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1319,6 +1319,52 @@ impl Http {
         .await
     }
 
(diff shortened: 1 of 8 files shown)


src/http/client.rs
```diff
@@ -1319,6 +1319,52 @@ impl Http {
...
+    ///
+    /// **Note**: Only the author of a message can modify it.
+    pub async fn edit_message_and_attachments(
+        &self,
+        channel_id: u64,
+        message_id: u64,
+        map: &Value,
+        new_attachments: impl IntoIterator<Item = AttachmentType<'_>>,
+    ) -> Result<Message> {
+        // TODO: Avoid duplication with send_files()?
```
arqunis: The code of `send_files` and this method is remarkably similar, but I would personally leave this be for now. I would attempt to de-duplicate code once it's used three times or more, as far as I can tell, this is just the second time.

kangalio: Yeah, it seems that's a [common role of thumb](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) in programming

arqunis (approved):

PR #1381, opened 2021-06-07
Fix deserialization of Interaction's Option fields

sbrocket:
Interaction's Deserialize implementation currently always fails because it does not handle the present-but-null values generated by the derived Serialize implementation for Option fields that equal None.

### DIFF
--- a/src/model/interactions.rs
+++ b/src/model/interactions.rs
@@ -38,21 +38,27 @@ pub struct Interaction {
     ///
     /// [`Ping`]: self::InteractionType::Ping
     /// [`kind`]: Interaction::kind
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub data: Option<InteractionData>,
     /// The message this interaction was triggered by, if
     /// it is a component.
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub message: Option<InteractionMessage>,
     /// The guild Id this interaction was sent from, if there is one.
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub guild_id: Option<GuildId>,
     /// The channel Id this interaction was sent from, if there is one.
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub channel_id: Option<ChannelId>,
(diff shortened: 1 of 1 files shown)


sbrocket:
@HarmoGlace Did you have a question or comment about the PR? Not sure what the meaning of that reaction is.

sbrocket:
OK, I can't run the workflows myself to verify but this should be good to go on top of the merged components support now.

Zaitara:
Sorry for the late answer but the deserialization is correct as it is implemented according to Discord documentation. You should only fix the serialization by skipping some fields if `None` with `#[serde(skip_serializing_if = "Option::is_none")]`

sbrocket:
> Sorry for the late answer but the deserialization is correct as it is implemented according to Discord documentation. You should only fix the serialization by skipping some fields if `None` with `#[serde(skip_serializing_if = "Option::is_none")]`

That was the first approach I used, but that will fail to deserialize any structs serialized before this change, which is why I switched to this. It’s a minor difference, but at the time I’d had some already serialized data. I certainly can add the skip_serializing_if attributes back too, though.

Where do the Discord docs say that you have to check field presence alone and error if the value happens to be “null”, instead of treating the null the same as not-present? That would be a surprising level of detail for their docs to go into.

Zaitara:
The docs write `user?` and `user object` (same for others fields) in the second column, which means the field might not be present in the payload (it's optional) but won't be nullable, it's like that on all their docs.

The best is to skip the serializing, I understand that previously serialized interaction won't work but it is a serialization issue, which needs to be fixed.

sbrocket:
Ah, yep, you’re totally correct; they do distinguish between optional and nullable fields. Interesting. https://discord.com/developers/docs/reference#nullable-and-optional-resource-fields

I’ll update this shortly.

sbrocket:
Whoops, forgot to go back and push this. There you go.

Zaitara (approved):

arqunis (approved):

Issue #1269, opened 2021-03-26
`Http::send_files` does not use the ratelimiter


drklee3:
`Http::send_files` makes a POST request using the reqwest client directly instead of using a `RatelimitedRequest` like other API requests use, bypassing the ratelimiter.

https://github.com/serenity-rs/serenity/blob/7e088780515dbd1210b2245b52bab297962084e8/src/http/client.rs#L2297-L2383

PR #1395, opened 2021-06-17
Add support for sticker packs, guild stickers, sticker routes

drklee3:
### Description

Adds the following features:

* Add/modify/delete guild stickers
* Fetch sticker information
* Fetch list of all nitro stickers
* Sending stickers in message (up to 3)

Breaking change as sticker module is moved out of the channel module and to the root model module.

This also adds an additional `mime` and `mime_guess` dependencies as Discord requires the mime type on the file part when creating a sticker via multipart/form-data requests otherwise it will return a 500 Internal Server Error. This dependency is also used by `reqwest` internally.

#### Relevant documentation:

* https://github.com/discord/discord-api-docs/pull/3127
* https://github.com/discord/discord-api-docs/pull/3128
* https://github.com/discord/discord-api-docs/pull/3171

### Tested

- Sending / receiving stickers in messages
- Get / create / edit / delete sticker


src/http/client.rs
```diff
@@ -688,6 +688,81 @@ impl Http {
...
+                let response = self.client.get(url).send().await?;
+                let mut bytes = response.bytes().await?;
+                let mut picture: Vec<u8> = vec![0; bytes.len()];
+                bytes.copy_to_slice(&mut picture[..]);
+                multipart =
+                    multipart.part("file", Part::bytes(picture).file_name(filename.to_string()));
+            },
+        }
+
+        multipart = multipart.text("payload_json", serde_json::to_string(&map)?);
```
Zaitara: This code seems to be repeated a lot in the library

drklee3: True. I was thinking if I can get #1269 fixed and removing the duplicate code with that fix since it's related, so it will probably be in a separate PR.

src/model/sticker.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// The unique ID of the pack the sticker is from.
+    pub pack_id: Option<StickerPackId>,
+    /// The name of the sticker.
+    pub name: String,
+    /// Description of the sticker
+    pub description: String,
+    /// For guild stickers, a unicode emoji representing the sticker's
+    /// expression. For nitro stickers, a comma-separated list of related
+    /// expressions.
+    pub tags: String,
```
Zaitara: You should make it an array to keep consistency and to make it easier to handle -which would mean that the array would be of length 1 if it is a guild owned one-

Zaitara (changes requested):

src/model/utils.rs
```diff
@@ -290,6 +290,48 @@ pub fn serialize_roles<S: Serializer>(
...
+    let vec = str_sequence.split(',').map(|item| item.trim().to_owned()).collect();
+
+    Ok(vec)
+}
+
+pub fn serialize_comma_separated_string<S: Serializer>(
+    vec: &Vec<String>,
+    serializer: S,
+) -> StdResult<S::Ok, S::Error> {
+    serializer.serialize_str(&vec.join(" "))
```
Zaitara:
```suggestion
    serializer.serialize_str(&vec.join(","))
```
To be consistent with the deserialization

Zaitara (changes requested):

src/model/utils.rs
drklee3: Oh yeah oops, thanks for the catch.

arqunis:
This has been manually merged: https://github.com/serenity-rs/serenity/commit/744bc1321a2c312b50da4c7ab1592944c6451168.

PR #1396, opened 2021-06-19
Add support for guild search members endpoint

drklee3:
### Description

Adds support for searching guild members whose username or nickname starts with a provided query.

Documentation: https://discord.com/developers/docs/resources/guild#search-guild-members

### Tested

`Guild::search_members()` correctly returns a vec of members, and an empty vec when none are found.

### DIFF
--- a/src/http/routing.rs
+++ b/src/http/routing.rs
@@ -597,6 +597,7 @@ impl Route {
     pub fn guild_members_search(guild_id: u64, query: &str, limit: Option<u64>) -> String {
         let mut s = format!(api!("/guilds/{}/members/search?"), guild_id);
 
+        #[allow(clippy::let_underscore_must_use)]
         let _ = write!(s, "&query={}", query);
 
         #[allow(clippy::let_underscore_must_use)]
--- a/src/http/routing.rs
+++ b/src/http/routing.rs
@@ -429,6 +429,7 @@ impl Route {
         api!("/channels/{}/messages/{}/reactions", channel_id, message_id)
     }
 
+    #[allow(clippy::let_underscore_must_use)]
     pub fn channel_message_reactions_list(
         channel_id: u64,
         message_id: u64,
@@ -442,7 +443,6 @@ impl Route {
         );
 
         if let Some(after) = after {
-            #[allow(clippy::let_underscore_must_use)]
             let _ = write!(uri, "&after={}", after);
         }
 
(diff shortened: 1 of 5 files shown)


src/model/guild/mod.rs
```diff
@@ -2030,6 +2050,29 @@ impl Guild {
...
+    ///
+    /// **Note**: Queries are case insensitive.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`Error::Http`] if the API returns an error.
+    ///
+    /// [`Error::Http`]: crate::error::Error::Http
+    #[inline]
+    pub async fn search_members(
```
Zaitara, line 2067: You should also add this method to `PartialGuild`

src/http/routing.rs
```diff
@@ -588,6 +594,18 @@ impl Route {
         format!(api!("/guilds/{}/members"), guild_id)
     }
 
+    pub fn guild_members_search(guild_id: u64, query: &str, limit: Option<u64>) -> String {
+        let mut s = format!(api!("/guilds/{}/members/search?"), guild_id);
+
+        #[allow(clippy::let_underscore_must_use)]
```
Zaitara: You can add it to the top of the function

src/http/routing.rs
```diff
@@ -588,6 +594,18 @@ impl Route {
...
+        let _ = write!(s, "&query={}", query);
+
+        #[allow(clippy::let_underscore_must_use)]
```
Zaitara: Ditto

Zaitara (changes requested):

src/model/guild/guild_id.rs
```diff
@@ -953,6 +953,27 @@ impl GuildId {
...
+    /// Optionally pass in the `limit` to limit the number of results.
+    /// Minimum value is 1, maximum and default value is 1000.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`Error::Http`] if the API returns an error.
+    ///
+    /// [`Error::Http`]: crate::error::Error::Http
+    #[inline]
+    pub async fn search_members(
```
kangalio, line 968: Would it be possible to take a CacheHttp here and search the cache first? And in Guild's and PartialGuild's version of this method too

Zaitara: The purpose of this method is to use the API feature, which will search by itself, especially for larger guilds. In my opinion it doesn't make sense to use the cache here and would make the mehod behavior inconsistent.

drklee3: Yeah I was thinking about that but another +1 to it might produce inconsistent/unexpected results.  Since if there are a few members cached that match, we still need to do an API request anyways for those additional non-cached that still can match.

It could work if it checked if all members are already cached (guild size == cached members size) to only use cache, but then at that point might as well use the other methods to be more explicit on intended behavior.

kangalio: I see. I thought the cache either contains either all or no members. If it can contain just a few users, that indeed seems too brittle to be worth it

arqunis (approved):

PR #1397, opened 2021-06-20
Clarify interaction docs

drklee3:
Minor changes to adjust slash command specific wording to include buttons and other interactions.  Prevents confusion that certain things are *only* for slash commands when they actually include other interactions.

### DIFF
--- a/src/client/event_handler.rs
+++ b/src/client/event_handler.rs
@@ -422,7 +422,8 @@ pub trait EventHandler: Send + Sync {
     ) {
     }
 
-    /// Dispatched when a user used a slash command.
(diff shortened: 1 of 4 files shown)


Zaitara:
It is already covered by #1356

drklee3:
> It is already covered by #1356

Ah missed that one. Seems like there are a few cases where it isn't updated like the event/event_handler which was the main part for me.  Will you update it in the linked PR?

Zaitara:
Yes I'll update it and fix them in my PR if it doesn't bother you

drklee3:
Sounds good, thanks 👍

Issue #1375, opened 2021-06-02
get_guilds should have optional pagination


tqwewe:
The get_guilds function has the following signature:
```rust
async fn get_guilds(&self, target: &GuildPagination, limit: u64) -> Result<Vec<GuildInfo>>
```
And according to the [Discord docs for guilds](https://discord.com/developers/docs/resources/user#get-current-user-guilds), the `before` and `after` fields are optional, but the Serenity Rust API forces me to pass one of them, but not none.

The API should be updated to:
```rust
async fn get_guilds(&self, target: Option<&GuildPagination>, limit: u64) -> Result<Vec<GuildInfo>>
```
Allowing an `before` and `after` to be optional.

tylerd008:
Mind if I take this? Looks pretty straightforward and shouldn't take much time at all to do.

drklee3:
Implemented in #1403

PR #1402, opened 2021-06-23
Implement Optional Pagination for get_guilds

tylerd008:
Fix for issue #1375.


Zaitara:
You need to target the `next` branch as it is a breaking change

src/model/user.rs
```diff
@@ -192,7 +192,9 @@ impl CurrentUser {
             let mut pagination = http
                 .as_ref()
                 .get_guilds(
-                    &GuildPagination::After(guilds.last().map_or(GuildId(1), |g: &GuildInfo| g.id)),
+                    Some(&GuildPagination::After(
```
Dinnerbone: I think this should be changed to `None`, as it's effectively trying to workaround this very issue.

tylerd008:
Oh whoops sorry, changed.

drklee3:
Maybe you should `cherry-pick` your commits onto a new branch based on `next` so you don't get all the other commits from `current` added in.

src/model/user.rs
tylerd008: Updated

tylerd008:
> Maybe you should `cherry-pick` your commits onto a new branch based on `next` so you don't get all the other commits from `current` added in.

Oh yeah sorry let me do that real quick

PR #1403, opened 2021-06-23
Implement optional pagination for get_guilds

tylerd008:
On the right branch this time

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2475,7 +2475,7 @@ impl Http {
     ///
     /// let guild_id = GuildId(81384788765712384);
     ///
-    /// let guilds = http.get_guilds(&GuildPagination::After(guild_id), 10).await?;
+    /// let guilds = http.get_guilds(Some(&GuildPagination::After(guild_id)), 10).await?;
     /// #     Ok(())
     /// # }
     /// ```
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2475,7 +2475,7 @@ impl Http {
     ///
     /// let guild_id = GuildId(81384788765712384);
     ///
-    /// let guilds = http.get_guilds(Some(&GuildPagination::After(guild_id)), 10).await?;
+    /// let guilds = http.get_guilds(Some(&GuildPagination::After(guild_id)), Some(10)).await?;
     /// #     Ok(())
     /// # }
     /// ```
@@ -2484,7 +2484,7 @@ impl Http {
     pub async fn get_guilds(
         &self,
         target: Option<&GuildPagination>,
-        limit: u64,
+        limit: Option<u64>,
     ) -> Result<Vec<GuildInfo>> {
         let (after, before) = match target {
             None => (None, None),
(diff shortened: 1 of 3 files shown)


src/http/client.rs
```diff
@@ -2475,16 +2475,23 @@ impl Http {
...
     ///
     /// [docs]: https://discord.com/developers/docs/resources/user#get-current-user-guilds
-    pub async fn get_guilds(&self, target: &GuildPagination, limit: u64) -> Result<Vec<GuildInfo>> {
-        let (after, before) = match *target {
-            GuildPagination::After(id) => (Some(id.0), None),
-            GuildPagination::Before(id) => (None, Some(id.0)),
+    pub async fn get_guilds(
+        &self,
+        target: Option<&GuildPagination>,
+        limit: u64,
```
Zaitara: The limit is also optional, according to [discord documentation](https://discord.com/developers/docs/resources/user#get-current-user-guilds)

Zaitara (changes requested):

src/http/client.rs
tylerd008: Updated

src/model/user.rs
```diff
@@ -190,13 +190,7 @@ impl CurrentUser {
...
-                .as_ref()
-                .get_guilds(
-                    &GuildPagination::After(guilds.last().map_or(GuildId(1), |g: &GuildInfo| g.id)),
-                    100,
-                )
-                .await?;
+            let mut pagination = http.as_ref().get_guilds(None, None).await?;
             let len = pagination.len();
             guilds.append(&mut pagination);
             if len != 100 {
```
Dinnerbone: This will break as None defaults to 200, vs the old hardcoded 100. If it gets 200 in response, it will break and not ask for the next page.

tylerd008: Oh yeah sorry I see now, should've paid more attention to what I was doing. I'll change it back to `Some(100)` real quick.

src/model/user.rs
```diff
@@ -190,13 +190,7 @@ impl CurrentUser {
...
         let mut guilds = Vec::new();
         loop {
-            let mut pagination = http
-                .as_ref()
-                .get_guilds(
-                    &GuildPagination::After(guilds.last().map_or(GuildId(1), |g: &GuildInfo| g.id)),
-                    100,
-                )
-                .await?;
+            let mut pagination = http.as_ref().get_guilds(None, Some(100)).await?;
```
arqunis: This has to be reverted back to the original code. The current version will only fetch at most 100 guilds and skip the rest if the bot is in more than 100 guilds (it may also enter an infinite loop if `Http::get_guilds` returns exactly 100 guilds).
```suggestion
            let mut pagination = http
                .as_ref()
                .get_guilds(
                    Some(&GuildPagination::After(guilds.last().map_or(GuildId(1), |g: &GuildInfo| g.id))),
                    Some(100),
                )
                .await?;
```

arqunis (changes requested):

src/model/user.rs
tylerd008: Oh I see, thanks. I'll change it back

arqunis (approved):

PR #1407, opened 2021-06-26
Improvements to embed methods on EditMessage

kangalio:
- Make embed methods not depend on each other
- Add add_embeds method
- Change set_embeds to actually _set_ embeds, not _add_ them
- Expand embed method documentation to explain differences to each other

I tested the embed(), set_embed(), and set_embeds() methods in a personal bot

### DIFF
--- a/src/builder/edit_message.rs
+++ b/src/builder/edit_message.rs
@@ -45,16 +45,6 @@ impl EditMessage {
         self
     }
 
-    /// Add an embed for the message.
-    pub fn add_embed<F>(&mut self, f: F) -> &mut Self
-    where
-        F: FnOnce(&mut CreateEmbed) -> &mut CreateEmbed,
-    {
-        let mut embed = CreateEmbed::default();
-        f(&mut embed);
-        self._add_embed(embed)
-    }
-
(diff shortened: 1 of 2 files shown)


src/builder/edit_message.rs
```diff
@@ -55,45 +70,51 @@ impl EditMessage {
...
 
     /// Set multiple embeds for the message.
+    ///
+    /// **Note**: This will replace all existing embeds. Use [`Self::add_embeds()`] to keep existing
+    /// embeds.
     pub fn set_embeds(&mut self, embeds: Vec<CreateEmbed>) -> &mut Self {
+        self.0.insert("embeds", Value::Array(Vec::new()));
         for embed in embeds {
             self._add_embed(embed);
         }
```
drklee3, line 120: Could be replaced with `self.add_embeds(embeds)`?  Though I'm not sure if it's something we should try to avoid since you mentioned "Make embed methods not depend on each other"

drklee3:
Should the same changes also be made to the CreateMessage builder?

kangalio:
> Should the same changes also be made to the CreateMessage builder?

Probably, yeah. I will add it to this PR

arqunis (approved):

PR #1408, opened 2021-06-27
Use ratelimiter when sending files in messages and webhooks

### LINKED ISSUE #1120: Rate Limits are not entirely followed.
The "10,000 requests per 10 minutes" http rate limit doesn't seem to be followed.

`thread 'main' panicked at 'Could not access application info: Http(UnsuccessfulRequest(ErrorResponse { status_code: 429, url: Url { scheme: "https", host: Some(Domain("discord.com")), port: None, path: "/api/v8/oauth2/applications/@me", query: None, fragment: None }, error: "You are being blocked from accessing our API temporarily due to exceeding our rate limits frequently. Please read our docs at https://discord.com/developers/docs/topics/rate-limits to prevent this moving forward." }))', src/main.rs:975:21`

My ip just got temporarily banned from using the HTTP api because of this. Though i actually have no idea what in my bot can even cause such rate limits
This appears to be an issue with reconnecting, as this temporal ban happened as the dns server i use died, and when it came back, it was unable to connect.

(btw i can read general chat, but i can't type on it :P)

### LINKED ISSUE #1121: Webhooks don't follow rate limit
The cause of the issue that happened on #1120 has been determinated to be webhooks not following the rate limits.
Both issues should still be fixed though.

### LINKED ISSUE #1269: `Http::send_files` does not use the ratelimiter

drklee3:
### Description

Uses the ratelimiter for sending multipart requests and removes duplicated code.

Minor breaking change as it changes `Request::build` to an async fn.

### Related Issues

fixes #1269

### DIFF
--- a/src/http/multipart.rs
+++ b/src/http/multipart.rs
@@ -18,7 +18,7 @@ pub struct Multipart<'a> {
 }
 
 impl<'a> Multipart<'a> {
-    pub async fn to_multipart_form(&mut self, client: &Client) -> Result<Form> {
+    pub(crate) async fn to_multipart_form(&mut self, client: &Client) -> Result<Form> {
         let mut multipart = Form::new();
 
         for (file_num, file) in self.files.iter_mut().enumerate() {
--- a/src/http/multipart.rs
+++ b/src/http/multipart.rs
@@ -13,8 +13,8 @@ use crate::internal::prelude::*;
 /// messages with attachments.
 #[derive(Clone, Debug)]
 pub struct Multipart<'a> {
-    pub(super) files: Vec<AttachmentType<'a>>,
-    pub(super) payload_json: Value,
+    pub files: Vec<AttachmentType<'a>>,
+    pub payload_json: Value,
 }
 
 impl<'a> Multipart<'a> {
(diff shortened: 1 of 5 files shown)


src/http/multipart.rs
```diff
@@ -0,0 +1,98 @@
...
+                    let mut f = File::open(path).await?;
+                    let mut buf = vec![];
+                    f.read_to_end(&mut buf).await?;
+
+                    // Modify current AttachmentType as saved Bytes as to prevent
+                    // the need for another disk read when retrying
+                    *file = AttachmentType::Bytes {
+                        data: buf.clone().into(),
+                        filename: filename.clone().unwrap_or_else(String::new),
+                    };
```
drklee3, line 57: Not entirely sure if it is worth making an extra copy of attachment bytes every time an attachment is sent in order to not read from disk / make a network request in the case the message send fails or is rate limited.

arqunis (approved):

PR #1410, opened 2021-06-27
Replace unnecessary guild cloning with cheaper alternatives

### LINKED ISSUE #1409: Unnecessary Guild cloning in several functions.
I noticed that there were large CPU spikes whenever a command was called from my bot in a large guild, especially in the help command. This is due to unnecessary cloning of guild objects, which can be quite large if there are large amounts of users associated with it. These guilds were being cloned when the data necessary could've otherwise been gotten by using the cache's `guild_field()` method. 

This can be seen in 4 places.

1. The `should_fail()` function in the standard framework's module file where the only thing necessary is the owner's ID, but the entire guild is cloned to get it. This function is called on every command call.
2. The `has_correct_permissions()` function in the same file. The entire guild is cloned just to get whether or not a user has sufficient permissions for a command. This is used in several places for the help command.
3. The `permissions()` method within `Member`. The entire guild for the member is cloned to call `member_permissions()` method within `Guild`. This is unnecessary because a separate pub(crate) function can be made within Guild to take a `Member` directly, meaning it doesn't have to be async, allowing its usage easily within the `permissions()` method using `guild_field()`. The `permissions()` function is a public function that can be used outside of the crate.
4. The `check_common_behaviour()` method within the `help_command` module within the standard framework. Here, a `HelpBehaviour` is returned and unnecessarily clones the entire guild to do it. It is called many times when the help command is called.

MelonShooter:
Fixes #1409

### DIFF
--- a/src/framework/standard/mod.rs
+++ b/src/framework/standard/mod.rs
@@ -35,7 +35,6 @@ use crate::cache::Cache;
 use crate::client::Context;
 #[cfg(feature = "cache")]
 use crate::model::channel::Channel;
-use crate::model::guild::Guild;
 #[cfg(feature = "cache")]
 use crate::model::guild::Member;
 use crate::model::{channel::Message, permissions::Permissions};
@@ -261,9 +260,7 @@ impl StandardFramework {
                     return Some(DispatchError::BlockedGuild);
                 }
 
-                let cache = ctx.cache.clone();
-                let guild_selector = |guild: &Guild| guild.owner_id;
-                let owner_id_option = cache.guild_field(guild_id, guild_selector).await;
+                let owner_id_option = ctx.cache.guild_field(guild_id, |guild| guild.owner_id).await;
 
                 if let Some(owner_id) = owner_id_option {
                     if self.config.blocked_users.contains(&owner_id) {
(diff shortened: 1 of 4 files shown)


src/framework/standard/mod.rs
```diff
@@ -835,32 +840,39 @@ pub(crate) async fn has_correct_permissions(
...
+            if channel_option.is_none() {
                 return false;
-            },
+            }
+
+            let member_option = guild.members.get(&message.author.id);
+
+            if member_option.is_none() {
+                return false;
+            }
```
arqunis: This can be condensed into one `match` expression just as it was before, removing the need to unwrap the `Option` later with `expect`.
```suggestion
            let member = match guild.members.get(&message.author.id) {
                Some(member) => member,
                None => return false,
            };

```

src/framework/standard/mod.rs
```diff
@@ -835,32 +840,39 @@ pub(crate) async fn has_correct_permissions(
...
-            Some(channel) => channel,
-            None => return false,
-        };
-        let member = match guild.members.get(&message.author.id) {
-            Some(member) => member,
-            None => return false,
-        };
+    } else {
+        let has_perms_accessor = |guild: &Guild| {
+            let channel_option = guild.channels.get(&message.channel_id);
```
arqunis: Ditto.

src/framework/standard/mod.rs
```diff
@@ -260,8 +261,12 @@ impl StandardFramework {
                     return Some(DispatchError::BlockedGuild);
                 }
 
-                if let Some(guild) = guild_id.to_guild_cached(&ctx.cache).await {
-                    if self.config.blocked_users.contains(&guild.owner_id) {
+                let cache = ctx.cache.clone();
+                let guild_selector = |guild: &Guild| guild.owner_id;
+                let owner_id_option = cache.guild_field(guild_id, guild_selector).await;
```
arqunis: I don't see the usefulness of placing the closure into its own variable here, it can just be inline with the `guild_field` call. The type of the parameter can also be omitted; the compiler will infer it. Furthermore, I also don't understand why the cache has to be cloned here. It is wrapped in an `Arc` for sure, which is cheap to clone, but nonetheless, it could be nice to avoid that.
```suggestion
                let owner_id_option = ctx.cache.guild_field(guild_id, |guild| guild.owner_id).await;
```

arqunis (changes requested):

src/framework/standard/mod.rs
LuckShiba: What about `member_option.unwrap_or(false)`?

src/framework/standard/mod.rs
MelonShooter: I have no idea why I cloned the `Arc` either. I'll change that. As for the separating of the closures, I'll inline them for this one and also all the other cases I did it. Generally, I like to separate closures from what they're being used for when they're really long because I think it's slightly cleaner. In this case with the owner ID, I have no idea why I did it though either :/

src/framework/standard/mod.rs
MelonShooter: Wow, I'm dumb. I'll change this.

arqunis (approved):

PR #1412, opened 2021-06-28
Change return signatures of help commands to return `Result`s

arqunis:
## Description

This changes the return type of the default help commands from `Option` to `Result` to allow the caller to decide how they want to present the error that occurred when sending the response.

## Type of Change

This is a breaking change in the framework, specifically its help system. The purpose of this change is to enhance the experience.

## How Has This Been Tested?

This has been tested by verifying that compilation succeeds under the new changes, in both the library and its tests.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -1291,7 +1291,12 @@ async fn send_error_embed(
 ///     .help(&MY_HELP);
 /// ```
 ///
+/// # Errors
+///
+/// Returns the same errors as [`ChannelId::send_message`].
+///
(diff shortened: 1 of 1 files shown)


src/framework/standard/help_commands.rs
```diff
@@ -1310,7 +1301,7 @@ pub async fn with_embeds(
     help_options: &HelpOptions,
     groups: &[&'static CommandGroup],
     owners: HashSet<UserId>,
-) -> Option<Message> {
+) -> Result<Message, Error> {
```
arqunis, line 1309: This should probably use a better error type to include the formatted help, just as the macro did before, but I'm not sure whether I should do that.

PR #1416, opened 2021-06-30
Configure project to be always compiled with native CPU support

### LINKED ISSUE #1415: `cargo test --all-features` fails on branch `next`
`cargo test --all-features` runs successfully on `current`, but fails to compile `simd-json` on `next`.

**Details**

The error is 
```sh
213 | fn please_compile_with_a_simd_compatible_cpu_setting_read_the_simdjonsrs_readme() -> ! {}
    |    ----------------------------------------------------------------------------      ^ expected `!`, found `()`

```

I tried looking for `simdjonsrs_readme`, but quickly noticed that they must mean `simdjsonrs`. [The readme](https://github.com/simd-lite/simd-json/blob/main/README.md) says that the error will go away if I specify a cpu target.

In `next`'s [`ci.yml`](https://github.com/serenity-rs/serenity/blob/next/.github/workflows/ci.yml) the cpu target is specified with
```
            - name: Test all features
              if: matrix.features == ''
              env:
                  RUSTFLAGS: -C target-feature=+sse4.2
              run: cargo test --all-features
```

On my ubuntu 20.04 I can specify the flags with
```sh
RUSTFLAGS="-C target-cpu=native" cargo test --all-features
```

With this change, the target cpu was specified and the tests ran perfectly fine. Please note that the above line will not work on Windows. They have their own ways of setting temporary `env` variables.

**What is the issue?**

When working on `next`, [the contributing guidelines](https://github.com/serenity-rs/serenity/blob/next/CONTRIBUTING.md) recommend running `cargo test --all-features`. Doing so will result in an error.

**Possible solutions**

- Add `RUSTFLAGS="-C target-cpu=<CPU-HERE>" cargo test --all-features` and the Windows equivalent to `CONTRIBUITNG.md`
- Remove `simd-json` tests (seems like a bad idea)
- Add a cargo config file as detailed in [simd-json README](https://github.com/simd-lite/simd-json/blob/main/README.md)

**Additional information**
```
Serenity version:
next

Rust version (`rustc -V`):
rustc 1.55.0-nightly (e8cb1a4a5 2021-06-27)

Minimal test case if possible:
git checkout next
cargo test --all-features
```

arqunis:
## Description

This adds a Cargo configuration file that defines flags for every invocation of `rustc` on Serenity and its dependencies. This removes the need to explicitly specify those flags when building the library, especially when testing `simd-json` code.

Fixes #1415.

## Type of Change

This is an improvement to the development of the library to reduce friction when testing all features, and in particular, `simdjson`.

## How Has This Been Tested?

This has been tested by running a lone `cargo build --all-features` (and `cargo test --all-features`) and ensuring that no errors arise, which they do not.


evieluvsrainbows (approved):
Looks good to me! (:

PR #1417, opened 2021-07-01
Separate interaction stuctures per kind

Zaitara:
This PR refactors the interactions structures by creating a structure per kind of interaction. It enforces the typings and make easier to handle it.

This has been tested.


src/model/interactions/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+    pub fn application_id(&self) -> ApplicationId {
+        match self {
+            Interaction::Ping(i) => i.application_id,
+            Interaction::ApplicationCommand(i) => i.application_id,
+            Interaction::MessageComponent(i) => i.application_id,
+        }
+    }
+
+    /// Gets the interaction token.
+    pub fn token(&self) -> &String {
```
arqunis: Is it useful to return a reference to a `String`? The only scenario this would be convenient is if the user is interested in the capacity of the allocated buffer, but I doubt there is someone that needs to know that. A simpler alternative would be to return `&str`:
```suggestion
    pub fn token(&self) -> &str {
```

arqunis (changes requested):

arqunis (approved):

src/model/interactions/message_component.rs
```diff
@@ -0,0 +1,689 @@
...
+    /// The custom id of the component.
+    pub custom_id: String,
+    /// The type of the component.
+    pub component_type: ComponentType,
+    /// The given values of the [`SelectMenu`]s
+    #[serde(default)]
+    pub values: Vec<String>,
+}
+
+// A component.
```
nickelc, line 385: @HarmoGlace, is there any reason why you kept the `Component` enum?
It seems to be no longer used beside the `From`/`TryFrom` implementations after the `Message::components` type  was changed to `Vec<ActionRow>`.

Zaitara: Indeed, it is no longer used in the library itself, but as the `ComponentType` structure still exists, I thought that it might be consistent to also add a wrapper to any component. Feel free to make a PR to remove it, it wouldn't bother me

nickelc: What is the use case you had in mind?

Zaitara: I implemented it for consistency : an enum for interaction exists, so an enum for each component might be interesting. It might be considered as a "util" structure.

PR #1419, opened 2021-07-02
Lazily clone structs only when collector filters are actually matched

sbrocket:
Adds a new internal LazyArc type that is used to delay cloning Messages,
Reactions, and Interactions until it's determined whether any filters
are actually matched or not


sbrocket:
By the way, the primary motivation behind this is preparation for my coming EventCollector PR. This will allow creating a generic EventCollector without cloning every single Event which an EventCollector/EventFilter exists.

arqunis (approved):

PR #1421, opened 2021-07-03
Add Cache::channel_messages_field

kangalio:
Sparked by [a discussion](https://discord.com/channels/381880193251409931/673965002805477386/860914066183553035) in the serenity Discord server, this PR adds a new method to Cache:

```rust
pub async fn channel_messages_field<T>(
    &self,
    channel_id: impl Into<ChannelId>,
    selector: impl FnOnce(MessageIterator<'_>) -> T,
) -> Option<T> 
```

The new method makes it possible to access cached messages for a channel, for example to collect messages by a specific user:
```rust
let messages_by_user = cache.channel_messages_field(7, |msgs| {
    msgs.filter(|m| m.author.id == 8).cloned().collect::<Vec<_>>()
});
```

The Iterator type passed to the closure is `MessageIterator`, a new type which wraps `std::collections::hash_map::Values`. Passing `MessageIterator` instead of the concrete `std::collections::hash_map::Values` makes it possible to swap out the actual iterator type in the future when cache implementation details change.

I tested the new method with a [small test bot](https://pastebin.com/VdmSpVuL).

### DIFF
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -352,7 +352,7 @@ impl Cache {
 
     /// This method allows to extract specific data from the cached messages of a channel by
     /// providing a `selector` closure picking what you want to extract from the messages
-    /// iterator of a given channell.
+    /// iterator of a given channel.
     ///
     /// ```rust,no_run
     /// # let cache: serenity::cache::Cache = todo!();
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -86,6 +86,23 @@ impl<F: FromStr> FromStrAndCache for F {
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/cache/mod.rs
```diff
@@ -333,6 +350,28 @@ impl Cache {
         None
     }
 
+    /// This method allows to extract specific data from the cached messages of a channel by
+    /// providing a `selector` closure picking what you want to extract from the messages
+    /// iterator of a given channell.
```
arqunis: Typo.
```suggestion
    /// iterator of a given channel.
```

arqunis (changes requested):

arqunis (approved):

PR #1422, opened 2021-07-03
Move workspace up to repo root, instead of under examples

sbrocket:
This allows rust-analyzer and other similar tools that operate based on
the Cargo package layout to work with the example source without having
to treat the examples directory as a separate root package.

The examples workspace was originally created so that common
dependencies were not rebuilt for each individual example, and this
property is maintained with this change.


sbrocket:
Ok, this should at least pass CI. I'm putting up a separate PR with all the automatic fixes clippy was able to make, but re-adding `--workspace` and getting clippy passing with all features is going to take some manual work. Hopefully unblocking this PR in this way is fine with you.

arqunis (approved):

PR #1425, opened 2021-07-04
Add guild commands into ApplicationCommand

SparkyPotato:
This allows guild-specific commands to be created, edited, and destroyed the same way global commands are.

### DIFF
--- a/src/model/interactions.rs
+++ b/src/model/interactions.rs
@@ -766,7 +766,7 @@ impl ApplicationCommand {
     /// list. If an outdated command data is sent by a user, discord will consider it as an error
     /// and then will instantly update that command.
     ///
-    /// As such, it is recommended that guild application commands be used for testing purposes.
+    /// As such, it is recommended that [`guild application commands`] be used for testing purposes.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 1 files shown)


Zaitara:
This is already supported, see https://serenity-rs.github.io/serenity/current/serenity/model/id/struct.GuildId.html#method.create_application_command

SparkyPotato:
That seems rather counterintuitive, with global commands being separated from guild commands and in a separate module. 

The global command docs don't even link to the guild ones.

On another note, the interactions example should include creating commands, not just responding to interactions.

Zaitara:
The example does include it, see https://github.com/serenity-rs/serenity/blob/695bbef18e06696b9b9d1f46225099e94001b6c8/examples/e14_slash_commands/src/main.rs#L74-L132
I understand but it is the same things for others structures ; a guild channels will be linked to a `GuildId`

SparkyPotato:
Oops, that was added after I last looked at the example. 

Also, the command creation only has to be done once, and not on every bot startup, which should probably be noted there as well.

The `create_global_application_command` docs should also link to the `GuildId` docs.

sbrocket:
I admit that this confused me initially too. I wouldn't add any new methods, and the example is probably enough, but it wouldn't hurt to add a pointer to the GuildId methods in ApplicationCommand's docs too.

Things like "command creation only has to be done once" might be useful too, though it doesn't seem like Serenity's docs in general aim to replace or include all of the nitty gritty detail of Discord's own API documentation

teotwaki:
@SparkyPotato Can this be closed then?

PR #1427, opened 2021-07-06
Add Event methods to get related user, guild, channel, and message IDs

sbrocket:
These methods will be used as part of the new generic EventCollector implementation, but this is long enough on its own and impacts a common type (Event) such that it warrants separate review while I finish EventCollector.

### DIFF
--- a/src/model/event.rs
+++ b/src/model/event.rs
@@ -1,5 +1,6 @@
 //! All the events this library handles.
 
+use std::convert::TryFrom;
 #[cfg(feature = "cache")]
 use std::mem;
 use std::{collections::HashMap, fmt};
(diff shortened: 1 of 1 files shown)


sbrocket:
By the way, specifically what these methods will be used for is to implement some simple user/guild/channel/message ID filters for the EventCollector, similar to what the existing collectors have, for use unless someone needs a more complex filter. These will make it easy to create a collector to, say, "give me all MessageUpdate events in this channel from this user".

sbrocket:
Pushing this back to draft, I switched these impls to be generated by a macro in later commits and am going to update the impl here to use that from the start.

arqunis (approved):

PR #1429, opened 2021-07-07
Create a generic EventCollector

sbrocket:
This creates a new collector impl that can be used to collect arbitrary event types, rather than the specific event types supported by the simpler collectors.

### DIFF
--- a/src/client/bridge/gateway/mod.rs
+++ b/src/client/bridge/gateway/mod.rs
@@ -69,7 +69,7 @@ use crate::gateway::ConnectionStage;
 /// A message either for a [`ShardManager`] or a [`ShardRunner`].
 // Once we can use `Box` as part of a pattern, we will reconsider boxing.
 #[allow(clippy::large_enum_variant)]
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub enum ShardClientMessage {
     /// A message intended to be worked with by a [`ShardManager`].
     Manager(ShardManagerMessage),
@@ -109,6 +109,7 @@ pub enum ShardManagerMessage {
     ShardInvalidGatewayIntents,
     /// If a connection has been established but privileged gateway intents
     /// were provided without enabling them prior.
+
     /// Emitted when a shard received a [`DisallowedGatewayIntents`] error.
     ///
     /// [`DisallowedGatewayIntents`]: crate::gateway::GatewayError::DisallowedGatewayIntents
--- a/src/client/bridge/gateway/mod.rs
+++ b/src/client/bridge/gateway/mod.rs
@@ -69,7 +69,7 @@ use crate::gateway::ConnectionStage;
 /// A message either for a [`ShardManager`] or a [`ShardRunner`].
 // Once we can use `Box` as part of a pattern, we will reconsider boxing.
 #[allow(clippy::large_enum_variant)]
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub enum ShardClientMessage {
     /// A message intended to be worked with by a [`ShardManager`].
     Manager(ShardManagerMessage),
--- a/src/client/bridge/gateway/mod.rs
+++ b/src/client/bridge/gateway/mod.rs
@@ -109,7 +109,6 @@ pub enum ShardManagerMessage {
     ShardInvalidGatewayIntents,
     /// If a connection has been established but privileged gateway intents
     /// were provided without enabling them prior.
-
     /// Emitted when a shard received a [`DisallowedGatewayIntents`] error.
     ///
     /// [`DisallowedGatewayIntents`]: crate::gateway::GatewayError::DisallowedGatewayIntents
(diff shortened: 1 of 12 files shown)


drklee3:
I believe bumping the MSRV would be a breaking change so this and #1430 would have to be retargeted to the `next` branch (no breaking changes on the `current` branch).

sbrocket:
I can replace the Iterator::reduce usage with Iterator::fold and avoid bumping the MSRV then. Didn't realize bumping the MSRV was considered a breaking change.

sbrocket:
Any comments or changes needed on this?

src/client/bridge/gateway/mod.rs
```diff
@@ -69,7 +69,7 @@ use crate::gateway::ConnectionStage;
 /// A message either for a [`ShardManager`] or a [`ShardRunner`].
 // Once we can use `Box` as part of a pattern, we will reconsider boxing.
 #[allow(clippy::large_enum_variant)]
-#[derive(Clone, Debug)]
+#[derive(Debug)]
```
Zaitara: Why have you removed it?

src/collector/event_collector.rs
```diff
@@ -0,0 +1,443 @@
...
+    time::Duration,
+};
+
+use futures::{future::BoxFuture, stream::Stream};
+use tokio::sync::mpsc::{
+    unbounded_channel,
+    UnboundedReceiver as Receiver,
+    UnboundedSender as Sender,
+};
+#[cfg(all(feature = "tokio_compat", not(feature = "tokio")))]
```
Zaitara, line 16:
```suggestion
#[cfg(not(feature = "tokio"))]
```

src/gateway/mod.rs
```diff
@@ -157,7 +157,7 @@ impl Display for ConnectionStage {
 /// [`client`]: crate::client
 /// [`gateway`]: crate::gateway
 /// [`voice`]: crate::model::voice
-#[derive(Clone, Debug)]
+#[derive(Debug)]
```
Zaitara: Same here

Zaitara (changes requested):

src/collector/event_collector.rs
```diff
@@ -0,0 +1,443 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EventFilter")
+            .field("event_types", &self.event_types)
+            .field("filter_limit", &self.filter_limit)
+            .field("collect_limit", &self.collect_limit)
+            .field("filter", &"Option<Arc<dyn Fn(&Arc<Event>) -> bool + 'static + Send + Sync>>")
+            .field("channel_id", &self.channel_id)
+            .field("guild_id", &self.guild_id)
+            .field("user_id", &self.user_id)
+            .field("user_id", &self.message_id)
```
drklee3:
```suggestion
            .field("message_id", &self.message_id)
```

src/client/bridge/gateway/mod.rs
sbrocket: This allows `EventFilter` and `event_collector::FilterOptions` to be `!Clone`. It turned out that these types that I removed the `Clone` derive from were never actually cloned, only moved. Removing the `Clone` bound allows the collector's filter closure to be `FnMut` without needing a lock, so that the closure can capture mutable data. Contrast with the existing collectors' respective `FilterOptions` types which use `Arc<dyn Fn<...>>`.

If it turns out that these types ever need to be cloned in the future `FilterOptions` can be updated from `Box<dyn FnMut<...>>` to `Arc<Mutex<FnMut<...>>>`, but for now that's not needed.

src/collector/event_collector.rs
sbrocket: Can you clarify why you're suggesting this? What I have currently matches the #[cfg] directive in all the existing collectors. Either way, having `tokio` disabled without `tokio_compat` enabled will cause a build failure, either because there's no imported `sleep` and `Sleep` (what I have now) or because this is trying to import something that doesn't exist (your suggestion).

Zaitara: It is to be consistent with the other cfg restriction below. But such an issue should not allowed in the `build.rs` file no?

sbrocket: build.rs only checks for a missing TLS backend, it doesn't check that you have one of either `tokio` or `tokio_compat` enabled.

I'm fine with changing this but I'd rather do it in a separate PR and change all the collectors at the same time if so, and leave this the same as the others for now. How does that sound?

src/collector/event_collector.rs
sbrocket: Fixed

sbrocket:
Comments addressed

sbrocket:
Format CI failed but for unrelated reasons, and it doesn't look like I can trigger a re-run

sbrocket:
Clone removal removed since it's a breaking change and filter closure switched to Fn, will propose updating to FnMut on next.

arqunis (approved):

PR #1444, opened 2021-07-17
Make cache Synchronous and use DashMap

vicky5124:
This PR aims to make the cache not require an asynchronous context to be used, along with removing locks for a way lower probability of a deadlock or poisoned locks, this was done by replacing `tokio::RwLock<HashMap>` with `DashMap` and `parking_lot::RwLock<...>` for other types that are neither Map nor Set.

Using parking_lot does not pull an extra dependency, as it's already a dependency of tokio. DashMap is the only new dependency, which itself doesn't depend on much.

This is a step forward to a library that depends less on tokio, so most of it features can compile to wasm, so they can be used with [wasm_plugin](https://github.com/alec-deason/wasm_plugin/) to allow for dynamic runtime functions and commands in the future.

~~Depends on https://github.com/xacrimon/dashmap/pull/152 to be merged, so `Clone` can be implemented on `Cache::MessageIterator`~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -151,6 +151,15 @@ version = "0.3.16"
 [dependencies.mime_guess]
 version = "2.0"
 
+[dependencies.dashmap]
+version = "4"
+features = ["serde"]
+optional = true
+
+[dependencies.parking_lot]
+version = "0.11"
+optional = true
+
 [dev-dependencies.http_crate]
 version = "0.2"
 package = "http"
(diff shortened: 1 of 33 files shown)


arqunis (approved):

vicky5124:
When the DashMap PR gets merged (because as of now, it looks like Acri is not looking at any PRs or Issues being done), another PR can be made to implement Clone to `Cache::MessageIterator`, as it will be non-breaking to do so.

This can probably be merged already.

CI is not passing though, but that's an issue on gh side

arqunis (approved):

PR #1447, opened 2021-07-26
Include Discord's error explanations in Display impl

kangalio:
This PR modified the std::fmt::Display implementation for serenity::http::Error to include the [recently added](https://github.com/serenity-rs/serenity/pull/1310) [detailed explanation messages](https://discord.com/developers/docs/reference#error-messages) for unsuccessful requests.

If a list of error explanations is present in the ErrorResponse, they are put in parentheses behind the normal error message, separated by commas. Example:
```diff
-Invalid Form Body
+Invalid Form Body (parse:["users"] and users: [ids...] are mutually exclusive.)
```

I am not sure if this is the best way to format the error. List of ideas I had:
- `Invalid Form Body (parse:["users"] and users: [ids...] are mutually exclusive.)`
- `Invalid Form Body ("parse:["users"] and users: [ids...] are mutually exclusive.")`
- `Invalid Form Body` with `parse:["users"] and users: [ids...] are mutually exclusive.` returned as a source error

### DIFF
--- a/src/http/error.rs
+++ b/src/http/error.rs
@@ -181,7 +181,7 @@ impl Display for Error {
                     f.write_str(&error.message)?;
                     for error in errors_iter {
                         f.write_str(", ")?;
-                        f.write_str(&error.message);
+                        f.write_str(&error.message)?;
                     }
                     f.write_str(")")?;
                 }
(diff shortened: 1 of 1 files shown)


arqunis (approved):
> I am not sure if this is the best way to format the error. List of ideas I had:

Well, I would choose option \#3. However, taking a quick glance, it appears the error is a combination of its message and explanations, which originate from `DiscordJsonError`. Considering how they come from a singular place, there is no "source" error, and it would be pointless to separate the message and the explanations, as they relate to one another. I say the current approach is fine.

PR #1469, opened 2021-08-13
Add user `banner` and `accent_color` fields

drklee3:
### Description

Adds support for new user profile banner and accent_color.

Relevant documentation: https://github.com/discord/discord-api-docs/pull/3448

### Tested

Fetching a user correctly has the banner hash and accent_color deserialized as a Colour struct.

### DIFF
--- a/src/model/user.rs
+++ b/src/model/user.rs
@@ -506,10 +506,18 @@ pub struct User {
     /// change if the username+discriminator pair becomes non-unique.
     #[serde(rename = "username")]
     pub name: String,
-    /// the public flags on a user's account
+    /// The public flags on a user's account
     pub public_flags: Option<UserPublicFlags>,
     /// Optional banner hash.
+    ///
+    /// **Note**: This will only be present if the user is fetched via Rest API,
+    /// e.g. with [`Http::get_user`].
     pub banner: Option<String>,
+    /// The user's banner color encoded as an integer representation of
+    /// hexadecimal color code
+    ///
+    /// **Note**: This will only be present if the user is fetched via Rest API,
+    /// e.g. with [`Http::get_user`].
     #[cfg(feature = "utils")]
     pub accent_color: Option<Colour>,
     #[cfg(not(feature = "utils"))]
@@ -627,6 +635,9 @@ impl User {
     /// Returns the formatted URL of the user's banner, if one exists.
     ///
     /// This will produce a WEBP image URL, or GIF if the user has a GIF banner.
+    ///
+    /// **Note**: This will only be present if the user is fetched via Rest API,
+    /// e.g. with [`Http::get_user`].
     #[inline]
     pub fn banner_url(&self) -> Option<String> {
         banner_url(self.id, self.banner.as_ref())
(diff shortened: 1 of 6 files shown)


src/model/user.rs
```diff
@@ -493,6 +495,9 @@ pub struct User {
     pub name: String,
     /// the public flags on a user's account
     pub public_flags: Option<UserPublicFlags>,
+    /// Optional banner hash.
+    pub banner: Option<String>,
+    pub accent_color: Option<Colour>,
```
arqunis: This will fail to compile if the library is compiled without `utils` support.
```suggestion
    #[cfg(feature = "utils")]
    pub accent_color: Option<Colour>,
    #[cfg(not(feature = "utils"))]
    pub accent_color: Option<u32>,
```

src/model/user.rs
```diff
@@ -49,6 +49,8 @@ pub struct CurrentUser {
     pub name: String,
     pub verified: Option<bool>,
     pub public_flags: Option<UserPublicFlags>,
+    pub banner: Option<String>,
+    pub accent_color: Option<Colour>,
```
arqunis:
```suggestion
    #[cfg(feature = "utils")]
    pub accent_color: Option<Colour>,
    #[cfg(not(feature = "utils"))]
    pub accent_color: Option<u32>,
```

src/model/user.rs
arqunis: Also, might want to change `color` to `colour` and use `#[serde(rename = "...")]` on the field. We prefer British English spelling.

arqunis (changes requested):

src/model/user.rs
Zaitara: > This will fail to compile if the library is compiled without `utils` support.
> ```suggestion
>     #[cfg(feature = "utils")]
>     pub accent_color: Option<Colour>,
>     #[cfg(not(feature = "utils"))]
>     pub accent_color: Option<u32>,
> ```

Maybe that it's time to put `Colour` outside of the util module?

arqunis: I've been considering truncating some of the features we have, `utils` being one. If you look at the contents of the module, the impact it has on compile time is inconsequential, because it's very small and does not depend on any dependency other than `serde`/`serde_json`. I don't any benefit in putting its code behind a feature, and `model` depends on `utils` in a lot of places anyway.

That is a plan for a major release, however, since removal of a feature would be a breaking change.

drklee3:
Seems like Discord doesn't provide the banner hash and accent_color unless its through the fetch user endpoint. So things like messages / member update would cause the user cache have them missing.

This means to get a user's banner you need to bypass the cache with `UserId::to_user(&ctx.http)` which isn't great🤔

Should we just wait until the feature is finalized in the documentation to see if Discord will add the banner and accent_color fields to events?

Flat:
In addition to the (possible) utils change, I believe we should wait for a final API to see if this is added to event updates, and if it is not move from there.

arqunis (approved):

PR #1479, opened 2021-08-27
Undocumented `#[required_permissions(perms)]` requirements

woongzeyi:
Ever since intents became mandatory, developers have been needed to add required intents while building client when using `#[required_permissions(perms)]` attribute from `group` and `command` macro, which is never documented. 
I fixed the misleading command framework example and added requirement docs to the macro attribute, which should help fixing this issue.

PR #1480, opened 2021-08-27
Cache members on more events

Splingush:
Hello!

This pull-request is about caching members when the necessary data is being received from various events.
So far, this pull-request caches the user+member in the `message-create` and `message-update`-event. The `message-reaction-add` and `typing-start` would also expose the member in the payload, but I would have to implement the `CacheUpdate`-trait for those events, and I do not know the semantics of the return-value of the `update`-function there.
Likewise, the `CacheUpdate` for the message-events are now not only doing a message-cache-update, but also a user/member-cache-update.
I have used similar code as in the `GuildMemberUpdateEvent` for the changes.

This also fixes an issue with the guild-member-update-event replacing the avatar by a None-value when one is available, due to an invalid deserialization.

The issue that https://github.com/serenity-rs/serenity/pull/1479 is solving by adding more intents to make the `#[required-permissions]` work would be solved with these changes differently, by having the member cached when messages are being received, and not only on GuildMember-related events or on startup (which needs the privileged `GUILD_PRESENCES`-intent).

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -150,7 +150,9 @@ impl DispatchEvent {
             Self::Model(Event::ReactionRemoveAll(ref mut event)) => {
                 update(cache_and_http, event).await;
             },
-            // TODO: ADD CACHE_UPDATE FOR typing_start
+            Self::Model(Event::TypingStart(ref mut event)) => {
+                update(cache_and_http, event).await;
+            },
             _ => (),
         }
     }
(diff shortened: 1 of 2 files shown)


Splingush:
This is now also keeping reactions on cached messages up to date. This does not necessarily keep it in the same order as Discord would send them on fetching messages.
If there is a better alternative to doing `cache.current_user_id().await` for getting the bot-user-id, please let me know.
If you think this should be a separate PR, please let me know as well.

Should I profile it, to see if these changes are causing performance to degrade? If so, do you have some recommended way of doing this?

Best Wishes

Splingush:
Hey, sorry about the long inactivity, life got in the way.
The PR currently works on multiple issues at once, and would have to be pulled apart. That's why it is still a draft.

However, I will not continue working on it. If any is still relevant for you, feel free to reopen or reuse code from these commits. Thank you!

PR #1482, opened 2021-08-28
Fix missing field `pack_id` decode error

Atakku:
Have been getting some errors saying ```"missing field `pack_id`"```
Discord [defines](https://discord.com/developers/docs/resources/sticker#sticker-object-sticker-structure) `pack_id` as optional. Changing pack_id to be `Option<StickerPackId>` seems to have resolved the issue


arqunis (approved):
This is technically a breaking change (changing the type of a field of the public API), but since the deserialisation of `Sticker` is broken without this, I'll make an exception.

Issue #1466, opened 2021-08-09
Audit log parsing fails


j2ghz:
It seems discord has added new audit log models (83..112), and the parsing function returns an error:
https://github.com/serenity-rs/serenity/blob/b2565da364dc7c50b10d9991968cb286d17a4f6d/src/model/guild/audit_log.rs#L380-L381
```rust
Http(Request(reqwest::Error { kind: Decode, source: Error("Unexpected action number: 111", line: 1, column: 318) }))
```
Regardless of updating it, would it make sense to change it to return UnkownAuditLogEntry for unknown numbers? To contain the original json, similarly to how Event and EventHandler handle unkown events.

nickelc:
Fixed by #1875

PR #1483, opened 2021-08-29
New audit log models

woongzeyi:
#1466 
Discord has added 3 new audit log models: `STAGE_INSTANCE`, `STICKER` & `THREAD`, which isn't supported by Serenity's audit log interface yet. 
This pull request adds these new model into Serenity (Three new enum variants for `Action`, and three new action enums).

### DIFF
--- a/examples/e05_command_framework/src/main.rs
+++ b/examples/e05_command_framework/src/main.rs
@@ -18,7 +18,7 @@ use std::{
 use serenity::prelude::*;
 use serenity::{
     async_trait,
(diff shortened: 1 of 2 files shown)


arqunis:
Sorry for the late reply. This looks good, but could you properly re-run `cargo fmt`? Your formatted code doesn't match our [rustfmt configuration](https://github.com/serenity-rs/serenity/blob/9bbb25aac4d651804286f333eb503a72d41e473b/rustfmt.toml). Make sure you are using the latest nightly, as that's the only version we support for `rustfmt` (we use unstable features).

arqunis (approved):

PR #1476, opened 2021-08-23
Add a simple dashboard example.

vicky5124:
This PR adds an example for a bot dashboard, using RillRate

	modified:   .github/workflows/ci.yml
	modified:   Cargo.toml
	modified:   Makefile.toml
	modified:   examples/README.md
	new file:   examples/e15_simple_dashboard/Cargo.toml
	new file:   examples/e15_simple_dashboard/Makefile.toml
	new file:   examples/e15_simple_dashboard/example video.mp4
	new file:   examples/e15_simple_dashboard/src/main.rs

PR #1495, opened 2021-09-07
Create SQLite example

kangalio:
This PR adds an example which shows using SQLx and SQLite to write a simple todo bot.

I added a README.md because using SQLx effectively requires a bit of background knowledge.


vicky5124:
Some small stuff need to be worked on before merging this: (1-2 break CI/CD)

1. Missing `offline` feature on `sqlx`.
2. Missing `sqlx-data.json` generated from `cargo sqlx prepare`
3. Example number 15 merge conflicts with #1476  and should be 16 to be safe
4. The example was not added to `/Makefile.toml` nor `/examples/README.md`
5. The Actions example check name is 14
6. It should probably be mentioned that `query!` automatically generates its own structs with the fields and types of the database, and that `query_as!` should be used if a custom structure is preferred.

kangalio:
The Makefiles for the new example may not work I think because they don't supply DATABASE_URL. I don't know enough about that Makefile-system to check or fix that though

vicky5124:
sqlx-cli reads `.env` so it's not needed, and the readme already says to use `.env` so it should work.

arqunis (approved):

PR #1502, opened 2021-09-13
Fix: spawn_named gives type error

Milo123459:
You would get an error like this:
image

This PR fixes it. Not sure how the CI didn't pick it up...

### DIFF
--- a/src/internal/tokio.rs
+++ b/src/internal/tokio.rs
@@ -6,7 +6,7 @@ where
     T: Send + 'static,
 {
     #[cfg(tokio_unstable)]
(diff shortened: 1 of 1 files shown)


arqunis (approved):
> Not sure how the CI didn't pick it up.

That is because we do not pass `--cfg tokio_unstable` anywhere in CI, which meant it did not compile the particular snippet gated behind `tokio_unstable`.

Milo123459:
> > Not sure how the CI didn't pick it up.
> 
> That is because we do not pass `--cfg tokio_unstable` anywhere in CI, which meant it did not compile the particular snippet gated behind `tokio_unstable`.

I realised now, haha

PR #1503, opened 2021-09-13
Better CI caching

Milo123459:
This PR modifies all workflows to add a better way of caching. If you need details why, check [here](https://github.com/Swatinem/rust-cache)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,15 +66,8 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libopus-dev
 
-      - name: Setup cache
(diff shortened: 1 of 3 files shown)


nickelc:
can you add an empty commit to see another ci run?

Milo123459:
> can you add an empty commit to see another ci run?

Sure. Hold on.

arqunis (approved):

arqunis:
Comparing the execution times of CI runs for before and after applying Swatinem's `rust-cache`, it's clearly evident that there are improvements across the board. For most jobs, a minute or two have been shaved off, which is great as it will encourage more prototyping and catching errors in builds or tests quicker.

PR #1507, opened 2021-09-17
Replace UserId with ApplicationId in some places and remove unused structures

kangalio:
Some structs used `UserId` for struct fields that should really be `ApplicationId`. While fixing this, I discovered two big structs that were totally unused - `ApplicationInfo` and `BotApplication` - so I removed those as well.

This is a breaking change because of the type change from `UserId` to `ApplicationId`.

Fun fact: the increased type safety actually uncovered a bug in one of the examples


kangalio:
Failing tests are false positives

arqunis:
[Manually squashed and merged](https://github.com/serenity-rs/serenity/commit/605b7f733fa64cbdfed65e4ec8bde483700ed0b8).

PR #1519, opened 2021-09-29
Implement deriving application id from token

### LINKED ISSUE #1457: Derive application ID from bot token
Currently, the bot's application ID must be provided manually if unstable_discord_api features are used.

However, the bot token contains the bot ID, which can be extracted via `serenity::utils::parse_token`. I think the bot ID is equal to the application ID, so it may be possible to remove the requirement of supplying the application ID manually.

Milo123459:
Fixes #1457

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -364,7 +364,7 @@ impl<'a> Future for ClientBuilder<'a> {
                     let application_id = parse_token(&self.token)
                         .expect("Couldn't parse token correctly")
                         .bot_user_id;
-                    self.application_id(application_id.as_u64());
+                    self.application_id(*application_id.as_u64());
                 }
             }
 
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -687,6 +687,14 @@ impl Client {
     /// [gateway docs]: crate::gateway#sharding
     #[instrument(skip(self))]
     pub async fn start(&mut self) -> Result<()> {
+        #[cfg(feature = "unstable_discord_api")]
+        {
+            if &self.application_id.is_none() {
+                let application_id =
+                    parse_token(&self.token).expect("Couldn't parse token correctly").bot_user_id;
+                self.application_id(application_id);
+            }
+        }
         self.start_connection([0, 0, 1]).await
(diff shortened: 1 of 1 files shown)


src/client/mod.rs
```diff
@@ -687,6 +691,14 @@ impl Client {
...
     #[instrument(skip(self))]
     pub async fn start(&mut self) -> Result<()> {
+        #[cfg(feature = "unstable_discord_api")]
+        {
+            if &self.application_id.is_none() {
+                let application_id =
+                    parse_token(&self.token).expect("Couldn't parse token correctly").bot_user_id;
+                self.application_id(application_id);
+            }
+        }
```
arqunis: Why not do this in the instantiation of the `Client` in the `ClientBuilder`'s `Future` implementation? The `Client` has other `start` methods. Doing this only here would introduce an inconsistency.

Milo123459: I guess that makes sense, just means that there wouldn't be any doc comments.

Milo123459:
I fixed most of the errors, but now I'm stuck on this one

Djazouli:
Hello, I just happen to find this issue, and I have some thoughts on this:

It looks like the only usage of `application_id` is to create the `http` attribute. But currently, the code is calling `self.application_id(*...)`after the extraction of `http` with `self.http.take()`. Thus, I feel like the code would always panic two lines later, on `panic!("Please provide an Application Id in order to use interactions features.");` if the application_id was not specified before-ahnd

So, it would feel more natural to make use of this default application_id before getting the http client. But on the other hand, if the http client was already set, we do not really want to override it, so we'd have to check for this.

Also, `ClientBuilder::application_id`consumes self, so it feels like it may be hard to call this method within the implementation of `poll`, which takes a `Pin<&mut Self>`. Maybe the easier would be to copy-paste the content of the `ClientBuilder::application_id`method within the `poll`implementation, but I feel like this is not the right thing to do.

I'd love to hear your thoughts about this !

Milo123459:
That probably makes more sense - yea. I'll rework my implementation soon™.

PR #1523, opened 2021-10-02
Fix support for raw identifiers in proc macros

### LINKED ISSUE #1454: `#[command]` proc_macro panics on raw identifiers
I have a command like this:
```rust
#[command]
... snip ...
async fn r#move(ctx: &Context, msg: &Message, mut args: Args) -. CommandResult { ... }
```
However the proc-macro is panicking on the raw identififer:
```rust
error: custom attribute panicked
    |
728 | #[command]
    | ^^^^^^^^^^
    |
    = help: message: `"R#MOVE_COMMAND_OPTIONS"` is not a valid identifier
```
It looks like it's trying to uppercase the "literal" string of `r#move` instead of taking the string of the identifier itself of `move`.

I can of course work around it by renaming the function to something like `move_` and adding `#[aliases("move")]` but it really shouldn't be panicking on valid identifiers and that leaves both a `move` and a `move_` in the help when I don't want the `move_` to show.

Friz64:
You can now declare commands and groups with raw identifiers and they will be referenced and used through the non-raw version of their name.

Fixes #1454

### DIFF
--- a/command_attr/src/lib.rs
+++ b/command_attr/src/lib.rs
@@ -104,7 +104,7 @@ pub fn command(attr: TokenStream, input: TokenStream) -> TokenStream {
     let _name = if !attr.is_empty() {
         parse_macro_input!(attr as Lit).to_str()
     } else {
-        fun.name.to_string()
+        fun.name.to_string_non_raw()
     };
 
     let mut options = Options::new();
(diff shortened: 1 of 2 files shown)


arqunis (approved):

arqunis:
Thanks!

PR #1537, opened 2021-10-07
Add channel types to slash commands

vicky5124:
https://github.com/discord/discord-api-docs/commit/0a4c2fe7ff57828dbacd9f57f39331cf510c56af
Limits the channels that show up on channel type arguments in slash commands.

	modified:   src/builder/create_application_command.rs
	modified:   src/model/channel/mod.rs
	modified:   src/model/interactions/application_command.rs

### DIFF
--- a/src/builder/create_application_command.rs
+++ b/src/builder/create_application_command.rs
@@ -134,7 +134,7 @@ impl CreateApplicationCommandOption {
     /// If the option is a [`Channel`], it will only be able to show these types.
     ///
     /// [`Channel`]: crate::model::interactions::application_command::ApplicationCommandOptionType::Channel
-    pub fn channel_types(&mut self, channel_types: Vec<ChannelType>) -> &mut Self {
+    pub fn channel_types(&mut self, channel_types: &[ChannelType]) -> &mut Self {
         self.0.insert(
             "channel_types",
             Value::Array(
--- a/src/builder/create_application_command.rs
+++ b/src/builder/create_application_command.rs
@@ -3,6 +3,7 @@ use std::collections::HashMap;
 use serde_json::{json, Value};
 
 use crate::internal::prelude::*;
+use crate::model::channel::ChannelType;
 use crate::model::interactions::application_command::{
     ApplicationCommandOptionType,
     ApplicationCommandType,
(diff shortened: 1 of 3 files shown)


src/builder/create_application_command.rs
```diff
@@ -129,6 +130,21 @@ impl CreateApplicationCommandOption {
 
         self
     }
+
+    /// If an option is a Channel type, it will only be able to show these types.
+    pub fn add_channel_types(&mut self, channel_types: Vec<ChannelType>) -> &mut Self {
```
arqunis: Name should be just `channel_types`, because we have established that `add_*` methods on builders will append data, not overwrite. The option type `Channel` should also be referenced, just like in the above method's documentation.
```suggestion
    /// If the option is a [`Channel`], it will only be able to show these types.
    ///
    /// [`Channel`]: crate::model::interactions::application_command::ApplicationCommandOptionType::Channel
    pub fn channel_types(&mut self, channel_types: Vec<ChannelType>) -> &mut Self {
```

src/model/interactions/application_command.rs
```diff
@@ -908,6 +908,9 @@ pub struct ApplicationCommandOption {
     /// [`SubCommandGroup`]: ApplicationCommandOptionType::SubCommandGroup
     #[serde(default)]
     pub options: Vec<ApplicationCommandOption>,
+    /// If an option is a Channel type, it will only be able to show these types.
```
arqunis:
```suggestion
    /// If the option is a [`Channel`], it will only be able to show these types.
    ///
    /// [`Channel`]: ApplicationCommandOptionType::Channel
```

arqunis (changes requested):

src/builder/create_application_command.rs
```diff
@@ -129,6 +130,23 @@ impl CreateApplicationCommandOption {
 
         self
     }
+
+    /// If the option is a [`Channel`], it will only be able to show these types.
+    ///
+    /// [`Channel`]: crate::model::interactions::application_command::ApplicationCommandOptionType::Channel
+    pub fn channel_types(&mut self, channel_types: Vec<ChannelType>) -> &mut Self {
```
kangalio: Could this take `&[ChannelType]` instead?

vicky5124: Because it gets mapped, this was possible to do in 55ce06b09

arqunis (approved):

PR #1538, opened 2021-10-10
Perf!: remove async in http builder

BastiDood:
# Summary
Previously, the `HttpBuilder` implemented the `Future` trait so that it may be constructed with `async`-`await`. However, this necessitated an extra heap allocation since the underlying future was stored inside a pinned `Box`. This issue may be resolved by wrapping the resultant `Http` client in a [`Ready` struct](https://doc.rust-lang.org/std/future/fn.ready.html) instead, thereby avoiding the heap allocation altogether.

However, I figured that since the `Http` client is already available for consumption, then `async`-`await` is unnecessary. That is, a synchronous builder is sufficient and less costly. This PR implements my proposal to remove the unnecessary extra work performed by the code below:

https://github.com/serenity-rs/serenity/blob/5a700f7ea40fc89ff1c5550e673dd37a3b739f06/src/http/client.rs#L198-L208

# Breaking Changes
Since `HttpBuilder` no longer implements `Future`, then the old example no longer works:

```rust
let http = HttpBuilder::new("token")
    .proxy("http://127.0.0.1:3000")
    .expect("Invalid proxy URL")
    .ratelimiter_disabled(true)
    .await
    .expect("Error creating Http");
```

Instead, we must now use the explicit `build` method:

```rust
let http = HttpBuilder::new("token")
    .proxy("http://127.0.0.1:3000")
    .expect("Invalid proxy URL")
    .ratelimiter_disabled(true)
    .build();
```

Note, however, that the current implementation panics when the client is built with invalid parameters (as did the old `Future` implementation). Eventually, we may add a fallible `try_build` method to avoid panicking, but that addition is beyond the scope of this PR since it may require the addition of an error enum.

Anyway, if there are any issues with these changes, please do let me know. In the event that I miss updating some documentation, I will gladly address them. Thanks!

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -63,18 +63,17 @@ use crate::model::prelude::*;
 ///     .expect("Error creating Http");
 /// # }
 /// ```
-pub struct HttpBuilder<'a> {
+pub struct HttpBuilder {
     client: Option<Arc<Client>>,
     ratelimiter: Option<Ratelimiter>,
     ratelimiter_disabled: Option<bool>,
     token: Option<String>,
     proxy: Option<Url>,
-    fut: Option<BoxFuture<'a, Result<Http>>>,
     #[cfg(feature = "unstable_discord_api")]
     application_id: Option<u64>,
 }
 
-impl<'a> HttpBuilder<'a> {
+impl HttpBuilder {
     fn _new() -> Self {
         Self {
             client: None,
@@ -82,7 +81,6 @@ impl<'a> HttpBuilder<'a> {
             ratelimiter_disabled: Some(false),
             token: None,
             proxy: None,
-            fut: None,
             #[cfg(feature = "unstable_discord_api")]
             application_id: None,
         }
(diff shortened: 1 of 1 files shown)


src/http/client.rs
```diff
@@ -166,49 +154,38 @@ impl<'a> HttpBuilder<'a> {
 
         Ok(self)
     }
-}
 
-impl<'a> Future for HttpBuilder<'a> {
-    type Output = Result<Http>;
+    /// Use the given configuration to build the `Http` client.
+    pub fn build(mut self) -> Http {
+        let token = self.token.take().unwrap();
```
arqunis: Since we are no longer behind `&mut self`, we can safely take ownership of each individual field. As a result, `Option::take` isn't necessary anymore.
```suggestion
        let token = self.token.unwrap();
```

src/http/client.rs
```diff
@@ -166,49 +154,38 @@ impl<'a> HttpBuilder<'a> {
...
+
+        #[cfg(feature = "unstable_discord_api")]
+        let application_id = self
+            .application_id
+            .expect("Expected application Id in order to use interacions features");
+
+        let client = self.client.take().unwrap_or_else(|| {
```
```suggestion
        let client = self.client.unwrap_or_else(|| {
```

src/http/client.rs
```diff
@@ -166,49 +154,38 @@ impl<'a> HttpBuilder<'a> {
...
+            let builder = configure_client_backend(Client::builder());
+            Arc::new(builder.build().expect("Cannot build reqwest::Client"))
+        });
+
+        let ratelimiter = self.ratelimiter.take().unwrap_or_else(|| {
```
```suggestion
        let ratelimiter = self.ratelimiter.unwrap_or_else(|| {
```

src/http/client.rs
```diff
@@ -166,49 +154,38 @@ impl<'a> HttpBuilder<'a> {
...
+            let client = Arc::clone(&client);
+            Ratelimiter::new(client, token.to_string())
+        });
 
-    #[allow(clippy::unwrap_used)]
-    #[instrument(skip(self))]
-    fn poll(mut self: Pin<&mut Self>, ctx: &mut FutContext<'_>) -> Poll<Self::Output> {
-        if self.fut.is_none() {
-            let token = self.token.take().unwrap();
+        let ratelimiter_disabled = self.ratelimiter_disabled.take().unwrap();
```
```suggestion
        let ratelimiter_disabled = self.ratelimiter_disabled.unwrap();
```

src/http/client.rs
```diff
@@ -166,49 +154,38 @@ impl<'a> HttpBuilder<'a> {
...
+        let proxy = self.proxy.take();
```
```suggestion
        let proxy = self.proxy;
```

arqunis (changes requested):
It is a nice proposal, because it simplifies the API/implementation. Besides a nitpick, my only request is to retarget the PR to the `next` branch. `current` is forbidden from having breaking changes.

BastiDood:
> It is a nice proposal, because it simplifies the API/implementation. Besides a nitpick, my only request is to retarget the PR to the `next` branch. `current` is forbidden from having breaking changes.

Sure! Let me just rebase everything. I will inform you as soon as the changes are ready. 👍

BastiDood:
@acdenisSK Alrighty, everything seems to be fixed now! I had some minor troubles with resolving the merge conflicts, but I think I've revised the Git history to be as clean as it can be. I have also applied the changes you requested. Hope this may be merged now!

src/http/client.rs
```diff
@@ -167,49 +161,37 @@ impl<'a> HttpBuilder<'a> {
 
         Ok(self)
     }
-}
 
-impl<'a> Future for HttpBuilder<'a> {
-    type Output = Result<Http>;
+    /// Use the given configuration to build the `Http` client.
+    pub fn build(self) -> Http {
```
arqunis, line 160: To prevent `clippy` from generating errors, we need this attribute. The `unwrap`s here are intentional.
```suggestion
    #[allow(clippy::unwrap_used)]
    pub fn build(self) -> Http {
```

arqunis (changes requested):

BastiDood:
Hmm... it seems that some of the checks are failing, but I'm not quite sure if it has anything to do with this PR. Is there anything I must do to resolve them?

arqunis (approved):
>  but I'm not quite sure if it has anything to do with this PR

It does not. CI on the `next` branch has been failing for a while. Anyway, this is good to be merged now. Thanks.

Issue #1358, opened 2021-05-24
Constants of the Colour struct needs to be updated


ghost:
Serenity provides some constants of the `Colour` struct (see [this in the docs](https://docs.rs/serenity/0.10.7/serenity/utils/struct.Colour.html#impl-1)).

- Some of them have been introduced across the years but I can't find out their relations with Discord.
- Other needs to be updated (`BLURPLE` for example) (see [Discord new colours](https://blog.discord.com/happy-blurpthday-to-discord-a-place-for-everything-you-can-imagine-fc99ee0a77c0)).
- Also, having the same red, green, orange or yellow that Discord is using for errors, warnings or successes might be useful.

```rust
let msg = msg.channel_id.send_message(&ctx.http, |m| {
    m.embed(|e| {
        e.title("Error");
        e.description("This is an error");
        e.colour(RED) // #ED4245 (new Discord's branding red).
        e
    });
    m
}).await;
```

Bellow the list of all colours currently in Serenity, then a list of colours used in Discord and finally the result of all colour that could be kept.

# Currently in Serenity

Currently, Serenity offers the 28 following colours:

| Name             | Hex       | Reason       |
| ---------------- | --------- | ------------ |
| `BLITZ_BLUE`     | `#6FC6E2` | ???          |
| `BLUE`           | `#3498DB` | role colours |
| `BLURPLE`        | `#7289DA` | old branding |
| `DARK_BLUE`      | `#206694` | role colours |
| `DARK_GOLD`      | `#C27C0E` | role colours |
| `DARK_GREEN`     | `#1F8B4C` | role colours |
| `DARK_GREY`      | `#607D8B` | role colours |
| `DARK_MAGENTA`   | `#AD1457` | role colours |
| `DARK_ORANGE`    | `#A84300` | role colours |
| `DARK_PURPLE`    | `#71368A` | role colours |
| `DARK_RED`       | `#992D22` | role colours |
| `DARK_TEAL`      | `#11806A` | role colours |
| `DARKER_GREY`    | `#546E7A` | role colours |
| `FABLED_PINK`    | `#FAB1ED` | ???          |
| `FADED_PURPLE`   | `#8882C4` | ???          |
| `FOOYOO`         | `#11CA80` | ???          |
| `GOLD`           | `#F1C40F` | role colours |
| `KERBAL`         | `#BADA55` | ???          |
| `LIGHT_GREY`     | `#979C9F` | role colours |
[...]

PR #1539, opened 2021-10-10
Add Discord's new branding, css & role colours

nickelc:
I grouped the colours by putting them into modules as a way to see where they are coming from.
This works around the different greens for **branding** `rgb(87, 242, 135)`,  **role** `rgb(46, 204, 113)`
and **positive** `hsl(139, 47.3%, 43.9%)` (positive is not an alias for branding green like #1358 stated).
Alternatively, the constants could be prefixed with `BRANDING_` / `ROLES_` and kept as associated constants of `Colour`. 

The current colour constants aren't changed beside removing the `colour!` macro.

see: #1358 

<details>
  <summary>weird flex but ok</summary>

image
</details>

### DIFF
--- a/src/utils/colour.rs
+++ b/src/utils/colour.rs
@@ -325,7 +325,6 @@ impl Default for Colour {
 }
 
 /// Colour constants used by Discord for their branding, role colour palette, etc.
-#[allow(dead_code)]
 pub mod colours {
     pub mod branding {
         use crate::utils::Colour;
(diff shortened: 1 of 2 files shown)


src/utils/colour.rs
```diff
@@ -335,6 +324,85 @@ impl Default for Colour {
     }
 }
 
+/// Colour constants used by Discord for their branding, role colour palette, etc.
+#[allow(dead_code)]
```
kangalio: Is the `#[allow(dead_code)]` needed?

nickelc: I added it because rust-analyzer complained about it for some reason.
Checked again and RA is now quiet..

arqunis (approved):

Issue #1423, opened 2021-07-03
Clippy fails when run with all features enabled


sbrocket:
I discovered this in #1422, because shifting the workspace that was previously under examples/ up to the repo root means that the `cargo clippy --workspace --tests` in the Lint action now runs clippy against serenity with all features enabled, rather than just default features.

Without that change, you can run `cargo clippy --all-features --tests` and see that it currently fails.

To unblock #1422 I'm going to update lint.yml to remove the `--workspace` arg, so that clippy is checking the same features as before. Readding it will require fixing all the outstanding clippy lints. I'm going to put up another PR will all the automatic fixes clippy was able to make, but there are other manual fixes that need to be made like missing doc sections.

PR #1543, opened 2021-10-10
Clean up Clippy errors

mdonoughe:
This fixes some of the errors preventing `cargo clippy --workspace --tests` (#1423).

I think this is the right thing to do? With the previous code, if the server sent something like a user ID that cannot be parsed to a `u64` or did not correspond to a known user, a panic would occur. With the new code, if the servers sends something like that the code will not panic and instead `options.resolved` will be left as `None`.

The only remaining error class is [missing_errors_doc](https://rust-lang.github.io/rust-clippy/master/index.html#missing_errors_doc). I'm not sure what to do about this. I was looking at the error handling code and for all the cases I checked there is no specific error handling so if anything goes wrong you just get `Error::Http`. Should I just copy and paste a generic message like "If there is an error, it will be either [`Error::Http`] or [`Error::Json`]." for those cases?

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,9 +30,7 @@ jobs:
         uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          # TODO(https://github.com/serenity-rs/serenity/issues/1423): Clippy currently fails with
-          # all features enabled. Re-add `--workspace` here once that's fixed.
-          args: --tests
+          args: --workspace --tests
 
   rustfmt:
     name: Format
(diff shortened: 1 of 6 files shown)


arqunis:
Sorry for the late reply, but regarding your question:
> Should I just copy and paste a generic message like "If there is an error, it will be either [`Error::Http`] or [`Error::Json`]." for those cases?

Yes, the message should be sufficient in most cases.

mdonoughe:
Okay that's mostly what I did. For one of them there was another similar function that had a slightly more specific errors section I adapted. I tried following some of the functions through to their official Discord API documentation and the errors aren't documented there either.

arqunis (approved):
The failing `Format` job appears to be bogus, so excluding that, this looks good. Thanks!

Edit: Nevermind, it was genuine. They have adjusted `rustfmt` to properly account for Rust code in documentation examples. Either way, that wasn't related to this PR.

PR #1544, opened 2021-10-11
Add update calls for dispatched thread events

AngelOnFira:
In #1465, @xMAC94x didn't add update calls to the event dispatcher for the threads. To be more clear, if a thread is created, Serenity receives an event, but doesn't properly update the cache. I think this is the solution, as it works for my needs. I took examples from how the `MessageCreate` event updates the cache.

Let me know if any changes are needed!


arqunis (approved):

arqunis:
Thanks!

PR #1550, opened 2021-10-13
Add ClientBuilder::get_token(&self) -> &str

kangalio:
This method is useful e.g. for a discord bot framework I'm writing, where the user can pass a ClientBuilder into the framework constructor. With access to the token, the framework can extract the bot user ID from the token which is needed for e.g. using bot mention as a prefix.

Anyways, I implemented this by replacing the two fields `token: Option<String>, http: Option<Http>` with just `http: Http`:
- the Http struct contains the token anyways
- removing the unneeded Option's means that get_token can return &str instead of Option<&str>

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -349,7 +349,7 @@ impl<'a> Future for ClientBuilder<'a> {
             let event_handler = self.event_handler.take();
             let raw_event_handler = self.raw_event_handler.take();
             let intents = self.intents;
-            let http = Arc::new(std::mem::replace(&mut self.http, Default::default()));
+            let http = Arc::new(std::mem::take(&mut self.http));
 
             #[cfg(feature = "unstable_discord_api")]
             if http.application_id == 0 {
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -76,12 +76,8 @@ pub use crate::CacheAndHttp;
 /// A builder implementing [`Future`] building a [`Client`] to interact with Discord.
 #[cfg(feature = "gateway")]
 pub struct ClientBuilder<'a> {
-    // FIXME: Remove this allow attribute once `application_id` is no longer feature-gated
-    // under `unstable_discord_api`.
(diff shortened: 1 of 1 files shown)


src/client/mod.rs
```diff
@@ -143,35 +138,27 @@ impl<'a> ClientBuilder<'a> {
...
-
-        self.token = Some(token.clone());
-
-        self.http = Some(Http::new_with_token(&token));
+        self.http = Http::new_with_token(token.as_ref());
 
         self
     }
 
+    pub fn get_token(&self) -> &str {
```
arqunis, line 155: There should be some documentation here.
```suggestion
    /// Gets the current token used for the [`Http`] client.
    ///
    /// [`Http`]: crate::http::Http 
    pub fn get_token(&self) -> &str {
```

src/client/mod.rs
```diff
@@ -359,7 +346,7 @@ impl<'a> Future for ClientBuilder<'a> {
             let event_handler = self.event_handler.take();
             let raw_event_handler = self.raw_event_handler.take();
             let intents = self.intents;
-            let http = Arc::new(self.http.take().unwrap());
+            let http = Arc::new(std::mem::replace(&mut self.http, Default::default()));
```
arqunis: [`std::mem::take`][take] can be used instead. It performs same work as this `replace` call here, but is simpler to use and shortens the code a bit. 
```suggestion
            let http = Arc::new(std::mem::take(&mut self.http));
```

[take]: https://doc.rust-lang.org/stable/std/mem/fn.take.html

arqunis (changes requested):

arqunis (approved):
[Seems like there's an extra space from my documentation suggestion](https://github.com/serenity-rs/serenity/pull/1550/checks?check_run_id=3908748445), sorry. Could you remove it? Otherwise, the PR is good to be merged.

kangalio:
Done

PR #1552, opened 2021-10-15
Make sure that guilds are removed from the cache if they were removed between disconnect and ready

ghost:
Whenever a bot is removed from a guild, they get an event and serenity correct handles it by removing the guild from the cache. If the bot is removed between the websocket disconnecting and reconnecting, however, it will never receive the event, and so the guild remains in the cache. This pull request makes it so that, upon receiving a ready event, all guilds (managed by that shard) that are cached but not in the ready event are removed from the cache.

Problems with current behavior:
- Potential very slow memory leak (after a few years, it could use a few megabytes!)
- If bots clean up data from guilds it's left, it'll never realize it's no longer in that guild. This also means that some processing could be going on in the background for the guild, which would cause errors, even though the cache contains the guild. This would be a pain to debug.

Problems with new behavior:
- If the cache is updated with a `Cache::update` that adds a new guild that the bot is not in on purpose, this will be cleared when reconnecting. This behavior would be much easier to debug than the errors with the old behavior, but is still something to consider.

Maybe this new behavior could be made perfect by adding a setting for this and exposing the cache settings somehow? Would that be too much for such a small behavioral change?

### DIFF
--- a/src/model/event.rs
+++ b/src/model/event.rs
@@ -1222,7 +1222,7 @@ impl CacheUpdate for ReadyEvent {
         let mut guilds_to_remove = vec![];
         let ready_guilds_hashset =
             HashSet::<GuildId>::from_iter(self.ready.guilds.iter().map(|status| status.id()));
-        let shard_data = self.ready.shard.unwrap_or_else(|| [1, 1]);
+        let shard_data = self.ready.shard.unwrap_or([1, 1]);
         for guild in cache.guilds.read().await.keys() {
             // Only handle data for our shard.
             if shard_id(guild.0, shard_data[1]) == shard_data[0]
--- a/src/model/event.rs
+++ b/src/model/event.rs
@@ -19,6 +19,8 @@ use crate::constants::OpCode;
 use crate::internal::prelude::*;
 #[cfg(feature = "unstable_discord_api")]
 use crate::model::interactions::{application_command::ApplicationCommand, Interaction};
+#[cfg(feature = "cache")]
+use crate::utils::shard_id;
 
 /// Event data for the channel creation event.
 ///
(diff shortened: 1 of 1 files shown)


Th3-M4jor:
Thanks for the PR, however I have one concern with the implementation.

Your usage of `iter().any()` within a `for` loop would be a rather expensive operation on a bot in many guilds. Might I suggest instead storing the guild-ids from the ready event in a `HashSet` and checking for existence that way?

ghost:
> Your usage of `iter().any()` within a `for` loop would be a rather expensive operation on a bot in many guilds. Might I suggest instead storing the guild-ids from the ready event in a `HashSet` and checking for existence that way?

Good suggestion! Added a commit to address this (and to address the compilation error I immediately introduced :P)

src/model/event.rs
```diff
@@ -1212,6 +1218,26 @@ impl CacheUpdate for ReadyEvent {
             }
         }
 
+        // We may be removed from some guilds between disconnect and ready, so we should handle that.
+        let mut guilds_to_remove = vec![];
+        let ready_guilds_hashset =
+            HashSet::<GuildId>::from_iter(self.ready.guilds.iter().map(|status| status.id()));
+        let shard_data = self.ready.shard.unwrap_or_else(|| [1, 1]);
```
arqunis: `*_else` variants on `Option` and `Result` are preferable when the default value is expensive to compute. In this case, however, `unwrap_or_else` can be substituted with `unwrap_or` since the value is a constant/isn't expensive to compute.
```suggestion
        let shard_data = self.ready.shard.unwrap_or([1, 1]);
```

arqunis (changes requested):

arqunis (approved):

PR #1553, opened 2021-10-15
Handle Unicode double-quote characters when parsing arguments

### LINKED ISSUE #1545: Support "Curly Quotes" for `Args.quoted`
### Overview
Users on Mavericks+ Apple platforms (eg iOS) by default have their quote keys input "Curly Quotes" rather than straight quotes (also known as "Smart Quotes" or "Smart Punctuation"). This causes arguments to fail parsing, as it appears [lexing checks against straight quotes only](https://docs.rs/serenity/0.10.9/src/serenity/framework/standard/args.rs.html#106-133). While users could be asked to [change their defaults](https://apple.stackexchange.com/questions/136402/how-to-disable-smart-quotes), it's an abrasive experience that they may not wish to endeavor; and - at least on iOS - is a global setting they may not wish to change for this one use case.

### Details

These Curly quotes appear to be [unicode characters](https://unicode-table.com/en/sets/quotation-marks/) `201C` and `201D`.

**Potential Solves**

- Bot owners could check themselves, and alert the user -- [kinda like the rust compiler](https://github.com/rust-lang/rust/issues/58436). This won't be very graceful, however, and would probably introduce some complexity around argument count validation when using the macros.
- Args could be standardized to replace various known quote shapes with their straight line forms before passed into lexing, but this may have unintended consequences.
- The lexer function (linked earlier) could check for multiple quote styles; the major downsides being if these quotes were intermixed it could be tricky to get right. It might be fair to assume the two types would not be mixed (as so: `"The “lazy” fox"`) or perhaps the first quote encountered can govern the style for the remainder of lexing.

Thoughts? I stumbled upon this in a guild with some friends, some of which use iPhone and others Android; some of whom are very not technical and it took some explaining to rectify what happened. But it's an annoying experience.

arqunis:
## Description

This adds support for "Smart Quotes", which is a feature on operating systems by Apple that automatically substitutes `"` with `”` when typing text. While innocent for display, it causes a problem with parsing arguments of commands. An argument may comprise of multiple words or numbers that should be grouped together. The `Args` struct handles these arguments by relying on quotes to determine the beginning and end of the argument. Currently, however, it does this by simply checking whether the current character is `"`, disregarding other quotation characters like `”`. This PR remedies that by checking for either `"` or `”`, but not both, to avoid ambiguities when both are used at the same time (e.g. `"The “lazy” fox"` should parse as `The “lazy” fox`).

Fixes #1545  

## Type of Change

This is a fix to the argument parsing process to account for Unicode quote characters like `”`, which a bot may encounter if the user is running an Apple OS (MacOS or iOS) and has the "Smart Quotes" feature on.

## How Has This Been Tested?

This has been tested by ensuring the ASCII `"` character still works for existing cases, and ensuring that `”` also works, and that using both in a single argument will not cancel the other (so that, again, `"The “lazy” fox"` should parse as `The “lazy” fox`, or vice versa).

### DIFF
--- a/src/framework/standard/args.rs
+++ b/src/framework/standard/args.rs
@@ -200,10 +200,7 @@ fn remove_quotes_only_if_present(s: &str) -> Option<&str> {
 }
 
 fn remove_quotes(s: &str) -> &str {
-    match remove_quotes_only_if_present(s) {
-        Some(unquoted) => unquoted,
-        None => s,
-    }
+    remove_quotes_only_if_present(s).unwrap_or(s)
 }
 
 #[derive(Debug, Clone, Copy)]
--- a/src/framework/standard/args.rs
+++ b/src/framework/standard/args.rs
@@ -180,7 +180,7 @@ fn lex(stream: &mut Stream<'_>, delims: &[Cow<'_, str>]) -> Option<Token> {
     Some(Token::new(TokenKind::Argument, start, end))
 }
 
-fn is_quoted(s: &str) -> Option<&str> {
+fn remove_quotes_only_if_present(s: &str) -> Option<&str> {
     // A quoted argument must have both quote characters. It cannot only contain a single quote, or
     // be empty.
     if s.len() < 2 {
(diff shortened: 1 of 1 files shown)


kadiwa4:
This works fine for English quotation mark usage, but for example with a German keyboard quotes look like „this“. Looking at [a list](https://en.wikipedia.org/wiki/Quotation_mark#Summary_table) of all the quotation mark styles, that might get too much though. Also, German quotes look a lot more distinct from ASCII quotes than the English ones do.

PR #1554, opened 2021-10-16
Remove `#[doc(cfg(...))]` attributes

arqunis:
## Description

This removes all uses of the `#[doc(cfg(...))]` attribute. The purpose of this attribute was to tell `rustdoc` to add an annotation, which would inform users what features an item (module, struct, trait, method, etc.) needed. However, in the latest Nightly, [this attribute is implicit when `#![feature(doc_cfg)]` is enabled](https://github.com/rust-lang/rust/pull/89596). Thus, our usage of this attribute is unnecessary and we can remove it.

## Type of Change

This is an enhancement to the library's code by removing clutter, which is related to documentation.

## How Has This Been Tested?

This has been tested by ensuring existing annotations haven't been removed.

### DIFF
--- a/src/builder/mod.rs
+++ b/src/builder/mod.rs
@@ -9,22 +9,17 @@ mod create_channel;
 mod create_embed;
 
 #[cfg(feature = "unstable_discord_api")]
-#[cfg_attr(docsrs, doc(cfg(feature = "unstable_discord_api")))]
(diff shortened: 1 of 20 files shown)


kangalio:
What a bliss not to have to duplicate all of the cfg's anymore :relieved:

PR #1559, opened 2021-10-21
Clean up ergonomics around setting bot prefixes

mkrasnitski:
`Configuration::prefix` already calls `to_string()` in its body, so to mirror the signature of `Configuration::prefixes`, I changed the parameter's type to `impl ToString` at no extra runtime cost. As well, to consolidate implementation differences, `Configuration::prefixes` now skips over empty strings in the passed-in Iterator, similar to the singular version.


arqunis (approved):

arqunis:
Thanks!

PR #1562, opened 2021-10-27
Add support for AutoComplete

Th3-M4jor:
I kept Autocomplete interactions the same inner type as regular `ApplicationCommands` since all it does is add a new field to `ApplicationCommandInteractionDataOption` and an optional field for creating `ApplicationCommand`s.


src/builder/create_interaction_response.rs
```diff
@@ -148,4 +148,55 @@ impl CreateInteractionResponseData {
         self.0.insert("components", Value::Array(components.0));
         self
     }
+
+    /// For autocomplete responses this sets their autocomlete suggestions.
```
kangalio: Typo "autocomlete"

kangalio (commented):
The changes look good to me, but my understanding of the internals of this part of the codebase is limited.

As a more thorough test, I'll play around with autocomplete options a bit once I get home in a few days.

kangalio:
After testing the PR, here's some suggestions:

- an `Interaction::autocomplete()` method, like already exists for the other variants?
- consistently spell it Auto**c**omplete (lowercase c), like Discord itself seems to do
- a designated type for autocomplete interactions? Those methods on ApplicationCommandInteraction don't make sense for autocomplete interactions

kangalio:
Noticed another thing: `CreateInteractionResponseData::add_int_choice` has an i32 parameter even though you're supposed to be able to pass values from -2^53 to 2^53. It should be i64

Th3-M4jor:
Requested changes as discussed in the Discord have been implemented, ready for review.

arqunis (approved):

PR #1565, opened 2021-10-31
Remove old version of `time` from dependency tree

kadiwa4:
chrono has a feature called `oldtime`, which adds a dependency to a 0.1 version of the time crate ([chrono docs](https://docs.rs/chrono/0.4.19/chrono/#duration)). serenity does not need it, I ran `cargo test` and `cargo check` and it worked just like before.


arqunis (approved):

arqunis:
Thanks!

Issue #1563, opened 2021-10-27
Adding an `Http::get_webhook_from_url` method


mkrasnitski:
I propose to add a method that allows users to copy-paste the url of their webhook straight from discord, like so:

    http.get_webhook_from_url("https://discord.com/api/webhooks/245037420704169985/ig5AO-wdVWpCBtUUMxmgsWryqgsW3DChbKYOINftJ4DCrUbnkedoYZD0VOH1QLr-S3sV")

There are multiple issues at play here that I feel warrant some discussion. I'll list them below:

1. Should we take the url as is, or deconstruct/parse it and reconstruct it later using existing methods? Since all other methods for getting a webhook only take the relevant parts (the id and/or the token), internally a string allocation is made to create the full webhook url. Passing in a full url at first seems like it would cause unnecessary allocations, however `Url::path_segments` returns a `Split`, meaning it's still only a single allocation. Also, this approach maintains a common api base url.
2. How should parsing be done? Take as an example, how [discord.py](https://github.com/Rapptz/discord.py/blob/29b808d33f02aa1519703f58823de091dc2fe370/discord/webhook/async_.py#L977) does it. They use a regex to match against the expected format of the url, when it is copy-pasted from Discord. Using regex would introduce a new dependency for just this one feature. On the other hand, we could just validate the format using methods on `Url`. This begs the question of how permissive the parsing should be. I would lean towards requiring the url is an actual valid webhook url, rather than just grabbing the id and token from the end.
3. How should we do error handling? Say the user passes in a url that doesn't correspond to a valid webhook. Currently, I don't think `HttpError` has any variants that accurately describe this. `HttpError::Url` takes a `url::ParseError` which applies more specifically to strings that aren't even *valid* urls, so none of its variants apply to our case either. I propose we add a variant to HttpError to communicate when a webhook url is invalid. We could also add this error handling to the other webhook methods of `Http`, since not all `u64`s are valid ids, and not all strings are valid tokens (though those would be considered breaking changes).

mkrasnitski:
If there are no objections, I'll start on a PR for this, and figure it out on the way.

mkrasnitski:
Closed by #1569.

PR #1569, opened 2021-11-06
Add `Http::get_webhook_from_url` method

mkrasnitski:
Implements #1563. Adds a dependency on `regex` if the `http` feature is enabled. Failing to parse a webhook is specific enough that I felt it warranted adding a separate `HttpError` variant.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2878,8 +2878,8 @@ impl Http {
 
     /// Retrieves a webhook given its Id.
     ///
-    /// This method requires authentication, whereas [`Self::get_webhook_with_token`] does
-    /// not.
+    /// This method requires authentication, whereas [`Self::get_webhook_with_token`] and
+    /// [`Self::get_webhook_from_url`] do not.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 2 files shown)


mkrasnitski:
Currently failing clippy CI because of use of unwraps. However, those unwraps will never panic given the circumstances. Should I remove them?

Flat (approved):

kangalio:
From the serenity discord:

> acdenisSK
 — 
Today at 7:04 PM
I agree with gonzo here. if we already included the regex crate, I wouldn't have a problem. however, I prefer not adding crates unless strictly necessary, and to me, adding regex for one function is overkill, especially if it can be done with methods from &str instead

To replace the regex, this code could be used:
```rust
fn parse_webhook_url(url: &url::Url) -> Option<(u64, &str)> {
    let valid_address = ["http", "https"].contains(&url.scheme())
        && ["discord.com", "discordapp.com"].contains(&url.domain()?);
    if !valid_address {
        return None;
    }

    let (webhook_id, token) = url.path().strip_prefix("/api/webhooks/")?.split_once("/")?;
    let webhook_id = webhook_id.parse::<u64>().ok()?;

    Some((webhook_id, token))
}
```
with this invocation:
```rust
let url = url.parse().map_err(|_| HttpError::InvalidWebhook)?;
let (webhook_id, token) = parse_webhook_url(&url).ok_or(HttpError::InvalidWebhook)?;
```

mkrasnitski:
Ah, seems like `split_once` was only stabilized in Rust 1.52, which is newer than the MSRV of 1.51. I'll try and find a workaround.

Flat (approved):

arqunis (approved):

PR #1576, opened 2021-11-12
Add message component example

pascalharp:
This adds an example on how to use and structure message components.

### DIFF
--- a/examples/e17_message_components/src/main.rs
+++ b/examples/e17_message_components/src/main.rs
@@ -175,7 +175,7 @@ struct Handler;
 #[async_trait]
 impl EventHandler for Handler {
     async fn message(&self, ctx: Context, msg: Message) {
-        if !msg.content.eq("animal") {
+        if msg.content != "animal" {
             return;
         }
 
--- a/examples/e17_message_components/src/main.rs
+++ b/examples/e17_message_components/src/main.rs
@@ -62,9 +62,7 @@ impl FromStr for Animal {
             "dog" => Ok(Animal::Dog),
             "horse" => Ok(Animal::Horse),
             "alpaca" => Ok(Animal::Alpaca),
-            _ => Err(ParseComponentError {
-                0: s.to_string(),
-            }),
+            _ => Err(ParseComponentError(s.to_string())),
         }
     }
 }
@@ -167,9 +165,7 @@ impl FromStr for Sound {
             "woof" => Ok(Sound::Woof),
             "neigh" => Ok(Sound::Neigh),
             "hoooooooonk" => Ok(Sound::Honk),
-            _ => Err(ParseComponentError {
-                0: s.to_string(),
-            }),
+            _ => Err(ParseComponentError(s.to_string())),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


examples/e17_message_components/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+}
+
+struct Handler;
+
+#[async_trait]
+impl EventHandler for Handler {
+    async fn message(&self, ctx: Context, msg: Message) {
+        if !msg.content.eq("animal") {
```
kangalio: Why not use `!=`?

examples/e17_message_components/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "cat" => Ok(Animal::Cat),
+            "dog" => Ok(Animal::Dog),
+            "horse" => Ok(Animal::Horse),
+            "alpaca" => Ok(Animal::Alpaca),
+            _ => Err(ParseComponentError {
+                0: s.to_string(),
+            }),
```
arqunis: While valid, it would be better to use the normal tuple constructor syntax.
```suggestion
            _ => Err(ParseComponentError(s.to_string())),
```

examples/e17_message_components/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "meow" => Ok(Sound::Meow),
+            "woof" => Ok(Sound::Woof),
+            "neigh" => Ok(Sound::Neigh),
+            "hoooooooonk" => Ok(Sound::Honk),
+            _ => Err(ParseComponentError {
+                0: s.to_string(),
+            }),
```
arqunis: Ditto.
```suggestion
            _ => Err(ParseComponentError(s.to_string())),
```

arqunis (changes requested):

kangalio (commented):
The builder closures are very spread out... Is that intended? Usually, the builders are filled directly within the closure if possible

pascalharp:
> The builder closures are very spread out... Is that intended? Usually, the builders are filled directly within the closure if possible

Personal preference I guess. I tend to use the same components in multiple different places and this way it is less repetitive. But I am happy to rework it, so that everything is done in the closure, if that is preferred?

kangalio:
> Personal preference I guess. I tend to use the same components in multiple different places and this way it is less repetitive. But I am happy to rework it, so that everything is done in the closure, if that is preferred?

Not sure which style is preferred here, maybe someone more experienced with serenity knows? Personally, I think the static methods could definitely be inlined; the `&self` methods would be a bit trickier to phase out[.](https://pastebin.com/iUXn6pqk)

Anyways, this is not so important anyways, just something I noticed and was curious/confused about. I very much like the example btw, it's fun to play around with in Discord :)

arqunis (approved):

PR #1577, opened 2021-11-13
Correcting Timestamp Code

TheOceanBreeze:
I noticed that the current code for setting a timestamp wouldn't work on Serenity v0.10.9. Thanks to anden3 on Discord, I was able to fix the code by instead making and setting a reference to a variable to get the time.

### DIFF
--- a/examples/e09_create_message_builder/src/main.rs
+++ b/examples/e09_create_message_builder/src/main.rs
@@ -38,7 +38,8 @@ impl EventHandler for Handler {
 
                         // Add a timestamp for the current time
                         // This also accepts a rfc3339 Timestamp
-                        e.timestamp(chrono::Utc::now());
+                        let my_timestamp = chrono::Utc::now();
+                        e.timestamp(&my_timestamp);
 
                         e
                     });


nickelc:
I don't see the issue with the current code.

`From<DateTime<Utc>>` is implemented for `Timestamp`.
https://github.com/serenity-rs/serenity/blob/37e33bd693dfc4e0ca1ff70ff476b4aee571371e/src/builder/create_embed.rs#L493-L502

And compiling the example works for me without an error.
```
$ git checkout v0.10.9
$ cargo check -p e06_create_message_builder
```

arqunis:
> I don't see the issue with the current code.

This is because the issue is already fixed on the `current` branch, but not on `v0.10.9`, where it does not have the `From<DateTime<Tz>>` impl. That is why it's broken for @ReasonablySelenium.

Additionally, the reason the example compiled when you switched to the `v0.10.9` tag is due to [the lack of the `.timestamp(chrono::Utc::now())` line in `v0.10.9`][ex9], which was only introduced in https://github.com/serenity-rs/serenity/commit/b90555404a4334110d52c6fa92b0bde04059c7c4.

I apologise Selenium, but I must close this. The example functions as expected, and the `From<DateTime<Tz>>` impl already exists to allow `.timestamp(chrono::Utc::now())`, so there is no need for fixing anything.

[ex9]: https://github.com/serenity-rs/serenity/blob/v0.10.9/examples/e09_create_message_builder/src/main.rs

nickelc:
oh yeah, i didn't look into the example's code at v0.10.9.
My bad.

TheOceanBreeze:
Me too. Apologies.

PR #1580, opened 2021-11-17
add steps to install cargo-make

kfnx:
imo it's better to make the instruction explicit. Instead of letting newbie (like me) figuring out why the command doesn't work and realize it need to be installed if it hasn't


arqunis (approved):

arqunis:
Thanks!

PR #1586, opened 2021-11-19
Change matches where the match expression is a bool to be more idiomatic

nickelc:
This changes the deserialization pattern of `match map.contains_key(key)` followed by
`map.remove(key).ok_or_else(|| "expected key"). ...` if the key exists.

Before
```rust
let value: Option<X> = match map.contains_key("value") {
    true => map
        .remove("value")
        .ok_or_else(|| DeError::custom("expected value"))
        .and_then(deserialize_value)
        .map_err(DeError::custom)?,
    false => None,
};
```

After
```rust
let value: Option<X> = map
    .remove("value")
    .map(deserialize_value)
    .transpose()
    .map_err(DeError::custom)?;
```

### DIFF
--- a/src/model/interactions/application_command.rs
+++ b/src/model/interactions/application_command.rs
@@ -347,15 +347,11 @@ impl<'de> Deserialize<'de> for ApplicationCommandInteraction {
             .and_then(ApplicationCommandInteractionData::deserialize)
             .map_err(DeError::custom)?;
 
-        let guild_id = match map.contains_key("guild_id") {
-            true => Some(
-                map.remove("guild_id")
-                    .ok_or_else(|| DeError::custom("expected guild_id"))
-                    .and_then(GuildId::deserialize)
-                    .map_err(DeError::custom)?,
-            ),
-            false => None,
-        };
+        let guild_id = map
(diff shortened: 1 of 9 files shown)


src/model/interactions/application_command.rs
```diff
@@ -652,30 +626,22 @@ impl<'de> Deserialize<'de> for ApplicationCommandInteractionDataOption {
...
-                map.remove("value")
-                    .ok_or_else(|| DeError::custom("expected value"))
-                    .and_then(Value::deserialize)
-                    .map_err(DeError::custom)?,
-            ),
-            false => None,
-        };
+        // TODO: Does deserializing a `Value` with `Value::deserialize` make sense?
+        let value =
+            map.remove("value").map(Value::deserialize).transpose().map_err(DeError::custom)?;
```
nickelc: @HarmoGlace wouldn't it be enough to just call `map.remove("value")`
```suggestion
        let value = map.remove("value");
```

Zaitara: Yeah indeed, I do not remember why I wrote this

arqunis (approved):

PR #1587, opened 2021-11-19
Fix #[help] generating help options var name

JohnTheCoolingFan:
`#[help]` macro generated name for static var help options that is not in upper snake case. For example, with function name `my_help` it would generate `MY_HELP__OPTIONS`.


arqunis (approved):

arqunis:
Thanks!

PR #1599, opened 2021-11-24
Move `User`'s discriminator de-/serialization into the `model::user` module

nickelc:
The serde functions are now contained in the `discriminator` and
`discriminator::option` modules and can be used with `#[serde(with)]`.

```
#[derive(Deserialize, Serialize)]
struct User {
    #[serde(with = "discriminator")]
    discriminator: u16,
}

#[derive(Deserialize, Serialize)]
struct PresenceUser {
    #[serde(with = "discriminator::option")]
    discriminator: Option<u16>,
}
```

**BREAKING CHANGE:** The discriminator is now serialized as a string and
aligns with the type in the Discord docs.

### DIFF
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -585,7 +585,7 @@ pub struct PresenceUser {
     pub id: UserId,
     pub avatar: Option<String>,
     pub bot: Option<bool>,
-    #[serde(deserialize_with = "deserialize_opt_u16")]
+    #[serde(with = "discriminator::option")]
     pub discriminator: Option<u16>,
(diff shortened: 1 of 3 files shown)


vicky5124:
About the breaking change, the API provides most integers as strings because of JS numbers being inaccurate at values over 53 bit; discriminator should be kept as an u16

nickelc:
The reasoning for the string snowflakes is clear and their reasoning for the string discriminator is the leading zeroes.

I just think the serialization should be aligned with what is returned by Discord's API.

arqunis (approved):

Issue #1540, opened 2021-10-10
Remove redundant guild ID parameter from several EventHandler methods


kangalio:
The Member parameter of guild_member_addition already contains the guild ID, so the guild ID parameter on guild_member_addition is redundant.

Similarly, the VoiceState parameter of voice_state_update already contains the guild ID, so the guild ID parameter on voice_state_update is redundant.

vicky5124:
Same issue with `guild_role_create`, `guild_role_update` and `guild_stickers_update`, where guild_id is available on `Role` and `Sticker`

PR #1607, opened 2021-12-04
Streamline `EventHandler` by removing redundant `guild_id` parameters

nickelc:
This removes the `guild_id` parameter from `guild_role_create/update`,
`guild_member_addition` and `voice_state_update` which is already provided
by the `Member`, `Role` and `VoiceState` parameters.

In addition to that the `guild_id` field is moved from the
`PresenceUpdateEvent` struct to the `Presence` struct and `Presence` is
passed directly to `EventHandler::presence_update`.

Closes: #1540

https://github.com/serenity-rs/serenity/issues/1540#issuecomment-939504866 mentions `guild_stickers_update` as another candidate but i think that's impractical because the data passed is a hashmap and isn't as accessible as a single entity.

### DIFF
--- a/examples/e11_gateway_intents/src/main.rs
+++ b/examples/e11_gateway_intents/src/main.rs
@@ -3,7 +3,10 @@ use std::env;
 use serenity::{
     async_trait,
     client::bridge::gateway::GatewayIntents,
-    model::{channel::Message, event::PresenceUpdateEvent, gateway::Ready},
+    model::{
+        channel::Message,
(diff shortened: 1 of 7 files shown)


arqunis (approved):

arqunis:
Code is fine, but before I merge this, could you fix the example code for `CreateEmbed::timestamp`? The breaking change to the `guild_member_addition` handler has caused it to fail.

image

nickelc:
I fixed the doc example

PR #1612, opened 2021-12-07
Deserialize the `GuildMembersChunkEvent` struct manually

nickelc:
Instead of using `JsonMap` to inject the `GuildId` to the member data
before deserializing the member list, a serde `Visitor` is implemented to
deserialize event and the `Member`s with the help of an iterim struct.

The correct `GuildId` is assigned afterwards.

### DIFF
--- a/src/model/guild/member.rs
+++ b/src/model/guild/member.rs
@@ -60,6 +60,46 @@ pub struct Member {
     pub avatar: Option<String>,
 }
 
+/// Helper for deserialization without a `GuildId` but then later updated to the correct `GuildId`.
+///
+/// The only difference to `Member` is `#[serde(default)]` on `guild_id`.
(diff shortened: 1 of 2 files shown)


src/model/guild/member.rs
```diff
@@ -60,6 +60,46 @@ pub struct Member {
     pub avatar: Option<String>,
 }
 
+/// Helper for deserialization without a `GuildId` but then later updated to the correct `GuildId`.
+///
+/// The only difference to `Member` is `#[serde(default)]` on `guild_id`.
+#[derive(Deserialize)]
+pub(crate) struct MemberHelper {
```
nickelc: I'm open to suggestions for a better name.

arqunis (approved):

PR #1613, opened 2021-12-07
Expose more json fields for simd

vicky5124:
Exposes the Error typo of both simd-json and serde-json, this is needed for songbird to properly support simd-json

	modified:   src/json.rs


vicky5124:
https://github.com/serenity-rs/serenity/commit/2bb9481cabd1360534fb69e16ce561bbbd665872
Merged

PR #1614, opened 2021-12-10
Move the event deserialization of `Role` creation and update into a helper function

nickelc:
This adds an utility function to deserialize the `GuildRoleCreateEvent` and
`GuildRoleUpdateEvent` structs with the help of interim structs instead of
using `JsonMap`.

### DIFF
--- a/src/model/event.rs
+++ b/src/model/event.rs
@@ -14,7 +14,7 @@ use serde::de::{Error as DeError, IgnoredAny, MapAccess};
 use serde::ser::{Serialize, SerializeSeq, Serializer};
 
 use super::prelude::*;
-use super::utils::{emojis, stickers};
+use super::utils::{emojis, roles, stickers};
 #[cfg(feature = "cache")]
 use crate::cache::{Cache, CacheUpdate};
 use crate::internal::prelude::*;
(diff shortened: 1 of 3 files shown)


kakarot-dev:
um.. if i may ask, whats the disadvantage to change JsonMap to explictly setting the values [here](https://github.com/serenity-rs/serenity/pull/1614/files#diff-22a83da2fe40c751277f8f5dc2c56a8406504021d9869b1a72ae88e311f6107b)

nickelc:
> um.. if i may ask, whats the disadvantage to change JsonMap to explictly setting the values [here](https://github.com/serenity-rs/serenity/pull/1614/files#diff-22a83da2fe40c751277f8f5dc2c56a8406504021d9869b1a72ae88e311f6107b)

Due to serenity's current handling of the gateway events, the passed `deserializer` is a `JsonMap`.
So, the data is deserialized from `JsonMap` to `JsonMap` then `"guild_id"` is allocated and inserted with the `guild_id` into the `"role"` data to be finally deserialized as `Role`.

With this change, the `deserializer` creates the interim structs, the data is moved to the `Role` struct and then the correct `guild_id` is assigned.

arqunis (approved):

PR #1624, opened 2021-12-22
add support for timeout feature

kristopherbullinger:
Hello. Discord released a new feature recently allowing moderators/admins to time out users for a specified amount of time. Timing out a user requires a [new permission](https://discord.com/developers/docs/topics/permissions#permissions) and is part of the [Modify Guild Member](https://discord.com/developers/docs/resources/guild#modify-guild-member) API.

Thank you for your time reading and reviewing this pull request. I am happy to address any issues if you find that there are any.

### DIFF
--- a/src/builder/edit_member.rs
+++ b/src/builder/edit_member.rs
@@ -1,7 +1,10 @@
 use std::collections::HashMap;
+use std::fmt::Display;
 
 use crate::internal::prelude::*;
 use crate::model::id::{ChannelId, RoleId};
+
+use chrono::{DateTime, TimeZone};
 
 /// A builder which edits the properties of a [`Member`], to be used in
 /// conjunction with [`Member::edit`].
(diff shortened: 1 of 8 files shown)


src/client/mod.rs
```diff
@@ -369,7 +369,6 @@ impl<'a> Future for ClientBuilder<'a> {
 
             self.fut = Some(Box::pin(async move {
                 let url = Arc::new(Mutex::new(http.get_gateway().await?.url));
-
```
nickelc: This change is unrelated.

src/builder/edit_member.rs
```diff
@@ -87,4 +90,40 @@ impl EditMember {
...
+        self.0.insert("communication_disabled_until", Value::String(timestamp));
+        self
+    }
+
+    /// Remove a user's timeout, if there is one.
+    ///
+    /// Requires the [Moderate Members] permission.
+    ///
+    /// [Moderate Members]: crate::model::permission::Permissions::MODERATE_MEMBERS
+    pub fn remove_timeout(&mut self) -> &mut Self {
```
GnomedDev: Seems like a weird name, maybe the other methods should take Option to follow the API or this should be renamed to `enable_communication`

GnomedDev (commented):
This PR also doesn't seem to add the new `communication_disabled_until` field to Member, however it may be missing something.

src/client/mod.rs
kristopherbullinger: I will undo this change. My mistake, left over from some debugging during local testing.

src/builder/edit_member.rs
kristopherbullinger: I will update this method name.

kristopherbullinger:
In total, I have:
- Updated `Member` and `PartialMember` to contain a new `communication_disabled_until: Option<DateTime<Utc>>` field.
- Added `disable_communication_until_datetime` and `enable_communication` methods to the `Member` struct.
- Added three new methods to `EditMember`: `disable_communication_until`, `disable_communication_until_datetime`, and `enable_communication`.

I am wondering if it is better for the `EditMember::disable_communication_until_datetime` to accept as a parameter `DateTime<Utc>` rather than `DateTime<Tz> where Tz: TimeZone, Tz::Offset: Display` to simplify the API.

src/model/guild/member.rs
```diff
@@ -92,7 +97,7 @@ impl Member {
                 self.roles.push(role_id);
 
                 Ok(())
-            },
+            }
```
kristopherbullinger: These changes kept getting applied every time I ran RustFmt. Let me know if you want me to undo them.

arqunis: These commas are meant to be there, as [defined by our rustfmt configuration](https://github.com/serenity-rs/serenity/blob/current/rustfmt.toml#L11). Which version of rustfmt are you using? We only support nightly, because we use options not yet supported in stable rustfmt.

kristopherbullinger: I see. I was not using nightly -- I was letting an editor plugin do the formatting for me, which was using the default rustfmt channel. Running `cargo +nightly fmt` resolved the commas and other formatting discrepancies. Thank you.

arqunis (approved):

arqunis:
Thanks!

PR #1634, opened 2021-12-26
Add get_ methods for many ClientBuilder fields

kangalio:
Specifically: application_id, type map, cache update timeout, cache_settings, framework, voice_manager, intents, event_handler, raw_event_handler. These methods are in line with the existing `get_token`

Why: my bot framework for serenity ([poise](https://github.com/kangalioo/poise/)) registers itself as an event handler in a ClientBuilder provided by the user. If the user has registered their own event handlers, the framework needs to store them in order to forward events to it.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -239,6 +239,7 @@ impl<'a> ClientBuilder<'a> {
     pub fn get_cache_settings(&self) -> &CacheSettings {
         // unwrap() is ok because cache_settings will only ever be None in the middle of being
         // .await'ed
+        #[allow(clippy::unwrap_used)]
         self.cache_settings.as_ref().unwrap()
     }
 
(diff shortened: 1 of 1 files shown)


src/client/mod.rs
```diff
@@ -222,6 +234,14 @@ impl<'a> ClientBuilder<'a> {
         self
     }
 
+    /// Gets the cache settings. See [`Self::cache_settings`] for more info.
+    #[cfg(feature = "cache")]
+    pub fn get_cache_settings(&self) -> &CacheSettings {
+        // unwrap() is ok because cache_settings will only ever be None in the middle of being
+        // .await'ed
+        self.cache_settings.as_ref().unwrap()
+    }
```
arqunis: Clippy unfortunately disagrees here, so it must be informed that it is okay.
```suggestion
    /// Gets the cache settings. See [`Self::cache_settings`] for more info.
    #[cfg(feature = "cache")]
    #[allow(clippy::unwrap_used)]
    pub fn get_cache_settings(&self) -> &CacheSettings {
        // unwrap() is ok because cache_settings will only ever be None in the middle of being
        // .await'ed
        self.cache_settings.as_ref().unwrap()
    }
```

arqunis (changes requested):

arqunis (approved):

PR #1638, opened 2021-12-27
Remove manual deserialization of `DiscordJsonError`

nickelc:
The serde attribute `#[serde(deserialize_with = "..")]` is now used to
deserialize the `errors` field.

The `http::utils` contains only internal deserialization functions and
was accidently made public.

**BREAKING CHANGE:** The `http::utils` module is no longer public.

### DIFF
--- a/src/http/error.rs
+++ b/src/http/error.rs
@@ -4,13 +4,11 @@ use std::{
 };
 
 use reqwest::{header::InvalidHeaderValue, Error as ReqwestError, Response, StatusCode, Url};
-use serde::de::{Deserialize, Deserializer, Error as DeError};
 use url::ParseError as UrlError;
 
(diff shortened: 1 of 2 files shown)


src/http/error.rs
```diff
@@ -19,40 +17,10 @@ pub struct DiscordJsonError {
     pub message: String,
     /// The full explained errors with their path in the request
     /// body.
+    #[serde(deserialize_with = "deserialize_errors")]
```
arqunis: This should probably call `Default::default` when the list of errors is absent, as suggested by the `unwrap_or_default` call in the manual deserialisation code.

```suggestion
    #[serde(default, deserialize_with = "deserialize_errors")]
```

arqunis (approved):

PR #1641, opened 2021-12-28
Refactor features combination `builder,unstable_discord_api`

Gentoli:
similar to https://github.com/serenity-rs/serenity/pull/1548, this PR makes `builder,unstable_discord_api` work.
`check --no-default-features --features "builder unstable_discord_api"`

Moves `AttachmentType` from http to model. It is _breaking_ for direct `AttachmentType` imports.

currently introduces this warning. Moving it into the feature flag or adding a `allow` could resolve it, but not sure which is better.

```
warning: associated function is never used: `build_application_command`
   --> ./serenity/src/model/interactions/application_command.rs:906:19
    |
906 |     pub(crate) fn build_application_command<F>(f: F) -> Map<String, Value>
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

```

### DIFF
--- a/src/model/channel/attachment_type.rs
+++ b/src/model/channel/attachment_type.rs
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
-use tokio::fs::File;
 use std::path::{Path, PathBuf};
+
+use tokio::fs::File;
 
 /// Enum that allows a user to pass a [`Path`] or a [`File`] type to [`send_files`]
 ///
--- a/src/model/channel/attachment_type.rs
+++ b/src/model/channel/attachment_type.rs
@@ -1,10 +1,10 @@
 use std::borrow::Cow;
+#[cfg(not(feature = "tokio"))]
+use std::fs::File;
 use std::path::{Path, PathBuf};
 
 #[cfg(feature = "tokio")]
 use tokio::fs::File;
-#[cfg(not(feature = "tokio"))]
-use std::fs::File;
 
 /// Enum that allows a user to pass a [`Path`] or a [`File`] type to [`send_files`]
 ///
--- a/src/model/channel/attachment_type.rs
+++ b/src/model/channel/attachment_type.rs
@@ -1,0 +1,80 @@
+use std::borrow::Cow;
+use tokio::fs::File;
+use std::path::{Path, PathBuf};
+
+/// Enum that allows a user to pass a [`Path`] or a [`File`] type to [`send_files`]
+///
+/// [`send_files`]: crate::model::id::ChannelId::send_files
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub enum AttachmentType<'a> {
+    /// Indicates that the [`AttachmentType`] is a byte slice with a filename.
+    Bytes { data: Cow<'a, [u8]>, filename: String },
+    /// Indicates that the [`AttachmentType`] is a [`File`]
+    File { file: &'a File, filename: String },
+    /// Indicates that the [`AttachmentType`] is a [`Path`]
+    Path(&'a Path),
+    /// Indicates that the [`AttachmentType`] is an image URL.
+    Image(&'a str),
+}
+
+impl<'a> From<(&'a [u8], &str)> for AttachmentType<'a> {
+    fn from(params: (&'a [u8], &str)) -> AttachmentType<'a> {
+        AttachmentType::Bytes {
+            data: Cow::Borrowed(params.0),
+            filename: params.1.to_string(),
+        }
+    }
+}
+
+impl<'a> From<&'a str> for AttachmentType<'a> {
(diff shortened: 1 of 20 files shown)


src/model/interactions/application_command.rs
```diff
@@ -622,6 +626,7 @@ pub struct ApplicationCommandInteractionDataOption {
     pub focused: bool,
 }
 
+#[cfg(all(feature = "unstable_discord_api"))]
```
arqunis:
```suggestion
#[cfg(feature = "unstable_discord_api")]
```

src/model/utils.rs
```diff
@@ -11,7 +11,7 @@ use super::prelude::*;
 use crate::cache::Cache;
 #[cfg(feature = "cache")]
 use crate::internal::prelude::*;
-#[cfg(all(feature = "unstable_discord_api", feature = "model"))]
+#[cfg(all(feature = "unstable_discord_api"))]
```
arqunis:
```suggestion
#[cfg(feature = "unstable_discord_api")]
```

src/model/channel/attachment_type.rs
```diff
@@ -0,0 +1,81 @@
+use std::borrow::Cow;
+use std::fs::File;
+use std::path::{Path, PathBuf};
+
+/// Enum that allows a user to pass a [`Path`] or a [`File`] type to [`send_files`]
+///
+/// [`send_files`]: crate::model::id::ChannelId::send_files
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub enum AttachmentType<'a> {
```
arqunis, line 22: Changing the location of a type is a major breaking change, which is forbidden on the `current` branch. You will have to retarget your PR to the `next` branch.

arqunis (changes requested):

src/model/channel/attachment_type.rs
```diff
@@ -0,0 +1,84 @@
+use std::borrow::Cow;
+#[cfg(not(feature = "tokio"))]
+use std::fs::File;
+use std::path::{Path, PathBuf};
+
+#[cfg(feature = "tokio")]
+use tokio::fs::File;
```
arqunis: This has caused a compilation error in the [multipart code of the HTTP client](https://github.com/serenity-rs/serenity/runs/4662315938?check_suite_focus=true#step:8:143). Personally, this should always be `tokio`'s `File`. I don't see a point branching the types here, considering `tokio` will always be available if the `http` (or `gateway`) feature is enabled, and the HTTP client is the only place where `AttachmentType` is used.

arqunis (changes requested):

src/model/channel/attachment_type.rs
arqunis: This actually made me wonder, why has it been moved to `serenity::model::channel`? How is it related to channels? Does it have to be moved from the `serenity::http` module?

Gentoli: `AttachmentType` is used in various places in `builder` as a member. 
https://github.com/serenity-rs/serenity/blob/4f7dad7c96b0b2111502325090147beacc04b5ce/src/builder/create_interaction_response_followup.rs#L11-L14
The `builder` feature don't select `http` or `gateway`, if the file is included as `File` in `AttachmentType`, it will have to be forked, unless `tokio` is added to `builder` - which wont make sense.
Or we can use something else for `CreateInteractionResponseFollowup`

arqunis: > `AttachmentType` is used in various places in `builder` as a member. 

I apologise, I should've used a better word. When I said "...and the HTTP client is the only place where `AttachmentType` is used." I meant it's the sole location where `AttachmentType` is used to alter behaviour, i.e. *where it is consumed*. If you disabled the `http` feature, it would lose its purpose.

That is why I don't agree moving `AttachmentType` to `serenity::model::channel`. It should stay where it is.

> The `builder` feature don't select `http` or `gateway`, if the file is included as `File` in `AttachmentType`, it will have to be forked, unless `tokio` is added to `builder` - which wont make sense.

True. Ideally, `builder` wouldn't have to depend on either of those, but there is no other elegant way of defining attachments in a message other than `AttachmentType`, and I'd like to avoid introducing another type for attachments just for builders. As a compromise, the builders that use `AttachmentType` should be feature-gated behind `http`. Subsequently, we wouldn't have to choose between `std::fs::File` or `tokio::fs::File` depending on whether `tokio` is present or not.

Gentoli: The reason for this PR is to make the lib usable with other request implementations. In such case, models without the `model` feature and request builders are the only components used. Therefore `AttachmentType` is needed without the `http` feature selected.

arqunis (approved):

PR #1642, opened 2021-12-28
Move GatewayIntents to model

Gentoli:
`GatewayIntents` is a model. moving it allow it to be used without the `client` feature.  

This is a breaking change for `GatewayIntents` imports, maybe we can add a deprecated alias for it.


src/model/gateway.rs
```diff
@@ -728,3 +730,316 @@ pub struct ActivityTimestamps {
...
+/// Once the bot is in 100 guilds or more, [the bot must be verified] in
+/// order to use privileged intents.
+///
+/// [gateway intent]: https://discord.com/developers/docs/topics/gateway#privileged-intents
+/// [Privileged Intents]: https://discord.com/developers/docs/topics/gateway#privileged-intents
+/// [the bot must be verified]: https://support.discord.com/hc/en-us/articles/360040720412-Bot-Verification-and-Data-Whitelisting
+/// [`GatewayIntents::GuildPresences`]: serenity::client::bridge::gateway::GatewayIntents::GUILD_PRESENCES
+/// [`GatewayIntents::GuildMembers`]: serenity::client::bridge::gateway::GatewayIntents::GUILD_MEMBERS
+#[derive(Copy, PartialEq, Eq, Clone, PartialOrd, Ord, Hash)]
+pub struct GatewayIntents {
```
arqunis, line 568: Changing the location of a type is a major breaking change, which is forbidden on the `current` branch. You will have to retarget your PR to the `next` branch.

arqunis (changes requested):

arqunis (approved):

PR #1648, opened 2021-12-31
Refactor examples 03 and 13 to better use the builder pattern

mkrasnitski:
As a follow up to #1636, I refactored examples 03 and 13 in the same manner.


mkrasnitski:
Accidental push of an unrelated commit, sorry.

arqunis (approved):

PR #1649, opened 2021-12-31
Move `parse_webhook_url` to `utils` module

mkrasnitski:
Moved the function and its corresponding test into `utils/mod.rs` since all other internal parsing functions live there. Also changed the signature to elide some unnecessary lifetimes, and for better error handling.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -433,6 +433,20 @@ pub fn parse_quotes(s: impl AsRef<str>) -> Vec<String> {
     args
 }
 
+/// Parses the id and token from a webhook url. Expects a [`reqwest::Url`] object rather than a [`&str`].
+///
+/// # Examples
+///
+/// ```rust
+/// use serenity::utils;
+///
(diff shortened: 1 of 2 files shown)


src/utils/mod.rs
```diff
@@ -432,6 +433,21 @@ pub fn parse_quotes(s: impl AsRef<str>) -> Vec<String> {
     args
 }
 
+pub fn parse_webhook(url: &Url) -> Option<(u64, &str)> {
```
arqunis, line 450: This should be given documentation. For example:
```suggestion
/// Parses the ID and token from a webhook URL. 
pub fn parse_webhook(url: &Url) -> Option<(u64, &str)> {
```
You should probably also add an example.

arqunis (changes requested):

src/utils/mod.rs
mkrasnitski: Done.

arqunis (approved):

Issue #1564, opened 2021-10-29
Implement role icons


ks129:
Discord now supports role icons that can be modified and received using API, so Serenity should support this too.

I would like to do it myself.

arqunis:
Implemented in https://github.com/serenity-rs/serenity/commit/be5cb544b62562b1d66a48ade8a8882b9bcbae3d.

PR #1667, opened 2022-01-17
Add role icons

Rstar284:


Rstar284:
ignore the last 3 commits, git is being bad.

Rstar284:
Closes #1564

drklee3 (commented):
Seems to work when just logging some roles with icons, as shown -- Should also be a good idea to add role icons to the edit role builder.

```rust
        Role {
            ...
            icon: None,
            unicode_emoji: Some(
                "🍣",
            ),
```

```rust
        Role {
            ...
            icon: Some(
                "3bdf3b182270daf9740bbb26d11313c4",
            ),
            unicode_emoji: None,
```

Note about the first 3 commits, you can avoid those by using feature branches instead of your `next` branch and reusing it. Usually I don't touch the target branches in my forks and only use them to base new branches off of. After the feature branch is merged, feature branch is deleted, and upstream target is fetched/pulled to update it to the latest version (then repeat as needed).

Rstar284:
Hmm i dont know how i would add them to the builder. Its confusing tbh

PR #1675, opened 2022-01-19
Change the return type of `Route::stage_instances` to `&'static str`

nickelc:
Functions without parameters should return the static string directly.

**BREAKING CHANGE:** The function `Route::stage_instances` now returns
a `&'static str` instead of a `String`.

### DIFF
--- a/src/http/routing.rs
+++ b/src/http/routing.rs
@@ -1031,8 +1031,8 @@ impl Route {
         api!("/applications/{}/guilds/{}/commands/permissions", application_id, guild_id)
     }
 
-    pub fn stage_instances() -> String {
-        api!("/stage-instances").to_string()
+    pub fn stage_instances() -> &'static str {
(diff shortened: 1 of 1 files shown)


arqunis (approved):

src/http/routing.rs
```diff
@@ -1031,8 +1031,8 @@ impl Route {
         api!("/applications/{}/guilds/{}/commands/permissions", application_id, guild_id)
     }
 
-    pub fn stage_instances() -> String {
-        api!("/stage-instances").to_string()
+    pub fn stage_instances() -> &'static str {
```
Zaitara, line 1034: It is inconsistent with others methods in this file

nickelc: no, it's not. see [Route::gateway](https://docs.rs/serenity/0.10.9/serenity/http/routing/enum.Route.html#method.gateway)

PR #1683, opened 2022-01-21
Replace the use of `DateTime<Utc>` with a newtype struct `Timestamp`

nickelc:
Adds support for using the `time` crate for Discord's timestamp fields
and the builders instead of `chrono`.

The `Timestamp` newtype avoids all the needed `#[cfg(feature)]` and `#[serde(with)]` attributes on every field.

The main difference between `chrono` and `time` is that `time` returns a result while `chrono` might panic or just return a malformed formatted string.

`time` as default implementation would bump the MSRV to 1.53 or if optional, only if it's enabled.

**BREAKING CHANGE:** `DateTime<Utc>` fields and `builder::Timestamp` are
replaced by `model::Timestamp`.

### DIFF
--- a/src/model/timestamp.rs
+++ b/src/model/timestamp.rs
@@ -1,0 +1,184 @@
+use std::fmt;
+use std::ops::Deref;
+
+#[cfg(not(feature = "time"))]
+use chrono::{DateTime, TimeZone, Utc};
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "time")]
+use time::format_description::well_known::Rfc3339;
+#[cfg(feature = "time")]
+use time::serde::rfc3339;
+#[cfg(feature = "time")]
+use time::{Duration, OffsetDateTime};
+
+/// Discord's epoch starts at "2015-01-01T00:00:00+00:00"
+const DISCORD_EPOCH: u64 = 1_420_070_400_000;
+
+/// Representation of a Unix timestamp.
+///
+/// The struct implements the `std::fmt::Display` trait to format the underlying type as
+/// an RFC 3339 date and string such as `2016-04-30T11:18:25.796Z`.
+///
+/// ```
+/// # use serenity::model::id::GuildId;
+/// # use serenity::model::Timestamp;
+/// #
+/// let timestamp: Timestamp = GuildId(175928847299117063).created_at();
+/// assert_eq!(timestamp.unix_timestamp(), 1462015105);
+/// assert_eq!(timestamp.to_string(), "2016-04-30T11:18:25.796Z");
+/// ```
+#[cfg(not(feature = "time"))]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Deserialize, Serialize)]
+#[serde(transparent)]
+pub struct Timestamp(DateTime<Utc>);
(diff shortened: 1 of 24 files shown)


src/model/timestamp.rs
```diff
@@ -0,0 +1,184 @@
...
+
+impl<'a> From<&'a str> for Timestamp {
+    /// Parses an RFC 3339 date and time string such as `2016-04-30T11:18:25.796Z`.
+    ///
+    /// Panics on invalid value.
+    fn from(s: &'a str) -> Self {
+        #[allow(clippy::unwrap_used)]
+        Timestamp::parse(s).unwrap()
+    }
+}
```
arqunis, line 165: Could also offer `TryFrom` variants of these that do not panic upon a `ParseError`. It would make said error type more useful. Currently, it just sits being present in the documentation, but it's nowhere used (i.e. through an interface exposed to the user).

src/model/timestamp.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    fn parse(input: &str) -> Result<Timestamp, ParseError> {
+        DateTime::parse_from_rfc3339(input)
+            .map(|d| Self(d.with_timezone(&Utc)))
+            .map_err(|_| ParseError)
+    }
+}
+
+#[derive(Debug)]
+pub struct ParseError;
```
arqunis, line 102: This should have documentation stating what it represents and when it is created. For example:
```rust
/// Signifies the failure to parse the `Timestamp` from an rfc3339 string.
```

arqunis (changes requested):

src/model/timestamp.rs
nickelc: i didn't really think of adding `TryFrom` because of how the current timestamp type is used in the builders (`Into<Timestamp>`).

arqunis: I see, but it'd be nice to add those. Otherwise, why make `ParseError` public?

nickelc: I didn't make the `timestamp` module public.

arqunis: Ah, you are right. I was only focusing in the scope of this module. You only re-export the `Timestamp` struct from `model/mod.rs`, so users never actually see `ParseError`. In that case, you can ignore this.

nickelc: I tried to add `TryFrom`. I've seen this error before.
```
error[E0119]: conflicting implementations of trait `std::convert::TryFrom<std::string::String>` for type `model::timestamp::Timestamp`
   --> src/model/timestamp.rs:157:1
    |
157 | impl TryFrom<String> for Timestamp {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: conflicting implementation in crate `core`:
            - impl<T, U> TryFrom<U> for T
              where U: Into<T>;
```

src/model/timestamp.rs
nickelc: The `timestamp` module is currently not public but it's good add to it if the module gets exported in the future.

arqunis (approved):

PR #1695, opened 2022-01-24
Update `JOIN_MESSAGES` constant to be up to date

arqunis:
The new list of messages is based on the commit from [discord.py][0]. A lot of the current join messages are not used anymore.

[0]: https://github.com/Rapptz/discord.py/commit/de7c96333ced03fa7b392b2437d0702dc0508890


src/constants.rs
```diff
@@ -34,45 +34,19 @@ pub const USER_AGENT: &str = concat!(
 /// List of messages Discord shows on member join.
 #[allow(clippy::non_ascii_literal)] // allow for discord join messages
```
jelni: This line is no longer needed

PR #1705, opened 2022-01-31
refactor: don't use Arc for reqwest clients

vaporoxx:
Taken from https://docs.rs/reqwest/latest/reqwest/struct.Client.html:

> You do not have to wrap the Client in an Rc or Arc to reuse it, because it already uses an Arc internally.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -54,7 +54,7 @@ use crate::utils;
 /// # }
 /// ```
 pub struct HttpBuilder {
-    client: Option<Arc<Client>>,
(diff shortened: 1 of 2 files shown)


arqunis (changes requested):
Code is fine, but this needs to be retargeted to the `next` branch, since `Arc<Client>` is used in the public API (e.g. `HttpBuilder::client`).

vaporoxx:
Alright, I retargeted and solved conflicts now

arqunis (approved):

Issue #1711, opened 2022-02-03
Make everything in the embed model optional


Chronophylos:
[`model::channel::Embed`](https://github.com/serenity-rs/serenity/blob/current/src/model/channel/embed.rs) requires `kind` while the discord docs mark the field as optional. The same goes for the other embed objects.

PR #1720, opened 2022-02-07
Make fields optional

Chronophylos:
Some fields in the embed model were not optional, while the discord documentation stated otherwise. I changed all of these to an `Option`. In the case of the `inline` field of `EmbedField` I marked it with `#[serde(default)]` to prevent having to add another `Option`. Technically this should've been an `Option` as well.

I sorted some of the structs the same way they were formatted in the discord API documentation.

`EmbedVideo` was missing the `proxy_url` field which I added.

Additionally I updated the description of some of the `url` fields according to the discord API documentation.

Some lines in `create_embed.rs` were automatically formatted with rustfmt.


Chronophylos:
This closes https://github.com/serenity-rs/serenity/issues/1711

vicky5124:
This is a breaking change, which according to the contributing guide, should be based of and merged to the `next` branch rather than `current`

PR #1731, opened 2022-02-10
Derive `Copy` and `Clone` traits for the audit action enums

nickelc:
This change simplifies the `Action::num` method and makes the other
`num` methods unnecessary.

It also includes a fix for the swapped values of `ActionEmoji::Update`
and `ActionEmoji::Delete` and adds tests for all variant values.

BREAKING CHANGE: The `num` method is removed from the sub action enums.
Use `T::Create as u8` instead.

### DIFF
--- a/src/model/guild/audit_log.rs
+++ b/src/model/guild/audit_log.rs
@@ -7,7 +7,7 @@ use crate::internal::prelude::*;
 use crate::model::prelude::*;
 
 /// Determines the action that was done on a target.
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug)]
 #[non_exhaustive]
 pub enum Action {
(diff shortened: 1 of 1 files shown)


src/model/guild/audit_log.rs
```diff
@@ -459,3 +330,62 @@ mod action_handler {
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn action_value() {
+        macro_rules! assert_action {
+            ($action:expr, $num:literal) => {{
+                assert_eq!($action.num(), $num);
+            }};
+        }
```
nickelc, line 344: This macro will be more important in a following PR.

arqunis (approved):

PR #1737, opened 2022-02-14
Add support to set a single child builder for several parent builders

natto1784:
- Implement a simple 'embed' function for 'CreateInteractionResponseData' and 'CreateInteractionResponseFollowup' builders, similar to 'set_embed' for 'CreateMessage' builder
- Implement a simple 'set_action_row' function for 'CreateComponents' builder, similar to 'set_embed' for 'CreateMessage' builder

### DIFF
--- a/src/builder/create_components.rs
+++ b/src/builder/create_components.rs
@@ -32,6 +32,14 @@ impl CreateComponents {
         self
     }
 
+    /// Set a single action row.
+    /// Calling this will overwrite all action rows.
+    pub fn set_action_row(&mut self, mut row: CreateActionRow) -> &mut Self {
+        self.0 = vec![row.build()];
+
+        self
+    }
+
     /// Sets all the action rows.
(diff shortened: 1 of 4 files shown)


natto1784:
Also how likely is it that functions can be renamed to be uniform for similar builders? For eg, the function names differ a lot for `CreateInteractionResponseData` when compared to `CreateMessage`, this can cause confusion sometimes.

arqunis:
> Also how likely is it that functions can be renamed to be uniform for similar builders? For eg, the function names differ a lot for `CreateInteractionResponseData` when compared to `CreateMessage`, this can cause confusion sometimes.

Under normal circumstances, that kind of change would need to be put on the `next` branch. Fortunately, in the case of interactions, they are allowed to be even on the `current` branch, as code marked under the `unstable_discord_api` feature can disregard the prohibition of breaking changes on `current`. If you want, you can make the methods of interaction builders be consistent with builders like `CreateMessage`.

arqunis (approved):

PR #1738, opened 2022-02-15
Add support for API v10

ghost:
Completed:
- [x] Version change
- [x] Add message content intent
- [x] `application.summary` now returns an empty string. This field will be removed in `v11`
- [x] `/channels/:channel_id/threads/active` has been removed
- [x] Existing `attachments` must be specified when `PATCH`ing messages with new attachments. Any attachments not specified will be removed and replaced with the specified list
- [x] Audit log `reason` as a body/query parameter is no longer supported, and you should instead use the `X-Audit-Log-Reason` header.
- [x] Message routes accept `embeds` rather than `embed`
- [x] Requests to `v10` and higher will no longer be supported on `discordapp.com`

### DIFF
--- a/src/builder/edit_message.rs
+++ b/src/builder/edit_message.rs
@@ -6,6 +6,7 @@ use crate::builder::CreateComponents;
 use crate::internal::prelude::*;
 use crate::json::{self, from_number};
 use crate::model::channel::{AttachmentType, MessageFlags};
+use crate::model::id::AttachmentId;
 
 /// A builder to specify the fields to edit in an existing message.
 ///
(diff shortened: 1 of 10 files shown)


ghost:
For the edit builder: how should we specify removing attachments? Should we just assume anything not explicitly specified should be removed? Maybe using something similar to an Iter's filter() method?

evieluvsrainbows:
@acdenisSK what do you think?

evieluvsrainbows (approved):
LGTM but there might be changes denis or other maintainers will request 👍🏼

arqunis (approved):

PR #1744, opened 2022-02-20
Cache result from to_* functions

Milo123459:
This just adds the result of `to_user` and `to_channel` to the cache.


src/model/channel/channel_id.rs
```diff
@@ -414,7 +414,16 @@ impl ChannelId {
             }
         }
 
-        cache_http.http().get_channel(self.0).await
+        let channel = cache_http.http().get_channel(self.0).await?;
+        #[cfg(feature = "cache")]
+        {
+            if let Some(cache) = cache_http.cache() {
+                if let Some(guild_channel) = channel.clone().guild() {
```
kangalio: Could replace this with `if let Channel::Guild(guild_channel) = &channel` and clone only if the if succeeds, to get rid of a useless clone() on non-guild channels. Although performance-wise it won't matter

kangalio (approved):

Milo123459:
~~Waiting for: https://github.com/moka-rs/moka/issues/92 and https://github.com/moka-rs/moka/issues/93~~

kangalio:
> Waiting for: https://github.com/moka-rs/moka/issues/92 and https://github.com/moka-rs/moka/issues/93

Implying that you want to rewrite serenity's whole cache system to moka in this PR?

arqunis:
> Implying that you want to rewrite serenity's whole cache system to moka in this PR?

No, as was discussed on Discord, the intent is to use moka to temporarily cache channels and users retrieved via the HTTP API.

Milo123459:
The `temp_users` field has a TTL of 1 hour, same with `temp_channels`. Do you think we should be caching for longer on channels?

src/cache/mod.rs
```diff
@@ -33,8 +33,11 @@
 use std::collections::{hash_map::Entry, HashMap, HashSet, VecDeque};
 use std::default::Default;
 use std::str::FromStr;
+use std::time::Duration;
 
 use async_trait::async_trait;
+#[cfg(feature = "cache")]
```
kangalio: The entire cache module is gated behind the cache feature so this cfg isn't needed

Cargo.toml
```diff
@@ -126,6 +126,12 @@ features = ["std"]
 [dependencies.percent-encoding]
 version = "2.1"
 
+[dependencies.moka]
+git = "https://github.com/moka-rs/moka"
```
kangalio: Note that you can't publish to crates.io with git dependencies, so this dependency needs to be changed to a crates.io dependency before 0.11 release

Milo123459: Yeah, I know. I'm waiting for [0.8.0 for moka.](https://github.com/moka-rs/moka/milestone/12)

src/cache/mod.rs
```diff
@@ -122,6 +124,10 @@ pub struct Cache {
     /// received and processed by the cache, the relevant channels are also
     /// removed from this map.
     pub(crate) channels: RwLock<HashMap<ChannelId, GuildChannel>>,
+    /// Cache of channels that have been fetched via to_channel.
+    ///
+    /// Each value has a maximum TTL of 1 hour.
+    pub(crate) temp_channels: RwLock<DashCache<ChannelId, GuildChannel>>,
```
tatsuya6502:
```suggestion
    pub(crate) temp_channels: DashCache<ChannelId, GuildChannel>,
```

You do not need to wrap `DashCache` with `RwLock` because `DashCache` is thread-safe and provides interior mutability. After making this change, you will also need to change all `temp_channels.read().await.something()` and `temp_channels.write().await.something()` to `temp_channels.something()` to fix compile errors.

The same things apply to `temp_users`: remove `RwLock` and change `temp_users.read().await` etc.

Milo123459: [`76b983e` (#1744)](https://github.com/serenity-rs/serenity/pull/1744/commits/76b983edf4ca8aa5b7ebb847f75fac2e44827868)

arqunis (approved):
Looks all right to me, but this will stay open till `moka` releases the `v0.8.0` version.

Cargo.toml
```diff
@@ -126,6 +126,12 @@ features = ["std"]
...
+branch = "master"
```
tatsuya6502: [Moka v0.8.0](https://crates.io/crates/moka/0.8.0) has been released.

```suggestion
version = "0.8"
```

src/cache/mod.rs
```diff
@@ -341,6 +351,8 @@ impl Cache {
         if let Some(channel) = self.channels.read().await.get(&id) {
             let channel = channel.clone();
             return Some(Channel::Guild(channel));
+        } else if let Some(channel) = self.temp_channels.get_if_present(&id) {
```
tatsuya6502: I renamed `get_if_present` method to `get`. Sorry for the last minutes change.

```suggestion
        } else if let Some(channel) = self.temp_channels.get(&id) {
```

src/cache/mod.rs
```diff
@@ -878,7 +890,11 @@ impl Cache {
     }
 
     async fn _user(&self, user_id: UserId) -> Option<User> {
-        self.users.read().await.get(&user_id).cloned()
+        if let Some(user) = self.users.read().await.get(&user_id) {
+            Some(user.clone())
+        } else {
+            self.temp_users.get_if_present(&user_id)
```
tatsuya6502:
```suggestion
            self.temp_users.get(&user_id)
```

tatsuya6502 (changes requested):
Hi @Milo123459 — I have published Moka v0.8.0 to crates.io. Before the release, I renamed `get_if_present` method to `get`, so please update your code. Sorry for the last minutes change.

Thanks!

Milo123459:
Thanks for the update!

cc @acdenisSK - should all be ready now!

PR #1758, opened 2022-03-02
Add components support to ExecuteWebhook and EditWebhookMessage

mkrasnitski:
Adds `ExecuteWebhook::{components, set_components}`, as well as `EditWebhookMessage::components`, which all have the same implementation as the corresponding methods on `CreateMessage` and others. Note, per the Discord API these are only valid to call on an application-owned webhook, so I introduced code to verify at runtime whether or not the usage is valid by looking at the webhook's kind.

### DIFF
--- a/src/model/webhook.rs
+++ b/src/model/webhook.rs
@@ -340,7 +340,7 @@ impl Webhook {
         let mut execute_webhook = ExecuteWebhook::default();
         f(&mut execute_webhook);
 
-        let map = json::hashmap_to_json_map(execute_webhook.0);
+        let map = self.verify_map(execute_webhook.0)?;
 
         if !execute_webhook.1.is_empty() {
             http.as_ref()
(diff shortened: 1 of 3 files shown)


src/model/webhook.rs
```diff
@@ -449,6 +449,21 @@ impl Webhook {
         let token = self.token.as_ref().ok_or(ModelError::NoTokenSet)?;
         Ok(format!("https://discord.com/api/webhooks/{}/{}", self.id, token))
     }
+
+    fn verify_map(&self, map: HashMap<&'static str, Value>) -> Result<JsonMap> {
+        // Having components is only valid for application-owned webhooks
+        if map.get("components").is_some() {
```
arqunis:
```suggestion
        if map.contains_key("components") {
```

arqunis (changes requested):

arqunis (approved):

PR #1768, opened 2022-03-16
Make the `model::guild::audit_log` module public

nickelc:
The `audit_log` module contains a lot of types which clutter the
`guild` module but are probably not used by most serenity users.

The types are still re-exported by the `guild` module but hidden from
the documentation.

### DIFF
--- a/src/model/guild/mod.rs
+++ b/src/model/guild/mod.rs
@@ -1,6 +1,6 @@
 //! Models relating to guilds and types that it owns.
 
-mod audit_log;
+pub mod audit_log;
 mod emoji;
 mod guild_id;
 mod guild_preview;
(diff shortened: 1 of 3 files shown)


src/model/guild/mod.rs
```diff
@@ -22,6 +22,8 @@ use tracing::error;
 #[cfg(all(feature = "model", feature = "cache"))]
 use tracing::warn;
 
+#[doc(hidden)]
+#[deprecated(note = "import the types from the `audit_log` module")]
```
nickelc, line 26: This doesn't trigger deprecation warnings because rustc currently ignores deprecated re-exports. It's more like comment here :).

arqunis (approved):

PR #1770, opened 2022-03-16
Enable compilation of builder/interaction code without a backend

FallenWarrior2k:
I have a somewhat odd use case. I wanted to create an Interactions webhook receiver using Cloudflare Workers. Their platform does not support native executables; rather you get a web-ish JS/WASM environment. This means bringing your own networking tools isn't in the cards, as there is no interface to use.

Still, I didn't want to reinvent the wheel and re-type (the parts I needed out of) the Discord API. Instead, I wanted to use the existing types provided by Serenity, even if I couldn't actually use the HTTP or gateway code (I didn't need those for this project). However, I found that selecting just the features I wanted, i.e. `builder` and `unstable_discord_api`, would result in compiler errors throughout the code. I worked through those and added (in my eyes) appropriate feature gates to ensure no change to existing users, while enabling my intended configuration to compile without error.

In future, I would like to perhaps explore the possibility of using Reqwest's support for the Fetch API and a manual rewrite of the WebSocket code to enable the full feature set of Serenity in an environment like that.

### DIFF
--- a/src/model/utils.rs
+++ b/src/model/utils.rs
@@ -11,7 +11,7 @@ use super::prelude::*;
 use crate::cache::Cache;
 #[cfg(feature = "cache")]
 use crate::internal::prelude::*;
-#[cfg(all(feature = "unstable_discord_api", feature = "model"))]
+#[cfg(feature = "unstable_discord_api")]
 use crate::model::interactions::application_command::*;
 
 pub fn default_true() -> bool {
@@ -84,7 +84,7 @@ pub fn deserialize_members<'de, D: Deserializer<'de>>(
     Ok(members)
 }
 
-#[cfg(all(feature = "unstable_discord_api", feature = "model"))]
+#[cfg(feature = "unstable_discord_api")]
 pub fn deserialize_partial_members_map<'de, D: Deserializer<'de>>(
     deserializer: D,
 ) -> StdResult<HashMap<UserId, PartialMember>, D::Error> {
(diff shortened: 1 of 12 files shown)


src/model/utils.rs
```diff
@@ -215,7 +215,7 @@ fn try_resolve(
     }
 }
 
-#[cfg(all(feature = "unstable_discord_api", feature = "model"))]
+#[cfg(all(feature = "unstable_discord_api"))]
```
nickelc:
```suggestion
#[cfg(feature = "unstable_discord_api")]
```

FallenWarrior2k: Ah drat, I thought I'd properly removed all of them. My bad; will fix as soon as I can.

arqunis (approved):

arqunis:
Seems all right, but before I merge, would you mind running `cargo fmt` one more time?

FallenWarrior2k:
From what I saw, everything that `cargo fmt` picked up was unrelated to my changes, so I didn't want to pollute the PR with it. If you want to include it anyway, I can do it.

arqunis:
> was unrelated to my changes

Ah, that may be because you are using the stable toolchain. We depend on features only present in the nightly version of `rustfmt`, which the stable version may ignore and hence produce a lot of changes. You may choose to either run the nightly toolchain or [manually fix the non-compliant code](https://github.com/serenity-rs/serenity/runs/5594605936?check_suite_focus=true), since it is small.

kangalio:
> You may choose to either run the nightly toolchain

`cargo +nightly fmt` by the way

FallenWarrior2k:
Done. Sorry for the delay.

arqunis (approved):

PR #1772, opened 2022-03-16
feat: set `Http::application_id` in ready event handler

vaporoxx:
This PR modifies `Shard` so that it overwrites the HTTP application id when it receives a ready payload. This is useful so that you do not have to save the application id in the environment too instead of just the token. I'm not sure if it is the best implementation but I have tested it and it seems to work.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -276,6 +276,10 @@ impl Http {
         }
     }
 
+    fn try_application_id(&self) -> Result<u64> {
+        self.application_id().ok_or_else(|| HttpError::ApplicationIdMissing.into())
+    }
+
     pub fn set_application_id(&self, application_id: u64) {
         self.application_id.store(application_id, Ordering::Relaxed);
     }
@@ -501,7 +505,7 @@ impl Http {
             multipart: None,
             headers: None,
             route: RouteInfo::CreateFollowupMessage {
-                application_id: self.application_id().ok_or(HttpError::ApplicationIdMissing)?,
+                application_id: self.try_application_id()?,
                 interaction_token,
             },
         })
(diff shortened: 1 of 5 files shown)


src/http/client.rs
```diff
@@ -269,13 +270,29 @@ impl Http {
...
+        let base = Self::new_with_token(token);
 
-        base.application_id = application_id;
+        base.set_application_id(application_id);
 
         base
     }
 
+    #[cfg(feature = "unstable_discord_api")]
+    pub fn application_id(&self) -> Option<u64> {
```
kangalio, line 269: All 25 callers of this function have `.ok_or(HttpError::ApplicationIdMissing)`. Maybe this function should return `Result<u64, HttpError>` directly?

kangalio:
Nice PR, this is pretty useful.

1. Could you fix the compilation errors in the doc tests?
2. This PR makes an Http instance mandatory for using the gateway, not sure if that's ok
3. Serenity has an ApplicationId struct, might be nice to use this here instead of raw u64

src/http/client.rs
vaporoxx: I'm not sure if that makes sense from a user perspective since the method doesn't return any other error variants, maybe a new private method would fit better?

kangalio: Good idea. Name suggestions: `try_application_id`, `application_id_or_err`

vaporoxx:
> This PR makes an Http instance mandatory for using the gateway, not sure if that's ok

That's a reasonable concern (although the docs strongly advise against creating a new shard manually), I'll wait for maintainer feedback on this before updating all the doc tests

> Serenity has an ApplicationId struct, might be nice to use this here instead of raw u64

The `Http` client uses raw numbers everywhere, I think changing it to `ApplicationId` would be too inconsistent

nickelc:
@vaporox I rebased your PR onto the current `next` branch.

kangalio (approved):

arqunis (approved):

PR #1774, opened 2022-03-19
Bump to 0.11.0

arqunis:
This prepares for the release of `0.11.0` by updating the `CHANGELOG.md` file and updating the version in relevant places.

This marks the end of the `v0.10.x` series and the beginning of the `v0.11.x` series. It includes a multitude of breaking changes, small and huge, and lots more of other additions, fixes, and removals.

With this, development on the `current` branch has stalled and won't be accepting additions for `v0.10.x`, instead they will be redirected to `v0.11.0`. After `v0.11.0` is released on https://crates.io, the `current` and `next` branches will wrap around, with `next` becoming the new `current` and `current` the new `next`.

Notable changes include:

- Removal of deprecated and unused things
- Removal of tokio `0.2` compatiblity
- Changing the cache to be synchronous
- Increasing the MSRV to 1.53
- Reorganisation of types into separate modules
- Improving perfomance and quality of deserialise implementations
- Requiring specifying gateway intents
- Making `chrono` an optional dependency (but is still turned on by default)
- Simplifying the `Http` API
- Stabilising application commands (i.e. `unstable_discord_api` is no longer necessary)
- Support for attachments when creating and updating followup messages

The full, unsullied changelog for the release that will be put on Github's releases is:

---
Thanks to the following for their contributions:

- [@AldanTanneo]
- [@AngelOnFira]
- [@AnnikaCodes]
- [@acdenisSK]
- [@almeidx]
- [@Bloectasy]
- [@ben-brook]
- [@Chronophylos]
- [@devtomio]
- [@drklee3]
- [@FallenWarrior2k]
- [@Gentoli]
- [@GnomedDev]
- [@gradiuscypher]
- [@HarmoGlace]
- [@hybras]
- [@JellyWX]
- [@kangalioo]
- [@Lakelezz]
- [@Licenser]
- [@MathyouMB]
- [@Milo123459]
- [@mkrasnitski]
- [@natto1784]
- [@nickelc]
- [@pascalharp]
- [@rasm47]
- [@Rstar284]
- [@Some-Dood]
- [@squili]
- [@TheBlackfurGuy]
- [@tylerd008]
- [@vaporox]
- [@vicky5124]
- [@xMAC94x]
- [@xfix]

### Added

- [ci] Add build job to test the `native-tls` backend ([@nickelc]) [c:440f0fa]
- [model] Add new command data field `guild_id` ([@nickelc]) [c:4dbf3d0]
[...]

PR #1778, opened 2022-03-20
Update serde_json for -Z minimal-versions builds

sugar700:
serenity has incorrect serde version bounds. Those aren't being caught to criterion's higher `serde-json` version requirement, however if the following lines were to be removed from `voice-model/Cargo.toml` the problem would be immediately apparent when using `cargo check`.

```toml
[dev-dependencies]
criterion = "0.3"
serde_test = "1"
```

This matters because crates that use `serenity` don't use its `dev-dependencies`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ members = ["examples/*"]
 
 [dependencies]
 bitflags = "1.1"
-serde_json = "1"
+serde_json = "1.0.25"
 async-trait = "0.1.9"
 
 [dependencies.tracing]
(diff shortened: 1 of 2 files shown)


nickelc:
This was already partial done in the `next` branch (#1669, #1222) and the `next` is about to get merged into `current`.

The best would be if you could change the base of the PR to the `next` branch and bump `serde_json` in the voice-model to 1.0.75.

sugar700:
Oh, fair enough, I didn't notice #1774. Closing.

PR #1780, opened 2022-03-20
Remove tokio 0.2 compatibility

sugar700:
Whether it's 0.11, 0.12 or 1.0, I think it makes sense to drop compatibility for an old version of tokio. At this point pretty much nobody supports tokio 0.2, and there are a few security vulnerabilities when using tokio 0.2: https://rustsec.org/advisories/RUSTSEC-2021-0124, https://rustsec.org/advisories/RUSTSEC-2021-0079, https://rustsec.org/advisories/RUSTSEC-2021-0078.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,6 @@ jobs:
           - no gateway
           - time
           - unstable Discord API features
-          - rustls tokio 0.2
-          - native-tls tokio 0.2
           - simd-json
 
         include:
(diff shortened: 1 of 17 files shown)


arqunis (approved):

arqunis:
Thanks. I've already been considering removing it; it has overstayed its welcome for too long. It might inconvenience people who might still be depending on `0.2` (usually because of another dependency demanding `0.2`), but we have to move on. It's been a year since they last released a version for `0.2.x`.

PR #1689, opened 2022-01-22
Implement missing fields, methods, endpoints, and other values that are documented

vicky5124:
Implements all the stuff that were documented on the discord API docs but not implemented in serenity.

The doc commits that are being implemented in this PR are:
https://github.com/discord/discord-api-docs/commit/53ee0109b874bde5ec2316379bf266bb332630a0
https://github.com/discord/discord-api-docs/commit/95816a58aec3b04798e77079dee2b271d243ecdd
https://github.com/discord/discord-api-docs/commit/30cc2ea5efc83e2398c72d2a47f5aeccd5e21a84
https://github.com/discord/discord-api-docs/commit/9114a1bcb882619f1c42dd1c577f22a80e82b125
https://github.com/discord/discord-api-docs/commit/7dbe9c760513daa2d3f88d52ba650a97ccb3daf9
https://github.com/discord/discord-api-docs/commit/d61166446ec954d2b8f561db0d87701df5de69d9
https://github.com/discord/discord-api-docs/commit/0f09124859874a48a44dadfea8197bfc7ff8b2fe
https://github.com/discord/discord-api-docs/commit/7ab5731e97262bcfdeb9278f8add0d555b98478d
https://github.com/discord/discord-api-docs/commit/ae09c938d05bd0ad2b5f2359ec3a08736f9a3b2c
https://github.com/discord/discord-api-docs/commit/aa5022364275072f775d162777cd5d3c348957ff
https://github.com/discord/discord-api-docs/commit/7684d7953bcdc50dd8c5ceffbd77c35b7d8a7fd5
https://github.com/discord/discord-api-docs/commit/52f1824451df49ad23bed19a706cbe124c7e7550
https://github.com/discord/discord-api-docs/commit/77ee0f906a3d9d2aff0bb3c285913e9db25d98dd

The commits that are not implemented in this PR and why are:
https://github.com/discord/discord-api-docs/commit/84f3290eb8155dac0a479e748f524542f2eda882 - Feature too big, should get its own PR
https://github.com/discord/discord-api-docs/commit/a80c210df2fba79ab95458d82f68633db7aaead6 - Requires previous point to be implemented.
https://github.com/discord/discord-api-docs/commit/ab9e3855ca2346075ee4d982a661cfa69a547260 - I have no idea how serenity handles this.

	modified:   src/builder/create_components.rs
	modified:   src/builder/create_embed.rs
	modified:   src/builder/create_message.rs
	modified:   src/builder/edit_message.rs
[...]

PR #1781, opened 2022-03-20
fix(GatewayIntents): add missing stuff

vaporoxx:
- Brings back `GatewayIntents::GUILD_SCHEDULED_EVENTS` which was added to current in #1689 but not to next due to the struct being moved
- Adds `MESSAGE_CONTENT` to the privileged intents
- Adds missing methods for `MESSAGE_CONTENT` and `GUILD_SCHEDULED_EVENTS`

### DIFF
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -685,11 +685,11 @@ bitflags! {
 
 #[cfg(feature = "model")]
 impl GatewayIntents {
-    /// Gets all of the intents that don't are considered privileged by Discord.
+    /// Gets all of the intents that aren't considered privileged by Discord.
     pub const fn non_privileged() -> GatewayIntents {
         // bitflags don't support const evaluation. Workaround.
         // See: https://github.com/bitflags/bitflags/issues/180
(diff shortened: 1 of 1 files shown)


src/model/gateway.rs
```diff
@@ -689,15 +697,15 @@ impl GatewayIntents {
...
     }
 
     /// Checks if any of the included intents are privileged
     ///
     /// [GUILD_MEMBERS]: #associatedconstant.GUILD_MEMBERS
     /// [GUILD_PRESENCES]: #associatedconstant.GUILD_PRESENCES
     pub fn is_privileged(self) -> bool {
-        self.guild_members() || self.guild_presences()
+        self.guild_members() || self.guild_presences() || self.message_content()
     }
```
kangalio, line 706: `Self::privileged().contains(self)` may be better so we don't have to keep is_privileged in sync with privileged

kangalio (approved):

src/model/gateway.rs
vaporoxx: Included that in the latest commit along with some other changes that seemed too small to open a separate PR for

PR #1795, opened 2022-03-30
Use `str::split_once` for parsing webhooks and user tags

mkrasnitski:
Now that Serenity's MSRV is on 1.53, we can replace some parsing logic with `str::split_once` for better maintainability/readability. There was some discussion about maybe reverting the MSRV back to 1.51, but I thought I'd submit this now since that discussion seems to have stalled.


mkrasnitski:
Rebased onto next.

kangalio (approved):

arqunis (approved):

PR #1804, opened 2022-03-31
Make chrono a default optional feature

sugar700:
chrono is insecure. In #1683, time was introduced as a feature to avoid increasing MSRV. Now that MSRV was increased this is no longer a concern.

This avoids `cargo audit` complaining about `chrono` dependency. Even when `time` feature was enabled `chrono` was still in the dependency tree as it wasn't an optional feature.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,8 @@ jobs:
             features: model http rustls_backend
           - name: time
             features: time
+          - name: chrono
+            features: chrono
           - name: unstable Discord API features
             features: default unstable_discord_api
             dont-test: true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,14 +54,15 @@ version = "0.13"
 optional = true
 version = "1.0.5"
 
-[dependencies.time]
-optional = true
+[dependencies.dep_time]
+package = "time"
 version = "0.3.6"
 features = ["formatting", "parsing", "serde-well-known"]
 
 [dependencies.chrono]
 default-features = false
 features = ["clock", "serde"]
+optional = true
 version = "0.4.10"
 
 [dependencies.flate2]
@@ -131,6 +132,9 @@ optional = true
 version = "0.12"
 optional = true
 
+[dependencies.cfg-if]
+version = "1.0.0"
+
 [dev-dependencies.http_crate]
 version = "0.2"
 package = "http"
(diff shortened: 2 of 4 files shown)


nickelc:
> chrono is insecure.
 
That's debatable. https://github.com/chronotope/chrono/issues/602#issuecomment-1075915577

Additional to that, Serenity doesn't use any effected `Local` timezone code from `chrono`.

> In #1683, time was introduced as a feature to avoid increasing MSRV. Now that MSRV was increased this is no longer a concern.

The `Timestamp` newtype was introduced with the `time` feature to let the users of serenity decide which date & time crate they wanna use. The higher MSRV with `time` was not really a concern but rather a side note. `chrono` is the current default because of it's previous usage in serenity.

kangalio:
Removing chrono would be a big breaking change. Chrono was the only way to interact with datetime in serenity until recently, so the vast majority of bots are presumably using chrono.

Chrono also recently got a new maintainer, so it's not a dead package. As long as chrono is still maintained and popular, I'd say we should keep support for it

sugar700:
Understandable. What about making chrono a default optional feature so that people who don't want `cargo audit` to report security vulnerabilities could remove it?

kangalio:
Good idea (not sure why this wasn't done before to be honest)

nickelc:
> Good idea (not sure why this wasn't done before to be honest)

The datetime fields were not the only places where `chrono` was used by serenity. See #1723.

Milo123459:
Me and the other maintainer of chrono are *almost* ready to publish a new version of Chrono without breaking changes, fixing the vulnerability. Once we release, I'm not sure this PR would have a real purpose, because chrono would be 100% secure on a stable release.

kangalio:
> Me and the other maintainer of chrono are almost ready to publish a new version of Chrono without breaking changes, fixing the vulnerability.

Nice!

This PR would still be useful for people who prefer time over chrono for other reasons, like interoperability with other crates that use time. Those people don't need chrono in their dependency tree

sugar700:
The way this PR is written it should be possible to change back to `chrono` as a default with no features enabled without it being a breaking change. The particular implementation that is used for `Timestamp` is only exposed when you enable any of the features. Once `chrono` fixes the vulnerability it should be fine to change back to `chrono` as a default implementation when none of the features are enabled.

Milo123459:
When we publish a new version, Chrono should be the default. It makes sense.

arqunis (approved):
Looks good. Your changes reduce the impact, but I marked them as breaking regardless. Someone might have `default-features = false` and be left with no `Timestamp` type if they don't enable either `chrono` or `time`.

PR #1813, opened 2022-04-02
Place temporary caching functionality under a separate feature flag

arqunis:
`moka` pulls in a lot of transitive dependencies, which may extend build times by a big margin. Furthermore, not everyone might utilise temporary caching. As such, to kill two birds with one stone, temporary caching is put behind the `tempcaching` feature (this name is subject to bikeshedding).

**BREAKING CHANGE**: Users who desire temporary caching must now enable `tempcaching` for Serenity, otherwise their code might not work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -160,7 +160,7 @@ default_no_backend = [
 ]
 
 builder = ["utils"]
-cache = ["moka", "dashmap", "parking_lot"]
+cache = ["dashmap", "parking_lot"]
 collector = ["gateway", "model"]
 client = ["http", "typemap_rev"]
 extras = []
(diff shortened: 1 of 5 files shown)


Milo123459:
Whilst I agree, Moka does bring along a lot of dependencies I don't think this is the right way of dealing with the issue. This means that there is no caching on `to_user` & `to_channel` which is a massive downside in my opinion.

anden3:
It can be, they just have to enable the flag, whilst people who do don't need it don't have to enable the feature.

Milo123459:
~~What about including this in the cache feature flag?~~

Nevermind, looks good now. However, this should maybe be added in the doc comments to the functions.

kangalio:
> `tempcaching` feature (this name is subject to bikeshedding)

I suggest `temp_cache` instead. "cache" is consistent with the `cache` feature. And serenity's features use snake_case instead of flatcase (except `voice-model` and `simdjson` for some reason)

Milo123459 (approved):

PR #1817, opened 2022-04-03
Make it required to set gateway intents explicitly

kangalio:
As decided democratically in the serenity server...
Screenshot_20220403_120102
https://discord.com/channels/381880193251409931/673965002805477386/957727492880343101

...this PR changes ClientBuilder to ~~panic when gateway intents are not provided~~ take not just token but also gateway intents as required parameters. Examples were updated accordingly, I hope I didn't miss any

I also removed the explicit application ID settings from some examples in favor of #1772. We need #1772 anyways because otherwise almost all examples wouldn't run anymore (due to stabilization of application commands).

### DIFF
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -237,7 +237,7 @@ impl CreateEmbed {
     ///     }
     /// }
     ///
-    /// let mut client = Client::builder("token").event_handler(Handler).await?;
+    /// let mut client = Client::builder("token", Default::default()).event_handler(Handler).await?;
     ///
     /// client.start().await?;
     /// #     Ok(())
@@ -285,7 +285,7 @@ impl CreateEmbed {
     ///     }
     /// }
     ///
-    /// let mut client = Client::builder("token").event_handler(Handler).await?;
+    /// let mut client = Client::builder("token", Default::default()).event_handler(Handler).await?;
     ///
     /// client.start().await?;
     /// #     Ok(())
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -237,7 +237,8 @@ impl CreateEmbed {
     ///     }
     /// }
     ///
-    /// let mut client = Client::builder("token", Default::default()).event_handler(Handler).await?;
+    /// let mut client =
+    ///     Client::builder("token", GatewayIntents::default()).event_handler(Handler).await?;
     ///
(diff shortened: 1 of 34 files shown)


nickelc:
If the intents are required then they should be passed as parameter to the `Client::builder` method.
This also makes the panic unnecessary.

kangalio:
I see the point, but on the other hand:

1. other fields use the panicky approach too (`framework`, `application_id`)

2. the gateway intents expression is usually long and would make formatting kinda ugly (all builder methods would be indented to base level by rustfmt which looks off)

3. existing bots that provide `.intents()` properly have some needless churn

Milo123459:
Just to check, will the following snippet panic?
```rs
.intents(GatewayIntents::empty())
```

kangalio:
No, it will set `intents` to `Some(GatewayIntents::empty())`. The e14_slash_commands example actually does this exact thing

nickelc:
> I see the point, but on the other hand:
> 
> 1. other fields use the panicky approach too (`framework`, `application_id`)
> 
> 2. the gateway intents expression is usually long and would make formatting kinda ugly (all builder methods would be indented to base level by rustfmt which looks off)
> 
> 3. existing bots that provide `.intents()` properly have some needless churn

1. `application_id` panic is gone, `framework` is special because of all the code that gets activated and expects a framework instance instead of handling it with an enum (`Set/No-op`,`Some`/`None`). Setting the gateway intents is required in all cases.
2. The intents can be initialized separately before.
3. Some churn is necessary to make the API more explicit and serenity's release schedule is relative conservative with releasing major breaking changes not more than once year.

kangalio:
You convinced me

src/builder/create_embed.rs
```diff
@@ -237,7 +237,7 @@ impl CreateEmbed {
     ///     }
     /// }
     ///
-    /// let mut client = Client::builder("token").event_handler(Handler).await?;
+    /// let mut client = Client::builder("token", Default::default()).event_handler(Handler).await?;
```
nickelc: It would be better to use `GatewayIntents` here to make the required parameter type clear.
```suggestion
    /// let mut client = Client::builder("token", GatewayIntents::default()).event_handler(Handler).await?;
```

nickelc: Thank you

arqunis (approved):

PR #1823, opened 2022-04-05
Change matches where the match expression is a bool to be more idiomatic

nickelc:
This changes the deserialization pattern of `match map.contains_key(key)` followed by
`map.remove(key).ok_or_else(|| "expected key"). ...` if the key exists.


nickelc:
Yes, i did this before. #1586

arqunis (approved):

PR #1828, opened 2022-04-07
Fix some clippy warnings

Milo123459:


nickelc:
Btw, these warnings are already fixed in the `next` branch. (#1606, #1696)

Milo123459:
Oh. Is it worth keeping this open?

PR #1832, opened 2022-04-08
Fix serenity not building with `tokio_unstable`

### LINKED ISSUE #1829: serenity fails to build with --cfg tokio_unstable
If I setup a .cargo/config.toml with:
```toml
[target.x86_64-unknown-linux-gnu]
rustflags = ["--cfg", "tokio_unstable"]
```

Then running `cargo build` on the latest commit on current branch gives:
```
error[E0433]: failed to resolve: could not find `Builder` in `task`
 --> src/internal/tokio.rs:9:18
  |
9 |     tokio::task::Builder::new().name(&*format!("serenity::{}", name)).spawn(future)
  |                  ^^^^^^^ not found in `tokio::task`
  |
help: consider importing one of these items
  |
1 | use std::thread::Builder;
  |
1 | use tokio::runtime::Builder;
  |
```

Sounds like we need to move this into a separate feature that enables the required tokio feature rather than directly detecting the tokio_unstable flag.

Milo123459:
Fixes #1829

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,7 +108,7 @@ version = "0.5"
 version = "1"
 default-features = true
 optional = true
-features = ["fs", "macros", "rt", "sync", "time", "tracing"]
+features = ["fs", "macros", "rt", "sync", "time"]
 
 # Tokio v0.2
 [dependencies.tokio_compat]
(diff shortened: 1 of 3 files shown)


nickelc:
We also have to add the feature to the cfg guards of the `internal::tokio::spawn_named` function.

Milo123459:
Added

PR #1839, opened 2022-04-10
Remove redundant `ClientBuilder::application_id` field

nickelc:
The field was used to return the ID as `ApplicationId` but it can also be
retrieved from `Http` and mapped to `ApplicationId`.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -74,7 +74,6 @@ pub struct ClientBuilder {
     http: Option<Http>,
     fut: Option<BoxFuture<'static, Result<Client>>>,
     intents: GatewayIntents,
-    application_id: Option<ApplicationId>,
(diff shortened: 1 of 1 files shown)


src/client/mod.rs
```diff
@@ -143,8 +141,6 @@ impl ClientBuilder {
 
     /// Sets the application id.
     pub fn application_id(mut self, application_id: u64) -> Self {
-        self.application_id = Some(ApplicationId(application_id));
-
         if let Some(http) = &mut self.http {
```
vaporoxx: `self` doesn't need to be mutable anymore now:

```suggestion
    pub fn application_id(self, application_id: u64) -> Self {
        if let Some(http) = &self.http {
```

arqunis (approved):

PR #1843, opened 2022-04-10
Revert #1759

mkrasnitski:
Now that #1811 has been merged, serenity allows message components to be used with regular webhooks, so the documentation should be reverted back to its previous state.


GnomedDev:
Keep in mind they don't show up unless it's application owned OR the edgecase that the webhook is owned by the bot account who sent it, that should probably be documented

mkrasnitski:
Ah nevermind then. I'll open a new PR.

PR #1855, opened 2022-04-16
Fix a docstring in `http/routing.rs`

mkrasnitski:


src/http/routing.rs
```diff
@@ -43,15 +43,15 @@ pub enum Route {
     /// The data is the relevant [`ChannelId`].
     ///
     /// [`ChannelId`]: crate::model::id::ChannelId
```
nickelc: Can you move this line above the `[Rate Limits]:` link?

mkrasnitski: Fixed

arqunis (approved):

PR #1871, opened 2022-04-25
Improve binary message deserialization time by optimizing convert_ws_message

MelonShooter:
I did some profiling earlier using callgrind then qcachegrind to analyze the output on a small program I wrote that mimics the startup of a bot. To test this, I used my own bot which has all privileged intents and the cache enabled and is in several servers with a few 100 people along with one server that has over 69,000 people.  Using this, I realized that a massive chunk of startup cost (a little over half by instruction count) was from the decompression of several rather large zlib-compressed binary payloads sent over the gateway for the Guild Create event. This occurred because the current implementation of ``convert_ws_message`` (assuming the simdjson feature is not enabled) uses serde's ``from_reader`` function. The reader passed into the function is a reader returned by flate2 to decompress the payload. However, ``from_reader`` appears to call the underlying reader's read function way more times than is necessary (10s of millions of times in my bot's case), which makes binary message deserialization within this function incredibly slow. ~~Unfortunately, there doesn't appear to be a way to reduce the amount of calls being made to the read function while still using ``from_reader``~~ (See other comments below), so this PR makes it so that ``convert_ws_message`` allocates a String and reads the reader into it. Then, instead of using ``from_reader``, ``from_str`` is used. Because this was the only usage of ``from_reader ``, this PR also removes that function, which is pub(crate). The callgrind outputs showed a reduction in instruction count by a factor of roughly 3 for the entirety of the aforementioned program as a result of this optimization.

If anyone is curious about the program I used to mimic startup, here it is:
``` rust
use std::sync::Arc;

use once_cell::sync::OnceCell;
use serenity::client::bridge::gateway::ShardManager;
use serenity::client::{Context, EventHandler};
use serenity::framework::StandardFramework;
use serenity::model::id::GuildId;
use serenity::prelude::{GatewayIntents, Mutex};
use serenity::{async_trait, Client};
use simplelog::{ConfigBuilder, LevelFilter, SimpleLogger};

struct TestEventHandler;

static SHARD_MANAGER: OnceCell<Arc<Mutex<ShardManager>>> = OnceCell::new();

#[async_trait]
impl EventHandler for TestEventHandler {
    async fn cache_ready(&self, _ctx: Context, _guilds: Vec<GuildId>) {
        let mut lock = SHARD_MANAGER.get().unwrap().lock().await;

        lock.shutdown_all().await;
    }
}

#[tokio::main]
async fn main() -> serenity::Result<()> {
    let token = include_str!(".env");
    let framework = StandardFramework::new();
    let mut client = Client::builder(token, GatewayIntents::all())
        .framework(framework)
        .event_handler(TestEventHandler)
        .await
        .expect("Couldn't build client.");

    SHARD_MANAGER
        .try_insert(client.shard_manager.clone())
        .unwrap();

    SimpleLogger::init(LevelFilter::Warn, ConfigBuilder::new().build()).unwrap();

    client.start().await
}
```

I've also attached the callgrind outputs when I ran it using my bot. Both callgrind outputs take into account the changes made in #1870 even though it hasn't been merged yet.  

Before this change: [callgrind.out-before.txt](https://github.com/serenity-rs/serenity/files/8551436/callgrind.out-before.txt)

After this change: [callgrind.out-after.txt](https://github.com/serenity-rs/serenity/files/8551438/callgrind.out-after.txt)

Final notes: This will likely increase the peak memory usage of ``serenity`` on startup because decompressed websocket binary messages are allocated into a temporary String. For a Create Guild event when all privileged intents are enabled, which should be the worst case scenario in terms of payload size, this could be up to roughly 30 MBs for a guild that is near the capacity before guild data must be chunked.

### DIFF
--- a/src/internal/ws_impl.rs
+++ b/src/internal/ws_impl.rs
@@ -1,3 +1,5 @@
+use std::io::Read;
+
 use async_trait::async_trait;
 use async_tungstenite::tungstenite::Message;
 use flate2::read::ZlibDecoder;
@@ -9,7 +11,7 @@ use url::Url;
 
 use crate::gateway::{GatewayError, WsStream};
 use crate::internal::prelude::*;
-use crate::json::{from_reader, from_str, to_string};
+use crate::json::{from_str, to_string};
 
 #[async_trait]
 pub trait ReceiverExt {
@@ -57,9 +59,19 @@ impl SenderExt for WsStream {
 
 #[inline]
 pub(crate) fn convert_ws_message(message: Option<Message>) -> Result<Option<Value>> {
+    const DECOMPRESSION_MULTIPLIER: usize = 3;
+
     Ok(match message {
         Some(Message::Binary(bytes)) => {
-            from_reader(ZlibDecoder::new(&bytes[..])).map(Some).map_err(|why| {
+            let mut decompressed = String::with_capacity(bytes.len() * DECOMPRESSION_MULTIPLIER);
+
+            ZlibDecoder::new(&bytes[..]).read_to_string(&mut decompressed).map_err(|why| {
+                warn!("Err decompressing bytes: {:?}; bytes: {:?}", why, bytes);
+
+                why
+            })?;
+
+            from_str(decompressed.as_mut_str()).map(Some).map_err(|why| {
                 warn!("Err deserializing bytes: {:?}; bytes: {:?}", why, bytes);
 
                 why
--- a/src/json.rs
+++ b/src/json.rs
@@ -93,24 +93,6 @@ where
     Ok(simd_json::from_str(s)?)
 }
 
-#[cfg(all(feature = "gateway", not(feature = "simd-json")))]
-pub(crate) fn from_reader<R, T>(r: R) -> Result<T>
-where
-    T: DeserializeOwned,
-    R: std::io::Read,
-{
-    Ok(serde_json::from_reader(r)?)
-}
-
-#[cfg(all(feature = "gateway", feature = "simd-json"))]
-pub(crate) fn from_reader<R, T>(r: R) -> Result<T>
-where
-    T: DeserializeOwned,
-    R: std::io::Read,
-{
-    Ok(simd_json::from_reader(r)?)
-}
-
 #[cfg(not(feature = "simd-json"))]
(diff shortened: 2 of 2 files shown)


MelonShooter:
On second thought, there may be a better way to do this by using a buffered reader on top of flate2's reader instead of writing it to a String, which I'm going to try right now. I had initially tried this out by trying to use a flate2::bufread::ZlibDecoder instead of a flate2::read::ZlibDecoder and performance didn't improve at all, so I instead opted for using a String.  The issue with using Strings is that memory usage within this function may be significant even though it's deallocated when the function exits. Initially, I had for some reason assumed that the bufread::ZlibDecoder struct would buffer the decompression, but in reality, it just takes in a buffered reader without buffering the decompression.  I'm going to turn this PR to a draft until then...

MelonShooter:
So, it appears that using a buffered reader that has an internal buffer with a capacity of the minimum between the compressed size of the binary message and 1MiB does significantly improve performance, but not nearly as much as using a String. The reason for the performance difference between this and using a String is because of the additional overhead serde has parsing the bytes in the small increments it does over such large payloads as opposed to having a single string that it can parse. I don't believe there's a way to further optimize it so that using a reader is just as efficient as using a String. Using a buffered reader would keep the peak memory usage of this function roughly the same as it was prior to this PR though. However, the peak memory usage might not be an issue considering how this only really is a potential issue when you have all privileged intents enabled with a cache because you're expected to have a good amount of memory in that case. Here's the callgrind output when using the buffered reader described above and the program in my original post.

**ADDITIONAL NOTE**: When the ``simdjson`` feature is enabled, whether a buffered reader is used or a string is used doesn't really matter because ``simd-json``'s implementation of ``from_reader`` will read the entire reader into a vector anyways, resulting in similar performance.

[callgrind.out-after-bufreader.txt](https://github.com/serenity-rs/serenity/files/8551928/callgrind.out-after-bufreader.txt)

Considering there are pros and cons to using a buffered reader or a String when ``simdjson`` is not enabled, I'm going to wait for a response as to see which way is preferred first. For reference, here's a side-by-side comparison of instruction counts of the original implementation (left), the implementation using a buffered reader (middle), and the implementation using a String (right). If using a buffered reader is preferred, I can commit that implementation to this PR instead.

combine_images

arqunis (approved):

PR #1876, opened 2022-04-28
Use `Cow` in `utils/content_safe.rs` to save on allocations

mkrasnitski:
Some unnecesary allocations were being done converting `&'static str` to `String` in `clean_mention`. Changing the return type to `Cow` lets us avoid these. Since content_safe might potentially be a hot code path in some cases, saving on allocations is a good thing.

### DIFF
--- a/src/utils/content_safe.rs
+++ b/src/utils/content_safe.rs
@@ -225,7 +225,7 @@ fn clean_mention(
         },
         Mention::Role(id) => id
             .to_role_cached(&cache)
-            .map_or_else(|| "@deleted-role".into(), |role| format!("@{}", role.name).into()),
+            .map_or(Cow::Borrowed("@deleted-role"), |role| format!("@{}", role.name).into()),
         Mention::User(id) => {
             if let Some(guild_id) = options.guild_reference {
                 if let Some(guild) = cache.guild(&guild_id) {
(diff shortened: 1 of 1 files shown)


src/utils/content_safe.rs
```diff
@@ -211,19 +213,19 @@ fn clean_mention(
...
+                format!("#{}", channel.name).into()
             } else {
-                "#deleted-channel".to_string()
+                "#deleted-channel".into()
             }
         },
         Mention::Role(id) => id
             .to_role_cached(&cache)
-            .map_or_else(|| "@deleted-role".to_string(), |role| format!("@{}", role.name)),
+            .map_or_else(|| "@deleted-role".into(), |role| format!("@{}", role.name).into()),
```
vaporoxx:
```suggestion
            .map_or(Cow::Borrowed("@deleted-role"), |role| format!("@{}", role.name).into()),
```

src/utils/content_safe.rs
```diff
@@ -243,13 +246,14 @@ fn clean_mention(
...
                 }
+                .into()
             };
             cache
                 .user(id)
                 .as_ref()
                 .map(get_username)
                 .or_else(|| users.iter().find(|u| u.id == id).map(get_username))
-                .unwrap_or_else(|| "@invalid-user".to_string())
+                .unwrap_or_else(|| "@invalid-user".into())
```
vaporoxx:
```suggestion
                .unwrap_or(Cow::Borrowed("@invalid-user"))
```

vaporoxx (commented):
You can avoid using the `_else` variants here by using the enum variant directly instead of function calls

arqunis (approved):

PR #1882, opened 2022-05-01
Implement Ord and PartialOrd for Timestamp

### LINKED ISSUE #1881: Implement Ord and PartialOrd for Timestamp
In simple terms, I run a bot that outputs a list containing message timestamps, and the user can specify the sorting of that list, which includes sorting by timestamp. Since the Timestamp type was introduced, that is no longer possible, since Timestamp does not implement Ord and PartialOrd.

I'm not super clear on what the implications are, but adding the following to `src/model/timestamp.rs` has my bot compiling locally again:
```rust
impl PartialOrd for Timestamp {
    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
        Some(self.unix_timestamp().cmp(&other.unix_timestamp()))
    }
}

impl Ord for Timestamp {
    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
        self.unix_timestamp().cmp(&other.unix_timestamp())
    }
}
```

This makes sense to me, since the unix timestamps are i64s and those can sensibly be compared. I will submit a PR with that addition.

tedtramonte:
Closes #1881

### DIFF
--- a/src/model/timestamp.rs
+++ b/src/model/timestamp.rs
@@ -54,7 +54,7 @@ cfg_if::cfg_if! {
         /// assert_eq!(timestamp.unix_timestamp(), 1462015105);
         /// assert_eq!(timestamp.to_string(), "2016-04-30T11:18:25.796Z");
         /// ```
-        #[derive(Debug, Clone, Copy, Eq, PartialEq, Deserialize, Serialize)]
+        #[derive(Debug, Clone, Copy, Eq, PartialEq, Deserialize, Serialize, Ord, PartialOrd)]
         #[serde(transparent)]
         pub struct Timestamp(DateTime<Utc>);
 
@@ -130,7 +130,7 @@ cfg_if::cfg_if! {
         /// assert_eq!(timestamp.unix_timestamp(), 1462015105);
         /// assert_eq!(timestamp.to_string(), "2016-04-30T11:18:25.796Z");
         /// ```
-        #[derive(Debug, Clone, Copy, Eq, PartialEq, Deserialize, Serialize)]
+        #[derive(Debug, Clone, Copy, Eq, PartialEq, Deserialize, Serialize, Ord, PartialOrd)]
         #[serde(transparent)]
         pub struct Timestamp(#[serde(with = "rfc3339")] OffsetDateTime);
 


src/model/timestamp.rs
```diff
@@ -282,6 +282,18 @@ impl From<&Timestamp> for Timestamp {
     }
 }
 
+impl PartialOrd for Timestamp {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.unix_timestamp().cmp(&other.unix_timestamp()))
```
arqunis: Since the implementation is the same, we can implement `PartialOrd` in terms of `Ord`:
```suggestion
        Some(self.cmp(other))
```

arqunis (changes requested):

tedtramonte:
Thanks for the tip. I changed that and squashed, adding you as a co-author.

vaporoxx:
Since both possible inner structs implement `Ord` (see [`chrono`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#impl-Ord) / [`time`](https://docs.rs/time/latest/time/struct.OffsetDateTime.html#impl-Ord)), I believe you should be able to just derive it instad of manually implementing

tedtramonte:
> Since both possible inner structs implement `Ord` (see [`chrono`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#impl-Ord) / [`time`](https://docs.rs/time/latest/time/struct.OffsetDateTime.html#impl-Ord)), I believe you should be able to just derive it instad of manually implementing

I thought the same, but Clippy was SCREAMING at me for even thinking of doing that. That's how I tried it originally. Perhaps I was doing something wrong, though.

nickelc:
> > Since both possible inner structs implement `Ord` (see [`chrono`](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#impl-Ord) / [`time`](https://docs.rs/time/latest/time/struct.OffsetDateTime.html#impl-Ord)), I believe you should be able to just derive it instad of manually implementing
> 
> I thought the same, but Clippy was SCREAMING at me for even thinking of doing that. That's how I tried it originally. Perhaps I was doing something wrong, though.

I don't know. clippy is quiet for me when deriving `PartialOrd` and `Ord`.

Can you post the error message?

tedtramonte:
@nickelc Sorry, not clippy complaining, it's the compiler.

```
error: lifetime may not live long enough
    --> src\model\guild\mod.rs:1634:33
     |
1630 |           &self,
     |           - let's call the lifetime of this reference `'1`
1631 |           prefix: &str,
     |                   - let's call the lifetime of this reference `'2`
...
1634 |       ) -> Vec<(&Member, String)> {
     |  _________________________________^
1635 | |         fn starts_with(prefix: &str, case_sensitive: bool, name: &str) -> bool {
1636 | |             case_sensitive && name.starts_with(prefix)
1637 | |                 || !case_sensitive && starts_with_case_insensitive(name, prefix)
...    |
1666 | |         members
1667 | |     }
     | |_____^ associated function was supposed to return data with lifetime `'1` but it is returning data with lifetime `'2`
```

There's like 7 of these errors. Could be a Windows thing? I'm using the nightly toolchain.

nickelc:
This looks unrelated.

tedtramonte:
I agree, but it compiles with the manual implementations and won't with the derives. If the derives will work, I'll switch to that and force push. Doesn't both me one way or the other!

nickelc:
I can't reproduce the error. windows@nightly build job is green.
https://github.com/serenity-rs/serenity/compare/current...nickelc:test-timestamp
https://github.com/nickelc/serenity/runs/6249536227?check_suite_focus=true

tedtramonte:
Good enough for me. Added the derives and you as a co-author.

arqunis (approved):

nickelc:
> @nickelc Sorry, not clippy complaining, it's the compiler.
> 
> ```
> error: lifetime may not live long enough
>     --> src\model\guild\mod.rs:1634:33
>      |
> 1630 |           &self,
>      |           - let's call the lifetime of this reference `'1`
> 1631 |           prefix: &str,
>      |                   - let's call the lifetime of this reference `'2`
> ...
> 1634 |       ) -> Vec<(&Member, String)> {
>      |  _________________________________^
> 1635 | |         fn starts_with(prefix: &str, case_sensitive: bool, name: &str) -> bool {
> 1636 | |             case_sensitive && name.starts_with(prefix)
> 1637 | |                 || !case_sensitive && starts_with_case_insensitive(name, prefix)
> ...    |
> 1666 | |         members
> 1667 | |     }
>      | |_____^ associated function was supposed to return data with lifetime `'1` but it is returning data with lifetime `'2`
> ```
> 
> There's like 7 of these errors. Could be a Windows thing? I'm using the nightly toolchain.

I now had a similar error in unrelated `standard` framework code after resolving merge conflicts.
The issues was gone after `cargo clean`. So, i think it's an incremental compilation issue.

PR #1884, opened 2022-05-02
Guild Scheduled Events support

### LINKED ISSUE #1620: Discord API - Scheduled Guild Events
Back in October, Discord released a new feature called [Scheduled Guild events](https://support.discord.com/hc/en-us/articles/4409494125719-Scheduled-Events) that allows guilds to schedule meetings so members can gather together, online and offline! As a fast follow they released [documentation over that API](https://discord.com/developers/docs/resources/guild-scheduled-event). It would be great to implement these into serenity (and I'd be happy to take the time to implement them if I can receive feedback around my modeling as it comes)

Primarily it looks like there are about 6 APIs they offer:
1. List Scheduled Events for Guild
2. Create Guild Scheduled Event
3. Get Guild Scheduled Event
4. Modify Guild Scheduled Event
5. Delete Guild Scheduled Event
6. Get Guild Scheduled Event Users

I am not certain exactly what modeling we'd need to do but nothing feels too spicy. We'd definitely need to add more models but I don't think there are any extending of existing models (e.g. users) we'd have to do.

mkrasnitski:
Closes #1620. This is a pretty large PR, so I've broken it into a few sub-commits that should be more digestible. The first one adds initial support, i.e. struct definitions, and also implements the two Get endpoints. After that, I added builders for the Create/Modify endpoints, plus added support for the Delete endpoint. Finally, the last commit adds support for the Get Users endpoint. 

Because of the size of the PR, it's likely I've missed a few mistakes in a few places, either in docs or maybe even in code. Please let me know if anything here doesn't match the [official docs](https://discord.com/developers/docs/resources/guild-scheduled-event).

Speaking of which, the permissions story for Scheduled Events is a complicated one. Depending on the scheduled event's kind, the various endpoints require different combinations of permissions. It's detailed [here](https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-permissions-requirements). Right now I only document that the `MANAGE_EVENTS` permission is required, but should docs be added that spell out the specific requirements, or maybe link to the official docs?


src/model/id.rs
```diff
@@ -122,6 +122,12 @@ pub struct MessageId(#[serde(with = "snowflake")] pub u64);
 )]
 pub struct RoleId(#[serde(with = "snowflake")] pub u64);
 
+/// An identifier for a Role
```
nickelc:
```suggestion
/// An identifier for a scheduled event
```

mkrasnitski: fixed and squashed into the first commit.

mkrasnitski:
GitHub made an ugly commit when resolving the merge conflict so I resolved it manually and force-pushed.

arqunis (approved):

PR #1886, opened 2022-05-02
Make serenity pass the majority of clippy::pedantic lints

GnomedDev:
I was bored.. okay?


src/framework/standard/help_commands.rs
```diff
@@ -298,7 +300,7 @@ async fn check_command_behaviour(
 // their sub-commands, trying to find `name`.
 // Similar commands will be collected into `similar_commands`.
 #[cfg(all(feature = "cache", feature = "http"))]
-#[allow(clippy::too_many_arguments)]
+#[allow(clippy::too_many_arguments, clippy::needless_match)]
```
nickelc: `clippy::needless_match` is only available since rust 1.61

arqunis (approved):

PR #1894, opened 2022-05-05
Split up a few functions to minimize monomorphization-bloat

elkowar:
This PR splits up the large and commonly used functions create_message, edit, create_interaction_response and create_followup_message, to minimize the length of the function that is generic over the closure argument.

In most cases, closures count as unique types, and thus every use of these functions _can_ cause a new copy of that function to be generated during monomorphization.
By moving most of the function body into separate, non-generic functions, we can minimize the impact of this monomorphization-bloat.

In my mid-sized discord bot, this change alone managed to reduce the size of the LLVM-codegen output (via `cargo-llvm-lines`) by about 8%, which _is_ noticable in build times, especially incremental builds.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -326,21 +326,20 @@ impl Message {
     where
         F: for<'b> FnOnce(&'b mut EditMessage<'a>) -> &'b mut EditMessage<'a>,
     {
-        let mut builder = self._prepare_edit_builder(cache_http.cache())?;
-        f(&mut builder);
-        self._send_edit(cache_http.http(), builder).await
-    }
-
-    fn _prepare_edit_builder<'a>(&self, cache: Option<&crate::Arc<Cache>>) -> Result<EditMessage<'a>> {
         #[cfg(feature = "cache")]
         {
-            if let Some(cache) = cache {
+            if let Some(cache) = cache_http.cache() {
                 if self.author.id != cache.current_user_id() {
                     return Err(Error::Model(ModelError::InvalidUser));
                 }
             }
         }
+        let mut builder = self._prepare_edit_builder()?;
+        f(&mut builder);
+        self._send_edit(cache_http.http(), builder).await
+    }
 
+    fn _prepare_edit_builder<'a>(&self) -> Result<EditMessage<'a>> {
         let mut builder = EditMessage::default();
 
         if !self.content.is_empty() {
@@ -369,7 +368,6 @@ impl Message {
             .await?;
         Ok(())
     }
-
 
     pub(crate) fn transform_content(&mut self) {
         match self.kind {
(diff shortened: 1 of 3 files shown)


src/model/channel/message.rs
```diff
@@ -326,9 +326,15 @@ impl Message {
     where
         F: for<'b> FnOnce(&'b mut EditMessage<'a>) -> &'b mut EditMessage<'a>,
     {
+        let mut builder = self._prepare_edit_builder(cache_http.cache())?;
```
```suggestion
        let mut builder = self._prepare_edit_builder()?;
```

src/model/channel/message.rs
```diff
@@ -326,9 +326,15 @@ impl Message {
...
+        f(&mut builder);
+        self._send_edit(cache_http.http(), builder).await
+    }
+
+    fn _prepare_edit_builder<'a>(&self, cache: Option<&crate::Arc<Cache>>) -> Result<EditMessage<'a>> {
```
```suggestion
    fn _prepare_edit_builder<'a>(&self) -> Result<EditMessage<'a>> {
```

src/model/channel/message.rs
```diff
@@ -326,9 +326,15 @@ impl Message {
...
         #[cfg(feature = "cache")]
         {
-            if let Some(cache) = cache_http.cache() {
+            if let Some(cache) = cache {
```
```suggestion
            if let Some(cache) = cache_http.cache() {
```

src/model/channel/message.rs
mkrasnitski: As it stands this requires a feature gate on the `cache` feature, which doesn't work, since the function contains code that isn't gated. I recommend instead putting the `#[cft(feature = "cache")]` block back into `Message::edit`, before calling `_prepare_edit_builder`. I don't think it'll add too much compile-time bloat, though do note that the `cache` feature is enabled by default, so you may want to perform some benchmarks.

mkrasnitski (changes requested):
Disabling the `cache` feature currently breaks what you've added.

src/model/channel/message.rs
elkowar: solved!

src/model/channel/message.rs
```diff
@@ -334,7 +334,12 @@ impl Message {
                 }
             }
         }
+        let mut builder = self._prepare_edit_builder()?;
+        f(&mut builder);
+        self._send_edit(cache_http.http(), builder).await
+    }
 
+    fn _prepare_edit_builder<'a>(&self) -> Result<EditMessage<'a>> {
```
```suggestion
    fn _prepare_edit_builder<'a>(&self) -> EditMessage<'a> {
```

src/model/channel/message.rs
```diff
@@ -347,21 +352,20 @@ impl Message {
         for attachment in &self.attachments {
             builder.add_existing_attachment(attachment.id);
         }
+        Ok(builder)
```
```suggestion
        builder
```

src/model/channel/message.rs
```diff
@@ -334,7 +334,12 @@ impl Message {
                 }
             }
         }
+        let mut builder = self._prepare_edit_builder()?;
```
```suggestion
        let mut builder = self._prepare_edit_builder();
```

arqunis (changes requested):

Milo123459:
I presume that this method of slitting up the code can be applied to any function that takes a closure?

PR #1897, opened 2022-05-06
Make typing.stop() to return bool instead of option, allow ignoring result.

Miezhiko:
Mostly we don't want (I think) to check return valur of `stop()`

It doesn't tell if there was error or not but instead if there was typing to stop or not, and there is nothing to do usually if typing wasn't active unless very specific cases.

### DIFF
--- a/src/http/typing.rs
+++ b/src/http/typing.rs
@@ -92,6 +92,7 @@ impl Typing {
     /// Returns false in case if typing has already been stopped.
     ///
     /// [`Channel`]: crate::model::channel::Channel
+    #[allow(clippy::must_use_candidate)]
     pub fn stop(self) -> bool {
         self.0.send(()).is_ok()
     }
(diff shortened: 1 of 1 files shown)


src/http/typing.rs
```diff
@@ -89,10 +89,11 @@ impl Typing {
     ///
     /// This should be used to stop typing after it is started using [`Typing::start`].
     /// Typing may persist for a few seconds on some clients after this is called.
+    /// Returns false in case if typing has already been stopped.
```
mkrasnitski:
```suggestion
    /// Returns false if typing has already stopped.
```

arqunis:
I [have rebased this onto `next`](https://github.com/serenity-rs/serenity/commit/818a26fc597d4bea16c89b5abe017caa56fede0f), since it contains a breaking change.

PR #1900, opened 2022-05-12
Implement announcement channel following

GnomedDev:
Quite surprised this wasn't implemented. I'm not the best doc writer and I'm probably inconsistent.
https://discord.com/developers/docs/resources/channel#follow-news-channel

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1766,7 +1766,11 @@ impl Http {
     }
 
     /// Follow an Announcement Channel to send messages to a target channel.
-    pub async fn follow_news_channel(&self, news_channel_id: u64, target_channel_id: u64) -> Result<Webhook> {
+    pub async fn follow_news_channel(
+        &self,
+        news_channel_id: u64,
+        target_channel_id: u64,
+    ) -> Result<FollowedChannel> {
         let map = json!({ "webhook_channel_id": target_channel_id });
         let body = to_vec(&map)?;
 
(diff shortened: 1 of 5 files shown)


src/http/client.rs
```diff
@@ -1765,6 +1765,26 @@ impl Http {
         .await
     }
 
+    /// Follow an Announcement Channel to send messages to a target channel.
+    pub async fn follow_news_channel(
```
Milo123459, line 1769: Should this be named `follow_news_channel` or `follow_announcement_channel`?

GnomedDev: The API calls them news channels, but I saw existing references to announcement channels in serenity. I thought for the raw http module I would follow the api name then could just call it `follow` on Channel(Id) and avoid the naming bikeshedding.

Milo123459: Maybe we should rename the existing references to news channels?

GnomedDev: Should this been done in a different PR to next since that would be breaking?

arqunis: We should strive to be consistent as possible, so yes, a pull request targeting `next` should be opened to replace instances of "announcement channel" with "news channel".

Milo123459: Why not just target this PR at next and replace all references in this PR?

kangalio: Probably best to keep this PR like it is and maybe make a follow up PR

Small PRs are good anyways

arqunis (approved):

PR #1907, opened 2022-05-21
Replace magic number for interaction response type

nickelc:


arqunis (approved):

nickelc:
The beta failure is unrelated
```
error: 4th rule of macro `define_related_ids_for_event_type` is never used
Error:     --> src/model/event.rs:1725:5

error: 5th rule of macro `define_related_ids_for_event_type` is never used
Error:     --> src/model/event.rs:1728:5
```

PR #1909, opened 2022-05-21
Fix clippy warnings

GnomedDev:
Targeted at next because probably breaks api for Emoji::delete and Emoji::edit


nickelc:
Please create a separate PR for the breaking changes and change the target for this PR to the `current` branch with the breaking changes removed.

GnomedDev:
Changed the target to current and got rid of breaking changes

GnomedDev:
Explanation for the force push spam, this was failing tests so I tried rebasing it on next, forgetting it's meant to be on current

arqunis (approved):

PR #1910, opened 2022-05-21
refactor(Message): don't set message content not sent by the api

vaporoxx:
Problems with the current setup:

- The content of messages with these types is not documented, meaning that it can change anytime
- Not every message type is handled, e.g. boost or thread creation messages do not get their content added
- Overall more possibly unexpected behaviour and less consistency with the API

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -314,14 +314,9 @@ pub(crate) fn dispatch<'rec>(
 
 async fn dispatch_message(
     context: Context,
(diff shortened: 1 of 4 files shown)


arqunis (approved):

kangalio:
For later reference: Discord welcome messages (JOIN_MESSAGES) were removed in this PR (context: https://discord.com/channels/381880193251409931/673965002805477386/1150444905680863378)

Issue #1915, opened 2022-05-26
Add to EventHandler support for "Guild Scheduled Event" gateway events


A-Phamfam:
Scheduled Events models and builders have been added, but the gateway events have not been added yet.

These Discord gateway events should be added to the Serenity [EventHandler](https://docs.rs/serenity/0.11.2/serenity/client/trait.EventHandler.html#):
- [GUILD_SCHEDULED_EVENT_CREATE](https://discord.com/developers/docs/topics/gateway#guild-scheduled-event-create)
- [GUILD_SCHEDULED_EVENT_UPDATE](https://discord.com/developers/docs/topics/gateway#guild-scheduled-event-update)
- [GUILD_SCHEDULED_EVENT_DELETE](https://discord.com/developers/docs/topics/gateway#guild-scheduled-event-delete)
- [GUILD_SCHEDULED_EVENT_USER_ADD](https://discord.com/developers/docs/topics/gateway#guild-scheduled-event-user-add)
- [GUILD_SCHEDULED_EVENT_USER_REMOVE](https://discord.com/developers/docs/topics/gateway#guild-scheduled-event-user-remove)

This would be used for implementing functionality such as automated announcement messages upon scheduled event creation, special thread creation when scheduled events are created, pinging of users to channels or threads upon subscribing/adding a scheduled event, adding special roles when a user has subscribed/added a scheduled event, renaming roles or threads upon scheduled event update, or deletion of threads or roles upon scheduled event delete.

PR #1916, opened 2022-05-28
Use `#[serde(default)]` for the `ScheduledEventMetadata::location` field

nickelc:
This field is optional and isn't sent for stage & voice events when the
metadata is sent for gateway events in some circumstances.

Adding `#[serde(default)]` is a workaround instead of changing it to
`Option<String>`.

discord docs: [guild-scheduled-event-entity-metadata]

[guild-scheduled-event-entity-metadata]: https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-object-guild-scheduled-event-entity-metadata

### DIFF
--- a/src/model/guild/scheduled_event.rs
+++ b/src/model/guild/scheduled_event.rs
@@ -78,6 +78,8 @@ enum_number!(ScheduledEventType {
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ScheduledEventMetadata {
+    // TODO: Change to `Option<String>` in next version.
+    #[serde(default)]
     pub location: String,
 }
 


mkrasnitski:
Back when implementing this, I set the type as `String` because the metadata field itself is optional, and for now location is the only subfield there, so having it also as an `Option<String>` felt terribly unergonomic. But, I agree this does bring it closer in line with the API definition.

nickelc:
I have implemented the gateway events #1915 and you get an deserialization error (missing field `location`) for `GUILD_SCHEDULED_EVENT_UPDATE` when you start the event.

payload: `entity_metadata":{"speaker_ids":[]}`

mkrasnitski:
Ok, makes sense. Is `speaker_ids` an undocumented field?

nickelc:
It was initially documented but deprecated later.
https://github.com/discord/discord-api-docs/search?q=speaker_ids&type=issues

arqunis (approved):

PR #1925, opened 2022-05-31
Add helper methods to CommandDataOptionValue

### LINKED ISSUE #1533: `ApplicationCommandInteractionDataOptionValue` should have `as_...(): Option<T>` methods
I think it could be useful for `ApplicationCommandInteractionDataOptionValue` to have methods similar to `as_bool()` in serde's `Value` to avoid bulky code like : 
```rs
if let ApplicationCommandInteractionDataOptionValue::Integer(n) = v {
   Some(n)
} else {
   None
}}) 
```

GnomedDev:
Fixes #1533

Defines a private macro to generate the actual helper methods, thanks Yandros for help with `generate_doc`!

### DIFF
--- a/src/model/application/interaction/application_command.rs
+++ b/src/model/application/interaction/application_command.rs
@@ -686,7 +686,7 @@ macro_rules! generate_doc {
 macro_rules! impl_as_ty {
     ($($name:ident, $variant:ident, $type:ty, $value_op:expr;)*) => {
         $(
-            generate_doc!{
+            generate_doc! {
                 #[generate_doc($variant, $type)]
                 pub fn $name(&self) -> Option<$type> {
                     if let CommandDataOptionValue::$variant(value) = self {
--- a/src/model/application/interaction/application_command.rs
+++ b/src/model/application/interaction/application_command.rs
@@ -701,7 +701,7 @@ macro_rules! impl_as_ty {
     };
 }
 
-#[inline(always)]
+#[inline]
 fn deref<T: Copy>(v: &T) -> T {
     *v
 }
--- a/src/model/application/interaction/application_command.rs
+++ b/src/model/application/interaction/application_command.rs
@@ -1,5 +1,4 @@
 use std::collections::HashMap;
-use std::convert::identity;
 
 use serde::de::Error as DeError;
 use serde::{Deserialize, Deserializer};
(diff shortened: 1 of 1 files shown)


GnomedDev:
Thoughts on bumping msrv up one version to `1.54`, and retargetting this PR at next?

nickelc:
I don't like the doc generation macro. it's too much macro-fu for my taste.

And I also have a PR in the works that changes this and many other types in the `application_command` / `autocomplete` module.

GnomedDev:
I'll change the doc generation macro to simply generate the text, then do `#[doc=generate_doc!{$variant, $type}]`

kangalio:
MSRV compatible macro proposal:
```rust
macro_rules! impl_as_ty {
    ($( $doc:literal, $name:ident, $variant:ident, $type:ty $(, $transform:tt)?; )*) => { $(
        #[doc = $doc]
        #[must_use]
        pub fn $name(&self) -> Option<$type> {
            if let CommandDataOptionValue::$variant(value) = self {
                Some($($transform)? value)
            } else {
                None
            }
        }
    )* };
}

impl CommandDataOptionValue {
    impl_as_ty!(
        "If the value is [`Self::Attachment`], it returns the inner value as [`&Attachment`]",
        as_role, Role, &Role;
        "If the value is [`Self::String`], it returns the inner value as `&str`",
        as_string, String, &str;
        "If the value is [`Self::Number`], it returns the inner value as `f64`",
        as_number, Number, f64, *;
        "If the value is [`Self::Integer`], it returns the inner value as `i64`",
        as_integer, Integer, i64, *;
        "If the value is [`Self::Boolean`], it returns the inner value as `bool`",
        as_boolean, Boolean, bool, *;
        "If the value is [`Self::Channel`], it returns the inner value as `&PartialChannel`",
        as_channel, Channel, &PartialChannel;
        "If the value is [`Self::Attachment`], it returns the inner value as `&Attachment`",
        as_attachment, Attachment, &Attachment;
    );

    /// If the value is [`Self::User`], it returns the inner value as `(&User, &Option<PartialMember>)`
    #[must_use]
    pub fn as_user(&self) -> Option<(&User, &Option<PartialMember>)> {
        if let CommandDataOptionValue::User(user, member) = self {
            Some((user, member))
        } else {
            None
        }
    }
}
```

- MSRV compatible
- shorter
- simpler: one function and one macro less

nickelc:
To be honest, I don't see a value in the macro.
The public getters are small and could be implemented without adding the indirection of the macro.

PR #1931, opened 2022-06-02
Fix compile errors in builder when model disabled

FallenWarrior2k:
PR #1830 fixed a bunch of clippy pedantic level lints, which also unfortunately caused some regressions for the no-model use case, as I intentionally used bare Default in #1770 to automatically select the correct type without manually inspecting the feature flags again.

I admit my use case is somewhat outlandish, but would you mind if I added one or two regression tests to CI to avoid situations like this in the future?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,9 @@ jobs:
             dont-test: true
           - name: builder without model
             features: builder
+          - name: unstable Discord API (no default features)
+            features: unstable_discord_api
+            dont-test: true
 
     steps:
       - name: Checkout sources
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,7 @@ jobs:
             dont-test: true
           - name: builder without model
             features: builder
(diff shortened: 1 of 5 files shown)


arqunis (approved):

arqunis:
> but would you mind if I added one or two regression tests to CI to avoid situations like this in the future

Sure, go ahead. Nick had no way of knowing it would break compilation of the library without the `model` feature. With new CI jobs, they should prevent that scenario from happening again.

FallenWarrior2k:
I'm not blaming them. Can't expect everybody to manually check for every combination of flags there is.

FallenWarrior2k:
Does this look good to you?

arqunis:
Looks good. If you handle the unused imports, I'll merge this.

FallenWarrior2k:
On it.

FallenWarrior2k:
As explained in the commit message, I disabled the tests for builder without model for now because a lot of them currently fail to compile due to missing feature gates. I would address that in another PR to prevent this one from blowing up in scope even further.

PR #1936, opened 2022-06-03
Clean up previous changes to cache

mkrasnitski:
This replaces the implementation of `From<Ref>` for `CacheRef` with a private method in order to not expose the dashmap type as part of the public API. Also, replaces `.map(|c| c.clone())` with `.as_deref().cloned()` on types returned from the cache in code and docs. Finally, removes an extra clone in `User::nick_in`, and refactors example 5 slightly.

### DIFF
--- a/src/model/user.rs
+++ b/src/model/user.rs
@@ -1005,7 +1005,23 @@ impl User {
         cache_http: impl CacheHttp,
         guild_id: impl Into<GuildId>,
     ) -> Option<String> {
-        guild_id.into().member(cache_http, &self.id).await.ok().and_then(|member| member.nick)
+        let guild_id = guild_id.into();
+
+        // This can't be removed because `GuildId::member` clones the entire `Member` struct if
+        // it's present in the cache, which is expensive.
+        #[cfg(feature = "cache")]
+        {
+            if let Some(cache) = cache_http.cache() {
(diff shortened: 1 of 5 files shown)


src/model/user.rs
```diff
@@ -1005,24 +1005,7 @@ impl User {
...
-                }
-            }
-        }
-
-        guild_id
-            .member(&cache_http, &self.id)
-            .await
-            .ok()
-            .and_then(|member| member.nick.as_ref().map(String::clone))
+        guild_id.into().member(cache_http, &self.id).await.ok().and_then(|member| member.nick)
```
GnomedDev: This is cloning the whole member instance if it was in cache, `.as_ref().map(String::clone)` should just be removed.

mkrasnitski: Currently working on a separate PR that removes the cache clone on `member`, so then I'll have to add `member.nick.clone()` but in the meantime this is much less verbose.

GnomedDev: Removing the cache clone on GuildId::member is impossible, see the description of https://github.com/serenity-rs/serenity/pull/1934

mkrasnitski: Oh, ok. That's unfortunate.

arqunis (approved):

PR #1938, opened 2022-06-04
Rework manual deserialization implementations 

GnomedDev:
This performs multiple optimizations throughout the library:
- Moves chunks of logic used for deserializing to functions, to remove duplication
- Moves adding extra `guild_id` fields to a function, to remove duplication
- Deserializes straight into `GatewayEvent` instead of to Value
- Uses serde derive for deserialization of:
- -  `Event`, removes `GuildUnavailable` in the process
- -  `Guild`, moves preprocessing to `GuildCreateEvent
- -  `CommandDataOption`
- Deduplicates a load of logic in deserializing:
- - `Reaction`
- - `PartialGuild`
- - `CommandData`
- - `ModalSubmitInteraction`
- - `MessageComponentInteraction`
- - `ApplicationCommandInteraction`

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -721,7 +721,7 @@ async fn handle_event(
                 event_handler.typing_start(context, event).await;
             });
         },
-        DispatchEvent::Model(Event::Unknown) => warn!("An unknown event was recieved"),
+        DispatchEvent::Model(Event::Unknown) => warn!("An unknown event was received"),
         DispatchEvent::Model(Event::UserUpdate(mut event)) => {
             let _before = update(&cache_and_http, &mut event);
             let event_handler = Arc::clone(event_handler);
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -117,9 +117,6 @@ impl DispatchEvent {
             Self::Model(Event::GuildStickersUpdate(ref mut event)) => {
                 update(cache_and_http, event);
             },
-            Self::Model(Event::GuildUnavailable(ref mut event)) => {
-                update(cache_and_http, event);
-            },
             // Already handled by the framework check macro
             Self::Model(Event::MessageCreate(_)) => {},
             Self::Model(Event::MessageUpdate(ref mut event)) => {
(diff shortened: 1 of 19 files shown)


src/client/event_handler.rs
```diff
@@ -393,7 +392,7 @@ pub trait EventHandler: Send + Sync {
     /// Dispatched when an unknown event was sent from discord.
     ///
     /// Provides the event's name and its unparsed data.
-    async fn unknown(&self, _ctx: Context, _name: String, _raw: Value) {}
+    async fn unknown(&self, _ctx: Context) {}
```
arqunis: I don't suppose we can ask serde to at least supply us the unknown variant's name? Without any data supplied to this event, it's just unnecessary. What useful behaviour can you put here, even? Just a log saying, "hey! I just encountered an unknown event. not sure what it is, though"?

If not, maybe it would be good to remove this handler, since without either the name or the unparsed data, the documentation comment is outdated.

src/client/dispatch.rs
```diff
@@ -722,11 +722,7 @@ async fn handle_event(
             });
         },
         DispatchEvent::Model(Event::Unknown) => {
-            let event_handler = Arc::clone(event_handler);
-
-            spawn_named("dispatch::event_handler::unknown", async move {
-                event_handler.unknown(context).await;
-            });
+            // No point in handling this event, gives no info.
```
arqunis: No point in dispatching the event for users, but might be worth to log something, so that we know we're receiving an unknown event (as for which kind, we'll have to figure it out ourselves). A `tracing::warn!` should be good here.

src/client/dispatch.rs
```diff
@@ -721,9 +721,7 @@ async fn handle_event(
                 event_handler.typing_start(context, event).await;
             });
         },
-        DispatchEvent::Model(Event::Unknown) => {
-            // No point in handling this event, gives no info.
-        },
+        DispatchEvent::Model(Event::Unknown) => warn!("An unknown event was recieved"),
```
arqunis: Typo `recieved` -> `received`.

arqunis (changes requested):

arqunis (approved):

nickelc:
The commit contains several breaking changes that are not reflected in the commit message. :(

PR #1951, opened 2022-06-07
Fix deserialization errors

GnomedDev:
Fixes errors caused by #1938, found, fixed, and tested by running this branch on my bot. Also drops down unknown event log to debug level, as my bot got a lot of unknown events with no way to silence them.

### DIFF
--- a/src/model/event.rs
+++ b/src/model/event.rs
@@ -15,6 +15,7 @@ use super::utils::{
     remove_from_map_opt,
     roles,
     stickers,
+    ignore_input,
 };
 use crate::constants::OpCode;
 use crate::internal::prelude::*;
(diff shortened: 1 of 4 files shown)


src/model/event.rs
```diff
@@ -266,27 +267,43 @@ impl<'de> Deserialize<'de> for GuildMembersChunkEvent {
...
-#[derive(Clone, Debug, Deserialize, Serialize)]
+#[derive(Clone, Debug, Serialize)]
 #[non_exhaustive]
 pub struct GuildRoleCreateEvent {
-    #[serde(deserialize_with = "roles::deserialize_event")]
     pub role: Role,
 }
 
+impl<'de> Deserialize<'de> for GuildRoleCreateEvent {
+    /// Custom deserializer to pass entire payload to roles::deserialize_event, does not work using `deserialize_with`
```
nickelc, line 277: Why doesn't this work anymore?

GnomedDev: As `roles::deserialize_event` needs the **entire** payload for the entire event, instead of just the `role` field as would be passed using `deserialize_with`

nickelc: Ah, you're changing it back to the original code before you broke it with #1938.
:+1:

arqunis (approved):

PR #1955, opened 2022-06-10
Take Into<String> in more places

GnomedDev:
Fixes a couple more `ToString` usages I missed in #1903 by constructing the Value in `Embed::field` without the `json!` macro, and swaps out `Option<&str>` for `Option<impl Into<String>>`

### DIFF
--- a/src/builder/edit_guild.rs
+++ b/src/builder/edit_guild.rs
@@ -68,8 +68,8 @@ impl EditGuild {
     /// ```
     ///
     /// [`utils::read_image`]: crate::utils::read_image
-    pub fn icon(&mut self, icon: Option<impl Into<String>>) -> &mut Self {
-        self.0.insert("icon", icon.map(Into::into).map_or(NULL, Value::String));
+    pub fn icon(&mut self, icon: Option<String>) -> &mut Self {
+        self.0.insert("icon", icon.map_or(NULL, Value::String));
         self
     }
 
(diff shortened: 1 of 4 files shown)


src/builder/edit_guild.rs
```diff
@@ -68,8 +68,8 @@ impl EditGuild {
     /// ```
     ///
     /// [`utils::read_image`]: crate::utils::read_image
-    pub fn icon(&mut self, icon: Option<&str>) -> &mut Self {
-        self.0.insert("icon", icon.map_or_else(|| NULL, |x| Value::from(x.to_string())));
+    pub fn icon(&mut self, icon: Option<impl Into<String>>) -> &mut Self {
```
mkrasnitski: Taking `Option<impl Into<String>>` requires specifying a concrete type via turbofish whenever `None` is passed in (e.g. `None::<&str>`), so this doesn't work.

GnomedDev: hmm, is there a way I can specify a default trait bound to fix this?

mkrasnitski: As it stands, the user can accomplish the same thing by just saying `s.map(|val| val.to_string()).as_deref()` to get an `Option<&str>` from a concrete `Option<impl Into<String>>`. Not sure if that closure can be replaced with a simple method name.

GnomedDev: The closure could be replaced by `Into::into`, however that doesn't fix the `None` problem... if I can't set a default bound and `None::<&str>` is unacceptable I'd prefer it to be `Option<String>` (as then the preference for owned strings is clearly shown)

arqunis: I'd rather just set the parameter to `Option<String>` and be done with it. It would keep things simple and would be up to the user to either pass ownership or not.

arqunis (approved):

PR #1958, opened 2022-06-10
Add support for Discord's new AutoMod feature

GnomedDev:
This PR is very draft, as the feature hasn't even officially been released, although it is currently enabled in some guilds (including one I own). I'll be adding features to this as they are documented, but the main point of this PR at the moment is to make serenity be quiet about `Unknown MessageType value: 24`


nickelc:
see #2035

PR #1962, opened 2022-06-12
Rework builder module to get rid of all `HashMap<'static str, T>` 

GnomedDev:
Swaps all the `HashMap<'static str, T>` for concrete rust structs and make them `derive(serde::Serialize)`, this allows for possible const builders, gets rid of a ton of heap allocations, and cleans up code massively. Nested builders are also serialised while building the top level one, meaning that all serialisation is done in one step. This also actually removes all usage of `json::to_value` outside one usage in a test, so I've `cfg(test)` gated that (and it shows that we no longer need Value, at least for serialisation).

Since I basically did a once over the whole builder module, some arguments have been changed to be more consistent or efficient, the builders internally store the `Type` enums instead of the raw number, and some int sizes have been reduced.

This does make all builder fields private, although a user can still pass custom JSON to the API via the raw HTTP functions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,11 +156,11 @@ default_no_backend = [
     "cache",
     "chrono",
     "client",
-#    "framework",
+    "framework",
     "gateway",
     "model",
     "http",
-#    "standard_framework",
+    "standard_framework",
     "utils",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,11 +156,11 @@ default_no_backend = [
     "cache",
     "chrono",
     "client",
-    "framework",
+#    "framework",
     "gateway",
     "model",
     "http",
-    "standard_framework",
+#    "standard_framework",
     "utils",
 ]
 
--- a/src/builder/add_member.rs
+++ b/src/builder/add_member.rs
@@ -1,20 +1,28 @@
-use std::collections::HashMap;
-
-use crate::json::{from_number, Value};
 use crate::model::id::RoleId;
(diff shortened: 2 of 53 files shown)


mkrasnitski:
Are we sure this is something we want to do? Any new fields added by discord will require breaking changes unless we make all builders `#[non_exhaustive]`. An alternative is to just make the hashmap private, then the methods on the builder are all that are exposed.

arqunis:
@mkrasnitski This is not an issue if the fields of the builders are private. Rust allows you to construct `struct`s only if all of their fields are `pub`lic. If one of them is private, users of the crate cannot instantiate the `struct`. `#[non_exhaustive]` is necessary when you want to impose this restriction for `struct`s that *do* have all fields `pub`lic, but is not necessarily open for direct construction, besides the crate where it is defined in, because you may want to add new fields in the future.

mkrasnitski:
Due to a limitation of `async fn`, having `'static` lifetimes in the params for `Http::create_sticker` causes incorrect lifetime unification. I think a solution is to either annotate the `&self` and `audit_log_reason` params with `'a`, or explicitly desugar the return type as `impl Future<Output = Result<Sticker>>> + 'a`, making sure to add the lifetime bound at the end there. Let me know which one ends up working.

The relevant issue is rust-lang/rust #63033/ (not linked as to not create a cross-reference). Not sure why only the MSRV build ran into this issue, as I think this fails to compile even on 1.61.

arqunis (approved):

PR #1963, opened 2022-06-12
Swap out storing &[u8] to Vec<u8> in Request

GnomedDev:
Makes the Vec allocation occur earlier, and opens up the possibility of passing `Vec<u8>` directly into http methods.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -3896,7 +3896,7 @@ impl Http {
     /// let route_info = RouteInfo::CreateMessage { channel_id };
     ///
     /// let mut request = RequestBuilder::new(route_info);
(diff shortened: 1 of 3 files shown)


nickelc:
I don't get the reason for this change and now cloning the request.

GnomedDev:
The request clone is only going to clone the Vec, which was already previously cloned by going from `let body = to_vec(&map)` (owned Vec), `&body` (passed into request as `&[u8]`), then a new vec is defined using `Vec::from`.

I made this originally for passing `Vec<u8>` into http methods directly, although I realised I can avoid this change by taking `impl Serialize`, although Request should probably take the body directly instead still.

arqunis (approved):

Issue #1905, opened 2022-05-16
Improving `Http` and surrounding APIs


mkrasnitski:
There was some discussion about some areas of the API where there is currently room for improvement; things like taking more advantage of Rust's type system to ensure API compatibility, plus removing duplicated code or method signatures as well as in general decreasing API surface area to prevent future internal changes from being breaking. I've gone and listed some of them below, so that they don't become lost:

1. While the methods on `Http` might be considered the "canonical" wrappers around the Discord API, they also don't take advantage of Rust's type system at all, since they take a Json object, which is basically just a HashMap with Strings as keys. The user interacts with the `Http` struct quite often, mainly to facilitate requests to Discord endpoints on their behalf. So, they also have access to these methods. 
It would be nice to mark all such methods as `pub(crate)` instead of `pub`, as this would require users go through the various model types instead, which only allow specific fields to be set in the hashmap. Plus, if Discord changes how an endpoint works in some way, then the underlying `Http` implementation could be changed without introducing breaking changes, while methods on model types would remain unaffected (methods would only be added, not changed). This would allow features to release quicker and reduce technical debt.
    * Some of the methods on `Http` take an optional `audit_log_reason` parameter, which currently all the wrapper methods on model  types just set to `None`. Should we keep these methods exposed, or add methods to the model types to allow passing an optional `audit_log_reason`? Better yet, is it worth considering adding builders for these specific function calls? As in, take as argument a closure that constructs a builder struct. This might not be worth doing for methods with very few arguments.
2. The model types are full of duplicated method signatures. For example, the `GuildId`/`Guild`/`PartialGuild` types. Many methods in `GuildId` are wrapped by methods on the other two guild types. When combined with (1), the worry is that this code duplication could introduce inconsistency. So, it would be ideal if there were only a single possible callsite. The logical choice is to just keep as many methods on `GuildId` as possible. However, semantically, these methods operate on guilds, not on the id itself, so there's a discrepancy there.
The alternative would be to auto-generate wrappers, perhaps using a proc-macro. Many of these wrappers are simple thin wrappers. Others add some functionality though, such as checking the cache to validate user permissions. These too could probably be auto-impl'd using a proc-macro, as long as the required permissions are specified in each case.
[...]

PR #1967, opened 2022-06-13
[DRAFT] Builder paradigm overhaul

mkrasnitski:
This is a draft PR as this work will not be ready all at once, and I'm open to comments and bikeshedding as we attempt to get this right. The main goal is to start addressing point number 3 listed in #1905: to rework the builder API to reduce monomorphization bloat (resulting from generic closure parameters), as well to allow for re-using preconstructed builders, and also to enable another later pass of changes to enforce (to a greater extent) API compliance at compile time using the builder API (e.g. always-required, conditionally-required, and mutually exclusive fields).

Some notable changes I've done:
1. Methods on model types that used to take builder construction closures now instead return an empty builder object.
2. Builders have an additional `execute` method that consumes them and actually perform the request. Essentially the functionality from the model types has been moved into the builder itself.
3. Because `execute` takes ownership of `self`, then in order to call `Builder::new().method1().method2().execute()`, the rest of the chain must take ownership as well. Therefore, builders will change to take and return `Self` instead of `&mut Self`.

So far I've tried reworking `CreateSticker` and `CreateThread`, with some success. I do have some concerns though:
~~1. A good deal of builders are called from multiple places. For example, `GuildId`/`Guild`/`PartialGuild`, or `ChannelId`/`GuildChannel`. Since the implementations on guild usually do a cache lookup, to preserve all functionality I had to make the builder store a reference to a generic `GuildIdWrapper`, which I will likely also have to do for `ChannelId`. Is this acceptable? I'm not yet satisfied with it as the implementation is exactly duplicated for `Guild` and `PartialGuild`, but that's something I may solve later, once I'm sure that this fact holds for ***all*** relevant builders.~~
2. Do we want to factor out `execute` ~~and `execute_with_cache`~~ into their own traits, e.g. `Builder<T>` ~~and `BuilderWithCache<T>`~~? We could probably auto-import these so that the user wouldn't have to import them themselves.
3. The reason the `execute` method takes ownership is because `json::hashmap_to_json_map()` consumes the builder's hashmap. Potentially this is addressed by #1962, so I'll wait until that PR is complete before getting too deep into the weeds on this one.

Feedback very much wanted; I'm not great at writing docs so a review of that part is also greatly appreciated.

### DIFF
--- a/src/builder/create_scheduled_event.rs
+++ b/src/builder/create_scheduled_event.rs
@@ -154,8 +154,8 @@ impl CreateScheduledEvent {
     }
 
     #[cfg(feature = "model")]
-    async fn _execute(self, http: impl AsRef<Http>) -> Result<ScheduledEvent> {
-        http.as_ref().create_scheduled_event(self.id.into(), &self.fields, None).await
+    async fn _execute(self, http: &Http) -> Result<ScheduledEvent> {
+        http.create_scheduled_event(self.id.into(), &self.fields, None).await
     }
 }
 
--- a/src/builder/create_scheduled_event.rs
+++ b/src/builder/create_scheduled_event.rs
@@ -120,8 +120,8 @@ impl CreateScheduledEvent {
     ///
     /// # Errors
     ///
-    /// May error if the icon is a URL and the HTTP request fails, or if the image is a file
-    /// on a path that doesn't exist.
+    /// May error if a URL is given and the HTTP request fails, or if a path is given to a file
+    /// that does not exist.
     #[cfg(feature = "http")]
     pub async fn image<'a>(
         mut self,
@@ -131,6 +131,14 @@ impl CreateScheduledEvent {
         let image_data = image.into().data(&http.as_ref().client).await?;
         self.image = Some(encode_image(&image_data));
         Ok(self)
+    }
+
+    /// Sets the cover image for the scheduled event. Requires the input be a base64-encoded image
+    /// that is in either JPG, GIF, or PNG format.
+    #[cfg(not(feature = "http"))]
+    pub fn image(mut self, image: String) -> Self {
+        self.image = Some(image);
+        self
     }
 
     /// Creates a new scheduled event in the guild with the data set, if any.
--- a/src/builder/create_scheduled_event.rs
+++ b/src/builder/create_scheduled_event.rs
@@ -140,9 +140,7 @@ impl CreateScheduledEvent {
     /// # Errors
     ///
     /// If the `cache` is enabled, returns a [`ModelError::InvalidPermissions`] if the current user
-    /// does not have permission to manage scheduled events.
-    ///
-    /// Otherwise will return [`Error::Http`] if the current user does not have permission.
+    /// does not have permission to manage scheduled events. Otherwise, returns [`Error::Http`].
     ///
     /// [Manage Events]: Permissions::MANAGE_EVENTS
     #[cfg(feature = "http")]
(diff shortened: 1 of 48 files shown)


mkrasnitski:
I've rebased on `next` now that #1962 has been merged, and wanted to mention that point 1 above is no longer relevant, as we can avoid passing around a reference to whatever `GuildId/Guild/PartialGuild` object called into the builder: simply give the builder a concrete `GuildId`, then for permissions cache lookups, do an additional lookup to get the `Guild` from its id before doing the permissions check on it. This lets us also perform cache lookups even if the method is called on a `GuildId`, so therefore I've elected to unify the `execute` and `execute_with_cache` functions, so cache lookups happen every time. To be honest, I think this is preferable.

src/builder/create_scheduled_event.rs
```diff
@@ -99,18 +118,50 @@ impl CreateScheduledEvent {
...
+                        return Err(Error::Model(ModelError::InvalidPermissions(req)));
+                    }
+                }
+            }
+        }
+
+        self._execute(cache_http.http()).await
+    }
+
+    async fn _execute(self, http: impl AsRef<Http>) -> Result<ScheduledEvent> {
```
GnomedDev: This would be the perfect method to take concrete types on, to avoid it being monomorphized, eg: take http directly and do the as_ref call beforehand.

GnomedDev (commented):
The builders should be made `must_use` aswell, instead of each method on them, imo

src/builder/create_scheduled_event.rs
mkrasnitski: True, although its a private method, and will only ever be given an `Http`. I could remove the `AsRef` and it would still work. I'm planning on eventually factoring all the cache lookups into a function and replacing them all with a one-liner like `utils::check_perms(&cache_http, Permissions::MANAGE_EVENTS).await?`.

mkrasnitski:
With the PR making good progress up to this point, I thought I'd write down some comments and concerns about the current state of the code.

1. The current obvious pain point that's bothering me is that when I reworked `CreateMessage`, it caused `UserId::direct_message` to take a hit in ergonomics. Since it first has to potentially create a direct message channel, before then giving the user a builder which they execute, it now not only requires a double `.await`, but also requires duplicating error handling code. One ugly solution is to separate `CreateMessage`'s fields into `CreateMessageData`, then create a separate `CreateDmMessage` builder that wraps them. But, this would require moving the field method implementations into `CreateMessageData`, then adding wrappers to both `CreateMessage` and `CreateDmMessage`, resulting in a lot of code duplication.
2. Speaking of, a looming dilemma are the `Edit` builders. The implementations on model types (e.g. `GuildChannel`) assign the result of the `http` method to `*self`. Therefore, I'm probably going to have to give the builder a generic mutable reference to the caller struct to assign to. This does make a case for factoring the fields themselves into a `BuilderFields` to separate it out from metadata. Overall, having to carry execution metadata along with the builder while it's being constructed sort of sucks, but there doesn't seem to be a better alternative.
3. Maybe once every builder has been reworked here and the dust has settled, we could auto-generate these aforementioned wrapper impls using macros (my guess is it would require proc-macros), but that's a problem for later. For now, full steam ahead, and we'll see where this ends up.

mkrasnitski:
One more issue to note about the `Edit` builders. As described above, they have to hold on to a mutable reference to the caller in order to mutate it after executing themselves. However, this means that they can no longer implement `Clone`, since `&mut T` cannot be `Clone`. Not sure yet what to do about this, as ideally it would be great if all builders could be cloned for easy re-use.

mkrasnitski:
I've thought of an approach that solves these problems. Unfortunately it requires redoing a large part of what has already been pushed here, so I'm closing this and trying again in a new PR.

PR #1981, opened 2022-06-16
Remove usage of json! macro from websocket message sending

GnomedDev:
This also implements https://github.com/discord/discord-api-docs/commit/6755c355caaf2e1c321dc7de17e1e257d9aa2db5

### DIFF
--- a/src/gateway/ws.rs
+++ b/src/gateway/ws.rs
@@ -15,11 +15,69 @@ use url::Url;
 use crate::client::bridge::gateway::ChunkGuildFilter;
 use crate::constants::{self, Opcode};
 use crate::gateway::{CurrentPresence, GatewayError};
-use crate::json::{from_str, json, to_string, Value};
+use crate::json::{from_str, to_string};
 use crate::model::event::GatewayEvent;
-use crate::model::gateway::GatewayIntents;
-use crate::model::id::GuildId;
+use crate::model::gateway::{ActivityType, GatewayIntents};
+use crate::model::id::{GuildId, UserId};
 use crate::{Error, Result};
+
+#[derive(Serialize)]
+struct IdentifyProperties {
+    browser: &'static str,
+    device: &'static str,
+    os: &'static str,
+}
+
+#[derive(Serialize)]
+struct ActivityData<'a> {
+    #[serde(rename = "type")]
+    kind: ActivityType,
+    url: &'a Option<Url>,
+    name: &'a str,
+}
+
+#[derive(Serialize)]
+struct ChunkGuildMessage<'a> {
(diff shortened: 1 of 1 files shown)


src/gateway/ws.rs
```diff
@@ -15,12 +15,75 @@ use url::Url;
...
+    limit: u16,
+}
+
+#[derive(Serialize)]
+#[serde(untagged)]
+enum WebSocketMessageData<'a> {
+    Heartbeat(Option<u64>),
+    ChunkGuild(ChunkGuildMessage<'a>),
+    Identify {
+        v: u8,
```
vaporoxx: This doesn't actually exist anymore

src/gateway/ws.rs
```diff
@@ -15,12 +15,75 @@ use url::Url;
...
+        session_id: &'a str,
+        token: &'a str,
+        seq: u64,
+    },
+}
+
+#[derive(Serialize)]
+struct WebSocketMessage<'a> {
+    op: OpCode,
+    #[serde(borrow)]
```
vaporoxx: From looking at the serde docs it seems like `serde(borrow)` is used for deserialization only, which doesn't apply here (correct me if i'm wrong)

src/gateway/ws.rs
```diff
@@ -112,65 +175,63 @@ impl WsClient {
...
-
-        match filter {
-            ChunkGuildFilter::None => payload["d"]["query"] = json!(""),
-            ChunkGuildFilter::Query(query) => payload["d"]["query"] = json!(query),
-            ChunkGuildFilter::UserIds(user_ids) => {
-                let ids = user_ids.iter().map(|x| x.0).collect::<Vec<u64>>();
-                payload["d"]["user_ids"] = json!(ids);
-            },
+        let (query, user_ids) = match filter {
+            ChunkGuildFilter::None => (None, None),
```
vaporoxx: Discord docs state that either `query` or `user_ids` is required, so this should return an empty string for the query instead

src/gateway/ws.rs
```diff
@@ -184,19 +239,19 @@ impl WsClient {
...
-                "afk": false,
-                "since": now,
-                "status": status.name(),
-                "game": activity.as_ref().map(|x| json!({
-                    "name": x.name,
-                    "type": x.kind,
-                    "url": x.url,
-                })),
+        self.send_json(&WebSocketMessage {
+            op: Opcode::RequestGuildMembers,
```
vaporoxx:
```suggestion
            op: Opcode::PresenceUpdate,
```
Small mistake from fixing conflicts

arqunis (approved):

PR #1982, opened 2022-06-17
Fix clippy warnings & errors for all features enabled

### LINKED ISSUE #1423: Clippy fails when run with all features enabled

nickelc:
closes: #1423

### DIFF
--- a/src/http/ratelimiting.rs
+++ b/src/http/ratelimiting.rs
@@ -299,8 +299,8 @@ impl Ratelimit {
         }
 
         #[cfg(feature = "absolute_ratelimits")]
-        if let Some(_reset) = parse_header::<f64>(response.headers(), "x-ratelimit-reset")? {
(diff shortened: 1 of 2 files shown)


GnomedDev:
Could CI be added to catch this in future?

GnomedDev:
I'll take that as a no?... lol

nickelc:
Feel free to open a PR.

PR #1984, opened 2022-06-17
Rework shard types

GnomedDev:
This reduces the int type for shard ids from `u64` to `u32`, as you would probably run into other problems with serenity before trying to boot 3 billion guilds, and replaces the weird `[u64; 2]` fields with a `ShardInfo` struct.


src/model/gateway.rs
```diff
@@ -506,6 +507,41 @@ pub struct SessionStartLimit {
...
+        }
+    }
+}
+
+impl<'de> serde::Deserialize<'de> for ShardInfo {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> StdResult<Self, D::Error> {
+        Deserialize::deserialize(deserializer).map(|raw: [u32; 2]| ShardInfo {
+            id: raw[0],
+            total: raw[1],
+        })
```
nickelc:
```suggestion
        <(u32, u32)>::deserialize(deserializer).map(|(id, total)| ShardInfo {
            id,
            total,
        })
```

src/model/gateway.rs
```diff
@@ -506,6 +507,41 @@ pub struct SessionStartLimit {
...
+
+#[derive(Debug, Clone, Copy)]
+pub struct ShardInfo {
+    pub id: u32,
+    pub total: u32,
+}
+
+impl ShardInfo {
+    #[must_use]
+    pub fn new(id: u32, total: u32) -> Self {
```
kangalio: Could remove this method entirely. If the goal of ShardInfo was to be clearer on what the first and second number is by naming them, `ShardInfo::new(1, 1)` doesn't help

GnomedDev: This was mostly made for internal use (where we are passing variables in with the same names anyway), thoughts on changing to pub(crate) instead?

kangalio: `ShardInfo { id, total }` is shorter than

`ShardInfo::new(id, total)` :D

Plus, subjectively, I find it aesthetically pleasing to have few methods on plain-old-data types (structs with only public fields; that are just dumb containers of their data)

nickelc: `ShardInfo::new(1, 1)` could be replaced with a `Default` implementation.
But wouldn't `id=0 total=1` make more sense?

GnomedDev: `ShardInfo { id, total }` is expanded to this by `cargo +nightly fmt`
image
(sorry for image, github markdown was working weird)

src/cache/event.rs
```diff
@@ -597,11 +598,12 @@ impl CacheUpdate for ReadyEvent {
         let mut guilds_to_remove = vec![];
         let ready_guilds_hashset =
             self.ready.guilds.iter().map(|status| status.id).collect::<HashSet<_>>();
-        let shard_data = self.ready.shard.unwrap_or([1, 1]);
+        let shard_data = self.ready.shard.unwrap_or_else(|| ShardInfo::new(1, 1));
```
nickelc, line 601: The ids start with `0`, right?
```suggestion
        let shard_data = self.ready.shard.unwrap_or_else(|| ShardInfo::new(0, 1));
```

GnomedDev: I didn't want to change the existing behavior of the code, just refactor how it currently works, but if this is a bug I can get rid of new, impl default on shardinfo with `0, 1` and use that

src/model/gateway.rs
```diff
@@ -505,6 +506,41 @@ pub struct SessionStartLimit {
     /// The total number of session starts within the ratelimit period allowed.
     pub total: u64,
 }
+
+#[derive(Debug, Clone, Copy)]
```
nickelc:
```suggestion
#[derive(Clone, Copy, Debug)]
```

src/cache/event.rs
nickelc: You're right. It's better to keep it as it is.

src/model/gateway.rs
nickelc: I'm fine with keeping it for internal use (`pub(crate)`).

src/model/gateway.rs
```diff
@@ -505,6 +506,41 @@ pub struct SessionStartLimit {
...
+
+#[derive(Clone, Copy, Debug)]
+pub struct ShardInfo {
+    pub id: u32,
+    pub total: u32,
+}
+
+impl ShardInfo {
+    #[must_use]
+    pub fn new(id: u32, total: u32) -> Self {
```
nickelc:
```suggestion
    pub(crate) fn new(id: u32, total: u32) -> Self {
```

PR #1994, opened 2022-06-20
Fixing e12 count

DRuppFv:
The "owo" count was counting the bot messages, and it was going like this:

After the changes, it stopped counting the bot messages:

### DIFF
--- a/examples/e12_global_data/src/main.rs
+++ b/examples/e12_global_data/src/main.rs
@@ -80,19 +80,19 @@ struct Handler;
 impl EventHandler for Handler {
     async fn message(&self, ctx: Context, msg: Message) {
         // We are verifying if the bot id is the same as the message author id
-        if msg.author.id != ctx.cache.current_user_id()  {
-            if msg.content.to_lowercase().contains("owo") {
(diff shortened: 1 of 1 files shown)


examples/e12_global_data/src/main.rs
```diff
@@ -79,17 +79,20 @@ struct Handler;
...
-                let data_read = ctx.data.read().await;
-                data_read.get::<MessageCount>().expect("Expected MessageCount in TypeMap.").clone()
-            };
-
-            // Atomic operations with ordering do not require mut to be modified.
-            // In this case, we want to increase the message count by 1.
-            // https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.fetch_add
-            count.fetch_add(1, Ordering::SeqCst);
+        // We are verifying if the bot id is the same as the message author id
+        if msg.author.id != ctx.cache.current_user().id  {
```
vicky5124:
```suggestion
        if msg.author.id != ctx.cache.current_user_id()  {
```
for consistency with the library, as this method is used more than getting the entire user.

vicky5124 (changes requested):

PR #1998, opened 2022-06-20
Modifying the e12 "owo" counter

DRuppFv:
The counter was just increasing by 1 for each message that cointains "owo", and not by how many "owo" it had:

After the changes, it is counting in the right way:

It won't count "owo"s that intersect:

### DIFF
--- a/examples/e12_global_data/src/main.rs
+++ b/examples/e12_global_data/src/main.rs
@@ -79,7 +79,7 @@ struct Handler;
 #[async_trait]
 impl EventHandler for Handler {
     async fn message(&self, ctx: Context, msg: Message) {
-        // We are verifying if the bot id is the same as the message author id
+        // We are verifying if the bot id is the same as the message author id.
         if msg.author.id != ctx.cache.current_user_id()
             && msg.content.to_lowercase().contains("owo")
         {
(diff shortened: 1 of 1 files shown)


kangalio:
If the PRs continue, next month the owo counter will have become sentient :p

DRuppFv:
> If the PRs continue, next month the owo counter will have become sentient :p

What would the ```owo_counter``` command be if it weren't DRuppFv....

examples/e12_global_data/src/main.rs
```diff
@@ -89,10 +89,18 @@ impl EventHandler for Handler {
                 data_read.get::<MessageCount>().expect("Expected MessageCount in TypeMap.").clone()
             };
 
+            // To check how many "owo" there are in msg.content, we split the string with the "owo" pattern 
+            // and reduce the count of iterator returned by the split function by one.
+            let owo_in_msg = msg.content.to_lowercase().split("owo").count() - 1;
```
kangalio: Technical nit: to_ascii_lowercase(). Is more efficient and cannot cause false positives if some weird Unicode character has a lowercase mapping that ends with "o"

kangalio: > `.split("owo").count() - 1`

Seems obtuse. `.matches("owo").count()` is simpler

kangalio: Just noticed to_lowercase() is already run above. The lowercased text could be saved to avoid computing twice

examples/e12_global_data/src/main.rs
```diff
@@ -89,10 +89,18 @@ impl EventHandler for Handler {
...
+            
             // Atomic operations with ordering do not require mut to be modified.
             // In this case, we want to increase the message count by 1.
             // https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.fetch_add
-            count.fetch_add(1, Ordering::SeqCst);
+            if owo_in_msg == 0 {
```
kangalio: Why this special case? Can it even happen?

DRuppFv: Yes. If a message content be just "owo", there will be just one count, because when split try to split it between before "owo" + "owo", and after "owo",  it would return just one iter, and 1 - 1 = 0. I thought about adding a string " " in the msg content to when a message contet be just "owo" it be splitted to "owo" and " ", did you get it?

kangalio: In the case you describe owo_in_msg = 1
Screenshot_20220620_221519

DRuppFv: Yes, that's true, i could do it. But check the new commit.

PR #2004, opened 2022-06-21
documenting e16

DRuppFv:
There were only three comments, and they weren't even about the "new content" in this example.
It has a "README" file but I bet almost no one reads it, and maybe doesn't even see it.

Yes, some sql code are self-explanatory, but I think it's important to specify what you're doing, mainly to not familiarized people.


nickelc:
I think it would be better to use the SQL terminology "`userId` column" instead of "user_id(sql)".

DRuppFv:
> I think it would be better to use the SQL terminology "`userId` column" instead of "user_id(sql)".

done

arqunis (approved):

PR #2005, opened 2022-06-22
Make `MessageActivityKind` variants upper camel case

nickelc:
One of these things is not like the other.
image


jzeuzs:
I think "upper camel case" is just PascalCase

PR #2006, opened 2022-06-22
Add webhook example

NotNorom:
I'd like to add this example because I needed some time to figure out you actually don't need a token if you are only doing webhooks and so others won't have to.

### DIFF
--- a/examples/e18_webhook/src/main.rs
+++ b/examples/e18_webhook/src/main.rs
@@ -1,4 +1,5 @@
-use serenity::{http::Http, model::webhook::Webhook};
+use serenity::http::Http;
+use serenity::model::webhook::Webhook;
 
 #[tokio::main]
 async fn main() {
--- a/examples/e18_webhook/src/main.rs
+++ b/examples/e18_webhook/src/main.rs
@@ -1,0 +1,13 @@
+use serenity::http::HttpBuilder;
+
+#[tokio::main]
+async fn main() {
+    // You don't need a token when you are only dealing with webhooks
+    let http = HttpBuilder::new("").build();
+    let webhook = http.get_webhook_from_url("https://discord.com/api/webhooks/133742013374206969/hello-there-oPNtRN5UY5DVmBe7m1N0HE-replace-me-Dw9LRkgq3zI7LoW3Rb-k-q").await.expect("Replace the webhook with your own");
+
+    webhook
+        .execute(&http, false, |w| w.content("hello there").username("Webhook test"))
+        .await
+        .expect("Could not execute webhook.");
+}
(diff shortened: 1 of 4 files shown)


examples/e18_webhook/src/main.rs
```diff
@@ -0,0 +1,13 @@
+use serenity::http::HttpBuilder;
+
+#[tokio::main]
+async fn main() {
+    // You don't need a token when you are only dealing with webhooks
```
DRuppFv: add a "." here, in the ending of the comment.
add a "." here, in the ending of the comment.
add a "." here, in the ending of the comment.

examples/e18_webhook/src/main.rs
```diff
@@ -0,0 +1,13 @@
...
+    let http = HttpBuilder::new("").build();
```
mkrasnitski:
```suggestion
    let http = Http::new("");
```
You don't need to use `HttpBuilder` if you're only passing in a token. Will need to change the import as well.

examples/e18_webhook/src/main.rs
```diff
@@ -0,0 +1,13 @@
...
+    let webhook = http.get_webhook_from_url("https://discord.com/api/webhooks/133742013374206969/hello-there-oPNtRN5UY5DVmBe7m1N0HE-replace-me-Dw9LRkgq3zI7LoW3Rb-k-q").await.expect("Replace the webhook with your own");
```
mkrasnitski:
```suggestion
    let webhook = Webhook::from_url(&http, "https://discord.com/api/webhooks/133742013374206969/hello-there-oPNtRN5UY5DVmBe7m1N0HE-replace-me-Dw9LRkgq3zI7LoW3Rb-k-q").await.expect("Replace the webhook with your own");
```

More ergonomic to construct the Webhook directly. Will need to import it as well.

examples/e18_webhook/src/main.rs
```diff
@@ -0,0 +1,13 @@
+use serenity::{http::Http, model::webhook::Webhook};
```
DRuppFv: Making cargo format happy!!
```suggestion
use serenity::http::Http;
use serenity::model::webhook::Webhook;
```

arqunis (approved):

vicky5124:
The PR was merged without the example being added in the root `/Makefile.toml`, nor in `/.github/workflows/ci.yml`

nickelc:
> The PR was merged without the example being added in the root `/Makefile.toml`, nor in `/.github/workflows/ci.yml`

Fixed in [7cec025](https://github.com/serenity-rs/serenity/commit/7cec025afabd4e03cc40c9f350a779e81429aeee)

PR #2008, opened 2022-06-23
Preserve values for unknown enum variants

nickelc:
The `enum_number!` macro now takes the whole enum definition and generates
`From` trait implementations to convert a value to the enum and back.

The implementations can then be picked up by `serde` with
`#[serde(from = "u8", into = "u8")]` to (de)serialize the types.

```rust
enum_number! {
    #[derive(Clone, Copy, Debug, Deserialize, Serialize)]
    #[serde(from = "u8", into = "u8")]
    pub enum Foo {
        A = 1,
        B = 2,
        _ => Unknown(u8),
    }
}
```

**BREAKING CHANGE**: The `Unknown` variant now takes the unknown value
and the removed `fn num() -> u64` method can be replaced
with `let v = u8::from(kind)`.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -31,54 +31,41 @@ pub const USER_AGENT: &str = concat!(
     ")"
 );
 
-/// An enum representing the [gateway opcodes].
-///
-/// [gateway opcodes]: https://discord.com/developers/docs/topics/opcodes-and-status-codes#gateway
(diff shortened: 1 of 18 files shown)


nickelc:
`_ =>` before the `Unknown(u8)` is required because macros don't backtrack.

PR #2010, opened 2022-06-23
fixing say command

DRuppFv:
If you didn't give an arg to say, it would print a error in the terminal and would not say anything.


nickelc:
can you run `cargo +nightly fmt` to format the code?

DRuppFv:
> can you run `cargo +nightly fmt` to format the code?

When i had done this i ran it with the wrong fmt configuration and just noticed now. Fixing ~~~~

arqunis (approved):

PR #2011, opened 2022-06-23
Cleanup collector module by replacing duplication with generics

GnomedDev:
This prevents the duplicate `Future` and `Stream` implementations on each collector by de-duplicating them all into `Collector<Item>`, and removes duplication on the Builders for each collector by de-duplicating them all into `CollectorBuilder<Item>`

Points of note:
- `FilterOptions` (the structs in each collector) had to be made public, is this avoidable?
- Type aliases were made to prevent breaking too much, although should they be deprecated/removed?
- The weird `manager: impl AsRef<ShardManager>` -> `manager.as_ref().clone()` was removed, as just... why?

### DIFF
--- a/examples/e17_message_components/src/main.rs
+++ b/examples/e17_message_components/src/main.rs
@@ -180,14 +180,18 @@ impl EventHandler for Handler {
             .unwrap();
 
         // Wait for the user to make a selection
-        let mci =
-            match m.await_component_interaction(&ctx).timeout(Duration::from_secs(60 * 3)).await {
-                Some(ci) => ci,
-                None => {
-                    m.reply(&ctx, "Timed out").await.unwrap();
-                    return;
-                },
-            };
+        let mci = match m
+            .component_interaction_collector(&ctx.shard)
+            .timeout(Duration::from_secs(60 * 3))
+            .collect_single()
+            .await
+        {
+            Some(ci) => ci,
+            None => {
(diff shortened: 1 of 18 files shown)


examples/e17_message_components/src/main.rs
```diff
@@ -180,14 +180,17 @@ impl EventHandler for Handler {
...
-            match m.await_component_interaction(&ctx).timeout(Duration::from_secs(60 * 3)).await {
-                Some(ci) => ci,
-                None => {
-                    m.reply(&ctx, "Timed out").await.unwrap();
-                    return;
-                },
-            };
+        let mci = match m
+            .await_component_interaction(ctx.shard.clone())
+            .timeout(Duration::from_secs(60 * 3))
```
DRuppFv, line 185: Why cant it be just 180? Isn't it just one more calculation for the machine?
If the reason is to make clear how many minutes it is, couldn't it have a comment?
```suggestion
            .timeout(Duration::from_secs(180)) // 3 minutes
```
(Yes i saw it repeating below vvvv)

GnomedDev: This shows as changed in my PR due to formatting, I didn't change this.

DRuppFv: I know, i just noticed that.

arqunis (approved):

PR #2017, opened 2022-06-26
Make Member::edit in-place

GnomedDev:
This allows for the weirdness in add_role, add_roles, remove_role, and remove_roles to just pass through to `self.edit`.

This could cause monomorphization bloat? Although I doubt it, it's much less code in edit than was in all the methods anyway, although I could make a `_edit` which directly takes a Builder? That would need a `guild_id._edit_member` though, so not much point

### DIFF
--- a/src/model/guild/member.rs
+++ b/src/model/guild/member.rs
@@ -115,30 +115,11 @@ impl Member {
         http: impl AsRef<Http>,
         role_id: impl Into<RoleId>,
     ) -> Result<()> {
-        self._add_role(&http, role_id.into()).await
+        self.add_roles(http, &[role_id.into()]).await
     }
 
-    async fn _add_role(&mut self, http: impl AsRef<Http>, role_id: RoleId) -> Result<()> {
-        if self.roles.contains(&role_id) {
-            return Ok(());
-        }
-
-        match http
-            .as_ref()
-            .add_member_role(self.guild_id.get(), self.user.id.get(), role_id.get(), None)
-            .await
-        {
-            Ok(()) => {
-                self.roles.push(role_id);
-
-                Ok(())
-            },
-            Err(why) => Err(why),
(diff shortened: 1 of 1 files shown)


src/model/guild/member.rs
```diff
@@ -148,28 +129,8 @@ impl Member {
...
-        {
-            Ok(member) => Ok(member.roles),
-            Err(why) => {
-                self.roles.retain(|r| !role_ids.contains(r));
-
-                Err(why)
-            },
-        }
+    pub async fn add_roles(&mut self, http: impl AsRef<Http>, role_ids: &[RoleId]) -> Result<()> {
+        self.edit(http, |b| b.roles(role_ids)).await
```
arqunis: This does not add new roles anymore; it instead removes whatever roles the member currently has and replaces them with those in `role_ids`. You will have to append the member's roles with `role_ids`:
```suggestion
        self.edit(http, |b| b.roles(self.roles.iter().chain(role_ids.iter()))).await
```

src/model/guild/member.rs
```diff
@@ -557,24 +498,8 @@ impl Member {
...
-        {
-            Ok(member) => Ok(member.roles),
-            Err(why) => {
-                self.roles.extend_from_slice(role_ids);
-
-                Err(why)
-            },
-        }
+    ) -> Result<()> {
+        self.edit(http, |b| b.roles(role_ids)).await
```
arqunis: This has the same problem as `add_roles`. You will have to do the opposite: filter out member's roles that are in `role_ids`:
```suggestion
        self.edit(http, |b| b.roles(self.roles.iter().filter(|r| !role_ids.contains(r)))).await
```

arqunis (changes requested):

arqunis (approved):

PR #2019, opened 2022-06-27
Cleanup dispatch and unnecessary clones in client

GnomedDev:
- Deduplicates a bit of dispatch_event
- Removes unnecessary wrapping of `ws_url` in `Arc<Mutex>`
- Removes unnecessary wrapping of `CacheAndHttp` in `Arc`
- Replaces `Mutex<HashMap>` with `DashMap` for `runners`
- Fixed a TOCTOU in `shutdown_all`


nathaniel-daniel:
I think `ws_url` should remain an `Arc<Mutex<String>>` as <https://discord.com/developers/docs/topics/gateway#get-gateway> seems to suggest that libraries should try to refresh it on connection failures, which Serenity currently isn't doing.

GnomedDev:
I believe it should be made back into an `Arc<Mutex<String>>` or `Arc<RwLock<String>>` (or even arcswap) when this functionality is actually implemented, I don't believe this should be left wrapped to facilitate a possible future change, as I can so clearly imagine it never happening.

arqunis (approved):

PR #2021, opened 2022-06-28
Implement `audit_log_reason` argument

SenseiHiraku:
I've noticed, that adding a reason for the audit_log, as implemented in other Discord API libraries, is not yet fully implemented here.
It may be breaking change for many functions, but it's a necessary one in my eyes, basically an addition to #1354


mkrasnitski:
Thanks for the PR! Unfortunate timing on your part, I suppose. If you look at ~~#1967~~ #2024, we're actually in the middle of an overhaul to how the builders are constructed and executed, which I expect will be finished and merged sometime next month. But, with the new system, adding support for `audit_log_reason` ought to be pretty easy, either as an additional method on the `builder`, or as a parameter to the `execute` function being introduced with every builder. I'm in favor of the former, which implies adding an `audit_log_reason` field to the builder struct and just marking it `#[serde(skip)]`, and feature-flagging it behind `"http"`.

mkrasnitski:
@SenseiHiraku Now that #2024 has been merged, if you'd like, you can take a stab at re-implementing this (if not let me know and I can tackle it). I think the best way to do so would be to add an optional `audit_log_reason` field to every builder marked `#[serde(skip)]`, as well as a corresponding method on the builder to set its value. Unfortunately I think it's best to close this PR in its current form and open a new one, as the changes it makes are no longer compatible with the current codebase.

arqunis:
Completed by https://github.com/serenity-rs/serenity/pull/2105

PR #2024, opened 2022-06-29
Builder paradigm overhaul, second attempt

mkrasnitski:
After working for a while on #1967, I realized that there were problems with the approach I was taking, and that solving them would require rewriting a good portion of what had already been pushed. So, I'm going to take a fresh start and hopefully get it right this time. Again, the goal is reduce monomorphization bloat due to generic closure arguments, allow for re-using preconstructed builders, and pave the way for later changes down the line that bring the library more in line with Discord's official docs (e.g. enforcing required fields, implementing `audit_log_reason`, etc.).

Here is the new plan:
1. Model methods will remain `async` and their return types will be unchanged, but they will now take a concrete `Builder` type as argument instead of a generic closure like `FnOnce(&mut Builder) -> &mut Builder`.
2. To reduce code duplication, builders will have an `execute` method that performs the API request. This helps in cases where model methods duplicate functionality between `Guild`/`PartialGuild`/`GuildId`, or `GuildChannel`/`ChannelId`, etc. I am initially in favor of exposing this method to users, though we may want to mark it `pub(crate)` instead, depending on consensus.
3. To allow for instantiating a builder inline via `Builder::default().method1().method2()`, methods on builders will take and return `Self` instead of `&mut Self`.

### DIFF
--- a/src/builder/create_application_command.rs
+++ b/src/builder/create_application_command.rs
@@ -340,8 +340,8 @@ impl CreateApplicationCommand {
             (Some(guild_id), None) => {
                 http.create_guild_application_command(guild_id.into(), &self).await
             },
-            (None, Some(channel_id)) => {
-                http.edit_global_application_command(channel_id.into(), &self).await
+            (None, Some(command_id)) => {
+                http.edit_global_application_command(command_id.into(), &self).await
             },
             (None, None) => http.create_global_application_command(&self).await,
         }
--- a/src/builder/create_application_command.rs
+++ b/src/builder/create_application_command.rs
@@ -1,9 +1,12 @@
 use std::collections::HashMap;
 
-use crate::json::prelude::*;
+#[cfg(feature = "http")]
+use crate::http::Http;
+use crate::internal::prelude::*;
+#[cfg(feature = "http")]
+use crate::model::application::command::Command;
 use crate::model::application::command::{CommandOptionType, CommandType};
-use crate::model::channel::ChannelType;
-use crate::model::Permissions;
+use crate::model::prelude::*;
 
 #[derive(Clone, Debug, Serialize)]
 pub struct CommandOptionChoice {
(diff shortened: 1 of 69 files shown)


mkrasnitski:
Well then, I think this PR is now ready for review. I've gone through and reworked everything in the `builder` directory. There are a few stray builders in the `utils` directory, specifically `MessageBuilder` and `CustomMesage`, but the former is so huge that probably it makes sense to tackle it separately, if at all. The latter isn't used anywhere in the codebase, so I'm not sure what to do with it. Is it actually used by downstream users?

As I've gone through, any extra breaking changes in each builder, such as removal of methods, or changing field types, have been noted in the corresponding commit messages. I could compile them all together if desired.

I think this should land *after* v0.11.3 is released, because there are some changes in `current`, as well as in currently-open PR's that I would like to rebase on top of.

mkrasnitski:
Rebased on next now that #2061 was merged, and also reworked the new EditAutoModRule builder and the new Webhook example. Should be all set for review.

EDIT: For some reason, github's listing of the commits above has some out of order/duplicated. I'd recommend just using the commits tab to get a list of all commits.

src/builder/create_application_command.rs
```diff
@@ -335,10 +311,48 @@ pub struct CreateApplicationCommand {
...
+        match (guild_id, command_id) {
+            (Some(guild_id), Some(command_id)) => {
+                http.edit_guild_application_command(guild_id.into(), command_id.into(), &self).await
+            },
+            (Some(guild_id), None) => {
+                http.create_guild_application_command(guild_id.into(), &self).await
+            },
+            (None, Some(channel_id)) => {
+                http.edit_global_application_command(channel_id.into(), &self).await
+            },
```
arqunis: Typo, commands aren't channels.
```suggestion
            (None, Some(command_id)) => {
                http.edit_global_application_command(command_id.into(), &self).await
            },
```

arqunis (approved):

PR #2031, opened 2022-06-30
Merge application command & autocomplete interactions

nickelc:
The structs are almost identical except for the `focused` field for the
current autocompleted option and that the `resolved` data map is not
sent for autocomplete interactions.

The entities from the `CommandData::resolved` data map are no longer
cloned for the option values during deserialization. But the resolved
options and values can be accessed via the `CommandData::options()` method.

For autocomplete interactions `CommandData::autocomplete()` returns the
current autocompleted option + value.

**BREAKING CHANGES:**
- `ApplicationCommandInteraction` is now used for application commands
  and their autocomplete interaction.

- The `CommandDataOptionValue`'s variants no longer have the cloned
  entities from the `CommandData::resolved` data map.
  See the `CommandData::options` method.

- The fields `value`, `kind`, `options` and `focused` of `CommandDataOption`
  are streamlined into the `CommandDataOptionValue` enum with new variants
  for subcommands, subcommand groups and autocomplete.


src/model/application/interaction/application_command.rs
```diff
@@ -462,39 +610,334 @@ pub struct CommandDataResolved {
...
+            Type,
+            Value,
+            Options,
+            Focused,
+            Unknown(String),
+        }
+
+        struct Visitor;
+
+        impl<'de> serde::de::Visitor<'de> for Visitor {
```
GnomedDev, line 645: Why is a hand written Visitor used here, instead of an intermediate struct?

src/model/application/interaction/application_command.rs
```diff
@@ -462,39 +610,334 @@ pub struct CommandDataResolved {
...
+            _ => (true, false),
+        };
+        let len = 2 + usize::from(value_or_options) + usize::from(focused);
+
+        let mut s = serializer.serialize_struct("CommandDataOption", len)?;
+
+        s.serialize_field("name", &self.name)?;
+        s.serialize_field("type", &self.value.kind())?;
+
+        match &self.value {
```
GnomedDev, line 765: Couldn't this be replaced by serializing `CommandDataOptionValue` (by deriving serialize and using `serde(untagged)`)

src/model/application/interaction/application_command.rs
```diff
@@ -369,6 +398,103 @@ pub struct CommandData {
...
+                        } else {
+                            resolved.roles.get(&RoleId(id.0)).map(ResolvedValue::Role)
+                        };
+                        value.unwrap_or(ResolvedValue::Unresolved(Unresolved::Mentionable(*id)))
+                    },
+                    CommandDataOptionValue::User(id) => resolved
+                        .users
+                        .get(id)
+                        .map(|u| ResolvedValue::User(u, resolved.members.get(id)))
+                        .unwrap_or(ResolvedValue::Unresolved(Unresolved::User(*id))),
```
GnomedDev, line 478: Can map_or be used instead?

nickelc: sure, it makes the code less readable.
```rust
CommandDataOptionValue::User(id) => resolved
    .users
    .get(id)
    .map_or(ResolvedValue::Unresolved(Unresolved::User(*id)), |u| {
        ResolvedValue::User(u, resolved.members.get(id))
    }),
```

GnomedDev: is that how rustfmt formats it, wow....

nickelc: yes, that's why I have decided against using it.

src/model/application/interaction/application_command.rs
nickelc: No, because the code flattens the data of `CommandDataOption` and `CommandDataOptionValue`.

src/model/application/interaction/application_command.rs
nickelc: The implementation with 5 fields is straight forward and doesn't involve additional proc-macro compile time.

arqunis (approved):

PR #2032, opened 2022-06-30
Fix timestamp having fallible From impl

GnomedDev:
This swaps the `From` impl into a `TryFrom` implementation defined in terms of `FromStr`. It also removes the redundant `From` implementation on `String` that was a small footgun (consumed `String` when only needed `&str`).

This also adds a warn for the nursery lint `clippy::fallible_impl_from` to detect for this usage again, as this lint has existed for ages with no open issues referring to it on the clippy repo, so I consider it stable enough for usage here, although if wanted to, I can remove the warn.

### DIFF
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -226,7 +226,7 @@ impl CreateEmbed {
     ///     async fn message(&self, context: Context, mut msg: Message) {
     ///         if msg.content == "~embed" {
     ///             let timestamp: Timestamp =
-    ///                 "2004-06-08T16:04:23".parse().expect("Invalid timestamp!");
+    ///                 "2004-06-08T16:04:23Z".parse().expect("Invalid timestamp!");
     ///             let _ = msg
     ///                 .channel_id
     ///                 .send_message(&context.http, |m| {
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -57,7 +57,7 @@ pub struct CreateEmbed {
     #[serde(skip_serializing_if = "Option::is_none")]
     thumbnail: Option<HoldsUrl>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    timestamp: Option<String>,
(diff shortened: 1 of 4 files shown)


src/builder/create_embed.rs
```diff
@@ -233,10 +225,12 @@ impl CreateEmbed {
     /// impl EventHandler for Handler {
     ///     async fn message(&self, context: Context, mut msg: Message) {
     ///         if msg.content == "~embed" {
+    ///             let timestamp: Timestamp =
+    ///                 "2004-06-08T16:04:23".parse().expect("Invalid timestamp!");
```
mkrasnitski:
```suggestion
    ///                 "2004-06-08T16:04:23Z".parse().expect("Invalid timestamp!");
```

Missing a UTC offset at the end of the timestamp which causes it to fail parsing. Alternatively you could put `+04:00` or whatever.

mkrasnitski (changes requested):

src/builder/create_embed.rs
GnomedDev: Haha, that means it was broken beforehand and panicked if actually used, nice catch

src/builder/create_embed.rs
```diff
@@ -206,16 +206,7 @@ impl CreateEmbed {
...
-    ///
-    /// - `2017-01-03T23:00:00`
-    /// - `2004-06-08T16:04:23`
-    /// - `2004-06-08T16:04:23`
-    ///
-    /// This timestamp must be in ISO-8601 format. It must also be in UTC format.
-    ///
-    /// You can also pass an instance of `chrono::DateTime<Utc>`,
-    /// which will construct the timestamp string out of it.
+    /// See the documentation of [`Timestamp`] for more information.
```
DRuppFv, line 209: You could leave the documentation link here

DRuppFv (changes requested):

arqunis (approved):

PR #2034, opened 2022-07-02
Replace `json::from_number` with `into()`

nickelc:
Both `Value` types implement `From` for their number variants.


nickelc:
This is technically a breaking change because the `ToNumber` trait was public.

PR #2035, opened 2022-07-03
Add support for `Auto Moderation` feature

nickelc:
Implements the `Auto Moderation` feature based on the current [documentation].

Change log: https://discord.com/developers/docs/change-log#auto-moderation

[documentation]: https://discord.com/developers/docs/resources/auto-moderation


mkrasnitski:
As a note, this will require a follow-up PR to de-hashmap the new builders on `next`.

kangalio:
What work is left to do here? Testing all new features, or reviewing the code for maintainability, or something else?

nickelc:
The PR is up for review.
I've tested the code for myself with a guy joining my test server for dropping some slurs.

src/model/guild/automod.rs
```diff
@@ -0,0 +1,859 @@
...
+use serde_value::{DeserializerError, Value};
+
+use crate::model::id::{ChannelId, GuildId, MessageId, RoleId, RuleId, UserId};
+
+/// Configured auto moderation rule.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Rule {
+    /// ID of the rule.
+    pub id: RuleId,
+    /// ID of the guild this rule belonfs to.
```
kangalio: Typo

src/builder/edit_automod_rule.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Set the event context the rule should be checked.
+    pub fn event_type(&mut self, event_type: EventType) -> &mut Self {
+        self.0.insert("event_type", u8::from(event_type).into());
+
+        self
+    }
+
+    /// Set the type of content which can trigger the rule.
+    ///
+    /// **None**: The trigger type can't be edited. Only its values.
```
kangalio: EditAutoModRule is almost a full copy of CreateAutoModRole. Maybe the same type can be used for both? With basically just a small change here: "\*\*No**t**e\*\*: **When editing a rule,** the trigger type can't be edited. Only its values."

src/model/guild/automod.rs
```diff
@@ -0,0 +1,859 @@
...
+    ///
+    /// Maximum of 20.
+    pub exempt_roles: Vec<RoleId>,
+    /// Channels that should not be affected by the rule.
+    ///
+    /// Maximum of 50.
+    pub exempt_channels: Vec<ChannelId>,
+}
+
+impl<'de> Deserialize<'de> for Rule {
```
kangalio: Why is a manual Deserialization implementation needed?

src/model/guild/automod.rs
```diff
@@ -0,0 +1,859 @@
...
+                    exempt_channels,
+                })
+            }
+        }
+
+        deserializer.deserialize_any(Visitor)
+    }
+}
+
+impl Serialize for Rule {
```
kangalio: And a manual Serialize?

src/model/guild/automod.rs
```diff
@@ -0,0 +1,859 @@
...
+    /// Timeout user for a specified duration.
+    ///
+    /// Maximum of 2419200 seconds (4 weeks).
+    ///
+    /// A `Timeout` action can only be setup for [`Keyword`] rules.
+    /// [`Permissions::MODERATE_MEMBERS`] permission is required to use the `Timeout` action type.
+    ///
+    /// [`Keyword`]: TriggerType::Keyword
+    /// [`Permissions::MODERATE_MEMBERS`]: crate::model::Permissions::MODERATE_MEMBERS
+    Timeout(u64),
```
kangalio: `Timeout { duration_seconds: u64 }` perhaps? To make the unit clear (might be minutes or milliseconds as well)

src/model/guild/automod.rs
```diff
@@ -0,0 +1,859 @@
...
+}
+
+#[derive(Deserialize, Serialize)]
+#[serde(rename = "ActionMetadata")]
+struct Timeout {
+    #[serde(rename = "duration_seconds")]
+    duration: u64,
+}
+
+impl<'de> Deserialize<'de> for Action {
```
kangalio, line 323: Why manual Deserialize :upside_down_face: ?

src/model/guild/automod.rs
```diff
@@ -0,0 +1,859 @@
...
+                    ActionType::Unknown(unknown) => Ok(Action::Unknown(unknown)),
+                }
+            }
+        }
+
+        deserializer.deserialize_any(Visitor)
+    }
+}
+
+impl Serialize for Action {
```
kangalio, line 391: You know the drill

If there's a good reason, it would be nice to have it commented above the impls

src/model/guild/automod.rs
```diff
@@ -0,0 +1,859 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use serde_test::Token;
+
+    use super::*;
+
+    #[test]
+    fn rule_trigger_serde() {
```
kangalio, line 480: Isn't it simpler to test the (De)Serialize impls against JSON? I.e. create a JSON literal, feed it into Deserialize, compare against expected value. Then serialize to JSON and compare against literal from before

src/model/guild/automod.rs
nickelc: it's needed to combine the `trigger_type` and `trigger_metadata` fields into the `Trigger` enum.

kangalio: `#[serde(flatten)]` may be the solution

src/model/guild/automod.rs
nickelc: serde supports only strings for [internally tagged] enums then it could be replaced the `#[serde(tag = "type")]`.

[internally tagged]: https://serde.rs/enum-representations.html#internally-tagged

src/builder/edit_automod_rule.rs
nickelc: That can probably said about every Create/Edit builder combination in serenity.

`EditAutoModRule` doesn't set the `trigger_type`, only the metadata.

src/model/guild/automod.rs
kangalio: Okay so `#[serde(flatten)]` on the trigger field and `#[serde(tag = "trigger_type", content = "trigger_metadata")]` ([adjacently tagged enum](https://serde.rs/enum-representations.html#adjacently-tagged)) on the Trigger enum _would_ have been the perfect solution _but_ serde still doesn't support integer tags [cuz dtolnay is radio silent on https://github.com/serde-rs/serde/pull/2056 :( ]

src/model/guild/automod.rs
nickelc: Changed it to `Timeout(std::time::Duration)`

src/model/guild/automod.rs
nickelc: I'm hesitated to use `#[serde(flatten)]` because its open issues and internal buffering.

kangalio: What open issues and internal buffering are you referring to?

src/builder/edit_automod_rule.rs
kangalio: Alright I guess then let's keep the duplication for consistency lol

src/model/guild/automod.rs
nickelc: Just what I have seen in general and the inactivity in that area over time.
https://github.com/serde-rs/serde/search?q=flatten+enum&type=issues

src/model/guild/automod.rs
```diff
@@ -0,0 +1,859 @@
...
+}
+
+/// Characterizes the type of content which can trigger the rule.
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub enum Trigger {
+    Keyword(Vec<String>),
+    HarmfulLink,
+    Spam,
+    KeywordPresent(Vec<KeywordPresentType>),
```
kangalio: A bunch of coordinated typos: Present -> Preset

nickelc: oof :sweat:

kangalio: Haha don't feel bad it happens

nickelc: yeah, happens when writing/reading is replaced by autocomplete/pattern-matching.

src/model/guild/automod.rs
nickelc: Would you like to see the same comment from the `Deserialize` impl here?

kangalio: Nah its fine

src/model/guild/automod.rs
kangalio: Seems to work fine here though https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b66f9cf04363bcd31a3c3c3f66c325aa

nickelc: I've adopted your implementation.

src/model/guild/automod.rs
nickelc: `serde_test` limits the scope to the actual code and is implementation-agnostic.

I could simplify the tests, because `Trigger` now implements `Deserialize`/`Serialize`.

src/builder/edit_automod_rule.rs
nickelc: Ok, I had the impression the builders come in pairs (create/edit) but it's actually either `Create*` or `Edit*`. 
I was not looking for a pattern of when to use one or the other.
I rolled a dice and `Edit` stays.

mkrasnitski: I haven't looked at this closely, but I will say there's a precedent here. Specifically, `CreateInteractionResponseFollowup` is used for both creating and editing followup interaction messages. I don't know if there's any other builders that are used for both.

src/model/guild/automod.rs
kangalio: We could do the same approach as in the Rule (de)serialization impl to chop off another few dozen lines of code. Should I implement or will you?

nickelc: Go ahead

src/model/guild/automod.rs
```diff
@@ -0,0 +1,666 @@
...
+    use std::time::Duration;
+
+    use serde_test::Token;
+
+    use super::*;
+
+    #[test]
+    fn rule_trigger_serde() {
+        #[derive(Debug, PartialEq, Deserialize, Serialize)]
+        struct Rule {
```
kangalio, line 482: Btw why redefine Rule for the test instead of using the existing Rule struct?

src/model/guild/automod.rs
kangalio: What do you mean with implementation-agnostic?

The thing is just, you can replace e.g. the below 26 lines of code with just 11 - a 2.5x improvement:
```rust
assert_eq!(
    serde_json::to_value(Rule {
        trigger: Trigger::Keyword(vec![String::from("foo"), String::from("bar")]),
    }),
    serde_json::json!({
        "trigger_type": 1,
        "trigger_metadata": {
            "keyword_filter": Some(["foo", "bar"]),
        },
    }),
);
```

And, well, yeah this test assumes working serde_json, but on the other hand you're code assumes a working serde_test so we're not limited to just the scope of the actual code either way

nickelc: This tests `serde_json`, what about `simd-json`?

src/model/guild/automod.rs
nickelc: This saves a bunch of tokens.

kangalio: Oh right the full Rule has a bunch more fields

src/model/guild/automod.rs
kangalio: I see

This works for both

```rust
assert_eq!(
    json::to_string(Rule {
        trigger: Trigger::Keyword(vec![String::from("foo"), String::from("bar")]),
    }),
    r#"{"trigger_type":1,"trigger_metadata":{"keyword_filter":["foo","bar"]}}"#,
);
```

src/model/guild/automod.rs
kangalio: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=43b420753caeca183e7c7e8400135ec9

src/builder/edit_automod_rule.rs
kangalio: Should probably rename to `CreateAutoModRole` then for consistency

src/model/guild/automod.rs
kangalio: I implemented this approach in https://github.com/nickelc/serenity/pull/3

src/builder/edit_automod_rule.rs
mkrasnitski: Found a counter-example in `EditRole`, which is used for both creation and editing of roles. I think this can stay named as is, to not cause further friction.

arqunis (approved):

PR #2037, opened 2022-07-05
bump moka

Milo123459:
moka 0.9 comes with lots of improvements, but also removes some deps

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -113,7 +113,7 @@ features = ["std"]
 version = "2.1"
 
 [dependencies.moka]
-version = "0.8"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -113,7 +113,7 @@ features = ["std"]
 version = "2.1"
 
 [dependencies.moka]
-version = "0.8"
+version = "0.9"
```
nickelc, line 116: from the changelog:
> When using experimental dash cache, opting out of sync will reduce the number of dependencies.

```suggestion
version = "0.9"
default-features = false
```

PR #2038, opened 2022-07-05
Add optional min/max length fields for application command string options

nickelc:
image

### DIFF
--- a/src/builder/create_application_command.rs
+++ b/src/builder/create_application_command.rs
@@ -273,6 +273,24 @@ impl CreateApplicationCommandOption {
 
         self
     }
+
+    /// Sets the minimum permitted length for this string option.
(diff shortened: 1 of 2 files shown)


src/builder/create_application_command.rs
```diff
@@ -273,6 +273,24 @@ impl CreateApplicationCommandOption {
 
         self
     }
+
+    /// Sets the minimum permitted length for this string option.
+    ///
+    /// The value of `min_length` must be greater or equal to `0`.
```
kangalio, line 279: Isn't this a bit redundant? Unsigned types in Rust cannot go below zero no matter what

kangalio (approved):

arqunis (approved):

PR #2043, opened 2022-07-07
updating some dependencies

DRuppFv:
There are more dependencies not updated, as bitflags, serde_json, chrono and probably some others, but i just updated the really outdated or that had performance updates ones

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,11 +26,11 @@ version = "1.0.7"
 optional = true
 
 [dependencies.simd-json]
-version = "0.4.14"
+version = "0.5.1"
 optional = true
 
 [dependencies.tracing]
-version = "0.1.23"
+version = "0.1.35"
 features = ["log"]
 
 [dependencies.command_attr]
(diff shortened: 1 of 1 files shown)


GnomedDev:
The only actual update here is simd-json, the others are semver compatible so will automatically be updated when selected for Cargo.lock

DRuppFv:
> The only actual update here is simd-json, the others are semver compatible so will automatically be updated when selected for Cargo.lock

Oh. I didn't know that. It had performance updates, then the PR is still valid, i think?

kangalio:
The tracing and reqwest updates are not helpful because it forces needless dependency updates on people who are still running older (but semver-compatible) dependency versions

kangalio:
Actually, isn't this a breaking change? We publically expose simd_json types, for example in the `serenity::json` module

I believe the PR should be retargetted at next

DRuppFv:
> Actually, isn't this a breaking change? We publically expose simd_json types, for example in the `serenity::json` module
> 
> I believe the PR should be retargetted at next

What do you mean with "retargetted at next"?

GnomedDev:
This should be bumped to 0.6, now that is out.

kangalio:
I think we can close this PR, there hasn't been activity on bumping the simd-json to the new latest version so the PR is outdated

PR #2058, opened 2022-07-19
Temporarily fix GuildChannel::message_count in a non-breaking way

GnomedDev:
This is purely to prevent Guild caching errors on startup for people using current, and should be removed when the next serenity release happens

### DIFF
--- a/src/model/channel/guild_channel.rs
+++ b/src/model/channel/guild_channel.rs
@@ -126,7 +126,9 @@ pub struct GuildChannel {
     pub rtc_region: Option<String>,
     /// The video quality mode for a voice channel.
     pub video_quality_mode: Option<VideoQualityMode>,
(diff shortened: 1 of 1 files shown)


src/model/channel/guild_channel.rs
```diff
@@ -120,6 +129,7 @@ pub struct GuildChannel {
     /// An approximate count of messages in the thread, stops counting at 50.
```
kangalio: Could update this "An approximate count of messages in the thread. Currently saturated at 255 (this limitation is removed in the next breaking serenity version)."

kangalio (approved):

arqunis (approved):

Miezhiko:
@GnomedDev why just changing it to `u32` is breaking change? just curious

kangalio:
Simple example `let count: u8 = channel.message_count` breaks with a type change

PR #2061, opened 2022-07-20
Fix clippy warnings and apply next patches to new code

GnomedDev:
- De-hashmaps Automod builders
- Adds Eq to automod derives
- Other clippy stuff

### DIFF
--- a/src/builder/edit_automod_rule.rs
+++ b/src/builder/edit_automod_rule.rs
@@ -65,8 +65,11 @@ impl EditAutoModRule {
     /// Set channels that should not be affected by the rule.
     ///
     /// Maximum of 50.
-    pub fn exempt_channels(&mut self, channels: Vec<ChannelId>) -> &mut Self {
-        self.exempt_channels = Some(channels);
+    pub fn exempt_channels(
+        &mut self,
+        channels: impl IntoIterator<Item = impl Into<ChannelId>>,
+    ) -> &mut Self {
+        self.exempt_channels = Some(channels.into_iter().map(Into::into).collect());
         self
     }
 }
--- a/src/builder/edit_automod_rule.rs
+++ b/src/builder/edit_automod_rule.rs
@@ -4,18 +4,18 @@ use crate::model::id::{ChannelId, RoleId};
 #[derive(Clone, Debug, Serialize)]
 pub struct EditAutoModRule {
     event_type: EventType,
-    #[serde(skip_serializing_if="Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none")]
     name: Option<String>,
-    #[serde(flatten, skip_serializing_if="Option::is_none")]
+    #[serde(flatten, skip_serializing_if = "Option::is_none")]
     trigger: Option<Trigger>,
-    #[serde(skip_serializing_if="Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none")]
     actions: Option<Vec<Action>>,
-    #[serde(skip_serializing_if="Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none")]
     enabled: Option<bool>,
-    #[serde(skip_serializing_if="Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none")]
     exempt_roles: Option<Vec<RoleId>>,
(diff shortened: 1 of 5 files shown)


src/builder/edit_automod_rule.rs
```diff
@@ -1,133 +1,83 @@
...
-    pub fn exempt_roles<I>(&mut self, roles: I) -> &mut Self
-    where
-        I: IntoIterator<Item = RoleId>,
-    {
-        let ids = roles.into_iter().map(|id| id.0.to_string()).collect();
-
-        self.0.insert("exempt_roles", ids);
-
+    pub fn exempt_roles(&mut self, roles: Vec<RoleId>) -> &mut Self {
+        self.exempt_roles = Some(roles);
```
mkrasnitski:
```suggestion
    pub fn exempt_roles(
        &mut self,
        roles: impl IntoIterator<Item = impl Into<RoleId>>,
    ) -> &mut Self {
        self.exempt_roles = Some(roles.into_iter().map(Into::into).collect());
```
I think this method as well as `exempt_channels` can keep the `IntoIterator` stuff, but change the internal item to `impl Into<Id>`. It lets them be more flexible, whereas the `actions` method sees no benefit from taking an `impl IntoIterator` vs a `Vec`

src/builder/edit_automod_rule.rs
```diff
@@ -1,133 +1,83 @@
...
-    pub fn exempt_channels<I>(&mut self, channels: I) -> &mut Self
-    where
-        I: IntoIterator<Item = ChannelId>,
-    {
-        let ids = channels.into_iter().map(|id| id.0.to_string()).collect();
-
-        self.0.insert("exempt_channels", ids);
-
+    pub fn exempt_channels(&mut self, channels: Vec<ChannelId>) -> &mut Self {
+        self.exempt_channels = Some(channels);
```
mkrasnitski, line 66: See comment above.

mkrasnitski (changes requested):

src/builder/edit_automod_rule.rs
GnomedDev: I prefer `Vec` because it prevents an unnecessary clone if passing a `Vec` in, however there is generally an inconsistency throughout the builder module, some discussion would happen off this PR about if we want to change all to Vec, or all to IntoIterator -> Vec, or something else.

mkrasnitski: I think most often, the user won't actually be passing in a straight `Vec<Id>`, rather a `Vec<u64>` is the most likely use-case, in which case the user would either call `.into` on each entry as they construct the vec, or would be forced to call `.map(Into::into).collect()`, which incurs the clone anyway. So, the ergonomic hit would be encountered more often than the extra clone.

GnomedDev: That sounds fair, okay

mkrasnitski: That also applies to `excempt_channels`.

src/builder/edit_automod_rule.rs
```diff
@@ -1,133 +1,86 @@
...
```
mkrasnitski:
```suggestion
    pub fn exempt_channels(
        &mut self,
        channels: impl IntoIterator<Item = impl Into<ChannelId>>,
    ) -> &mut Self {
        self.exempt_channels = Some(channels.into_iter().map(Into::into).collect());
```

See above discussion.

mkrasnitski (changes requested):

src/builder/edit_automod_rule.rs
kangalio: Subjectively I prefer taking straight types if possible, that is `Vec<Id>`. Also, people will be passing in `Vec<Id>` plenty of times; like when the list of IDs is not hardcoded but comes from a struct or whatever.

Final tidbit: there would be no unnecessary clone either way because the standard library specializes `.into_iter().map(/* non-size-increasing transformation*/).collect()` to reuse the allocation

mkrasnitski: Oh wow, I didn't know that. Do you know where I can read up more about this? I'm curious to know how they do it.

mkrasnitski (approved):

arqunis (approved):

PR #2064, opened 2022-07-24
Reduce unknown event log to debug level

GnomedDev:
Turns out that with the current state of this (not printing the event name) it is completely useless and spammy, especially when running on a large bot receiving a lot of events. This PR just reduces the warning log to a debug log, to prevent it appearing on production builds of bots (hopefully).

I've had this change on my personal branch for a while, and am just upstreaming it now to reduce the effort I need to do to rebase it.

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -5,7 +5,7 @@ use std::sync::Arc;
 use futures::channel::mpsc::UnboundedSender as Sender;
 use futures::future::{BoxFuture, FutureExt};
 use tokio::sync::RwLock;
-use tracing::{instrument, warn};
+use tracing::{debug, instrument, warn};
 use typemap_rev::TypeMap;
 
 #[cfg(feature = "gateway")]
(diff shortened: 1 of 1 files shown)


kangalio:
Shouldn't production bots just disable warning logs from serenity?

GnomedDev:
Other than this, there are no other spammy warning logs like it, and the warnings I do get from serenity are useful (noticing the timeouts on shard shutdowns, websocket close messages, etc).

I specifically made Gnomeutils' logger allow warn and above messages from libraries.

kangalio:
I see

And even if this warning gave the event type, it wouldn't be useful?

GnomedDev:
Kinda? It would help for events that are documented on ddev docs but not added to serenity, as discord has said that undocumented events shouldn't be relied on. Even then, if a PR was made to show the event type, this PR could be reverted (although I don't believe it should be even then)

arqunis (approved):

PR #2065, opened 2022-07-24
Remove unnecessary generics

GnomedDev:
These generics were mistakenly left behind after #2024 and can be removed


mkrasnitski:
Whoops, I was hoping I got all of them. Is there a clippy lint for this? If so, it might be worth adding it to the `#[deny]` list.

GnomedDev:
Seems not, although there is one for unused lifetimes, not for unused generics.

mkrasnitski:
Actually, I think it's related to the fact that the function is marked `pub` and isn't called anywhere in the library itself. If it was, then it would actually error (if no turbofish was provided) because it's impossible to infer the type. As a result, it would be user code that triggered the compile error, so I'm quite surprised there isn't at least an optional lint for it.

arqunis (approved):

PR #2067, opened 2022-07-24
Make select menu values optional to fix deserialization of msg component interactions

### LINKED ISSUE #2062: Message deserialization errors if it has select menu
If I got the details correct: #1986 accommodated for Discord sending the `values` field on components in the wrong place. Discord seems to have fixed the problem now so deserialization fails due to missing `values` field.

This was noticed via get_interaction_response calls failing with a select menu present in the retrieved message https://discord.com/channels/381880193251409931/919310428344029265/999359158635741224

bumblepie:
Fixes #2062

### DIFF
--- a/src/model/application/component.rs
+++ b/src/model/application/component.rs
@@ -153,7 +153,7 @@ pub struct SelectMenu {
     #[serde(default)]
     pub options: Vec<SelectMenuOption>,
     /// The result location for modals
-    pub values: Vec<String>,
+    pub values: Option<Vec<String>>,
 }
 
 /// A select menu component options.
--- a/src/model/application/component.rs
+++ b/src/model/application/component.rs
@@ -153,7 +153,8 @@ pub struct SelectMenu {
     #[serde(default)]
     pub options: Vec<SelectMenuOption>,
     /// The result location for modals
-    pub values: Option<Vec<String>>,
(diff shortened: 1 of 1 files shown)


kangalio:
One person [already had success](https://discord.com/channels/381880193251409931/381912587505500160/1001540546113847346) with this PR

kangalio (approved):

ForeverAnApple:
+1 on this PR. Thanks so much.

src/model/application/component.rs
```diff
@@ -153,7 +153,7 @@ pub struct SelectMenu {
     #[serde(default)]
     pub options: Vec<SelectMenuOption>,
     /// The result location for modals
-    pub values: Vec<String>,
+    pub values: Option<Vec<String>>,
```
arqunis: This is good, but would you mind changing the type back to `Vec<String>` and adding `#[serde(default)]`? This change constitutes as breaking, which isn't typically allowed on `current` (in certain cases it is, but here it's more viable to fix it temporarily and then add the real fix on `next`; that is, using an `Option`).
```suggestion
    #[serde(default)]
    pub values: Vec<String>,
```

arqunis (changes requested):

arqunis (approved):

PR #2069, opened 2022-07-25
Fix some builders returning unsendable Futures

mkrasnitski:
Due to how the permissions lookups were being performed in #2024, the Futures were being marked `!Send` because `GuildRef` and `CacheRef` in general is also `!Send`. So, we unfortunately have to clone the guild we get from the cache. Also factored out the permissions lookup into a function.


mkrasnitski:
Related issue for a relevant clippy lint: https://github.com/rust-lang/rust-clippy/issues/6947

arqunis (approved):

PR #2071, opened 2022-07-26
[HTTP] Remove unnecessary clones by passing around owned data

GnomedDev:
A lot of the methods on Request and AttachmentType were only given references or mutable references to their data, instead of taking owned data. This was unnecessary and caused a lot of cloning. This PR fixes this


GnomedDev:
Looking at the tests that failed, they are related to the AttachmentType `From` impl changing, but taking a second look it seems to be quite ambiguous anyway, should I remove these conversions and make it explicit for the user?

src/model/channel/attachment_type.rs
```diff
@@ -19,75 +19,95 @@ use crate::error::{Error, Result};
...
     /// Indicates that the [`AttachmentType`] is a [`File`]
-    File { file: &'a File, filename: String },
+    File { file: &'a File, filename: Cow<'static, str> },
     /// Indicates that the [`AttachmentType`] is a [`Path`]
     Path(&'a Path),
     /// Indicates that the [`AttachmentType`] is an image URL.
     Image(Url),
 }
 
+async fn data_path(path: &Path) -> Result<Vec<u8>> {
```
mkrasnitski, line 31:
```suggestion
#[cfg(feature = "http")]
async fn data_path(path: &Path) -> Result<Vec<u8>> {
```

You should feature-gate these, as they're only used if `http` is enabled (would also fix the import error).

mkrasnitski:
```suggestion
#[cfg(feature = "http")]
async fn data_path(path: &Path) -> Result<Vec<u8>> {
```

You should feature-gate this, as it's only used if `http` is enabled (would also fix the import error).

src/model/channel/attachment_type.rs
```diff
@@ -19,75 +19,95 @@ use crate::error::{Error, Result};
...
+    tokio::fs::read(path).await.map_err(Into::into)
+}
+
+async fn data_file(file: &File) -> Result<Vec<u8>> {
```
mkrasnitski:
```suggestion
#[cfg(feature = "http")]
async fn data_file(file: &File) -> Result<Vec<u8>> {
```

src/model/channel/attachment_type.rs
```diff
@@ -19,75 +19,95 @@ use crate::error::{Error, Result};
...
+    let mut data_buf = Vec::new();
+    file.try_clone().await?.read_to_end(&mut data_buf).await?;
+    Ok(data_buf)
+}
+
+async fn data_image(client: &Client, url: Url) -> Result<Vec<u8>> {
```
mkrasnitski:
```suggestion
#[cfg(feature = "http")]
async fn data_image(client: &Client, url: Url) -> Result<Vec<u8>> {
```

mkrasnitski (changes requested):

src/http/multipart.rs
```diff
@@ -20,27 +20,17 @@ pub struct Multipart<'a> {
...
-            let filename = file.filename()?;
-
-            // Modify current AttachmentType to Bytes variant to prevent the
-            // need for another disk read or network request when retrying
-            if let AttachmentType::Path(_) | AttachmentType::Image(_) = file {
-                *file = AttachmentType::Bytes {
-                    data: data.clone().into(),
-                    filename: filename.clone().unwrap_or_default(),
-                };
-            }
```
mkrasnitski, line 45: As far as I understand it, nowhere in the lib is a multipart actually re-used, since everything takes `Request`s by ownership instead of borrowing?

src/model/channel/attachment_type.rs
```diff
@@ -107,11 +127,11 @@ impl<'a> From<&'a PathBuf> for AttachmentType<'a> {
...
+impl<'a> From<(&'a File, Cow<'static, str>)> for AttachmentType<'a> {
+    fn from((file, filename): (&'a File, Cow<'static, str>)) -> AttachmentType<'a> {
         AttachmentType::File {
-            file: f.0,
-            filename: f.1.to_string(),
+            file,
+            filename,
         }
     }
 }
```
mkrasnitski: Github won't let me suggest changes directly here, but you can change this to take an `Into<Cow>` as follows:

```rust
impl<'a, S> From<(&'a File, S)> for AttachmentType<'a>
where
    S: Into<Cow<'static, str>>,
{
    fn from((file, filename): (&'a File, S)) -> AttachmentType<'a> {
        AttachmentType::File {
            file,
            filename: filename.into(),
        }
    }
}
```

src/model/channel/attachment_type.rs
```diff
@@ -19,75 +19,95 @@ use crate::error::{Error, Result};
...
+impl From<(Cow<'static, [u8]>, Cow<'static, str>)> for AttachmentType<'static> {
+    fn from((data, filename): (Cow<'static, [u8]>, Cow<'static, str>)) -> Self {
         AttachmentType::Bytes {
-            data: Cow::Borrowed(params.0),
-            filename: params.1.to_string(),
+            data,
+            filename,
         }
     }
 }
```
mkrasnitski: Same as above, I'd recommend replacing as follows:
```rust
impl<'a, S> From<(Cow<'static, [u8]>, S)> for AttachmentType<'a>
where
    S: Into<Cow<'static, str>>,
{
    fn from((data, filename): (Cow<'static, [u8]>, S)) -> Self {
        AttachmentType::Bytes {
            data,
            filename: filename.into(),
        }
    }
}
```
Btw, why did you change it to `AttachmentType<'static>`, was it to coerce `AttachmentType::Bytes(&'static [u8], &'static str)` to be `'static` as well? I'd say with the new impl it's better to loosen the lifetime again.

mkrasnitski: Note that it's currently not possible to change the `Cow<'static, [u8]>` into an `impl Into<Cow>` because the compiler complains about conflicting implementations, citing:
```
upstream crates may add a new impl of trait `From<&tokio::fs::File>` for type `Cow<'static, [u8]>` in future versions
```
Now, I think it's obvious this won't happen, but I don't really like the ergonomics as is, where the user is forced to call `.into()` on their arguments (which normally would be passed inline). I'm in favor of taking a simple `&[u8]` here and collecting into a Vec, but that would mean that `Cow::<'static, u8>::Borrowed` never actually ends up being constructed, so better then to just change the `data` field of the `Bytes` variant to a `Vec<u8>` to avoid confusion.

src/http/multipart.rs
GnomedDev: Yes

kangalio:
Clean diff here: https://github.com/serenity-rs/serenity/compare/5edbf0b3a720c528394a5d88325cb2ec804ae31b...GnomedDev:serenity:cleanup-http

Currently blocked on @GnomedDev reviewing https://github.com/GnomedDev/serenity/pull/1 I believe?

GnomedDev:
Sorry for the close and reopen, was accidental.

I've just removed most of the public AttachmentType changes, apart from changing the `Cow<'a, [u8]>` to `Vec<u8>` as previously it was useless (but the From implementation still uses Cow, someone else can bikeshed them).

mkrasnitski (changes requested):
I think it's fine to leave it as a `Cow<'a, [u8]>` as it'll cause fewer merge conflicts with #2163. There's no harm in leaving it. Other than that, this looks fine.

GnomedDev:
There is harm, it is misleading currently as if it is passed through it goes into `impl Into<Cow<'static, [u8]>>` which means that the Cow is useless as-is.

mkrasnitski:
Not sure what you mean, could you give an example?

mkrasnitski:
@GnomedDev now that #2163 has been merged, could you rebase this on top and remove the attachments changes?

PR #2072, opened 2022-07-27
Guild CacheRef kick function fix

Bloectasy:
Fix the Guild CacheRef in the kick_with_reason function on the member struct

### DIFF
--- a/src/model/guild/member.rs
+++ b/src/model/guild/member.rs
@@ -390,7 +390,8 @@ impl Member {
         #[cfg(feature = "cache")]
         {
             if let Some(cache) = cache_http.cache() {
-                if let Some(guild) = cache.guilds.get(&self.guild_id) {
+                let guild_ = cache.guilds.get(&self.guild_id).as_deref().cloned();
+                if let Some(guild) = guild_ {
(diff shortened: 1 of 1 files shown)


Bloectasy:
Is there another place where a cacheref issue still exists? So I can add it to this pr

src/model/guild/member.rs
```diff
@@ -390,7 +390,8 @@ impl Member {
         #[cfg(feature = "cache")]
         {
             if let Some(cache) = cache_http.cache() {
-                if let Some(guild) = cache.guilds.get(&self.guild_id) {
+                let guild_ = cache.guilds.get(&self.guild_id).as_deref().cloned();
+                if let Some(guild) = guild_ {
```
mkrasnitski:
```suggestion
                let lookup = cache.guild(self.guild_id).as_deref().cloned();
                if let Some(guild) = lookup {
```
This is the correct fix, as it also prevents the chance of deadlocking (since `cache.guilds.get()` returns a different type than `cache.guild()`)

mkrasnitski (approved):

arqunis (approved):

PR #2076, opened 2022-07-27
Simplify timestamp.rs implementations

### LINKED ISSUE #1926: It's possible to enable neither `chrono` or `time`
Not sure how that can even work but with `default-features = false` and `chrono` or `time` not in the features, the Timestamp type becomes kinda unusable in an unintuitive way? Can't investigate further right now. See discussion https://discord.com/channels/381880193251409931/381912587505500160/981185223985348618

Polyhistorian:
Made Deref & From<DateTime<Tz> implementations obey the same logic as the Timestamp itself, as there was a scenario where you could end up with no implementations if you had defined neither Chrono nor Time

Fixes #1926 as that edge case now defaults to having the Time implementations instead of nothing

### DIFF
--- a/src/model/timestamp.rs
+++ b/src/model/timestamp.rs
@@ -113,6 +113,20 @@ cfg_if::cfg_if! {
                 f.write_str(&s)
             }
         }
+
+        impl std::ops::Deref for Timestamp {
+            type Target = DateTime<Utc>;
+
+            fn deref(&self) -> &Self::Target {
+                &self.0
+            }
+        }
+
+        impl<Tz: TimeZone> From<DateTime<Tz>> for Timestamp {
+            fn from(dt: DateTime<Tz>) -> Self {
+                Self(dt.with_timezone(&Utc))
+            }
+        }
     } else {
         use dep_time::format_description::well_known::Rfc3339;
         use dep_time::serde::rfc3339;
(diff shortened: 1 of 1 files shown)


ChargedByte:
It would make sense to me to remove the `time` feature entirely ([Cargo.toml#L188](https://github.com/serenity-rs/serenity/blob/69484dc115cfa1fec15af5202c0d2f9f8b6ef949/Cargo.toml#L188)) as it serves no real purpose anymore. Since `time` is always included anyway ([Cargo.toml#L61-L64](https://github.com/serenity-rs/serenity/blob/69484dc115cfa1fec15af5202c0d2f9f8b6ef949/Cargo.toml#L61-L64)) defaulting to it would make sense and switching to `chrono` only when that feature is explicitly requested.

Polyhistorian:
Could be done, just didn't want to make that change in the same request as currently it would change nothing about working configurations, only adding implementations in that edge case. 

Where as not having that (meta) feature anymore would require it being deprecated ([which is possible](https://github.com/serenity-rs/serenity/blob/current/build.rs)), and having the knock on effect of making chrono not overrideable anyone with time while using default features as chrono is included in [them](https://github.com/serenity-rs/serenity/blob/69484dc115cfa1fec15af5202c0d2f9f8b6ef949/Cargo.toml#L162).

arqunis (approved):

PR #2094, opened 2022-08-16
Remove Arc around CacheAndHttp

GnomedDev:
This was one section of #2019 which I am splitting up to avoid breakage and test separately


arqunis:
Manually merged in https://github.com/serenity-rs/serenity/commit/0af335ad471e41612cd3971f5a3511b95257d74f to fix compilation of example.

PR #2100, opened 2022-08-18
Fix `BucketBuilder::delay_action` being poorly documented

### LINKED ISSUE #2099: delay_action doc comment fails to mention await_ratelimits
Relevant source file: [`buckets.rs line 472`](https://github.com/serenity-rs/serenity/blob/b5d7973ad47a61a79225ddb4d59d6dbb3faa34f5/src/framework/standard/structures/buckets.rs#L472).

`delay_action` 's doc comment fails to mention that for it to run, you need to set await_ratelimits to a non-zero value. 
On top of that, it doesn't even mention that you need to make a closure that passes a future from a function you intend to run (probably should be rewritten to avoid that). 

It is important that we write our documentation clearly, so issues like this don't arise. An obvious fix for this would be to rewrite the doc comment so that it mentions the weird quirk and optionally has an example showing it in practice.

NinekoTheCat:
fixes #2099 

This pull request adds better documentation of the usage of delay_action by adding an example along with more clear documentation.

### DIFF
--- a/src/framework/standard/structures/buckets.rs
+++ b/src/framework/standard/structures/buckets.rs
@@ -520,7 +520,6 @@ impl BucketBuilder {
     /// #     Ok(())
     /// # }
     /// ```
-    /// You can use this to for example send a custom response when someone exceeds the amount of commands they're allowed to make.
     #[inline]
     pub fn delay_action(&mut self, action: DelayHook) -> &mut Self {
         self.delay_action = Some(action);
--- a/src/framework/standard/structures/buckets.rs
+++ b/src/framework/standard/structures/buckets.rs
@@ -476,20 +476,24 @@ impl BucketBuilder {
     /// # Examples
     ///
     /// ```rust
+    /// # async fn run() -> Result<(), Box<dyn std::error::Error>> {
     /// use serenity::framework::standard::macros::{command, group};
     /// use serenity::framework::standard::{CommandResult, StandardFramework};
     /// use serenity::model::channel::Message;
     /// use serenity::prelude::*;
+    ///
     /// #[command]
-    /// #[bucket="example_bucket"]
+    /// #[bucket = "example_bucket"]
     /// async fn example_command(ctx: &Context, msg: &Message) -> CommandResult {
-    ///     msg.reply(ctx,"Example message, You can only repeat this once every 10 seconds").await?;
+    ///     msg.reply(ctx, "Example message, You can only repeat this once every 10 seconds").await?;
     ///
     ///     Ok(())
     /// }
+    ///
     /// async fn example_overuse_response(ctx: &Context, msg: &Message) {
-    ///     msg.reply(ctx,"I told you that you can't call this command less than every 10 seconds").await?;
+    ///     msg.reply(ctx, "I told you that you can't call this command less than every 10 seconds!").await?;
     /// }
+    ///
     /// #[group]
     /// #[commands(example_command)]
     /// struct General;
@@ -498,24 +502,23 @@ impl BucketBuilder {
     ///
     /// let framework = StandardFramework::new()
     ///     .configure(|c| c.prefix("~"))
-    ///     //we initialise the bucket with the closure
-    ///     .bucket("example_bucket", |b|b.delay_action(|msg,ctx| {
-    ///     //we return a future for the function we want to run
-    ///         Box::pin(example_overuse_response(ctx,msg))
+    ///     .bucket("example_bucket", |b| {
+    ///         // We initialise the bucket with the function we want to run
+    ///         b.delay_action(|msg, ctx| {
+    ///             Box::pin(example_overuse_response(ctx,msg))
+    ///         })
     ///     })
-    ///         //we set the delay to 10 seconds
-    ///         .delay(10)
-    ///         //we override the default behavior so that the function actually gets ran
-    ///         .await_ratelimits(1)
-    ///     //we await the return of the bucket
-    ///     ).await
+    ///     .delay(10) // We set the delay to 10 seconds
+    ///     .await_ratelimits(1)).await // We override the default behavior so that the function actually gets run
     ///     .group(&GENERAL_GROUP);
     ///
     /// let mut client = Client::builder(&token, GatewayIntents::default())
     /// .framework(framework)
     /// .await?;
     ///
-    /// //presumably you'd then start the client
+    /// client.start().await?;
+    /// #     Ok(())
+    /// # }
     /// ```
     /// You can use this to for example send a custom response when someone exceeds the amount of commands they're allowed to make.
     #[inline]
--- a/src/framework/standard/structures/buckets.rs
+++ b/src/framework/standard/structures/buckets.rs
@@ -470,8 +470,10 @@ impl BucketBuilder {
     }
 
     /// This function is called when a user's command invocation is delayed when:
-    /// 1. await_ratelimits is set to a non zero value (the default is 0).
-    /// 2. user's message rests comfortably within await_ratelimits (ex. if you set it to 1 then it will only respond once when the delay is first exceeded).
+    /// 1. `await_ratelimits` is set to a non zero value (the default is 0).
+    /// 2. user's message rests comfortably within `await_ratelimits` (ex. if you set it to 1 then it will only respond once when the delay is first exceeded).
+    ///
+    /// You can use this to, for example, send a custom response when someone exceeds the amount of commands they're allowed to make.
     ///
     /// # Examples
     ///
--- a/src/framework/standard/structures/buckets.rs
+++ b/src/framework/standard/structures/buckets.rs
@@ -493,7 +493,7 @@ impl BucketBuilder {
     /// }
     ///
     /// async fn example_overuse_response(ctx: &Context, msg: &Message) {
-    ///     msg.reply(ctx, "I told you that you can't call this command less than every 10 seconds!").await?;
+    ///     msg.reply(ctx, "I told you that you can't call this command less than every 10 seconds!").await.unwrap();
     /// }
     ///
     /// #[group]
--- a/src/framework/standard/structures/buckets.rs
+++ b/src/framework/standard/structures/buckets.rs
@@ -469,7 +469,55 @@ impl BucketBuilder {
         self
     }
 
-    /// This function will be called once a user's invocation has been delayed.
+    /**
+    This function is called when a user's command invocation is delayed when:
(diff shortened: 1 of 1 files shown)


src/framework/standard/structures/buckets.rs
```diff
@@ -469,7 +469,55 @@ impl BucketBuilder {
...
+        .group(&GENERAL_GROUP);
+    
+    let mut client = Client::builder(&token, GatewayIntents::default())
+    .framework(framework)
+    .await?;
+
+    //presumably you'd then start the client
+    ```
+    You can use this to for example send a custom response when someone exceeds the amount of commands they're allowed to make.
+    */
```
arqunis: Rust does not use Java-style `/***...***/` documentation. You have to put `///` before each line for the text to count as the documentation for an item (in this case, the `delay_action` function).

arqunis (changes requested):

src/framework/standard/structures/buckets.rs
NinekoTheCat: Cargo doc didn't find issues with it and my IDE kept freaking out because of it, please elaborate

NinekoTheCat: Nevermind rust RFC doesn't like it. Will change

arqunis: Oh? It seems `/**` is an alternative form for documentation comments, which I didn't know before. Anyway, this is not the style we use. We do `///` per line. Like so:
```suggestion
    ///  This function is called when a user's command invocation is delayed when:
    ///  1. await_ratelimits is set to a non zero value (the default is 0).
    ///  2. user's message rests comfortably within await_ratelimits (ex. if you set it to 1 then it will only respond once when the delay is first exceeded).
    ///  Example:
    ///  ```rust
    ///  use serenity::framework::standard::macros::{command, group};
    ///  use serenity::framework::standard::{CommandResult, StandardFramework};
    ///  use serenity::model::channel::Message;
    ///  use serenity::prelude::*;
    ///  #[command]
    ///  #[bucket="example_bucket"]
    ///  async fn example_command(ctx: &Context, msg: &Message) -> CommandResult {
    ///      msg.reply(ctx,"Example message, You can only repeat this once every 10 seconds").await?;

    ///      Ok(())
    ///  }
    ///  async fn example_overuse_response(ctx: &Context, msg: &Message) {
    ///      msg.reply(ctx,"I told you that you can't call this command less than every 10 seconds").await?;
    ///  }
    ///  #[group]
    ///  #[commands(example_command)]
    ///  struct General;

    ///  let token = std::env::var("DISCORD_TOKEN")?;

    ///  let framework = StandardFramework::new()
    ///      .configure(|c| c.prefix("~"))
    ///      //we initialise the bucket with the closure
    ///      .bucket("example_bucket", |b|b.delay_action(|msg,ctx| {
    ///          //we return a future for the function we want to run
    ///          Box::pin(example_overuse_response(ctx,msg))
    ///      })
    ///          //we set the delay to 10 seconds
    ///          .delay(10)
    ///          //we override the default behavior so that the function actually gets ran
    ///          .await_ratelimits(1)
    ///      //we await the return of the bucket
    ///      ).await
    ///      .group(&GENERAL_GROUP);
    ///  
    ///  let mut client = Client::builder(&token, GatewayIntents::default())
    ///  .framework(framework)
    ///  .await?;

    ///  //presumably you'd then start the client
    ///  ```
    ///  You can use this to for example send a custom response when someone exceeds the amount of commands they're allowed to make.
```

NinekoTheCat: implemented the change

src/framework/standard/structures/buckets.rs
```diff
@@ -469,7 +469,53 @@ impl BucketBuilder {
         self
     }
 
-    /// This function will be called once a user's invocation has been delayed.
+    /// This function is called when a user's command invocation is delayed when:
+    /// 1. await_ratelimits is set to a non zero value (the default is 0).
+    /// 2. user's message rests comfortably within await_ratelimits (ex. if you set it to 1 then it will only respond once when the delay is first exceeded).
+    /// Example:
```
arqunis:
```suggestion
    ///
    /// # Examples
    ///
```

src/framework/standard/structures/buckets.rs
```diff
@@ -469,7 +469,53 @@ impl BucketBuilder {
...
+    ///         .await_ratelimits(1)
+    ///     //we await the return of the bucket
+    ///     ).await
+    ///     .group(&GENERAL_GROUP);
+    ///
+    /// let mut client = Client::builder(&token, GatewayIntents::default())
+    /// .framework(framework)
+    /// .await?;
+    ///
+    /// //presumably you'd then start the client
```
arqunis:
```suggestion
    /// # async fn run() -> Result<(), Box<dyn std::error::Error>> {
    /// use serenity::framework::standard::macros::{command, group};
    /// use serenity::framework::standard::{CommandResult, StandardFramework};
    /// use serenity::model::channel::Message;
    /// use serenity::prelude::*;
    ///
    /// #[command]
    /// #[bucket = "example_bucket"]
    /// async fn example_command(ctx: &Context, msg: &Message) -> CommandResult {
    ///     msg.reply(ctx, "Example message, You can only repeat this once every 10 seconds").await?;
    ///
    ///     Ok(())
    /// }
    ///
    /// async fn example_overuse_response(ctx: &Context, msg: &Message) {
    ///     msg.reply(ctx, "I told you that you can't call this command less than every 10 seconds!").await?;
    /// }
    ///
    /// #[group]
    /// #[commands(example_command)]
    /// struct General;
    ///
    /// let token = std::env::var("DISCORD_TOKEN")?;
    ///
    /// let framework = StandardFramework::new()
    ///     .configure(|c| c.prefix("~"))
    ///     .bucket("example_bucket", |b| {
    ///         // We initialise the bucket with the function we want to run
    ///         b.delay_action(|msg, ctx| {
    ///             Box::pin(example_overuse_response(ctx,msg))
    ///         })
    ///     })
    ///     .delay(10) // We set the delay to 10 seconds
    ///     .await_ratelimits(1)).await // We override the default behavior so that the function actually gets run
    ///     .group(&GENERAL_GROUP);
    ///
    /// let mut client = Client::builder(&token, GatewayIntents::default())
    /// .framework(framework)
    /// .await?;
    ///
    /// client.start().await?;
    /// #     Ok(())
    /// # }
```

src/framework/standard/structures/buckets.rs
```diff
@@ -469,7 +469,53 @@ impl BucketBuilder {
...
+    /// ```
+    /// You can use this to for example send a custom response when someone exceeds the amount of commands they're allowed to make.
```
arqunis:
```suggestion
```

src/framework/standard/structures/buckets.rs
```diff
@@ -469,7 +469,53 @@ impl BucketBuilder {
         self
     }
 
-    /// This function will be called once a user's invocation has been delayed.
+    /// This function is called when a user's command invocation is delayed when:
+    /// 1. await_ratelimits is set to a non zero value (the default is 0).
+    /// 2. user's message rests comfortably within await_ratelimits (ex. if you set it to 1 then it will only respond once when the delay is first exceeded).
```
arqunis:
```suggestion
    /// This function is called when a user's command invocation is delayed when:
    /// 1. `await_ratelimits` is set to a non zero value (the default is 0).
    /// 2. user's message rests comfortably within `await_ratelimits` (ex. if you set it to 1 then it will only respond once when the delay is first exceeded).
    ///
    /// You can use this to, for example, send a custom response when someone exceeds the amount of commands they're allowed to make.
```

arqunis (changes requested):

src/framework/standard/structures/buckets.rs
```diff
@@ -469,7 +469,59 @@ impl BucketBuilder {
...
+    /// let framework = StandardFramework::new()
+    ///     .configure(|c| c.prefix("~"))
+    ///     .bucket("example_bucket", |b| {
+    ///         // We initialise the bucket with the function we want to run
+    ///         b.delay_action(|msg, ctx| {
+    ///             Box::pin(example_overuse_response(ctx,msg))
+    ///         })
+    ///     })
+    ///     .delay(10) // We set the delay to 10 seconds
+    ///     .await_ratelimits(1)).await // We override the default behavior so that the function actually gets run
```
arqunis:
```suggestion
    ///     .bucket("example_bucket", |b| {
    ///         // We initialise the bucket with the function we want to run
    ///         b.delay_action(|msg, ctx| {
    ///             Box::pin(example_overuse_response(ctx, msg))
    ///         })
    ///         .delay(10) // We set the delay to 10 seconds
    ///         .await_ratelimits(1) // We override the default behavior so that the function actually gets run
    ///     })
    ///     .await
```

arqunis (changes requested):

src/framework/standard/structures/buckets.rs
```diff
@@ -469,7 +469,60 @@ impl BucketBuilder {
...
+    /// #[commands(example_command)]
+    /// struct General;
+    ///
+    /// let token = std::env::var("DISCORD_TOKEN")?;
+    ///
+    /// let framework = StandardFramework::new()
+    ///     .configure(|c| c.prefix("~"))
+    ///     .bucket("example_bucket", |b| {
+    ///         // We initialise the bucket with the function we want to run
+    ///         b.delay_action(|msg, ctx| {
```
arqunis:
```suggestion
    ///         b.delay_action(|ctx, msg| {
```

arqunis (changes requested):

src/framework/standard/structures/buckets.rs
```diff
@@ -469,7 +469,60 @@ impl BucketBuilder {
...
+    /// #[command]
+    /// #[bucket = "example_bucket"]
+    /// async fn example_command(ctx: &Context, msg: &Message) -> CommandResult {
+    ///     msg.reply(ctx, "Example message, You can only repeat this once every 10 seconds").await?;
+    ///
+    ///     Ok(())
+    /// }
+    ///
+    /// async fn example_overuse_response(ctx: &Context, msg: &Message) {
+    ///     msg.reply(ctx, "I told you that you can't call this command less than every 10 seconds!").await?;
```
arqunis:
```suggestion
    ///     msg.reply(ctx, "I told you that you can't call this command less than every 10 seconds!").await.unwrap();
```

arqunis (changes requested):

arqunis (approved):

PR #2110, opened 2022-08-28
Make `description` field optional when creating slash commands

mkrasnitski:
This field is not allowed when creating user (context) commands, but is required for all other slash commands. Best to just make it optional again.

### DIFF
--- a/examples/e14_slash_commands/src/main.rs
+++ b/examples/e14_slash_commands/src/main.rs
@@ -74,13 +74,13 @@ impl EventHandler for Handler {
                 .expect("GUILD_ID must be an integer"),
         );
 
-        let c1 = CreateCommand::new("ping", "A ping command");
+        let c1 = CreateCommand::new("ping").description("A ping command");
 
-        let c2 = CreateCommand::new("id", "Get a user id").add_option(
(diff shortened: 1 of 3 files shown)


arqunis (approved):

examples/e14_slash_commands/src/main.rs
```diff
@@ -77,21 +77,25 @@ impl EventHandler for Handler {
         let c1 = CreateCommand::new("ping", "A ping command");
 
         let c2 = CreateCommand::new("id", "Get a user id").add_option(
-            CreateOption::new(CommandOptionType::User, "id", "The user to lookup").required(true),
+            CreateOption::new(CommandOptionType::User, "id")
```
arqunis: Shouldn't this change be applied on `CreateApplicationCommand` instead? I believe descriptions being required for options should be fine.

arqunis (changes requested):

examples/e14_slash_commands/src/main.rs
mkrasnitski: :facepalm: you're right. I got the two builders mixed up. The actual fix is now pushed :upside_down_face:

arqunis (approved):

PR #2118, opened 2022-09-02
Remove `CacheAndHttp`, plus cleanup of dispatch

mkrasnitski:
The `CacheAndHttp` struct has too similar a name to the `CacheHttp` trait, which might cause confusion. Therefore, I've removed the struct, and inlined its fields where it was used before. In addition, cleaned up some of `client/dispatch.rs`, for example the `dispatch` and `handle_event` functions now take a `Context` as parameter (to reduce cloning), and `dispatch` is now a regular `async fn`, rather than returning a `BoxFuture`. Also introduced an impl of `Debug` to `Context` so that the tracing instrumentation on `handle_event` remains unchanged.

### DIFF
--- a/src/client/bridge/gateway/shard_manager.rs
+++ b/src/client/bridge/gateway/shard_manager.rs
@@ -16,16 +16,18 @@ use super::{
     ShardQueuerMessage,
     ShardRunnerInfo,
 };
+#[cfg(feature = "cache")]
+use crate::cache::Cache;
 #[cfg(feature = "voice")]
 use crate::client::bridge::voice::VoiceGatewayManager;
 use crate::client::{EventHandler, RawEventHandler};
 #[cfg(feature = "framework")]
 use crate::framework::Framework;
 use crate::gateway::PresenceData;
+use crate::http::Http;
 use crate::internal::prelude::*;
 use crate::internal::tokio::spawn_named;
 use crate::model::gateway::GatewayIntents;
-use crate::CacheAndHttp;
 
 /// A manager for handling the status of shards by starting them, restarting
 /// them, and stopping them when required.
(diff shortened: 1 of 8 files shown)


kangalio:
Is it worth the churn? [referring to the "removing CacheAndHttp" part]

GnomedDev:
This PR also seems to be multi-purpose, and last time I tried cleaning up this code in a multi-purpose commit everything broke. I believe this should be split up into separate PRs (cleaning up dispatch, possibly removing cacheandhttp, etc)

mkrasnitski:
I made my best attempt not to touch any of the caching logic, just simplifying and cleaning things up. I tested with some example bots and it seems to work fine, but you're welcome to clone the branch and test some more substantial code with it. 

As for justification, I think being able to just say `ctx.{http,cache}` instead of `ctx.cache_and_http.{http,cache}` when implementing methods for the `EventHandler` trait is enough of a reason to do this. These two fields were wrapped (I presume) to hide the feature-flag on the cache, in an attempt to please the compiler when building without cache enabled. But the struct itself is rather meaningless otherwise.

mkrasnitski:
I made my best attempt not to touch any of the caching logic, just simplifying and cleaning things up. I tested with some example bots and it seems to work fine, but you're welcome to clone the branch and test some more substantial code with it.

As for justification, these two fields were wrapped (I presume) to hide the feature-flag on the cache, in an attempt to please the compiler when building without cache enabled. But the struct itself is rather meaningless otherwise, and only adds mental overhead. Plus, the `CacheHttp` trait abstracts over pretty much the exact same thing, but is useful in more places.

kangalio:
For #2138, I was thinking of storing CurrentUser in CacheAndHttp as a third field. Could we keep CacheAndHttp for now to keep that possibility open?

mkrasnitski:
Going to split this up into two separate PRs. Closing for now.

Issue #2119, opened 2022-09-04
Please tell me how to add the Message gotten by ChannelId::message() to the Cache.


WinLinux1028:
ChannelId::message() does not use cache to get the Message, and does not cache the Message it gets.
Therefore, I need to add the Message to the Cache somehow, but I don't know how to do that.

PR #2120, opened 2022-09-05
Changed ChannelId::message() to use Cache

WinLinux1028:
I have looked for a solution to [this issue](https://github.com/serenity-rs/serenity/issues/2119) but it does not seem to exist.
Therefore, I have fixed this issue.
However, this PR breaks compatibility when passing Http directly.


WinLinux1028:
Sorry, I have to send this PR to next branch, not current.

PR #2121, opened 2022-09-05
Changed to add Message to Cache when ChannelId::message() gets Message from API.

WinLinux1028:
This is a re-submission of [this PR](https://github.com/serenity-rs/serenity/pull/2120)

I have looked for a solution to [this issue](https://github.com/serenity-rs/serenity/issues/2119) but it does not seem to exist.
Therefore, I have fixed this issue.


kangalio:
If the message is edited or deleted after having been cached, the cache will be outdated. This is why the main cache is only filled with data from gateway, because the gateway is continually updated with everything that happens, so a gateway-filled cache isn't outdated.

But a secondary cache was recently introduced under the temp_cache flag which caches HTTP-retrieved data. Cached data from temp_cache is purged after some data to prevent the issue with outdated data. You could cache retrieved messages in the temp_cache. See the fields of Cache. It currently only temp-caches channels and users but you could add `temp_messages`

WinLinux1028:
> If the message is edited or deleted after having been cached, the cache will be outdated.

hmm... I think that is incorrect.
I made [a simple BOT](https://github.com/WinLinux1028/discordbot_cache_test) to test it and it seemed fine.
image

I think the cache is updated when the MessageUpdate event occurs, is that wrong?

kangalio:
That's true, but only if the GUILD_MESSAGES intent is enabled

kangalio:
Clean diff: https://github.com/serenity-rs/serenity/compare/5392c16a881ede6e1519761c8d846181e34b7ff8...WinLinux1028:serenity:add_msg_to_cache

kangalio:
Closing due to 5 month inactivity

Issue #1513, opened 2021-09-24
[Next] Message.category_id() and Message.Channel.category() always return None


tiritto:
_**EDIT:** It looks like the issue is not directly caused by the category_id() function itself, but rather something further down the line. See my 2nd comment for more information. I've left my original message intact to not disturb the flow for those, who might have seen the issue prior to my edit._

It appears that `category_id()` from `serenity::model::channel::Message` doesn't work. I've been trying to access it in multiple ways, but it just always returns "None". 

**Current behavior:**
The function returns `None` regardless whether channel is placed within a category or not.

**Expected behavior:**
The function returns `ChannelId` of its parent category, if any.

**Affected branches:** `next`

**Additional information:**
- Cache is properly initialized and does contain categories in question;
- `rustc -V`: rustc 1.54.0 (a178d0322 2021-07-26)

**Affected code:**
```rust
struct Handler;

#[async_trait]
impl EventHandler for Handler {
    async fn message(&self, ctx: Context, msg: Message) {
        let current_category = msg.category_id(ctx.cache); // Also tested with &ctx.cache
        println!("{:#?}", &current_category); // <- Returns "None"
        match current_category {
            Some(id) => { println!("This category has ID: {}", id.as_u64()) },
            None => println!("This channel has no category!") // <- Always triggered
        }
    }
}

#[tokio::main]
async fn main() {
    let mut client =
        Client::builder(TOKEN) // declaration removed on purpose
            .intents(GatewayIntents::GUILD_MEMBERS
                | GatewayIntents::GUILDS
                | GatewayIntents::GUILD_MESSAGES)
            .event_handler(Handler)
            .await
            .expect("Err creating client");
    if let Err(why) = client.start().await {
        println!("Client error: {:?}", why);
    }
}
```

tiritto:
After trying to come out with some workaround while this issue persists, I've found out that the issue is doesn't come from `Message.category_id()` function directly, but from something else along the way. It looks like it, because after trying to get `ChannelCategory` via `Message.Channel.category()` _(simplified for readability)_ I get the exact same problem of always being left out with `None`.

**Affected code fragment:**
```rust
    async fn message(&self, ctx: Context, msg: Message) {
        let category = &msg.channel(&ctx).await.map_or(None, |channel| channel.category());

        match category {
            Some(channel) => { println!("This category has ID: {}", channel.id.as_u64()) },
            _ => println!("This channel has no category!") // <- Always gets triggered
        }
    }
[...]

PR #2128, opened 2022-09-08
Remove Channel::Category

### LINKED ISSUE #1352: Cache and HTTP use different types for channel categories
When you request channel data about a channel category via its channel ID, serenity yields a different object depending on whether it uses HTTP or the cache. HTTP yields a `Channel::Category` and the cache yields a `Channel::Guild` with the `kind` field set to `Category`.

Both of these options kind of make sense but it would be ideal if one was decided upon. Having two separate representations of channel categories makes them harder to handle correctly ([example](https://discord.com/channels/381880193251409931/381912587505500160/844965959273545788)).

I was able to reproduce the problem with
```rust
// Replace with the ID of any channel category of a server your bot is on
let channel = ChannelId(736186039571316827);
dbg!(channel.to_channel(&ctx).await.unwrap()); // HTTP
dbg!(channel.to_channel(&ctx).await.unwrap()); // cache
```
([full code](https://pastebin.com/5vG1RGyJ))

kangalio:
Fixes #1352

Channel categories had two different representations: Channel::Guild(c) with c.kind == ChannelType::Category, and Channel::Category(c). Different parts of serenity used either of these two. This PR unifies serenity to the former.

(I chose the first representation because it results in a lot shorter code, less duplication, is closer to Discord's API, is consistent with how other channel types don't have their own Channel enum variant, makes Channel enum variants be just Guild and Private which is a clean and obvious differentiation.)


kangalio:
Presumably also fixes #1513

src/cache/event.rs
```diff
@@ -223,21 +186,12 @@ impl CacheUpdate for GuildDeleteEvent {
...
-                            // Remove the channel's cached messages.
-                            cache.messages.remove(channel_id);
-                        },
-                        Channel::Category(_) => {
-                            // Remove the category from the cache
-                            cache.categories.remove(channel_id);
-                        },
-                        _ => {},
-                    }
+                for (channel_id, _) in &guild.1.channels {
```
arqunis: Could you fix the clippy warning here?

src/model/guild/member.rs
```diff
@@ -196,7 +196,7 @@ impl Member {
         let member = guild.members.get(&self.user.id)?;
 
         for channel in guild.channels.values() {
-            if let Channel::Guild(channel) = channel {
+            if channel.kind != ChannelType::Category {
```
arqunis: Ditto.

arqunis (changes requested):

arqunis (approved):

PR #2138, opened 2022-09-10
Store CurrentUser in Context

### LINKED ISSUE #1953: cache and msg.is_own()
Apologies, this is not a bug report (maybe docs or feature request).

I upgraded serenity from a very old (2018ish) version to 0.11 and I had to change this code:

```
- use serenity::CACHE;
[...]
- if msg.author.id == CACHE.read().unwrap().user.id {
+ if msg.is_own(&context.cache) {
```

Long story short, I tried to make sense of the docs and source and I found no easier way to discern if the currently received message was sent by the bot itself than using `cache` because even the old version didn't feel right to me.

Now quoting the cache docs:

> If you’re asking this, the answer is likely “definitely yes” or “definitely no”; any in-between tends to be “yes”. If you are low on RAM, and need to run on only a couple MB, then the answer is “definitely no”.

So my actual question is more like: Am I just not looking at the right place? Because surely for this simple thing (is this my own message?) I shouldn't have to enable the cache?

The only thing in the examples I found would be `ready.user.name|id` from the `Handler::ready` callback, but is that the intended way to check this? And following up, wouldn't that be good to add to the basic example?

kangalio:
Aims to fix #1953 

We currently don't even store the bot user ID when cache is disabled, even though the bot user ID is very useful. This PR adds a new field `bot_user: CurrentUser` to `Context`. Not sure if this should be expanded to include `PartialCurrentApplicationInfo`, or maybe even the entire `Ready` struct.

But I'm quite unhappy with CurrentUser being in Context:
1. it's a breaking change so @Nogamara would need to wait until next breaking release (and me too, in poise, where this PR would be useful too)
2. serenity functions take `impl CacheHttp`, so they can't actually make use of the new data
    - we'd need to change the parameters to `impl AsRef<Context>`, but then people can't supply `CacheAndHttp` anymore which is also a dealbreaker
    - can we somehow smuggle CurrentUser into CacheAndHttp?

As demonstration, I changed Message::is_own to take `impl AsRef<Context>` to fix the original issue #1953. But `impl AsRef<Context>` is not a long-term solution.

**Ideas welcome!**


kangalio:
Instead of storing CurrentUser directly, there should probably be a new struct `BotContext` or `BotInfo` or similar, with fields for `CurrentUser` (useful for bot ID, name and discriminator), `CurrentApplicationInfo` (useful for application ID), `shard` (useful for [InviteGuild, GuildId]::shard_id) and potentially more

kangalio:
I'm thinking about an alternative way to achieve the goals: have `cache` be always on and instead add configuration for how much data is cached. That way, the absolute bare minimum (current user ID) is always cached, but people on resource constrained devices can just turn off caching of the rest to stay as lightweight as before. This would also simplify the codebase by removing all the `#[cfg(feature = "cache")]` special handling

kangalio:
This PR is never gonna be ready in its current state, and #2210 is a better way of achieving the goals anyways

PR #2139, opened 2022-09-11
Channel id name getch

### LINKED ISSUE #1911: ChannelId::name() returns None
https://github.com/serenity-rs/serenity/blob/current/src/model/channel/channel_id.rs#L538 This line will return `None` if the channel isn't in the cache. Thing is I'm pretty sure this is wrong. I expect this API to first check the cache, then if that fails populate the cache with data from Discord. If it's unable to populate with data from Discord, return `Result::Err(SerenityError)`. I didn't expect this to just fail if the cache hadn't been populated beforehand.

kangalio:
Fixes #1911 

I have a git stash where I adapt dozens of cache-only functions to cache+HTTP. But to be honest some of those seem ridiculous; e.g. I had to change `Message::is_own()` to call `http.get_current_user()` if cache is disabled, but realistically nobody with cache disabled would want that behavior because it's way too expensive making a request for every is_own() check... So... not sure where to draw the line where it's worth it to support HTTP fallback. So I just kept this PR small to fix only #1911. Feel free to tell if I should publish my other changes as well

### DIFF
--- a/src/model/channel/channel_id.rs
+++ b/src/model/channel/channel_id.rs
@@ -393,7 +393,10 @@ impl ChannelId {
     /// required permissions the HTTP-request will be issued. Additionally, you might want to
     /// enable the `temp_cache` feature to cache channel data retrieved by this function for a
     /// short duration.
-    #[allow(clippy::missing_errors_doc)]
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::Http`] if the channel retrieval request failed.
     #[inline]
     pub async fn to_channel(self, cache_http: impl CacheHttp) -> Result<Channel> {
         #[cfg(feature = "cache")]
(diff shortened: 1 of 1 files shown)


arqunis (approved):

arqunis:
> e.g. I had to change `Message::is_own()` to call `http.get_current_user()` if cache is disabled, but realistically nobody with cache disabled would want that behavior because it's way too expensive making a request for every is_own() check

Functions like `Message::is_own` should stay as "cacheful"; a HTTP request is computationally expensive to make, and if it is expected the function will be called often, requests should be minimised so that they're ideally not done at all. 

Also, in this case, `Message::is_own` is a function that's very simple - it uses the id of the author and the id of the bot to compare whether the message was produced by the bot or by someone else. The only reason it needs the cache is because user data about the bot isn't available locally, so it must be gotten from another location. If users do not like this, they can always make their own version of `Message::is_own` and caching of bot data, since as I said, it's simple to implement, and it only exists as a convenience function.

So my opinion would be to publish the cache-only -> cache+http changes only for functions where a HTTP request could be tolerated. For `Message::is_own` et al., they can be dropped.

Issue #2117, opened 2022-09-02
Establish a policy for the project's MSRV


mkrasnitski:
Serenity does not currently have a concrete policy for determining its MSRV. The general consensus seems to be that bumping MSRV is a breaking change, and should only happen on the next branch. I generally agree with this sentiment, but this aversion to bumping the MSRV has led to the current MSRV (1.53) having been released well over a year ago, while in the meantime a new language edition has been released. In fact, satisfying this MSRV requires the user pin some of serenity's dependencies to older versions, as those libraries have since released minor versions that bumped MSRV. Without such pinning, our de facto MSRV is actually 1.59.

To remedy this, I propose the following:
* The `current` branch should maintain a stable MSRV. This will occasionally mean that dependencies may break this, in effect raising our own MSRV, but in that case users may opt to pin those libraries' versions if they so desire. However, serenity will commit to never bump MSRV between minor releases, unless it is unavoidable.
* The `next` branch should track the latest Rust release as its MSRV. This will allow for swift development as new features are added to the language (such as the forthcoming GATs stabilization, or let_chains), and will reduce technical debt in the long run. When a major release is cut, then `current` will bump its MSRV to that of `next`, and that MSRV will be supported until the next major release is cut.

GnomedDev:
+1, having the CI having to pin deps to old versions to pass the MSRV test is not exactly a good situation. I'm not sure about only supporting the latest rust version, but it would help with development.

mkrasnitski:
Having CI pin those dependencies is a compromise I think we should make, because if instead the library tracked the highest MSRV among its dependencies, we would be denying users the ability to perform the pinning themselves, or in fact they'd have to pin serenity's version as well, which might preclude them from features and fixes.

I think tracking the latest Rust release on `next` is fine, because that branch can be considered bleeding-edge anyways. When a major release is cut, the MSRV will indeed be quite recent, but users can simply wait a few months to upgrade, at which point the MSRV will be sufficiently old enough that their own codebases might bump to it. I don't think it's worth it to track a Rust release, say, 2 versions behind the current one, because at the point when a major release is cut, it will already be a few months behind on new language features, and committing to supporting that for an extended period of time would introduce a fair bit of technical debt. Besides, not every Rust version will include features we can necessarily use, so this would actually be a sort of soft-MSRV, with the actual exact MSRV being some unspecified version between the MSRV on `current`, and the current Rust release.

kangalio:
Good proposal, I'm in favor

PR #2145, opened 2022-09-17
Bump MSRV to 1.63

mkrasnitski:
This bumps MSRV to 1.63, as an implementation of the policy proposed in #2117. This lets us use some new methods that have been stabilized since 1.53, and do some cleanup in the process. Also, I elected to remove the `backports` module and the dependency on `rustversion`, as going forward, backports will only ever need to happen on `current`, and would be much rarer. Therefore, we should only add this dependency back if absolutely necessary in the future.

Also, changed the MSRV build job to use `--all-features`, and removed any dependency pinning it did.


mkrasnitski:
#2157 should be merged first and `next` then rebased on `current`, then I can rebase this PR on top.

mkrasnitski:
Closing this and will open a new PR in a few days when 1.64 releases.

PR #2154, opened 2022-09-17
Add more fields to MessageUpdateEvent

### LINKED ISSUE #1183: Consider all fields of a message when receiving `MESSAGE_UPDATE` events
[Currently](https://github.com/serenity-rs/serenity/blob/current/src/model/event.rs#L910), the `MessageUpdateEvent` struct doesn't contain all fields of a message object, such as reactions. This results in out-of-date messages in the cache when unspecified fields are updated.

kangalio:
Fixes #1183

### DIFF
--- a/src/cache/event.rs
+++ b/src/cache/event.rs
@@ -480,48 +480,39 @@ impl CacheUpdate for MessageCreateEvent {
 impl CacheUpdate for MessageUpdateEvent {
     type Output = Message;
 
+    #[rustfmt::skip]
     fn update(&mut self, cache: &Cache) -> Option<Self::Output> {
-        if let Some(messages) = cache.messages.get_mut(&self.channel_id) {
-            if let Some(mut message) = messages.get_mut(&self.id) {
-                let item = message.clone();
+        // Destructure, so we get an `unused` warning when we forgot to process one of the fields
+        // in this method
+        #[allow(deprecated)] // yes rust, exhaustive means exhaustive, even the deprecated ones
+        let Self {
+            id, channel_id, content, edited_timestamp, tts, mention_everyone, mentions,
+            mention_roles, mention_channels, attachments, embeds, reactions, pinned, flags,
+            components, sticker_items,
 
-                if let Some(attachments) = self.attachments.clone() {
-                    message.attachments = attachments;
-                }
+            author: _, timestamp: _,  nonce: _, kind: _, stickers: _,  guild_id: _,
+        } = &self;
 
-                if let Some(content) = self.content.clone() {
-                    message.content = content;
-                }
+        let messages = cache.messages.get_mut(channel_id)?;
+        let mut message = messages.get_mut(id)?;
+        let old_message = message.clone();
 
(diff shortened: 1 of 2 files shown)


src/cache/event.rs
```diff
@@ -480,48 +480,39 @@ impl CacheUpdate for MessageCreateEvent {
...
-                if self.flags.is_some() {
-                    message.flags = self.flags;
-                }
-
-                return Some(item);
-            }
-        }
-
-        None
+        // Destructure, so we get an `unused` warning when we forgot to process one of the fields
```
arqunis:
```suggestion
        // Destructure, so we get an `unused` warning when we forget to process one of the fields
```

src/model/event.rs
```diff
@@ -406,22 +407,29 @@ pub struct MessageDeleteEvent {
...
+    pub nonce: Option<String>, // TODO: Is this a Message field that can even change?
+    pub pinned: Option<bool>,
+    pub kind: Option<MessageType>, // TODO: Is this a Message field that can even change?
     pub flags: Option<MessageFlags>,
+    pub components: Option<Vec<ActionRow>>,
+    #[deprecated(note = "deprecated by Discord")]
+    pub stickers: Option<Vec<StickerItem>>,
+    pub sticker_items: Option<Vec<StickerItem>>,
+
+    pub guild_id: Option<GuildId>, // TODO: Is this a Message field that can even change?
```
arqunis, line 432: Regarding these `TODO` comments, I do not think so? They are all fields that are filled in when the message is created, and should stay the way they are. Unless Discord adds the ability to, say, move messages from one place to another, which could change fields like `guild_id`. But that's unlikely to be added as a feature.

We can leave them in, though, for pedantic purposes (maybe Discord *could* go crazy and allow changing those fields). I would just remove the comments; they're redundant.

arqunis (changes requested):

src/model/event.rs
kangalio: > Regarding these TODO comments, I do not think so? They are all fields that are filled in when the message is created, and should stay the way they are. Unless Discord adds the ability to, say, move messages from one place to another, which could change fields like guild_id. But that's unlikely to be added as a feature.

I agree

> We can leave them in, though, for pedantic purposes (maybe Discord could go crazy and allow changing those fields).

If so, we should pull it through and add all of Message's fields. The reason I haven't done that is because then, for consistency, we should also process all those other fields in the CacheUpdate implementation, which is non-trivial. Also, it blows up the size of the type, and is potentially confusing to users who think those properties of a Message can actually change. What do you think? I can also open another PR to remove the author, timestamp, kind... fields on next.

arqunis: > Also, it blows up the size of the type, and is potentially confusing to users who think those properties of a Message can actually change. 

Yeah, when I was making that comment, I thought of the size of the struct that would get ridiculously large if we add all of the fields. If the fields can't be altered, perhaps it would just be best to remove them altogether. They can be re-added at a later date if Discord allows it in the future. You can remove them on `next`

arqunis (approved):

PR #2157, opened 2022-09-18
Add #2117-proposed MSRV policy to README.md

### LINKED ISSUE #2117: Establish a policy for the project's MSRV

kangalio:
Fixes #2117

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,20 +101,23 @@ serenity = "0.11"
 
 ## MSRV Policy
 
-Currently, serenity's minimum supported Rust version (MSRV) is Rust 1.53.
+Serenity's minimum supported Rust version (MSRV) is Rust 1.53.
 
-The `current` maintains a stable MSRV. This will occasionally mean that dependencies may break this,
-in effect raising our own MSRV, but in that case users may opt to pin those libraries' versions if
-they so desire. However, serenity will commit to never bump MSRV between minor releases, unless it
-is unavoidable.
+We opt to keep MSRV stable on the `current` branch. This means it will remain
+unchanged between minor releases. Occasionally, dependencies may violate SemVer
+and update their own MSRV in a breaking way. As a result, pinning their
+versions will become necessary to successfully build Serenity using an older
+Rust release. (**NOTE**: This is currently the case; building using Rust 1.53
+requires pinning `dashmap = "=5.2.0"`, `indexmap = "=1.8.2"`, and `time =
+"=0.3.9"`. If the `simd_json` feature is enabled, you must additionally pin
+`halfbrown = "=0.1.12"` and `value-trait = "=0.2.10"`. Without dependency
+pinning, **the de facto MSRV is Rust 1.59**.)
 
-The `next` branch should track the latest Rust release as its MSRV. This will allow for swift
-development as new features are added to the language (such as the forthcoming GATs stabilization,
-or let_chains), and will reduce technical debt in the long run. When a major release is cut, then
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,7 +99,20 @@ Add the following to your `Cargo.toml` file:
...
+they so desire. However, serenity will commit to never bump MSRV between minor releases, unless it
+is unavoidable.
+
+The `next` branch should track the latest Rust release as its MSRV. This will allow for swift
+development as new features are added to the language (such as the forthcoming GATs stabilization,
+or let_chains), and will reduce technical debt in the long run. When a major release is cut, then
+current will bump its MSRV to that of next, and that MSRV will be supported until the next major
+release is cut.
 
 **NOTE**: This is only achievable by pinning specific dependencies which have bumped their MSRV in a breaking manner. Specifically, you must pin `dashmap = "=5.2.0"`, `indexmap = "=1.8.2"`, and `time = "=0.3.9"`. If the `simd_json` feature is enabled, you must additionally pin `halfbrown = "=0.1.12"` and `value-trait = "=0.2.10"`. Without dependency pinning, **the de facto MSRV is Rust 1.59**.
```
```suggestion
## MSRV Policy

Serenity's minimum supported Rust version (MSRV) is Rust 1.53.

We opt to keep MSRV stable on the `current` branch. This means it will remain
unchanged between minor releases. Occasionally, dependencies may violate SemVer
and update their own MSRV in a breaking way. As a result, pinning their
versions will become necessary to successfully build Serenity using an older
Rust release. (**NOTE**: This is currently the case; building using Rust 1.53
requires pinning `dashmap = "=5.2.0"`, `indexmap = "=1.8.2"`, and `time =
"=0.3.9"`. If the `simd_json` feature is enabled, you must additionally pin
`halfbrown = "=0.1.12"` and `value-trait = "=0.2.10"`. Without dependency
pinning, **the de facto MSRV is Rust 1.59**.)

The `next` branch tracks the latest Rust release as its MSRV. This allows for
swift development as new languages features are stabilized, and reduces
technical debt in the long run. When a new major release is cut, the MSRV on
`current` will be updated to that of `next`, and we will commit to supporting
that MSRV until the following major release.
```

mkrasnitski (changes requested):
Good idea to document this for posterity. I did some editing: changed to present tense, moved some things around, and cleaned up wording for better clarity.

arqunis (approved):

PR #2162, opened 2022-09-18
Remove parse_token

### LINKED ISSUE #1906: token fails to parse creation time
note: i'm using commit `aad22f9981496e9d62b83e4d59c49ed8a634708e` but looking at the current source I expect the same issue would still occur.

I recently generated a discord token (regenerated it a few times and it happened consistently) where the second section doesn't parse correctly.

For example `OTc1MDAwMDAwMDAwMDAwMDA.G0mmRn.MDAw` (the first and last sections were edited)

This produces (results in just None being returned)
```rust
base64::decode_config(timestamp_base64, base64_config) = Err(
    InvalidLastSymbol(
        5,
        110,
    ),
)
```

Adding `let base64_config = base64_config.decode_allow_trailing_bits(true);` allows it to parse and the bot to run normally but the produced timestamp is `1751651526`/`2025-07-04T17:52:06` which isn't the actual creation time of the token.

A playground showing the issue
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3d7f33009749150372ab85ee726d09f0

kangalio:
Fixes #1906

### DIFF
--- a/src/utils/token.rs
+++ b/src/utils/token.rs
@@ -2,8 +2,6 @@
 
 use std::{fmt, str};
 
-use crate::model::id::UserId;
-
 /// Validates that a token is likely in a valid format.
 ///
 /// This performs the following checks on a given token:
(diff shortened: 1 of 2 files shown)


src/utils/token.rs
```diff
@@ -32,7 +30,18 @@ use crate::model::id::UserId;
...
-    parse(token).map(|_| ()).ok_or(InvalidToken)
+    // Tokens can be preceded by "Bot " (that's how the Discord API expects them)
+    let mut parts = token.as_ref().trim_start_matches("Bot ").split('.');
+
+    let is_valid = parts.next().map_or(false, |p| !p.is_empty())
+        && parts.next().map_or(false, |p| !p.is_empty())
+        && parts.next().map_or(false, |p| !p.is_empty())
+        && parts.next().is_none();
+
+    match is_valid {
```
arqunis: Could you fix the clippy warning here?

kangalio: Ah sadge we lint against that. I prefer the match because it's shorter and aligns the branches which is more pleasant to read. But oh well

Also I wonder why the clippy CI doesn't catch this. In general I feel like our CI is pretty leaky with regards to warnings

arqunis (approved):

PR #2163, opened 2022-09-18
Attachments overhaul

### LINKED ISSUE #1856: Cargo features should be additive
As documented in the Cargo reference [here][1], features that are not mutually exclusive should be additive. In other words, code that compiles with one set of features enabled should also compile when more features are added.

This is not always the case for the `serenity` crate, though. For example I've found these publicly exposed parts that may break things when features are added:
* The `EventHandler` trait has a few method signatures that change when the `cache` feature is enabled:
  * `channel_update`, `guild_create`, `guild_delete`, `guild_member_removal`, `guild_member_update`, `guild_role_delete`, `guild_role_update`, `guild_update`, `message_update`, `user_update`, and `voice_state_update`
* The `CurrentUser` and `User` types have a public field `accent_colour` whose type changes from `Option<u32>` to `Option<Colour>` if the `utils` feature is enabled. The `Role` and `Embed` types have a public field `colour` whose type changes from `u32` to `Colour` if the `utils` feature is enabled.
* The `ClientBuilder` type fails if the `framework` features is enabled, unless the `.framework` method is called.

This can introduce issues when using the crate, such as:
1. Non-additive features go against Rust conventions, which can lead to the library having surprising behavior.
1. Documentation being incomplete and/or wrong for the set of features actually being used, making the library harder to use correctly.
2. Library crates cannot reliably use types and traits from the `serenity` crate, since features could be enabled by other crates that would make them unable to compile, which means that code reuse becomes harder.

[1]: https://doc.rust-lang.org/cargo/reference/features.html#feature-unification

### LINKED ISSUE #2177: EditInteractionResponse is missing attachments support
Per the Discord documentation on [Edit Original Interaction Response](https://discord.com/developers/docs/interactions/receiving-and-responding#edit-original-interaction-response), the endpoint is identical to [Edit Webhook Message](https://discord.com/developers/docs/resources/webhook#edit-webhook-message) which supports attachments.

Without this function, the only way to achieve an equivalent functionality is by the following sequence of calls, where `T` is either `ApplicationCommandInteraction` or `MessageComponentInteraction` (only two I have checked):
```rs
T::create_interaction_response();
let message = T::get_interaction_response();
message.edit();
```
This introduces an extraneous API call which only serves as a workaround. If `EditInteractionResponse` supported attachments, the following code would suffice:
```rs
T::create_interaction_response();
T::edit_original_interaction_response();
```

kangalio:
Fixes #1856

**Edit: this PR evolved overtime. Here's the current description:**

- Change the AttachmentType enum to a simple struct that contains just the attachment data (ff5f952bea9fd00d188a735a0e16172c49cc76d5, 29ca1d414a198431c73019c439d598bbdd8e1288)
- The former Image, Path, File enums were converted to constructor methods on AttachmentType (ff5f952bea9fd00d188a735a0e16172c49cc76d5, c5dc7bcdd9b16b4f39f62cc724a50152d02ed1fc)
- AttachmentType renamed to CreateAttachment and moved to the builder model. It semantically fits perfect there (fce46c66c0be874d5a2dde97a7e6b0dd7c80089c)
- And the side effect of this clean up is that functions taking attachments don't need to evaluate the attachment themselves anymore, therefore not requiring the http feature anymore, therefore eliminating the need for `#[cfg(not(feature = "http"))]` alternative methods with incompatible signature (the original purpose of the PR, fixes #1856)
- Remove RequestBuilder, which completely duplicated Request (ec36f0e24223897538114a68eb5f8753719577a3)
- Fix EditMessage's attachment editing methods (d6472b582a63eee59906e9e2f0349e839b9384b5)
- Add `examples/testing` to test the above (d6472b582a63eee59906e9e2f0349e839b9384b5)
    - `examples/testing` could become a place for new PRs to put their testing code into
- Support editing interaction response attachments (fixes #2177) (fd77f9576efd1b3fd0f090b4e833f74e8e10a6fa)
- Integrate redundant _with_files and _and_attachments HTTP methods into the base methods (52815db47c0f9e48394dfbf668163084e0f02a2a)


mkrasnitski:
I originally added these methods for the "builder without model" usecase, as otherwise any builder methods that required the http feature would disappear, without a way to set the corresponding fields. I guess you can maybe get around this by serializing the builder to json first, then manually inserting the field before sending the request off.

kangalio:
> otherwise any builder methods that required the http feature would disappear

Why would you use builders with http feature disabled anyways? Isn't the builders' whole purposes to configure HTTP requests? (not all of them but the ones in this PR at least)

mkrasnitski:
If you're running serverless (i.e. Cloudflare workers), and using your own networking isn't an option, but modelling requests is still something you want to do. See #1770 for details on this.

kangalio:
Interesting, thank you

kangalio:
Update: I overhauled this PR again. Now, it does this:

- Change the AttachmentType enum to a simple struct that contains just the attachment data ([ff5f952](https://github.com/serenity-rs/serenity/pull/2163/commits/ff5f952bea9fd00d188a735a0e16172c49cc76d5))
- The former Image, Path, File enums were converted to constructor methods on AttachmentType ([ff5f952](https://github.com/serenity-rs/serenity/pull/2163/commits/ff5f952bea9fd00d188a735a0e16172c49cc76d5))
- AttachmentType renamed to CreateAttachment and moved to the builder model. It semantically fits perfect there ([fce46c6](https://github.com/serenity-rs/serenity/pull/2163/commits/fce46c66c0be874d5a2dde97a7e6b0dd7c80089c))

And the side effect of this clean up is that functions taking attachments don't need to evaluate the attachment themselves anymore, therefore not requiring the http feature anymore, therefore eliminating the need for `#[cfg(not(feature = "http"))]` alternative methods with incompatible signature.

mkrasnitski:
Nice job, I'd say this is definitely the right approach to take, and results in a much cleaner API. However there are a few concerns I have.

First, some tests are failing, and also some unused imports need to be feature flagged behind `http`. Second, the use of `std::fs::File` versus `tokio::fs::File` makes me wonder. Isn't it invalid to `.await` on methods called on a `std::fs::File`? I think the linter didn't catch that because of the import errors, but what would be the solution here? The `CreateAttachment::{path,file}` methods would have to be synchronous but only if the `http` feature is disabled. Or, maybe we provide separate synchronous versions of those methods.\

Also, I think the PR title should be changed to reflect the additional changes you've made.

kangalio:
#2177 fixed by fd77f9576efd1b3fd0f090b4e833f74e8e10a6fa

arqunis (approved):

PR #2167, opened 2022-09-18
Deny Clippy warnings in CI

kangalio:
With this, Rust and Clippy warnings should actually be caught in CI

### DIFF
--- a/examples/e16_sqlite_database/src/main.rs
+++ b/examples/e16_sqlite_database/src/main.rs
@@ -1,6 +1,8 @@
 // It is recommended that you read the README file, it is very important to this example.
 // This example will help us to use a sqlite database with our bot.
 
+use std::fmt::Write as _;
+
 use serenity::async_trait;
 use serenity::model::prelude::*;
 use serenity::prelude::*;
@@ -59,7 +61,7 @@ impl EventHandler for Bot {
 
             let mut response = format!("You have {} pending tasks:\n", todos.len());
             for (i, todo) in todos.iter().enumerate() {
-                response += &format!("{}. {}\n", i + 1, todo.task);
+                writeln!(response, "{}. {}", i + 1, todo.task).unwrap();
             }
 
             msg.channel_id.say(&ctx, response).await.unwrap();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,6 +73,7 @@
     clippy::missing_panics_doc, // clippy::unwrap_used
 )]
 #![deny(rustdoc::broken_intra_doc_links, rustdoc::private_intra_doc_links)]
+#![warn(clippy::unwrap_used, clippy::non_ascii_literal, clippy::let_underscore_must_use)]
 #![cfg_attr(test, allow(clippy::unwrap_used))]
 #![type_length_limit = "3294819"] // needed so ShardRunner::run compiles with instrument.
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -140,7 +140,6 @@ impl AsRef<Cache> for CacheAndHttp {
 #[cfg(feature = "client")]
 impl AsRef<Http> for CacheAndHttp {
     fn as_ref(&self) -> &Http {
-        let _ = 12_u64 as u8;
         &self.http
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,7 +52,7 @@
 //! [gateway docs]: crate::gateway
 #![doc(html_root_url = "https://docs.rs/serenity/*")]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
-#![deny(rust_2018_idioms)]
+#![warn(rust_2018_idioms)]
 #![warn(clippy::pedantic)]
 #![allow(
     // Allowed to avoid breaking changes.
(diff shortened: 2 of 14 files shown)


kangalio:
Ah look what's failing the CI: now-functioning Clippy warnings!

mkrasnitski:
Rust 1.64 brought some new lints, particularly one showstopper that I already addressed in #2174. But, as it stands, feel free to fix the rest with additional commits to this PR.

EDIT: Some of the lints were actually part of 1.65, my mistake. Right now, on 1.64, the current branch does not trigger clippy.

src/lib.rs
```diff
@@ -71,6 +71,8 @@
     clippy::cast_lossless,
     // Covered by other lints
     clippy::missing_panics_doc, // clippy::unwrap_used
+    // Callbacks with `+ 'static + Send + Sync`, trigger this pretty easily
+    clippy::type_complexity,
```
mkrasnitski:
```suggestion
```
I think you should instead define a type alias for the collector like:`type Filter<T> = Arc<dyn Fn(&Arc<T>) -> bool + 'static + Send + Sync>;`, and then just replace the uses of it. The next branch essentially has this, but souped up with GATs and the like.

src/model/guild/role.rs
```diff
@@ -307,7 +307,7 @@ impl FromStrAndCache for Role {
 ///
 /// [Discord docs](https://discord.com/developers/docs/topics/permissions#role-object-role-tags-structure).
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
-#[cfg_attr(test, derive(PartialEq))]
+#[cfg_attr(test, derive(PartialEq, Eq))]
```
mkrasnitski, line 310:
```suggestion
#[cfg_attr(test, derive(Eq, PartialEq))]
```
Keep these in alphabetical order.

mkrasnitski:
@kangalioo Thoughts on the suggested changes?

kangalio:
Totally forgot about them. Will do later

src/model/guild/role.rs
kangalio: I agree with the goal of having consistency, however, serenity doesn't sort derives in lots of places

Screenshot_20220927_154207

For consistency, all should be fixed, but in a separate PR. But even then, I think it's futile to manually order derives. The slight aesthetic satisfaction is not worth the laborious editing.

Eventually rustfmt should get support to automatically sort derives

src/lib.rs
```diff
@@ -71,18 +71,8 @@
...
 )]
 #![deny(rustdoc::broken_intra_doc_links, rustdoc::private_intra_doc_links)]
-#![deny(
-    clippy::unwrap_used,
-    clippy::non_ascii_literal,
-    clippy::missing_errors_doc,
-    clippy::let_underscore_must_use,
-    clippy::unused_async,
-    unused
-)]
```
mkrasnitski, line 82: I'm not sure we need to remove these entirely. AFAIK, `unwrap_used`, `non_ascii_literal`, and `let_underscore_must_use` are part of `clippy::restriction`, so better to keep them as warnings. Plus, I'm not sure if `clippy::unused_async` is implied by Rust's `unused`, since I think that's part of the compiler itself (do they interface at all?). We need to make sure we're not accidentally allowing any of these by removing them.

mkrasnitski (changes requested):

src/lib.rs
kangalio: You're totally correct, thanks 😅

mkrasnitski:
I think you missed `warn(unused)`, unless it's implied by something else.

src/lib.rs
```diff
@@ -72,14 +72,7 @@
...
 )]
-#![deny(
-    clippy::unwrap_used,
-    clippy::non_ascii_literal,
-    clippy::missing_errors_doc,
-    clippy::let_underscore_must_use,
-    clippy::unused_async,
-    unused
-)]
+#![warn(clippy::unwrap_used, clippy::non_ascii_literal, clippy::let_underscore_must_use, unused)]
```
mkrasnitski: One final nit, you may want to combine this with the `#![warn(clippy::pedantic)]` up above.

src/lib.rs
```diff
@@ -135,6 +140,7 @@ impl AsRef<Cache> for CacheAndHttp {
 #[cfg(feature = "client")]
 impl AsRef<Http> for CacheAndHttp {
     fn as_ref(&self) -> &Http {
+        let _ = 12_u64 as u8;
```
mkrasnitski: Missed a spot ;)

kangalio: Oh crap I wanted to verify my "else the blanket warnings would override the individual allow's" claim but forgot to take the test code out :sweat_smile: Thanks for your rigorous reviewing

examples/e16_sqlite_database/src/main.rs
```diff
@@ -1,6 +1,8 @@
 // It is recommended that you read the README file, it is very important to this example.
 // This example will help us to use a sqlite database with our bot.
 
+use std::fmt::Write as _;
```
mkrasnitski, line 4:
```suggestion
use std::fmt::Write;
```

Ok, last one for real

kangalio: Why? The Write trait is not referenced anywhere, we only need it in scope for `writeln!` to work

mkrasnitski: I think these have the same effect, right? Technically, the first one means you can't name the trait explicitly in code, but I don't think it's a problem, plus it looks a bit messy.

kangalio: I think this is entirely subjective

To make sure, I asked on the community server (https://discord.com/channels/273534239310479360/818964227783262209/1025179266209562654) and they could find no strong objective reason for one or the other. The tendency was towards `as _` to avoid conflicts if the user copy-pastes the example, adds file IO and imports `std::io::Write`.

If there are no other objections I'd keep this line as is?

mkrasnitski: Ok, makes sense. I won't belabor the point further.

mkrasnitski:
LGTM!

mkrasnitski (approved):

arqunis (approved):

PR #2170, opened 2022-09-19
Use Id types everywhere (instead of u64 or NonZeroU64)

kangalio:
Type safety, yay! This PR actually revealed a bug, maybe two:
- The message delete endpoint was rate limited using message ID as bucket, not Channel ID: https://github.com/serenity-rs/serenity/compare/next...kangalioo:id-types-everywhere?expand=1#diff-235176c24a85ae6dd7f1d22902684c132ba7e40fea30ae1420de3ad674359bdcR1831
- Many webhook-related endpoints share the same rate limit bucket; potentially causing the rate limit exceeding issues https://github.com/serenity-rs/serenity/issues/1121#issuecomment-1250652911

In addition, I found and removed an unused route in the HTTP routing code: LeaveGroup, ~~bots can't be in group DMs~~ (wrong, see below, thanks @GnomedDev)

### DIFF
--- a/src/builder/add_member.rs
+++ b/src/builder/add_member.rs
@@ -49,7 +49,7 @@ impl AddMember {
         guild_id: GuildId,
         user_id: UserId,
     ) -> Result<Option<Member>> {
-        http.as_ref().add_guild_member(guild_id.into(), user_id.into(), &self).await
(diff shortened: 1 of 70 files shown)


GnomedDev:
Bots can be in group DMs, it was an old oauth2 thing.

arqunis (approved):

PR #2172, opened 2022-09-20
Fix http and gateway feature

kangalio:
Previously, you couldn't even build serenity if you had just http or just gateway enabled. And, gateway used to depend on http and utils which is non sensical. Both these things are fixed now

There's room for more feature improvements, which I noted in two TODO's in Cargo.toml. Also, serenity is littered with #[cfg(feature = "")]'s everywhere, which can probably be greatly cleaned up by distributing code across modules better. I might make future PRs about that

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -173,25 +173,43 @@ default_no_backend = [
     "utils",
 ]
 
+# Enables builder structs to configure Discord HTTP requests. Without this feature, you have to
+# construct JSON manually at some places.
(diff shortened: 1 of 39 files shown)


kangalio:
MSRV failure is unrelated as far as I can tell

arqunis (approved):

kangalio:
Merge conflicts resolved! ^^ @acdenisSK

PR #2181, opened 2022-09-26
Automatically generate most GuildId wrapper methods

kangalio:
Related: #1905 (bullet point 2)

I made sure not to accidentally remove any methods in the porting process by diffing the method list from rustdoc before and after, for GuildId, PartialGuild and Guild each.

Guild has the following previously-missing wrapper methods added:
- get_preview, get_welcome_screen, get_widget, roles, sticker, stickers

PartialGuild has the following previously-missing wrapper methods added:
- add_member, create_scheduled_event, delete_scheduled_event, edit_scheduled_event, get_active_threads, get_preview, get_welcome_screen, get_widget, roles, scheduled_event, scheduled_event_users, scheduled_event_users_optioned, scheduled_events, sticker, stickers


mkrasnitski:
To be honest, this feels a little too magical and hacky for me, specifically around inserting cache checks. It's pretty ugly to decorate the function definitions with what looks like arbitrary syntax. IMO work should first be done to unify cache checks, because right now there's a lot of boilerplate related to them.

kangalio:
> IMO work should first be done to unify cache checks, because right now there's a lot of boilerplate related to them.

What do you have in mind about unification?

mkrasnitski:
We should factor cache checks into a function so we'd be able to just call `check_perms_cache(cache_http, permission)?`, preferably without decorating it with the cache feature gate as well. Then, more methods on `GuildId` should accept `impl CacheHttp`, such that the cache lookup would happen in `GuildId` methods and the wrapper calls wouldn't need to perform it themselves.

I remember trying a stab at this once, but there's several different ways serenity checks cache for permissions that it was hard to wrap my head around it. For all the perms checks on guilds, there already exists `utils::user_has_guild_perms` which I used as part of the builder rework. We could use that here, but keep in mind that it performs an extra cache lookup to retrieve the Guild (so that `GuildId` can also have access to it). There's also `utils::user_has_perms_cache` which is used for channels (this is a terrible function name btw, it should be changed). There's some other custom ways permissions checks are handled, but yeah. I think separate PRs should be filed first to get that all sorted, before auto-generating wrappers is attempted.

kangalio:
> We should factor cache checks into a function so we'd be able to just call check_perms_cache(cache_http, permission)?

Good idea, this really makes things cleaner

> cache lookup would happen in GuildId methods and the wrapper calls wouldn't need to perform it themselves

Then PartialGuild could no longer delegate to GuildId, because the PartialGuild's data is not available via cache

> there's several different ways serenity checks cache for permissions

All checks but one work the same way: check current user's permission in guild, compare against required permissions, throw InvalidPermissions. In the latest commit I factored out that common behavior. The only exception is guild.delete(), which I have kept as is

kangalio:
> It's pretty ugly to decorate the function definitions with what looks like arbitrary syntax.

I changed the syntax to be attribute-like in 861d077, maybe you like it better @mkrasnitski

mkrasnitski:
Having the cache check decorator at all feels ugly imo, especially if it's a large block.

> > cache lookup would happen in GuildId methods and the wrapper calls wouldn't need to perform it themselves
> 
> Then PartialGuild could no longer delegate to GuildId, because the PartialGuild's data is not available via cache
> 

Could you elaborate on this? Because this is exactly what I did during the builder overhaul. I moved the cache check into the `Builder::execute` method, which meant (like it would here) that`Guild` and `PartialGuild` now do an extra cache lookup using their id to get a full `Guild` out of cache.

kangalio:
> Guild and PartialGuild now do an extra cache lookup using their id to get a full Guild out of cache.

PartialGuild's permission checks are currently done using PartialGuild's data itself. So, with your proposed change, you lose the permission checks on PartialGuild if the guild is not cached

kangalio:
> Having the cache check decorator at all feels ugly imo, especially if it's a large block.

I'm open to alternative designs, if you have ideas

mkrasnitski:
It's the PartialGuild caching that I'd like more clarification on. Under what circumstances will we be dealing with a PartialGuild where the Guild itself has not been cached (there likely are some, I just want them listed because I'm not that familiar with serenity's caching architecture). Is it possible maybe to cache PartialGuilds and evict them when a full Guild with the same id comes in, and have cache lookups return either a partial or full guild depending on what's cached?

kangalio:
> Under what circumstances will we be dealing with a PartialGuild where the Guild itself has not been cached

When the cache feature is disabled and the user retrieved a PartialGuild with `PartialGuild::get()`, `GuildId::to_partial_guild()` etc. In the future, when the cache becomes more configurable, guilds may also potentially not be present in cache to limit RAM usage

kangalio:
What's the state of this PR?

mkrasnitski:
Any reason why `{PartialGuild, Guild}::{ban, ban_with_reason}` aren't autogenerated?

kangalio:
Good question

kangalio:
What's the state of this PR now?

src/model/guild/guild_id.rs
```diff
@@ -337,10 +351,13 @@ impl GuildId {
         builder.execute(cache_http, self).await
     }
 
-    /// Creates an emoji in the guild with a name and base64-encoded image.
+    /// Creates an emoji in the guild with a name and base64-encoded image. The
+    /// [`utils::read_image`] function is provided for you as a simple method to
+    /// read an image and encode it into base64, if you are reading from the
+    /// filesystem.
```
mkrasnitski: Rework this, as `utils::read_image` no longer exists.

src/model/guild/guild_id.rs
```diff
@@ -352,12 +369,11 @@ impl GuildId {
     ///
     /// # Errors
     ///
-    /// Returns [`Error::Http`] if the current user lacks permission,
-    /// if the name is too long, or if the image is too big.
+    /// Returns [`Error::Http`] if the current user lacks permission.
     ///
     /// [`EditProfile::avatar`]: crate::builder::EditProfile::avatar
+    /// [`utils::read_image`]: crate::utils::read_image
```
mkrasnitski: Same as above.

src/model/guild/guild_id.rs
```diff
@@ -203,9 +232,12 @@ impl GuildId {
...
     ///
     /// [Ban Members]: Permissions::BAN_MEMBERS
-    #[inline]
-    pub async fn ban(self, http: impl AsRef<Http>, user: impl Into<UserId>, dmd: u8) -> Result<()> {
-        self._ban(http, user.into(), dmd, None).await
+    #[cache_check = if let Some(cache) = cache_http.cache() {
+        self.require_perms(&cache_http, Permissions::BAN_MEMBERS).await?;
+        self.check_hierarchy(cache, user)?;
+    }]
+    pub async fn ban(self, cache_http: impl CacheHttp, user: UserId, dmd: u8) -> Result<()> {
```
mkrasnitski, line 239: A shame that `impl Into<UserId>` had to be changed to `UserId`. The cache check here is the reason for that change, right? Is there any way to possibly resolve this? If not, the signature change is causing a doctest to fail, so just change that callsite to pass `user.id` instead of `user`.

src/model/guild/guild_id.rs
```diff
@@ -214,35 +246,18 @@ impl GuildId {
...
-        reason: impl AsRef<str>,
-    ) -> Result<()> {
-        self._ban(http, user.into(), dmd, Some(reason.as_ref())).await
-    }
-
-    async fn _ban(
-        self,
-        http: impl AsRef<Http>,
+        cache_http: impl CacheHttp,
         user: UserId,
```
mkrasnitski, line 256: Same as above.

mkrasnitski (changes requested):

src/model/guild/guild_id.rs
kangalio: Oh, are the `impl Into<UserId>` something we want to keep?

src/model/guild/guild_id.rs
kangalio: `_ban` is an internal method so there's no concern for API convenience. In fact, `_ban` was using `UserId` directly before this PR already

mkrasnitski: Sorry, the diff there was kinda weird, but this was referring to `ban_with_reason`.

src/model/guild/guild_id.rs
mkrasnitski: Unsure, but quite a lot of other methods take `impl Into<UserId>`, including `kick` and `kick_with_reason`. The only ones that take a raw `UserId` are `audit_logs` and `_ban`.

PR #2182, opened 2022-09-26
Take &CacheAndHttp instead of &Context in argument_convert

kangalio:
That way, you can use it outside of gateway event contexts, like with the Interactions Endpoint URL


mkrasnitski:
Why not take `impl CacheHttp` instead? I'd like to ultimately sunset `CacheAndHttp`, as long as the issue around `CurrentUser` in #2138 can find a resolution; I'd rather not introduce further dependencies on it.

kangalio:
I think I was subconsciously driven to that decision by the desire to reduce monomorphization in serenity... but you're right it should be impl CacheHttp for consistency and to keep existing invocations with Context working

arqunis (approved):

mkrasnitski:
@acdenisSK This was accidentally merged with an incorrect commit message. I meant to suggest a change to the PR name to reflect the change from taking `&CacheHttp` to taking `impl CacheHttp`, but forgot.

PR #2183, opened 2022-09-27
Deny rustdoc lints only in CI

kangalio:
Both those lints are warn-by-default, so all we need to do is deny warnings in CI to catch documentation errors

Setting those lints to deny always is quite annoying because if you're developing, and quickly wanna check out the docs of your current state of work, you're forced to fix up all links (context switch), which you'd normally do just once at the end of coding the PR

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -28,7 +28,7 @@ jobs:
 
       - name: Build docs
         env:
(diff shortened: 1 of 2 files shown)


kangalio:
Oh and also, with `#[deny()]`, you cannot generate documentation when some features are disabled
Screenshot_20220927_184819

arqunis (approved):

PR #2188, opened 2022-09-27
Fix token validation base64 config to be forgiving to Discord's non-standard encoding

### LINKED ISSUE #2187: Discord token base64 encoding uses non-standard encoding, resulting in occasional errors
I was trying to use the `utils::token::validate` function to validate my Discord tokens and noticed that it occasionally panicked due to base64 encoding errors. This seemed strange to me as the tokens were valid; I tracked it down and I think the solution is to use a forgiving base64 decode call.

I am also not sure if Discord tokens ever use padding (i.e. should `URL_SAFE_NO_PAD` be used?) but it doesn't seem to matter in practice because padding is optional for the `URL_SAFE` config.

zbuc:
Includes testcase

Fixes #2187

### DIFF
--- a/src/utils/token.rs
+++ b/src/utils/token.rs
@@ -22,6 +22,7 @@ use crate::model::id::UserId;
 ///
 /// // ensure a valid token is in fact a valid format:
 /// assert!(validate("Mjg4NzYwMjQxMzYzODc3ODg4.C_ikow.j3VupLBuE1QWZng3TMGH0z_UAwg").is_ok());
+/// assert!(validate("Mjg4NzYwMjQxMzYzODc3ODg4.GzRABc.j3VupLBuE1QWZng3TMGH0z_UAwg").is_ok());
 ///
 /// assert!(validate("Mjg4NzYwMjQxMzYzODc3ODg4").is_err());
 /// assert!(validate("").is_err());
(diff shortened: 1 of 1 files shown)


kangalio:
Are you sure the extracted data is correct? I was under the impression Discord had changed the format. Maybe the more forgiving base64 parsing is just reading wrong data now?

By the way, in the next branch (the next major release of serenity), the parse function was removed #2162

zbuc:
@kangalioo The timestamp for my test case is a bit in the future (2025) but seems valid-ish? I'm not sure what they're supposed to be, but that's exactly what was in the token Discord returned to me

kangalio:
The timestamp is supposed to be when that token was generated

zbuc:
That seems wrong, then. Do you know what library Discord might be using for base64 encoding? The value in the added test case came directly from a Discord-generated token. They seem to have some non-standard encoding.

kangalio:
They don't use a public library as far as I know; their token format is custom

kangalio:
Since the implementation of this PR is wrong, and we can't fix it because Discord doesn't publish token structure, and even if we reverse-engineered it Discord could change it beneath our feet like has happened recently - I think we should close this PR and keep the parse function removed (#2162)

PR #2194, opened 2022-09-30
Add cached messages to channel_delete event handler

### LINKED ISSUE #2180: [Request] Remove channel messages from cache after `ChannelDeleteEvent` is handled or provide cached messages to event handler
The [`CacheUpdate`](https://github.com/serenity-rs/serenity/blob/next/src/cache/event.rs#L80) implementation for `ChannelDeleteEvent` Removes all cached messages unlike when messages are specifically deleted. It could be useful to log messages from within a channel after is deleted and then remove them from the cache or if the cached messages could be provided as a `MessageIterator`  like `channel_messages_field` outputs or could just be the same output as:
```rs
ctx.cache.channel_messages_field(channel.id.0, |s| {
        s.filter_map(|m| {
            if m.channel_id.0 == channel.id.0 {
                Some(m.clone())
            } else {
                None
            }
        })
    .collect::<Vec<_>>()
});
```

An example usecase;

When a channel is deleted in my server I log the information of the channel and upload it to my own hastebin for archival but since messages are removed when the event fires there is obviously no messages to be put into the logged channel's info.

Example output of deleted ticket channel in current state: https://paste.buymymojo.net/afubewecudoveriz

~~This issue is a little messy, sorry!~~

louboecker:
The Reason I'm making this PR is issue #2180

If there is a better way to do this I'd be happy to implement that instead (especially if it doesn't introduce a breaking change). 

Updating the cache after the event is handled is not possible without having to clone `event.channel` which I didn't wanna do.

Closes #2180

### DIFF
--- a/src/cache/event.rs
+++ b/src/cache/event.rs
@@ -78,9 +78,9 @@ impl CacheUpdate for ChannelCreateEvent {
 }
 
 impl CacheUpdate for ChannelDeleteEvent {
-    type Output = ();
+    type Output = Vec<Message>;
 
-    fn update(&mut self, cache: &Cache) -> Option<()> {
+    fn update(&mut self, cache: &Cache) -> Option<Vec<Message>> {
         match &self.channel {
             Channel::Guild(channel) => {
                 let (guild_id, channel_id) = (channel.guild_id, channel.id);
@@ -97,9 +97,12 @@ impl CacheUpdate for ChannelDeleteEvent {
         };
 
         // Remove the cached messages for the channel.
-        cache.messages.remove(&self.channel.id());
-
-        None
+        match cache.messages.remove(&self.channel.id()) {
+            Some((_, messages)) => {
+                Some(messages.values().map(|msg| msg.to_owned()).collect::<Vec<_>>())
+            },
+            None => None,
+        }
     }
 }
 
--- a/src/cache/event.rs
+++ b/src/cache/event.rs
@@ -100,7 +100,7 @@ impl CacheUpdate for ChannelDeleteEvent {
         cache
             .messages
             .remove(&self.channel.id())
-            .map(|(_, messages)| messages.values().cloned().collect())
+            .map(|(_, messages)| messages.into_values().collect())
     }
 }
 
--- a/src/cache/event.rs
+++ b/src/cache/event.rs
@@ -97,12 +97,10 @@ impl CacheUpdate for ChannelDeleteEvent {
         };
 
         // Remove the cached messages for the channel.
-        match cache.messages.remove(&self.channel.id()) {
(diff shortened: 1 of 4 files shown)


kangalio:
CI fails if the cache is disabled. The code should only get the messages from cache if it is enabled.

For consistency, this should happen in the CacheUpdate implementation for ChannelDeleteEvent, like already done for member/guild/role/thread deletes. That also takes care of only running the code when cache is enabled

louboecker:
I'm confused, do I have to do something or is this alright now?

mkrasnitski:
You'll have to rebase your PR on top of next. I recommend doing `git remote add upstream git@github.com:serenity-rs/serenity.git` and `git fetch upstream next` if you haven't already, then rebase or cherry-pick your commits on top of it, then do a force push to your origin.

src/cache/event.rs
```diff
@@ -97,9 +97,12 @@ impl CacheUpdate for ChannelDeleteEvent {
...
         // Remove the cached messages for the channel.
-        cache.messages.remove(&self.channel.id());
-
-        None
+        match cache.messages.remove(&self.channel.id()) {
+            Some((_, messages)) => {
+                Some(messages.values().map(|msg| msg.to_owned()).collect::<Vec<_>>())
+            },
+            None => None,
+        }
```
mkrasnitski:
```suggestion
        cache
            .messages
            .remove(&self.channel.id())
            .map(|(_, messages)| messages.values().cloned().collect())
```
You can use `Option::map` and `Iterator::cloned` here, and `collect` can actually infer its return type in this case.

kangalio (approved):
LGTM. If a maintainer could run CI that'd be great

louboecker:
I always forget to format with nightly haha

src/cache/event.rs
```diff
@@ -97,9 +97,10 @@ impl CacheUpdate for ChannelDeleteEvent {
         };
 
         // Remove the cached messages for the channel.
-        cache.messages.remove(&self.channel.id());
-
-        None
+        cache
+            .messages
+            .remove(&self.channel.id())
+            .map(|(_, messages)| messages.values().cloned().collect())
```
arqunis: To avoid cloning the messages redundantly, you can change this to either:
```suggestion
            .map(|(_, messages)| messages.into_iter().map(|(_, m)| m).collect())
```
or, since we are bumping the MSRV to 1.63, to:
```suggestion
            .map(|(_, messages)| messages.into_values().collect())
```
(`HashMap::into_values` was added in 1.54)

arqunis (changes requested):

arqunis (approved):

PR #2197, opened 2022-10-01
Replace usage of `CacheAndHttp` in dispatch with `Context`

mkrasnitski:
This is part of an effort to split #2118 up to address some of the concerns raised there. This PR changes dispatch such that it passes a context around instead of constructing one over and over, simplifying method signatures and reducing the number of clones (the clones themselves were cheap anyways because it was just Arc cloning, which means just incrementing a refcount).

The whole `DispatchEvent::update_cache` thing is still really ugly, as its only used in the absence of an non-raw event handler. Ideally I'd like to get rid of it somehow.

### DIFF
--- a/src/client/bridge/gateway/shard_runner.rs
+++ b/src/client/bridge/gateway/shard_runner.rs
@@ -14,7 +14,7 @@ use super::{ShardClientMessage, ShardId, ShardManagerMessage, ShardRunnerMessage
 #[cfg(feature = "voice")]
 use crate::client::bridge::voice::VoiceGatewayManager;
 use crate::client::dispatch::{dispatch, DispatchEvent};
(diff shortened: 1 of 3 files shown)


kangalio:
FYI this is a bit annoying to review due to many changes done in one commit. I've pulled them apart here: 1. e7f63dbb059cccaba2c693e21b764a999ffbc9c2, 2. 332257072fc9a008dcc8660c32da248d8ae12590

mkrasnitski:
Ah yeah, sorry. The dispatch function itself didn't change much, I just removed the calls to `context()` and changed the `update()` calls.

kangalio (approved):

arqunis (approved):

PR #2203, opened 2022-10-03
Simplify dispatch

kangalio:
This PR is based on #2197 and should be merged after

This PR progressively simplifies src/client/dispatch.rs in individually reviewable commits.

The problem @mkrasnitski mentioned is also solved:

> The whole DispatchEvent::update_cache thing is still really ugly, as its only used in the absence of an non-raw event handler. Ideally I'd like to get rid of it somehow.

Previously, the issue was that, depending on whether an EventHandler was configured, the return value of the CacheUpdate implementations was either used or not used. So there had to be two completely separate branches; one to run CacheUpdate and discarding the results; one to run CacheUpdate and feed the results into EventHandler. This PR solves the issue by runnning CacheUpdate and unconditionally creating an instance of the new FullEvent enum. The FullEvent is then conditionally dispatched onto the EventHandler with a simple `if ... { full_event.dispatch(event_handler).await }`.

A FullEvent enum type also opens the door to more flexibility with events. Instead of being constrained to handle events in EventHandler immediately, you can now pass them around. This is very useful for e.g. poise, which needs to pass events around in order to call its `listener` callback. Currently, poise has a [big macro](https://github.com/kangalioo/poise/blob/a613e745b752cdf829fecee2d1d744489bd43bf8/src/event.rs) to create its own FullEvent enum and funnel EventHandler invocations in and out of it. Moving this macro to serenity itself ensures that it keeps up to date, and that other users than poise can make use of it.

### DIFF
--- a/src/client/event_handler.rs
+++ b/src/client/event_handler.rs
@@ -45,7 +45,7 @@ macro_rules! event_handler {
             /// # use serenity::client::{Context, FullEvent};
             /// # fn _foo(ctx: Context, event: FullEvent) {
             /// if let FullEvent::Message { .. } = &event {
-            ///     assert_eq!(event.snake_case_name(), "cache_ready");
+            ///     assert_eq!(event.snake_case_name(), "message");
             /// }
             /// # }
             /// ```
--- a/src/client/event_handler.rs
+++ b/src/client/event_handler.rs
@@ -10,9 +10,23 @@ use crate::model::application::interaction::Interaction;
 use crate::model::guild::automod::{ActionExecution, Rule};
 use crate::model::prelude::*;
 
-/// The core trait for handling events by serenity.
-#[async_trait]
-pub trait EventHandler: Send + Sync {
+macro_rules! event_handler {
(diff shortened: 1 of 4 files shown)


src/client/event_handler.rs
```diff
@@ -10,38 +10,91 @@ use crate::model::application::interaction::Interaction;
...
+        )* }
+
+        impl FullEvent {
+            /// Returns the name of this event as a snake case string
+            ///
+            /// ```rust,no_run
+            /// # use serenity::client::{Context, FullEvent};
+            /// # fn _foo(ctx: Context, event: FullEvent) {
+            /// if let FullEvent::Message { .. } = &event {
+            ///     assert_eq!(event.snake_case_name(), "cache_ready");
```
arqunis: Shouldn't this be `message` if the event is `FullEvent::Message`?

```suggestion
            ///     assert_eq!(event.snake_case_name(), "message");
```

arqunis (approved):

PR #2205, opened 2022-10-03
Box big model type fields

kangalio:
Before:
image

After:
Screenshot_20221003_183650

Guild unfortunately cannot reasonably be shrunken by just boxing its fields because it just contains a huge amount of small fields (HashMap takes 48 bytes each, Vec/String 24 bytes each).


kangalio:
Out of curiosity I tested this on 32 bit:
Screenshot_20221008_104156

So yeah, people with <2GB RAM should run with 32 bit to cut down RAM usage significantly. (And 32 bit math is faster on modern CPUs too)

arqunis (approved):

PR #2208, opened 2022-10-04
MSRV-related cleanup

mkrasnitski:
Remove any workarounds contingent on the MSRV, did some misc cleanup, plus applied the `if_then_some_else_none` lint to the codebase (without actually enabling it). Some of the resultant changes to using `bool::then` are debatable in terms of readability, so feedback welcome.

### DIFF
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -509,15 +509,15 @@ fn nested_group_command_search<'rec, 'a: 'rec>(
                     .checks
                     .iter()
                     .chain(group.options.checks.iter())
-                    .filter_map(|check| check.display_in_help.then(|| check.name.to_string()))
+                    .filter(|check| check.display_in_help)
+                    .map(|check| check.name.to_string())
                     .collect();
 
                 let sub_command_names: Vec<String> = options
                     .sub_commands
                     .iter()
-                    .filter_map(|cmd| {
-                        cmd.options.help_available.then(|| cmd.options.names[0].to_string())
-                    })
+                    .filter(|cmd| cmd.options.help_available)
+                    .map(|cmd| cmd.options.names[0].to_string())
                     .collect();
 
                 return Ok(CustomisedHelpData::SingleCommand {
--- a/src/framework/standard/help_commands.rs
+++ b/src/framework/standard/help_commands.rs
@@ -509,24 +509,14 @@ fn nested_group_command_search<'rec, 'a: 'rec>(
                     .checks
                     .iter()
                     .chain(group.options.checks.iter())
-                    .filter_map(|check| {
-                        if check.display_in_help {
-                            Some(check.name.to_string())
-                        } else {
-                            None
-                        }
-                    })
+                    .filter_map(|check| check.display_in_help.then(|| check.name.to_string()))
                     .collect();
 
                 let sub_command_names: Vec<String> = options
                     .sub_commands
                     .iter()
                     .filter_map(|cmd| {
-                        if cmd.options.help_available {
-                            Some(cmd.options.names[0].to_string())
-                        } else {
-                            None
-                        }
+                        cmd.options.help_available.then(|| cmd.options.names[0].to_string())
                     })
                     .collect();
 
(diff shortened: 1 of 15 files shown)


src/framework/standard/help_commands.rs
```diff
@@ -509,24 +509,14 @@ fn nested_group_command_search<'rec, 'a: 'rec>(
...
                     .iter()
                     .chain(group.options.checks.iter())
-                    .filter_map(|check| {
-                        if check.display_in_help {
-                            Some(check.name.to_string())
-                        } else {
-                            None
-                        }
-                    })
+                    .filter_map(|check| check.display_in_help.then(|| check.name.to_string()))
```
kangalio: Use .filter() and .map() separately here, in my opinion

src/framework/standard/help_commands.rs
```diff
@@ -509,24 +509,14 @@ fn nested_group_command_search<'rec, 'a: 'rec>(
...
                 let sub_command_names: Vec<String> = options
                     .sub_commands
                     .iter()
                     .filter_map(|cmd| {
-                        if cmd.options.help_available {
-                            Some(cmd.options.names[0].to_string())
-                        } else {
-                            None
-                        }
+                        cmd.options.help_available.then(|| cmd.options.names[0].to_string())
```
kangalio: Ditto

src/model/channel/guild_channel.rs
```diff
@@ -1038,15 +1038,10 @@ impl GuildChannel {
...
-                    if self
-                        .permissions_for_user(cache, e.0)
+                    self.permissions_for_user(cache, e.0)
                         .map(|p| p.contains(Permissions::VIEW_CHANNEL))
                         .unwrap_or(false)
-                    {
-                        Some(e.1.clone())
-                    } else {
-                        None
-                    }
```
kangalio, line 1049: Ditto

src/model/channel/message.rs
```diff
@@ -448,22 +448,15 @@ impl Message {
...
-        // points.
-        let count = content.chars().count();
-
-        if count > constants::MESSAGE_CODE_LIMIT {
-            Some(count - constants::MESSAGE_CODE_LIMIT)
-        } else {
-            None
+        match utils::check_overflow(content.chars().count(), constants::MESSAGE_CODE_LIMIT) {
+            Ok(()) => None,
+            Err(overflow) => Some(overflow),
```
kangalio: Can use .err() here

src/model/guild/member.rs
```diff
@@ -518,8 +518,7 @@ impl Member {
                 .guild(self.guild_id)?
                 .roles
                 .iter()
-                .filter(|(id, _)| self.roles.contains(id))
-                .map(|(_, v)| v.clone())
+                .filter_map(|(id, role)| self.roles.contains(id).then(|| role.clone()))
```
kangalio: I don't think this is more readable. Instead of two operations in sequence, there's now nesting which you have to pull apart mentally

src/model/guild/mod.rs
```diff
@@ -1664,17 +1667,12 @@ impl Guild {
...
-                            if starts_with(prefix, case_sensitive, nick) {
-                                Some((member, nick.clone()))
-                            } else {
-                                None
-                            }
-                        },
+                        Some(nick) => starts_with(nick, prefix, case_sensitive)
+                            .then(|| (member, nick.clone())),
                         None => None,
                     }
```
kangalio, line 1677: If you wanna go shorter and more functional, try `.filter(|nick| starts_with(...))).map(...)`

src/model/guild/mod.rs
```diff
@@ -1727,28 +1725,18 @@ impl Guild {
...
                     match &member.nick {
-                        Some(nick) => {
-                            if contains(substring, case_sensitive, nick) {
-                                Some((member, nick.clone()))
-                            } else {
-                                None
-                            }
-                        },
+                        Some(nick) => contains(nick, substring, case_sensitive)
+                            .then(|| (member, nick.clone())),
```
kangalio, line 1739: Can this be deduplicated with the code above?

src/model/guild/mod.rs
mkrasnitski: This doesn't quite work, because `(member, nick.clone())` would only be returned on `Some(true)`, so you have to map using `bool::then` anyways.

kangalio (approved):

arqunis (approved):

PR #2211, opened 2022-10-05
Simplify serde (de)serialization implementations and remove serde-value dependency

kangalio:


GnomedDev:
Have you seen the [`from` container attribute](https://serde.rs/container-attrs.html#from)? This could probably be used for `CommandDataOption`

kangalio:
Nice to know. In this case `#[serde(try_from = "...")]` would work but I think I'll not use it here. It's just a single line shorter, adds indentation, adds a redundant clone, and `impl TryFrom` is not as nicely symmetric with `fn option_from_raw` as `fn option_to_raw` was.

arqunis (approved):

PR #2212, opened 2022-10-05
Cleanup dependencies

kangalio:
- Removes dependencies: cfg-if, mime, tokio-test
- Makes dependencies optional: fxhash, mime_guess
- Puts each dependency in Cargo.toml on a single line
- Put dependency attributes in consistent order (version, package/path, default-features, features, optional)
- Group dependencies by category

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1,5 +1,7 @@
 #![allow(clippy::missing_errors_doc)]
 
+use std::borrow::Cow;
+use std::convert::TryFrom;
 use std::fmt;
 use std::num::NonZeroU64;
 use std::str::FromStr;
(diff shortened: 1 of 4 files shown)


src/http/client.rs
```diff
@@ -186,11 +184,11 @@ fn parse_token(token: impl AsRef<str>) -> String {
...
-    let header_value = match Cow::from(utf8_percent_encode(reason, NON_ALPHANUMERIC)) {
-        Cow::Borrowed(value) => HeaderValue::from_str(value),
-        Cow::Owned(value) => HeaderValue::try_from(value),
-    }
-    .expect("Invalid header value even after percent encode");
+    // "The X-Audit-Log-Reason header supports 1-512 URL-encoded UTF-8 characters."
+    // https://discord.com/developers/docs/resources/audit-log#audit-log-entry-object
+    let header_value =
+        HeaderValue::from_str(&utf8_percent_encode(reason, NON_ALPHANUMERIC).to_string())
+            .expect("Invalid header value even after percent encode");
```
GnomedDev: Why do this change? This reverts a previous PR to specifically use `from_str` and `try_from` as an efficiency improvement....

kangalio: From a quick glance I got the impression there's no performance difference and both functions converted to &str internally, similar to #2202. And since this is never gonna be a bottleneck anyways I went ahead and simplified the code.

Looking again, they convert to String (basically), not &str. So we could hard-code to use HeaderValue::try_from without losing performance. But I just reverted the change because that's the lowest friction way at this point and the change doesn't really matter

arqunis (approved):

PR #2216, opened 2022-10-13
Fix embeds being cleared

jepcd:
https://discord.com/channels/381880193251409931/381912587505500160/1027735242787794996

### DIFF
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -55,11 +55,13 @@ impl CreateInteractionResponse {
                     .map_err(|overflow| Error::Model(ModelError::MessageTooLong(overflow)))?;
             }
 
-            check_overflow(data.embeds.len(), constants::EMBED_MAX_COUNT)
-                .map_err(|_| Error::Model(ModelError::EmbedAmount))?;
+            if let Some(embeds) = &data.embeds {
+                check_overflow(embeds.len(), constants::EMBED_MAX_COUNT)
+                    .map_err(|_| Error::Model(ModelError::EmbedAmount))?;
 
-            for embed in &data.embeds {
-                embed.check_length()?;
+                for embed in embeds {
+                    embed.check_length()?;
+                }
             }
         }
         Ok(())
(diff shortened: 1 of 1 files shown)


src/builder/create_interaction_response.rs
```diff
@@ -160,13 +163,13 @@ impl CreateInteractionResponseData {
 
     /// Adds an embed to the message.
```
kangalio, line 164: Should probably add a paragraph like "If you're calling this function while editing a message, existing embeds on the message are cleared" to make it clear what happens to existing embeds (phrasing up for debate)

src/builder/create_interaction_response.rs
```diff
@@ -160,13 +163,13 @@ impl CreateInteractionResponseData {
...
     pub fn add_embed(mut self, embed: CreateEmbed) -> Self {
-        self.embeds.push(embed);
+        self.embeds.get_or_insert_with(Vec::new).push(embed);
         self
     }
 
     /// Adds multiple embeds for the message.
```
kangalio, line 172: Here too

src/builder/create_interaction_response.rs
jepcd: But doesn't it not clear the existing embeds?
`.add_embed` and `.add_embeds` just add additional embeds
`.embed` and `.embeds` clear existing embeds, and they already document that

kangalio: Say you posted a message with embed A. Editing without calling add_embed() will keep embed A. Editing with calling add_embed(B) will replace the embed with B. Editing with calling add_embed(B).add_embed(C) will replace the embed with B and C

kangalio: Discord interprets `None` in the embeds field as "no change" and `Some()` as "replace all existing embeds with these". That's why embeds were previously always cleared; there was no Option so Discord always received Some

jepcd: Ah ok I got it, added them

kangalio: Feel free to improve the phrasing. It wasn't clear to you so it won't be clear to many others as well

mkrasnitski: It's it's best to reword the comment into future tense, and remove the use of 2nd-person. E.g, "calling this while editing..."

jepcd: I updated it to 
```rs
/// Calling this while editing a message will overwrite existing embeds.
```

arqunis (approved):

PR #2217, opened 2022-10-14
Implement custom_ids filter for CollectorBuilder

GnomedDev:
Implements a built-in filter to avoid having to use the filter function for the common use of just checking custom IDs.
This also deduplicates calls of macro that generates the setter macros. 

Thanks jepcd#0837 for the idea on the serenity server.

### DIFF
--- a/src/collector/component_interaction_collector.rs
+++ b/src/collector/component_interaction_collector.rs
@@ -42,7 +42,7 @@ impl super::CollectorBuilder<'_, MessageComponentInteraction> {
     impl_channel_id!("Sets the channel on which the interaction must occur. If an interaction is not on a message with this channel ID, it won't be received.");
     impl_guild_id!("Sets the guild in which the interaction must occur. If an interaction is not on a message with this guild ID, it won't be received.");
     impl_message_id!("Sets the message on which the interaction must occur. If an interaction is not on a message with this ID, it won't be received.");
-    impl_custom_ids!("Sets acceptable custom IDs for the interaction. If an interaction does not contain one of the custom IDs, it won't be recieved.");
+    impl_custom_ids!("Sets acceptable custom IDs for the interaction. If an interaction does not contain one of the custom IDs, it won't be received.");
     impl_author_id!("Sets the required author ID of an interaction. If an interaction is not triggered by a user with this ID, it won't be received.");
 }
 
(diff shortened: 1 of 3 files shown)


src/collector/component_interaction_collector.rs
```diff
@@ -30,13 +35,15 @@ pub struct FilterOptions {
...
     message_id: Option<MessageId>,
+    custom_ids: Option<Vec<String>>,
 }
 
 impl super::CollectorBuilder<'_, MessageComponentInteraction> {
     impl_channel_id!("Sets the channel on which the interaction must occur. If an interaction is not on a message with this channel ID, it won't be received.");
     impl_guild_id!("Sets the guild in which the interaction must occur. If an interaction is not on a message with this guild ID, it won't be received.");
     impl_message_id!("Sets the message on which the interaction must occur. If an interaction is not on a message with this ID, it won't be received.");
-    impl_author_id!("Sets the required author ID of an interaction. If an interaction is not triggered by a user with this ID, it won't be received");
+    impl_custom_ids!("Sets acceptable custom IDs for the interaction. If an interaction does not contain one of the custom IDs, it won't be recieved.");
```
arqunis:
```suggestion
    impl_custom_ids!("Sets acceptable custom IDs for the interaction. If an interaction does not contain one of the custom IDs, it won't be received.");
```

src/collector/modal_interaction_collector.rs
```diff
@@ -30,13 +35,15 @@ pub struct FilterOptions {
...
     message_id: Option<MessageId>,
+    custom_ids: Option<Vec<String>>,
 }
 
 impl super::CollectorBuilder<'_, ModalSubmitInteraction> {
     impl_channel_id!("Sets the channel on which the interaction must occur. If an interaction is not on a message with this channel ID, it won't be received.");
     impl_guild_id!("Sets the guild in which the interaction must occur. If an interaction is not on a message with this guild ID, it won't be received.");
     impl_message_id!("Sets the message on which the interaction must occur. If an interaction is not on a message with this ID, it won't be received.");
-    impl_author_id!("Sets the required author ID of an interaction. If an interaction is not triggered by a user with this ID, it won't be received");
+    impl_custom_ids!("Sets acceptable custom IDs for the interaction. If an interaction does not contain one of the custom IDs, it won't be recieved.");
```
arqunis:
```suggestion
    impl_custom_ids!("Sets acceptable custom IDs for the interaction. If an interaction does not contain one of the custom IDs, it won't be received.");
```

arqunis (changes requested):

arqunis (approved):

PR #2219, opened 2022-10-17
Reduce extra clones in the http module

mkrasnitski:
This is a revamp of #2071 based on top of the new changes to next, as the old PR had gotten stale. We pass around owned data to reduce clones. Also, removing any last uses of `String` in `RouteInfo` lets us mark it `Copy`, improving ergonomics.


evieluvsrainbows (approved):
Seems fine to me 👍🏼

arqunis (approved):

PR #2230, opened 2022-10-21
Move `#[must_use]` to struct definition for CreateGuildWelcomeChannel

mkrasnitski:
This was a nit I had with #2225 that I didn't get the chance to voice before it was merged.


kangalio:
Oops another case of that accident where I blindly followed clippy and didn't realize you could much easier put the attribute on the entire struct

kangalio (approved):

arqunis (approved):

PR #2238, opened 2022-10-23
Add create_forum_thread() in a channel [WIP]

### LINKED ISSUE #2142: Add create_forum_thread() in a channel
Creating a thread in the new Forum channel types works differently, and the existing methods like create_public_thread() can't be used.
Discord API docs: https://discord.com/developers/docs/resources/channel#start-thread-in-forum-channel

trkohler:
resolve https://github.com/serenity-rs/serenity/issues/2142#issuecomment-1276635360

### DIFF
--- a/src/http/routing.rs
+++ b/src/http/routing.rs
@@ -1762,6 +1762,9 @@ pub enum RouteInfo<'a> {
         channel_id: u64,
         message_id: u64,
     },
+    CreateForumThread {
+        channel_id: u64,
+    }
 }
 
 impl<'a> RouteInfo<'a> {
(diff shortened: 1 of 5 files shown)


kangalio:
Is this PR ready to be reviewed?

src/http/routing.rs
```diff
@@ -2979,6 +2982,13 @@ impl<'a> RouteInfo<'a> {
                 Route::ChannelsIdPinsMessageId(channel_id),
                 Cow::from(Route::channel_pin(channel_id, message_id)),
             ),
+            RouteInfo::CreateForumThread {
+                channel_id
+            } => (
+                LightMethod::Post,
+                Route::ChannelsIdThreads(channel_id),
+                Cow::from(Route::channel_private_threads(channel_id)),
```
oSumAtrIX, line 2990: While this works, the [API specs](https://discord.com/developers/docs/resources/channel#start-thread-in-forum-channel) have a separate definition for creating threads in forums. The type of the created thread is PUBLIC_THREAD. Both routes are identical, which is why the current implementation works, but it is better to stick to the specification and create a separate route for creating threads in forums as they have distinct differences to creating a regular, private thread, which the current implementation does not imply. This route also has different JSON parameters, though this is also not considered in the route of creating a private thread.

oSumAtrIX (changes requested):

trkohler:
I understand that I can block work with my stale PR, so I would close it since it seems like I am not being good at my time management 😔 thanks for the team who took time to help me onboard.

PR #2239, opened 2022-10-23
Make fields mandatory on CreateModal

jepcd:
These fields should be mandatory. Discord throws an error if not present.


kangalio:
Already done in commit 9657b288d295646c4ef0ab044eeacbdc2a9b0a1b of #2236

jepcd:
ah my bad, didn't see that

PR #2240, opened 2022-10-24
Rewrite collectors

kangalio:
They were overengineered and a bit complicated to use. Now, each collector has just one corresponding type, with methods collect_stream() or collect_single(), depending on what you need.

The implementation is much simplified, too.

Previously, each collector type had a method on Context, a method on ShardMessenger, a variant on ShardRunnerMessage, a branch in ShardRunner::handle_filters and a branch in ShardRunner::handle_rx_value. But they all fundamentally do the same thing; so I replaced all of that with a single CollectorCallback system that all collectors use. This reduces the collector-related code in the gateway to like a couple lines of code.

collector.rs is now a single file, with a collector primitive function called `collect()` which is enough to implement any possible collector. The existing predefined collectors for messages, component interactions, modal interactions, reactions, and generic events still exist though. Those collector structs are generated using a macro, because their structure is very consistent (due to them being mostly dumb wrapper around ´collect()`)

I removed the `removed` filter on ReactionCollector because listening for removed reactions seems very niche, and (like anything else), can be achieved with the `collect()` primitive as well. So ReactionCollector now always collects added reactions. The `collect()` docs show an example of how to collect removed reactions.

I also removed the filters on EventCollector. Filtering for specific events is better done using `collect()` with a match statement over Event, like is shown in `collect()`'s example


kangalio:
New API:
image

Compare against [`current`](https://docs.rs/serenity/0.11.5/serenity/collector/index.html), [`next`](https://serenity-rs.github.io/serenity/next/serenity/collector/index.html)

kangalio:
I deprecated EventCollector because its approach of "first match event type, then later fallibly unwrap" is suboptimal. Same reason `if option.is_some() { option.unwrap() }` is discouraged. You can see this nicely in the e10_collector diff:
Screenshot_20221025_224927

kangalio:
I added some comments in collector.rs to explain how make_specific_collector works:

```rust
make_specific_collector!(
    // First line has name of the collector type, and the type of the collected items.
    ComponentInteractionCollector, ComponentInteraction,
    // This defines the extractor pattern, which extracts the data we want to collect from an Event.
    Event::InteractionCreate(InteractionCreateEvent {
        interaction: Interaction::Component(interaction),
    }) => interaction,
    // All following lines define built-in filters of the collector.
    // Each line consists of:
    // - the filter name (the name of the generated builder-like method on the collector type)
    // - filter argument type (used as argument of the builder-like method on the collector type)
    // - filter expression (this expressoin must return true to let the event through)
    author_id: UserId => interaction.user.id == *author_id,
    channel_id: ChannelId => interaction.channel_id == *channel_id,
    guild_id: GuildId => interaction.guild_id.map_or(true, |x| x == *guild_id),
    message_id: MessageId => interaction.message.id == *message_id,
    custom_ids: Vec<String> => custom_ids.contains(&interaction.data.custom_id),
);

kangalio:
What's blocking this PR?

arqunis (approved):

PR #2241, opened 2022-10-25
Refactor http module

kangalio:
This PR contains 5 small refactor commits, but the bulk of this PR is the final commit (7628216860688c8829d3d516a85e60b5ab2a277d), which changes the required code for each Http method from

```rust
impl Http {
    pub async fn add_member_role(
        &self,
        guild_id: GuildId,
        user_id: UserId,
        role_id: RoleId,
        audit_log_reason: Option<&str>,
    ) -> Result<()> {
        self.wind(204, Request {
            body: None,
            multipart: None,
            headers: audit_log_reason.map(reason_into_header),
            route: RouteInfo::AddMemberRole {
                guild_id,
                role_id,
                user_id,
            },
        })
        .await
    }

    // ...
}

impl Route {
    pub fn guild_member_role(guild_id: GuildId, user_id: UserId, role_id: RoleId) -> String {
        api!("/guilds/{}/members/{}/roles/{}", guild_id, user_id, role_id)
    }

    // ...
}

pub enum RouteInfo {
    AddMemberRole {
        guild_id: GuildId,
        role_id: RoleId,
        user_id: UserId,
    },
    // ...
}

impl RouteInfo {
    pub fn deconstruct(self) -> (LightMethod, Route, Cow<'static, str>) {
        match self {
            RouteInfo::AddMemberRole {
                guild_id,
                role_id,
                user_id,
            } => (
                LightMethod::Put,
                Route::GuildsIdMembersIdRolesId(guild_id),
                Cow::from(Route::guild_member_role(guild_id, user_id, role_id)),
            ),
            // ...
        }
    }
}
```
to
```rust
impl Http {
    pub async fn add_member_role(
        &self,
        guild_id: GuildId,
        user_id: UserId,
        role_id: RoleId,
        audit_log_reason: Option<&str>,
    ) -> Result<()> {
        self.wind(204, Request {
            body: None,
            multipart: None,
            headers: audit_log_reason.map(reason_into_header),
            method: LightMethod::Put,
            bucket: RatelimitBucket::GuildsIdMembersIdRolesId(guild_id),
            path: api!(self, "/guilds/{}/members/{}/roles/{}", guild_id, user_id, role_id),
        })
        .await
    }
    // ...
}
```


src/http/client.rs
```diff
@@ -200,53 +188,20 @@ fn reason_into_header(reason: &str) -> Headers {
...
-    pub ratelimiter_disabled: bool,
-    pub proxy: Option<Url>,
+    pub ratelimiter: Option<Ratelimiter>,
+    pub api: String,
+    pub status_api: String,
     pub token: String,
     application_id: AtomicU64,
 }
 
-impl fmt::Debug for Http {
```
arqunis, line 208: This manual `Debug` implementation exists to hide the `token` in the debug representation to prevent someone accidentally leaking it. This shouldn't be removed.

kangalio: Ohh I see. Thank you, I'll add it back with a comment so future readers don't fall into the same trap as me

mkrasnitski:
Do you think it's worth having an `api` method on `Http` to automatically prepend `self.api` when forming routes? that way it's easier to not mess up when adding new routes.

mkrasnitski:
I don't super like the separation of `RatelimitBucket` from the actual full route, but I guess I suppose it's necessary since the ratelimiting only happens one level deep (meaning, e.g. the ratelimit bucket only cares about the guild id, whereas the actual request has extra parameters like a channel id and a message id). I guess it's the same reason that `Route` and `RouteInfo` needed to be separate? Maybe a custom `Eq` (or `Hash`) implementation could solve this?

kangalio:
> I don't super like the separation of `RatelimitBucket` from the actual full route, but I guess I suppose it's necessary since the ratelimiting only happens one level deep (meaning, e.g. the ratelimit bucket only cares about the guild id, whereas the actual request has extra parameters like a channel id and a message id). I guess it's the same reason that `Route` and `RouteInfo` needed to be separate?

Yep, indeed.

> Maybe a custom Eq (or Hash) implementation could solve this?

I thought about this too. If we still had RouteInfo, then we could implement rate limit buckets using a Eq/Hash impl on RouteInfo instead of the auxiliary Route enum, yes. But then we're back to Http methods implementations being scattered in 4 places (Http method, RouteInfo enum variant, RouteInfo match statement for URL, RouteInfo Eq+Hash implementation).

kangalio:
> Do you think it's worth having an api method on Http to automatically prepend self.api when forming routes? that way it's easier to not mess up when adding new routes.

I doubt this would happen; you'd copy-paste an existing route and adjust it anyways, plus you'd test the route afterwards. But I can do it

mkrasnitski:
> I thought about this too. If we still had RouteInfo, then we could implement rate limit buckets using a Eq/Hash impl on RouteInfo instead of the auxiliary Route enum, yes. But then we're back to Http methods implementations being scattered in 4 places (Http method, RouteInfo enum variant, RouteInfo match statement for URL, RouteInfo Eq+Hash implementation).

I think it would actually be better. We'd avoid duplicating route strings for Create/Edit/Delete endpoints (where the `LightMethod` is the only difference). The `Route` enum would store _all_ of the relevant ids (instead of just one) and the url could fully be derived from it using a simple match on `self` (instead of dedicated methods). Another method could match on `self` to return the one field used for ratelimiting, and the custom Eq/Hash impls could make use of that.

kangalio:
> I think it would actually be better. We'd avoid duplicating route strings for Create/Edit/Delete endpoints (where the LightMethod is the only difference)

In principle I'm all for reducing duplication. However, in this case, Discord itself writes the URLs separately for each endpoint. [For example](https://discord.com/developers/docs/resources/emoji#create-guild-emoji):
Screenshot_20221025_224000

So, by writing the URLs separately ourselves, we don't just simplify the implementation a lot, we also match Discord's semantics.

mkrasnitski:
I mean, how else would they be expected to document it? Since the Http methods are sorted alphabetically, the Create/Edit/Delete endpoints are quite far apart in the source, which is risky imo.

src/http/client.rs
```diff
@@ -206,6 +206,17 @@ pub struct Http {
...
+    ($self:ident, $string:literal $($args:tt)*) => {
+        format!(concat!("{}", $string), $self.api $($args)*)
+    };
+}
+macro_rules! status {
+    ($self:ident, $string:literal $($args:tt)*) => {
+        format!(concat!("{}", $string), $self.status_api $($args)*)
+    };
+}
+
```
mkrasnitski: I think hardcoding the use of `self` here is better than passing it in every time as the first argument.

kangalio: That is not possible due to macro hygiene rules

kangalio:
> I mean, how else would they be expected to document it? Since the Http methods are sorted alphabetically, the Create/Edit/Delete endpoints are quite far apart in the source, which is risky imo.

Could you list actual specific potential bug sources with the separated URLs approach? That would make the discussion less abstract

mkrasnitski:
If a new endpoint is added, or an endpoint url is changed (this is less likely), then there's three places to add/change the url rather than just one. Typos in strings won't get caught, whereas a typo in an enum variant will.

kangalio:
> If a new endpoint is added, or an endpoint url is changed (this is less likely), then there's three places to add/change the url rather than just one.

I see the concern. In the end, it's a trade-off. The Route enum approach has the advantage you mentioned, but the disadvantage that the code becomes scattered and indirect again:

<details>

```rust
enum Route {
    RouteA { guild: GuildId, user: UserId },
    RouteB { channel: ChannelId, message: MsesageId },
    // ...
}

impl Route {
    fn url(&self) {
        match self {
            RouteA { ... } => "...",
            RouteB { ... } => "...",
            // ...
        }
    }

    fn bucket(&self) -> impl Hash + Eq {
        let id = match self {
            RouteA { guild, .. } => guild.0,
            RouteB { channel, .. } => channel.0,
            // ...
        };
        (std::mem::discriminant(self), id)
    }
}

impl Http {
    fn endpoint_a(&self) {
        Route::RouteA { ... }
    }

    fn endpoint_b(&self) {
        Route::RouteB { ... }
    }

    // ...
}
```
</details>

Compare against:

<details>

```rust
enum RatelimitBucket {
    RouteA { guild: GuildId },
    RouteB { channel: ChannelId },
    // ...
}

impl Http {
    fn endpoint_a(&self) {
        let url = "...";
        let bucket = RatelimitBucket::RouteA { guild };
    }

    fn endpoint_b(&self) {
        let url = "...";
        let bucket = RatelimitBucket::RouteB { channel };
    }

    // ...
}
```

</details>

In practice, I don't think we'll run into issues with having multiple instances of the same URL. When adding new endpoints, you'll copy-paste the URL or the entire method body anyways, so the URLs will be in sync. And when editing routes (which has [happened before](https://discord.com/developers/docs/change-log#api-and-gateway-v8), apparently), you'd Ctrl+F for the route anyways, which shows you all the places you have to modify.

Yes, it's not optimal, but I believe it's still a better trade-off than having to add code in 4 different places to add a new route.

mkrasnitski:
I actually prefer the first example. It's a bit hard to verbalize why, but I'll try and write down some reasons. 

1. I dislike in the second example that the url and bucket are passed along together yet separately, even though they're derived from the same source, so to speak. 
2. I also don't like the RatelimitBucket struct itself:
    - It feels too separate from the rest of the routing code, enough so that I think it belongs in `ratelimiting.rs` rather than `routing.rs` (which would then become empty as a result).
    - The variant names are kind of bad. They were inherited from the old `Route` enum, but the point still stands that they're long and a bit noisy (only so that they're distinguishable from each other).
3. If only the bucket struct and strings are used for routing, then serenity loses the comprehensive list of endpoints and their parameters that was once stored in `RouteInfo`.

On the other hand, the enum in the first example handles both the url and bucket, has variant names that match the actual endpoints, has a very simple solution for ratelimiting (I didn't know about `std::mem::discriminant` before!), and has a complete parameter list for each endpoint, which addresses all of the above concerns.

kangalio:
> I actually prefer the first example. It's a bit hard to verbalize why, but I'll try and write down some reasons.
> 
> 1. I dislike in the second example that the url and bucket are passed along together yet separately, even though they're derived from the same source, so to speak.
> 
> 2. I also don't like the RatelimitBucket struct itself:
>        
>    * It feels too separate from the rest of the routing code, enough so that I think it belongs in `ratelimiting.rs` rather than `routing.rs` (which would then become empty as a result).
>    * The variant names are kind of bad. They were inherited from the old `Route` enum, but the point still stands that they're long and a bit noisy (only so that they're distinguishable from each other).
> 
> 3. If only the bucket struct and strings are used for routing, then serenity loses the comprehensive list of endpoints and their parameters that was once stored in `RouteInfo`.
> 
> 
> On the other hand, the enum in the first example handles both the url and bucket, has variant names that match the actual endpoints, has a very simple solution for ratelimiting (I didn't know about `std::mem::discriminant` before!), and has a complete parameter list for each endpoint, which addresses all of the above concerns.

Thank you for articulating your worries so clearly.

I thought of a new approach which you can see in the WIP commit 3e67ac6f7233f31a4cd61469f0f3a1d17b5fb1ec. (it's 3:30 AM and I'll be gone tomorrow so I'm pushing out a WIP instead of finishing)

mkrasnitski:
I understand it's WIP, but I think doing this with a macro makes it way too over-engineered. I think a good first step is to implement it **without macros**, then in a later PR trim it down if it's *really* needed. As it stands the macro is completely unmaintainable, especially for adding new endpoints. I know I've mentioned this before, but having declarative macros take tokens that aren't valid rust syntax feels very wrong to me. It just feels way too hacky for my taste, and it's pretty inaccessible in terms of readability.

kangalio:
I made the macro much lighter, removed the ratelimit policy macro trickery, added comments.

A nice thing about this macro representation is, it can be converted into any other representation by just tweaking the macro definition. So if we ever want to change how route paths are generated, or ratelimit buckets are handled, or how the Route enum variants look, or how proxy URLs are passed, etc. it's a trivial change.

>  I think a good first step is to implement it without macros, then in a later PR trim it down if it's really needed

We can trivially go from the macro way to the old way via `cargo expand`, but the other way around is not.

> As it stands the macro is completely unmaintainable, especially for adding new endpoints

I'd disagree. How is 
```rust
routes!('a
    Channel { channel_id: ChannelId }
    => api!("/channels/{}", channel_id),
    Ratelimit::ByRouteAndId(channel_id.0);

    ChannelInvites { channel_id: ChannelId }
    => api!("/channels/{}/invites", channel_id),
    Ratelimit::ByRouteAndId(channel_id.0);

    ChannelMessage { channel_id: ChannelId, message_id: MessageId }
    => api!("/channels/{}/messages/{}", channel_id, message_id),
    Ratelimit::ByRouteAndId(channel_id.0);
);
```
so egregiously bad in comparison to
```rust
pub enum Route<'a> {
    Channel { channel_id: ChannelId },
    ChannelInvites { channel_id: ChannelId },
    ChannelMessage { channel_id: ChannelId, message_id: MessageId },
}
impl Route<'_> {
    pub fn path(&self) -> String {
        match self {
            Self::Channel { channel_id } => api!("/channels/{}", channel_id),
            Self::ChannelInvites { channel_id: ChannelId } => api!("/channels/{}/invites", channel_id),
            Self::ChannelMessage { channel_id: ChannelId, message_id: MessageId } => api!("/channels/{}/messages/{}", channel_id, message_id),
        }
    }
    pub fn bucket(&self) -> RatelimitBucket {
        let ratelimit = match self {
            Self::Channel { channel_id } => Ratelimit::ByRouteAndId(channel_id.0),
            Self::ChannelInvites { channel_id: ChannelId } => Ratelimit::ByRouteAndId(channel_id.0),
            Self::ChannelMessage { channel_id: ChannelId, message_id: MessageId } => Ratelimit::ByRouteAndId(channel_id.0),
        };
        // ...
    }
}
```

mkrasnitski:
Alright, the new macro is a massive improvement. And apologies for the curtness of my previous comment. A few things though:

1. Having that lifetime at the start of the macro invocation feels a bit weird (not sure how to get rid of it tbh).
2. I think `Ratelimit::Dont` should be renamed `Ratelimit::None`
3. The macro should allow unit-struct variants by wrapping the `$arg_name: $arg_type` block in `$(...)?`, so you don't have to construct it like `Route::Gateway {}` (instead just `Route::Gateway`).
4. You left some calls to `api!` with self as the first argument in `client.rs`. Since those take into account proxies, not sure how you'd deal with those. Also, does having proxy support for all routes make sense?
5. Query params are missing but I'm assuming are forthcoming.

kangalio:
1. I don't think it is possible, unfortunately 
2. I thought about it, especially because the old route struct called it None. But I honestly find "Dont" clearer. Not sure why that is, perhaps because None is overloaded from Rust's Option
3. Alright
4. I don't know how to deal with those yet. We could go back to the old way and hardcode discord.com while generating URLs and `.replace("discord.com", ...)` later. This is ugly but easy. Alternatively, we could have api_url and status_url as parameters to path() and make use of them in the format strings...
5. Yes, I'll add a `params` field to Request. I philosophized about it and came to the conclusion that makes much more sense than having it be part of the route

mkrasnitski:
1. Maybe then just put a comma after it, and/or put the full list of endpoints in curly braces.
2. `Route::None` already used to be a thing. I think about it this way, by asking the question "what kind of ratelimiting does this endpoint perform?". The answer is either "There is ratelimiting by id", "There is ratelimiting", or "There is none." Using "Dont" is strictly worse than "DoNot", but then you see that there is a discrepancy between the english grammar of the variants. Here's my crack at naming them. 
```rust
enum Ratelimit {
    RouteWithId(NonZeroU64),
    Route,
    None
}
```

   - If `Ratelimit::Route` were changed instead to just be a word of positive confirmation (like `Some`, but not that word. Something that reads as "yes, do ratelimiting for this endpoint"), then you could rename `RouteWithId` to just `WithId`.

4. By the way, I think `Route::path` should return `Cow<'static, str>`, since using `api!` without format strings produces a `&'static str` rather than a `String`.
5. I agree, hopefully it ends up being simpler that way. A lot of endpoints will just take `None` in the Request, but that's fine.

kangalio:
1. Ok
2. I see the enum and variant name as an English question and answer: `Ratelimit? By route and ID.` `Ratelimit? Don't.`
3. 
4. I suppose. I didn't do it initially because only a couple of endpoints have no arguments anyways, and the cost of allocating a dozen bytes is minuscule.

mkrasnitski:
In my opinion being declarative is better, especially if the value is returned from a function and taken as argument to other functions. Also had a think and not sure if `Ratelimit::Route` is better than `Ratelimit::ByRoute`; it's a toss-up.

kangalio:
I currently don't have the energy to continue on this PR as-is

mkrasnitski:
Sorry to hear that. I hope the bikeshedding wasn't a large factor in that decision. If it's ok, I can pick it up and carry the torch if you're so inclined.

kangalio:
Sure, you can :)

> I hope the bikeshedding wasn't a large factor in that decision

Eh, slightly, but bikeshedding is normal and good. My main problem was that the HTTP code was redesigned multiple times and I fear to have introduced a mistake somewhere along the way. If I were to pick this up again, I'd start anew and document the refactoring process to make sure I don't miss anything.

I'm just slightly sad that the 5 small refactor commits get lost too, now ([diff](https://github.com/kangalioo/serenity/compare/cf38735d1b92363e6b391e21bed6855da06df9bf...63c38642b2aacc60df49d3293af7b47af9670460)). I tried to extract them into a separate PR three times but it was hell and it's easier to re-implement them

PR #2246, opened 2022-10-25
Make absolute ratelimits a runtime config instead of feature

kangalio:
Screenshot_20221026_011244

One feature down

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,11 +76,11 @@ jobs:
 
       - name: Build all features
         if: matrix.features == ''
-        run: cargo build --all-features
+        run: cargo build --features full
 
       - name: Test all features
         if: matrix.features == ''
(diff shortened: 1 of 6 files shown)


arqunis (approved):

arqunis (changes requested):
That was a premature approval (not that the changes are bad). Could you also remove the `absolute_ratelimits` feature, or add a warning to [build.rs](https://github.com/serenity-rs/serenity/blob/next/build.rs) that it's deprecated and you must now configure `Http` at runtime?

kangalio:
Done
Screenshot_20221027_184355

kangalio:
Good catch, by the way, thanks :)

arqunis (approved):

kangalio:
The CI failure is unrelated to this PR. It happens on upstream `next` too and is due to a bad interaction of #2254 and #2256

PR #2261, opened 2022-10-31
Avoid cloning entire cache fields by adding ReadOnlyMapRef

GnomedDev:
Reimplements MaybeMap as ReadOnlyMapRef except holding a reference to the inner DashMap and not providing methods for mutation, allowing cache fields to be directly referenced (without risk of mutation) instead of entirely cloned.

Currently it is not easily possible to implement ToOwned without TAIT or other changes, so that can be implemented in a followup PR.

### DIFF
--- a/src/cache/wrappers.rs
+++ b/src/cache/wrappers.rs
@@ -40,7 +40,7 @@ impl<K: Eq + Hash, V> MaybeMap<K, V> {
     }
 }
 
-#[derive(Debug)]
+#[derive(Clone, Copy, Debug)]
 /// A wrapper around a reference to a MaybeMap, allowing for public inspection of the underlying map
 /// without allowing mutation of internal cache fields, which could cause issues.
 pub struct ReadOnlyMapRef<'a, K: Eq + Hash, V>(Option<&'a DashMap<K, V, FxBuildHasher>>);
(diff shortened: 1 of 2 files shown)


src/cache/wrappers.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub fn len(&self) -> usize {
+        self.0.as_ref().map_or(0, |map| map.len())
+    }
+
+    pub(crate) fn as_read_only(&self) -> ReadOnlyMapRef<'_, K, V> {
+        ReadOnlyMapRef(self.0.as_ref())
+    }
+}
+
+#[derive(Debug)]
```
kangalio: Should implement Copy (and Clone) because it's conceptually a reference

kangalio:
guild_channels still returns an owned DashMap by cloning the stored map

GnomedDev:
guild_channels cannot be made to return a reference as the channels are stored in the guild

kangalio:
I see. Though we can return `impl Deref<Target = [GuildChannel]` by using https://docs.rs/dashmap/latest/dashmap/mapref/one/struct.Ref.html#method.map on the Guild Ref

kangalio:
> I see. Though we can return `impl Deref<Target = [GuildChannel]` by using https://docs.rs/dashmap/latest/dashmap/mapref/one/struct.Ref.html#method.map on the Guild Ref

I guess this can also be a follow-up PR. Then this PR could be merged already to finally unblock #2263 and #2264

arqunis (approved):

PR #2262, opened 2022-10-31
Fix checks

GnomedDev:
Fixed the checks by running `cargo clippy --features collector,temp_cache --fix --allow-dirty` after fixing `examples/testing`


kangalio:
Outdated by now

PR #2266, opened 2022-11-03
fix NotAuthor error when deleting messages

Chronophylos:
Sometimes Discord does not include `guild_id` in message objects. Because `Message::is_private` checks if `guild_id` is set `Message::delete` returns `Error::Model(ModelError::NotAuthor)` if `guild_id` is not set.
This means the function can fail when it shouldn't.

This PR removes the faulty check.


mkrasnitski:
FYI, we use nightly `rustfmt` options that aren't stable yet. You should rerun `cargo +nightly fmt` to fix those differences.

arqunis (approved):

PR #2270, opened 2022-11-05
Rework Framework trait 

kangalio:
So it's useful for more than just the built-in text command framework. New Framework trait API:

```rust
pub trait Framework: Send + Sync {
    async fn init(&mut self, client: &Client);
    async fn dispatch(&self, event: FullEvent);
}
```

Previous:

```rust
pub trait Framework: Send + Sync {
    async fn dispatch(&self, _: Context, _: Message);
}
```

This PR was developed in tandem with porting poise to it. Here's the corresponding poise code: https://github.com/kangalioo/poise/tree/better-framework


arqunis (approved):

kangalio:
Light speed, thanks

PR #2275, opened 2022-11-08
Make messages() type-safe

kangalio:
Retrieving messages from a channel can happen on a most-recent basis, or before a specific message ID, or after a specific message ID, or around a specific message ID. You can also provide the number of messages to retrieve, which falls back to 50.

This feature was modeled in serenity using a builder with .after(), .before(), .around() functions which were mutually exclusive and fell back to most-recent search if none was selected. The builder also took an optional message count.

It's simpler to model the search filter using a four-variant `MessageFilter` enum. And the message count should be a required parameter, since it's clearer to explicitly write `50` instead of quietly relying on some arbitrary default value. Therefore, the API of the messages() function on channel-like types is now:

```rust
pub async fn messages(
    self,
    http: impl AsRef<Http>,
    filter: MessageFilter,
    limit: u32,
) -> Result<Vec<Message>> {}

pub enum MessageFilter {
    After(MessageId),
    Around(MessageId),
    Before(MessageId),
    MostRecent,
}
```


kangalio:
Closing because nobody said if this is desirable and by now the PR is outdated anyways

PR #2280, opened 2022-11-10
Properly escape all reactions

mkrasnitski:
Previously, `DeleteReactionMessageEmoji` and `GetReactionUsers` endpoints would fail with the #️⃣ emoji because they forgot to escape it. This unifies the escaping to always happen.

### DIFF
--- a/src/model/channel/reaction.rs
+++ b/src/model/channel/reaction.rs
@@ -439,7 +439,7 @@ impl ReactionType {
             } => {
                 format!("{}:{}", name.as_ref().map_or("", String::as_str), id)
             },
-            ReactionType::Unicode(ref unicode) => unicode.clone(),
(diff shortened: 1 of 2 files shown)


src/model/channel/reaction.rs
```diff
@@ -439,7 +439,7 @@ impl ReactionType {
             } => {
                 format!("{}:{}", name.as_ref().map_or("", String::as_str), id)
             },
-            ReactionType::Unicode(ref unicode) => unicode.clone(),
+            ReactionType::Unicode(ref unicode) => unicode.replace('#', "%23"),
```
kangalio: Cleaner fix would be to use the percent-encoding which we already depend on anyways

mkrasnitski: Done.

arqunis (approved):

PR #2285, opened 2022-11-15
Miscellaneous http refactoring

mkrasnitski:
Some small http refactors, mainly getting rid of cruft and moving some things around:

 1. Instead of having a `ratelimiter_disabled` flag, simply wrap `Http::ratelimiter` in `Option`.
 2. Use `HttpBuilder` in `Http::new` instead of duplicating construction logic
 3. Unify `ban` and `ban_with_reason`, and likewise with the kick methods. The model methods API remains unchanged.
 4. Introduce a `SecretString` type to prevent leaking the token, e.g. via `Debug`.
 5. Change `Http::proxy` to `Option<String>`, because url parsing is already done before a request is sent out. This changes the error to happen at request execution-time rather than when the `Http` is built, which may or may not be desirable. Thoughts?

This contains work adapted from #2241.

### DIFF
--- a/examples/testing/src/main.rs
+++ b/examples/testing/src/main.rs
@@ -11,6 +11,7 @@ const IMAGE_URL: &str = "https://raw.githubusercontent.com/serenity-rs/serenity/
 const IMAGE_URL_2: &str = "https://rustacean.net/assets/rustlogo.png";
 
 async fn message(ctx: &Context, msg: Message) -> Result<(), serenity::Error> {
+    let guild_id = msg.guild_id.unwrap();
     let channel_id = msg.channel_id;
     let guild_id = msg.guild_id.unwrap();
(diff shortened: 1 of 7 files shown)


mkrasnitski:
Just realized there's a typo in commit message number 5. `Request::proxy` should be `Http::proxy`.

kangalio (approved):

kangalio:
Nice changes ;)

arqunis (approved):

cheesycod:
proxy is kinda broken and now needs a trailing slash @mkrasnitski @kangalioo

cheesycod:
otherwise it panics with invalid port

mkrasnitski:
@cheesycod A hotfix for this has been pushed in #2323. Can you confirm that this has been fixed for you?

PR #2293, opened 2022-11-19
Add back lost special deserialization behavior

kangalio:
https://discord.com/channels/381880193251409931/381912587505500160/1043283742254448820

> i forgot to port some of the derive-deviating behavior to the new Deserialize impls in https://github.com/serenity-rs/serenity/pull/2287
> namely, the fact that `user` is actually not a required field and is provided by serenity in guilds by cloning `member.user`

### DIFF
--- a/src/model/application/interaction/application_command.rs
+++ b/src/model/application/interaction/application_command.rs
@@ -61,6 +61,7 @@ pub struct CommandInteraction {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub member: Option<Box<Member>>,
     /// The `user` object for the invoking user.
+    #[serde(default)]
     pub user: User,
     /// A continuation token for responding to the interaction.
     pub token: String,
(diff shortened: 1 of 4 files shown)


src/model/application/interaction/application_command.rs
```diff
@@ -242,6 +242,11 @@ impl<'de> Deserialize<'de> for CommandInteraction {
     fn deserialize<D: Deserializer<'de>>(deserializer: D) -> StdResult<Self, D::Error> {
         let mut interaction = Self::deserialize(deserializer)?; // calls #[serde(remote)]-generated inherent method
         if let Some(guild_id) = interaction.guild_id {
+            if let Some(member) = &mut interaction.member {
+                member.guild_id = guild_id;
+                // If `member` is present, `user` wasn't sent and is still filled with default data
+                interaction.user = member.user.clone();
+            }
             interaction.data.resolved.roles.values_mut().for_each(|r| r.guild_id = guild_id);
```
mkrasnitski, line 251: After taking a closer look at #2287, I noticed this line was missing from the implementations for `ComponentInteraction` and `ModelInteraction`, even though they both previously called `add_guild_id_to_resolved`. Was this intentional on your part?

kangalio: Yes, that was actually intentional. add_guild_id_to_resolved did two things:
- insert guild_id into Member
- insert guild_id into CommandData::resolved::roles

The latter only applies to CommandInteraction's so I just inlined the former into ComponentInteraction and ModalInteraction.

kangalio:
Can this PR be merged? I've now run into the bug myself while testing the `/modal` command of my test bot :sweat_smile: 
Screenshot_20221120_171421

arqunis (approved):

PR #2295, opened 2022-11-20
Actually call Framework::init() and add ShardManager::intents()

kangalio:
Follow-up PR to #2270. This should be all functionality for poise to fully use the Framework trait.

## Implementation details

`Framework::init()` takes `&Client` which contains ShardManager which contains `Option<Arc<dyn Framework>>`, so we had a circular dependency. To break this up, I changed the `Option<Arc<dyn Framework>>` in ShardManager to `Arc<OnceCell<Arc<dyn Framework>>>`.

The OnceCell type is from the once_cell crate which has been added to Cargo.toml. once_cell already was a transitive dependency anyways, via dashmap, so no dependencies are added in total.

## New poise usage

Poise port to this PR here: https://github.com/serenity-rs/poise/commit/b7bf25c4c52e11b50d32b838de5884cd7189fef4

### Comparison of poise framework setup

Old:

```rust
poise::Framework::builder()
    .token(token)
    .user_data_setup(user_data_setup)
    .options(options)
    .intents(intents)
    .run()
    .await?;
```

New:

```rust
serenity::Client::builder()
    .framework(poise::Framework::new(options, user_data_setup))
    .token(token)
    .intents(intents)
    .await?
    .start()
    .await?;
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ mime_guess = { version = "2.0", optional = true }
 dashmap = { version = "5.1.0", features = ["serde"], optional = true }
 parking_lot = { version = "0.12", optional = true }
 ed25519-dalek = { version = "1.0.1", optional = true }
+once_cell = { version = "1.16.0", optional = true }
 # Serenity workspace crates
 command_attr = { version = "0.4.1", path = "./command_attr", optional = true }
 serenity-voice-model = { version = "0.1.1", path = "./voice-model", optional = true }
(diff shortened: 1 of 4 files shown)


arqunis:
Could you remove instances of `framework_arc` from the documentation, since it has been deleted?

kangalio:
Done

arqunis (approved):

kangalio:
Thanks! Pushed the commit that makes use of this in poise to `serenity-next` branch

PR #2297, opened 2022-11-21
Add serenity::all module

kangalio:
Quoting the module docs:

> Special module that re-exports most public items from this crate.
> 
> Useful, because you don't have to remember the full paths of serenity items.
> 
> Not exported:
> - `crate::json`: it's a general-purpose JSON wrapper, not intrinsic to serenity
> - `crate::framework::standard`: has many standard_framework-specific items that may collide with items from the rest of serenity

Poise has had this module [since the beginning](https://github.com/serenity-rs/poise/blob/97bb8aceb9b717336c43fffe2b04ffabfa08be3f/src/lib.rs#L19-L27) for convenience. I think there's no reason it can't be upstreamed.

The re-exports within the module have been imported from poise, but broken apart in order to add the feature gates.

Poise branch that switches from its own serenity_prelude to this PR's serenity::all: https://github.com/serenity-rs/poise/commit/46baede3d4e5b38ed6556221a796922ebbcb5696

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -179,7 +179,7 @@ pub mod all {
     pub use crate::cache::*;
     #[cfg(feature = "voice")]
     #[doc(no_inline)]
-    pub use crate::client::bridge::gateway::voice::*;
+    pub use crate::client::bridge::voice::*;
     #[cfg(feature = "client")]
     #[doc(no_inline)]
     pub use crate::client::{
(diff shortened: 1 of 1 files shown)


arqunis (approved):

kangalio:
Thanks! Pushed the commit that makes use of this in poise to `serenity-next` branch

PR #2305, opened 2022-12-09
Add files in the edit original interaction response

dpytaylo:
I have implemented methods for working with files in the edit original interaction response.
Example:
```rust
interaction.edit_original_interaction_response(http, |response|
    response
        .add_file("image.png")
).await.unwrap();
```


mkrasnitski:
Apologies for not responding sooner. This is already possible if you use the  `next` branch. The corresponding functions were added in #2163, but what you want to use is `EditInteractionResponse::new_attachment`. Note that on `next`, the builder API no longer uses closures, and the attachments API is also changed, so you'll have to port your old code over.

dpytaylo:
Thanks you for your replay! So, I will close this pull request in favor of #2163 in this case.

Issue #2306, opened 2022-12-10
Scheduled event listening no longer working on next


marcantoinem:
Scheduled event listening no longer work on the branch `next`. Other events still work, but not `guild_scheduled_event_create`, `guild_scheduled_event_delete`, `guild_scheduled_event_update`, `guild_scheduled_event_user_add` and `guild_scheduled_event_user_remove`.

Serenity version: branch: `next`, more specifically since PR #2232 (ea8651aa569c1365afa2bdc939e1a8c086b1fd7e)

Rust version : 1.65.0

Minimal test case:
```rust
use serenity::{async_trait, model::prelude::*, prelude::*};
use std::env;

struct Handler;

#[async_trait]
impl EventHandler for Handler {
    async fn guild_scheduled_event_create(&self, _: Context, _: ScheduledEvent) {
        println!("This should output on event create!");
    }
    async fn guild_scheduled_event_delete(&self, _: Context, _: ScheduledEvent) {
        println!("This should output on event delete!");
    }
    async fn guild_scheduled_event_update(&self, _: Context, _: ScheduledEvent) {
        println!("This should output on event update!");
    }
    async fn message(&self, _ctx: Context, _: Message) {
        println!("This output on new message.");
    }
    async fn ready(&self, _: Context, _: Ready) {
        println!("This output when ready.");
    }
}

#[tokio::main]
async fn main() {
    let token = env::var("DISCORD_TOKEN").expect("Expected a token in the environment");
    let intents = GatewayIntents::non_privileged() | GatewayIntents::GUILD_SCHEDULED_EVENTS;
    let mut client = Client::builder(token, intents)
        .event_handler(Handler)
        .await
        .expect("Error creating client");

    if let Err(why) = client.start().await {
        println!("Client error: {:?}", why);
    }
}
```

marcantoinem:
Fixed with #2307

PR #2232, opened 2022-10-22
Shorten type names

kangalio:
Global find-and-replace:
- ApplicationCommand -> Command
- AutoModeration -> AutoMod
- GuildScheduledEvent -> ScheduledEvent
- MessageComponent -> Component

PR #2307, opened 2022-12-10
Fix scheduled event

marcantoinem:
This PR fix a regression on scheduled event with #2232 . (See bug #2306) It just replace `ScheduledEvent` -> `GuildScheduledEvent` where it was before #2232 .


mkrasnitski:
Hi, thanks for the PR! Note that we use some nightly-only `rustfmt` options. You should therefore run `cargo +nightly fmt` to fix some of the formatting here.

marcantoinem:
Thanks for the info, I've fixed that.

arqunis (approved):

kangalio:
Why did you rename ScheduledEvent back to GuildScheduledEvent?

marcantoinem:
The name of these structs seem to be relied to the call in the api for the event handler. (https://discord.com/developers/docs/resources/guild-scheduled-event) With the name `ScheduledEvent`, it no longer register any events for `GuildScheduledEvent`. (see the bug #2306) I'm not exactly sure how the name of the struct is directly connect to the call in the api, but reverting back to `GuildScheduledEvent` fixed it.

kangalio:
Ohh I see. Yes the enum variant name is used for JSON serialization which I didn't think about when making #2232. It could be fixed with a `#[serde(rename = "...")]` though not sure if I care enough about the more concise name to make yet another PR

PR #2308, opened 2022-12-11
Remove `CacheAndHttp`

mkrasnitski:
Supersedes #2118. As far as I can tell, any concerns around removing this have been addressed by this point. Therefore, removing this and inlining its fields is a net positive, mainly serving to avoid confusion with the `CacheHttp` trait.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,7 +39,7 @@
 //! serenity = "0.11"
 //! ```
 //!
-//! [`Cache`]: crate::caceh::Cache
+//! [`Cache`]: crate::cache::Cache
 //! [`Context`]: crate::client::Context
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -39,6 +39,7 @@
 //! serenity = "0.11"
 //! ```
 //!
+//! [`Cache`]: crate::caceh::Cache
```
arqunis:
```suggestion
//! [`Cache`]: crate::cache::Cache
```

arqunis (changes requested):

src/lib.rs
mkrasnitski: Lol, that typo has been sitting there for 2 months.

arqunis (approved):

PR #2319, opened 2022-12-22
Remove virtual events

GnomedDev:
This removes the "virtual" events `cache_ready`, `shards_ready`, and `ratelimit` to make sure serenity's concept of an event stays close to discord's events. This allows for `Context` to be taken out of `FullEvent` and be passed around separately. 

### Migration path for users
`cache_ready`: Differences in the examples show that it was mainly useless and could be re-implemented quite cleanly in user code with one or two lines.
`shards_ready`: Similar to `cache_ready`, however without an example showing the migration
`ratelimit`: Can be re-implemented in user code by getting access to `Ratelimiter` and using `set_ratelimit_callback`


kangalio:
Throwing this out here so I don't forget: we can probably do this in a 0.11 backwards compatible way. FullEvent is new anyways so it doesn't matter if we take away events from it. But we can still call the 3 old EventHandler methods (perhaps gated behind a runtime flag?). They'd be deprecated, of course, and removed in 0.13 or something

GnomedDev:
I don't really see the requirement for this to go through a deprecated cycle seeing that:

1. All the other changes required in this update mean that deprecated usage is probably going to updated at the same time
2. The ratelimit event is new to next, so there were no users of it anyway
3. `0.11` is probably going to stay usable for quite a while, seeing it is on the latest API version, so users can procrastinate updating for ages anyway.

I also strongly oppose a runtime flag being added to enable deprecated features... we should not be doing development work to help people who don't want to use the latest version of serenity.

kangalio:
1. What do you mean?
2. Good point, we'd only need to migrate ready and cache_ready then
3. Eh, just because the migration can be procrastinated doesn't mean it should be needlessly painful

> I also strongly oppose a runtime flag being added to enable deprecated features

I tend to agree, it's clunky to the point of not being worth it (I just put it in the comment brainstorm style)

> we should not be doing development work to help people who don't want to use the latest version of serenity

With this, are you saying that users should bring enough intrinsic motivation to use the latest version to endure a painful migration process? If so, I would disagree

GnomedDev:
1. Migrating to 0.12 is going to need a lot of required changes, so if a couple depreciation warnings are emitted they could easily be fixed at the same time (it's not like this is a massive change compared to everything else)

3. Yes... people should be motivated to update their libraries if they want updates??

kangalio:
1. Makes sense. Although if we actually remove the ready and cache_ready functions from EventHandler, we need to make it easy to figure out what the alternative is. With deprecation warnings it's easy because the compiler tells you directly, but if the methods are removed, the user needs to go search in some changelog or migration guide. This should be as painless as possible
2. If we go with this attitude and just willy nilly break backwards compatibility everywhere with no consideration for anything because "users will figure it out if they really want the update" we'll end up alienating users from this library

PR #2330, opened 2023-01-05
Add owner_id for group dms, threads and forum posts

WebbiiiCC:
Allows retrieving the creator/owner of a group dm or thread.
In code it just adds the owner_id (Option<UserId>) field for deserialization upon to_channel invocation.


arqunis:
I have manually merged this in https://github.com/serenity-rs/serenity/commit/a07410392915a6279c1b61e6a3f25c7d3fc37588 to fix the compilation of tests, and to exclude `owner_id` from `PrivateChannel`. Bots cannot be participants of Group channels, so it is unnecessary to include it there.

Issue #2324, opened 2022-12-27
Implement `Deserialize` on all builder types


AngelOnFira:
As I've been looking through types in the builder on the `next` branch, it seems that many are `Serialize`, but few are `Deserialize`. I'm hoping to have both traits implemented, as so I can store builders in my database to be used later. [For example](https://github.com/serenity-rs/serenity/blob/72c76467dab136f50c2b1739e1945736e4a7ff39/src/builder/create_components.rs#L151):

```rust
/// A builder for creating a [`SelectMenu`].
///
/// [`SelectMenu`]: crate::model::application::SelectMenu
#[derive(Clone, Debug, Serialize)]
#[must_use]
pub struct CreateSelectMenu {
  ...
}
```

I'm curious about what complications lie in just adding `Deserialize`? (Other than having to do it on any sub types).

AngelOnFira:
Also, if there are no large blockers on this, I wouldn't mind going through and setting this up best I can :)

AngelOnFira:
I made this commit on my fork so that I could get things working locally, and it *seems* to work on a surface level. Though I don't know if there are other footguns, since I've never really fooled around with Serde past the derive traits.

Commit link: https://github.com/AngelOnFira/serenity/commit/a7600e565240a9b520109ac34cb3334be337892b

kangalio:
Status: we need feedback from those users who'd like to store and load builders locally, in order to devise an alternative approach. What's your use case? If you could turn builder structs into a generic `Request` struct that could be stored locally and executed later, would that work for you?

(Using the existing Serialize impls is problematic because they are tuned specifically for the Discord API, and hence leave out some fields of the serialized payload (bad!) or require special formatting which would have to be replicated in the Deserialize impl (annoying!))

(this message was copied from the linked PR)

GnomedDev:
This is a bad idea, this is the same as making all builder fields public on a semver point of view. Let's not.

PR #2335, opened 2023-01-18
Add deserialize for create_components

AngelOnFira:
## This PR

This adds the `Deserialize` trait to several types in the `create_components` builder. It also adds tests where the deserialization logic is more custom.

The following types now implement `Deserialize`:
- `CreateActionRow`
- `CreateButton`
- `CreateSelectMenu`
- `CreateSelectMenuOption`
- `CreateInputText`

## Other items to note

- The deserialize logic for `CreateActionRow` will check if no components are being deserialized, and return an error if so. This can occur by accident if the `CreateActionRow` is of type `Buttons` and the interior `Vec` is empty, and this is serialized.

## Issues

This helps towards solving #2324

### DIFF
--- a/src/builder/create_components.rs
+++ b/src/builder/create_components.rs
@@ -574,7 +574,7 @@ mod test {
         let serialized_action_row: String = serde_json::to_string(&action_row).unwrap();
 
         let deserized_action_row: Result<CreateActionRow, _> =
-            serde_json::from_str(&serialized_action_row.replace("1", "2"));
+            serde_json::from_str(&serialized_action_row.replace('1', "2"));
 
         assert!(deserized_action_row.is_err());
 
--- a/src/builder/create_components.rs
+++ b/src/builder/create_components.rs
@@ -153,20 +153,6 @@ impl CreateButton {
         })
     }
 
-    /// Sets the custom id of the button, a developer-defined identifier. Replaces the current value
-    /// as set in [`Self::new`].
-    ///
-    /// Has no effect on link buttons.
-    pub fn custom_id(mut self, id: impl Into<String>) -> Self {
-        if let ButtonKind::NonLink {
-            custom_id, ..
-        } = &mut self.0.data
-        {
-            *custom_id = id.into();
-        }
-        self
-    }
-
     /// Sets the style of this button.
     ///
     /// Has no effect on link buttons.
--- a/src/builder/create_components.rs
+++ b/src/builder/create_components.rs
@@ -15,7 +15,7 @@ pub enum CreateActionRow {
 }
 
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


src/builder/create_components.rs
```diff
@@ -112,18 +194,18 @@ pub enum CreateSelectMenuKind {
...
     fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        #[derive(Serialize)]
-        struct Json<'a> {
-            #[serde(rename = "type")]
-            kind: u8,
-            options: Option<&'a [CreateSelectMenuOption]>,
-            channel_types: Option<&'a [ChannelType]>,
-        }
-
         #[rustfmt::skip]
```
AngelOnFira, line 200: Changing out this section might be a bit dirty, and it might be better to just have two separate types on each side if it is better to have slices while serializing. I read [here](https://users.rust-lang.org/t/problem-with-serde-serialize-of-slice-of-structs/28070/3) that it's probably best to use a `Vec`, though I don't really understand the rabbit hold of lifetimes/references while deserializing, and this was an easy fix.

src/builder/create_components.rs
```diff
@@ -14,26 +14,108 @@ pub enum CreateActionRow {
...
-                Self::SelectMenu(x) => serde_json::to_value(vec![x]).map_err(S::Error::custom)?,
-                Self::InputText(x) => serde_json::to_value(vec![x]).map_err(S::Error::custom)?,
-            }
-        })
+        let components: Vec<serde_json::Value> = match self {
+            Self::Buttons(x) => {
+                let mut buttons = Vec::new();
+
+                // This needs to be a for loop so that we can use `?` to return
+                // early on error
```
AngelOnFira: Don't know if these comments are helpful or not /shrug

AngelOnFira:
@kangalioo If you could take a look at this that would be great :smiley:

src/builder/create_components.rs
kangalio: They're limitedly helpful. They don't hurt though and can always be removed later if they do turn out to be annoyance

kangalio: Btw, you may be interested in the "collect to Result" pattern in the future: `let buttons = x.iter().map(|button| serde_json::to_value(button).map_err(S::Error::custom)?).collect::<Result<Vec<_>, _>>()?;`

kangalio (approved):

src/builder/create_components.rs
AngelOnFira: Thanks for the recommendation! I'll have to check out how that works exactly, but I swapped it out for now 👍

arqunis (approved):

PR #2337, opened 2023-01-21
Add deserialize trait for all builder types

### LINKED ISSUE #2324: Implement `Deserialize` on all builder types

AngelOnFira:
## This PR

This PR adds the `Deserialize` trait (and sometimes `Serialize`) to all types [found here](https://docs.rs/serenity/0.11.5/serenity/builder/index.html), barring several that don't seem to benefit from it.

## Notes

- #2335 should be merged first (I'll keep this a draft until that has been completed)

## Issues

This will close #2324

### DIFF
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -274,13 +274,13 @@ impl CreateInteractionResponseMessage {
     super::button_and_select_menu_convenience_methods!();
 }
 
-#[derive(Clone, Default, Debug, Serialize)]
+#[derive(Clone, Default, Debug, Serialize, Deserialize)]
 pub struct AutocompleteChoice {
     pub name: String,
     pub value: Value,
 }
 
-#[derive(Clone, Debug, Default, Serialize)]
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
 #[must_use]
 pub struct CreateAutocompleteResponse {
     choices: Vec<AutocompleteChoice>,
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -141,7 +141,7 @@ impl Builder for CreateInteractionResponse {
     }
 }
 
-#[derive(Clone, Debug, Default, Serialize)]
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
 #[must_use]
 pub struct CreateInteractionResponseMessage {
     #[serde(skip_serializing_if = "Option::is_none")]
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -54,22 +54,43 @@ pub enum CreateInteractionResponse {
     Modal(CreateModal),
 }
 
+#[derive(Serialize, Deserialize)]
+struct InteractionResponseJson {
+    #[serde(rename = "type")]
+    kind: u8,
+    data: serde_json::Value,
+}
+
 impl serde::Serialize for CreateInteractionResponse {
-    fn serialize<S: serde::Serializer>(&self, serializer: S) -> StdResult<S::Ok, S::Error> {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         use serde::ser::Error as _;
 
         #[allow(clippy::match_same_arms)] // hurts readability
-        serde_json::json!({
-            "type": match self {
-                Self::Pong { .. } => 1,
-                Self::Message { .. } => 4,
-                Self::Defer { .. } => 5,
-                Self::Acknowledge { .. } => 6,
-                Self::UpdateMessage { .. } => 7,
-                Self::Autocomplete { .. } => 8,
-                Self::Modal { .. } => 9,
+        InteractionResponseJson {
+            kind: match self {
+                Self::Pong {
+                    ..
+                } => 1,
+                Self::Message {
+                    ..
+                } => 4,
+                Self::Defer {
+                    ..
+                } => 5,
+                Self::Acknowledge {
+                    ..
+                } => 6,
+                Self::UpdateMessage {
+                    ..
+                } => 7,
+                Self::Autocomplete {
+                    ..
+                } => 8,
+                Self::Modal {
+                    ..
+                } => 9,
             },
-            "data": match self {
+            data: match self {
                 Self::Pong => serde_json::Value::Null,
                 Self::Message(x) => serde_json::to_value(x).map_err(S::Error::custom)?,
                 Self::Defer(x) => serde_json::to_value(x).map_err(S::Error::custom)?,
@@ -77,9 +98,33 @@ impl serde::Serialize for CreateInteractionResponse {
                 Self::UpdateMessage(x) => serde_json::to_value(x).map_err(S::Error::custom)?,
                 Self::Autocomplete(x) => serde_json::to_value(x).map_err(S::Error::custom)?,
                 Self::Modal(x) => serde_json::to_value(x).map_err(S::Error::custom)?,
-            }
-        })
+            },
+        }
         .serialize(serializer)
+    }
+}
+
+impl<'de> serde::Deserialize<'de> for CreateInteractionResponse {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        use serde::de::Error as _;
+
+        let InteractionResponseJson {
+            kind,
+            data,
+        } = serde::Deserialize::deserialize(deserializer)?;
+
+        let data = match kind {
+            1 => Self::Pong,
+            4 => Self::Message(serde_json::from_value(data).map_err(D::Error::custom)?),
+            5 => Self::Defer(serde_json::from_value(data).map_err(D::Error::custom)?),
+            6 => Self::Acknowledge,
+            7 => Self::UpdateMessage(serde_json::from_value(data).map_err(D::Error::custom)?),
+            8 => Self::Autocomplete(serde_json::from_value(data).map_err(D::Error::custom)?),
+            9 => Self::Modal(serde_json::from_value(data).map_err(D::Error::custom)?),
+            _ => return Err(D::Error::custom("invalid interaction response type")),
+        };
+
+        Ok(data)
     }
 }
 
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -361,7 +361,7 @@ impl Builder for CreateAutocompleteResponse {
     }
 }
 
-#[derive(Clone, Debug, Default, Serialize)]
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
 #[must_use]
 #[non_exhaustive]
 pub struct CreateModal {
--- a/src/builder/create_sticker.rs
+++ b/src/builder/create_sticker.rs
@@ -21,7 +21,7 @@ pub struct CreateSticker<'a> {
     name: String,
     tags: String,
     description: String,
-    
+
     #[serde(skip)]
     file: CreateAttachment,
     #[serde(skip)]
--- a/src/builder/create_sticker.rs
+++ b/src/builder/create_sticker.rs
@@ -15,7 +15,7 @@ use crate::model::prelude::*;
 /// - [`PartialGuild::create_sticker`]
 /// - [`Guild::create_sticker`]
 /// - [`GuildId::create_sticker`]
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 #[must_use]
 pub struct CreateSticker<'a> {
     name: String,
--- a/src/builder/create_sticker.rs
+++ b/src/builder/create_sticker.rs
@@ -21,6 +21,8 @@ pub struct CreateSticker<'a> {
     name: String,
     tags: String,
     description: String,
+    
+    #[serde(skip)]
     file: CreateAttachment,
     #[serde(skip)]
     audit_log_reason: Option<&'a str>,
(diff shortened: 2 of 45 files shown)


kangalio:
Hi :) Is this PR done (apart from the merge errors due to the rebase)?

AngelOnFira:
@kangalio Sorry for the delay, I fixed this branch thanks to @mkrasnitski :)

AngelOnFira:
I'll try to check out the failing checks though

src/builder/create_interaction_response.rs
```diff
@@ -54,35 +54,80 @@ pub enum CreateInteractionResponse {
...
+                } => 6,
+                Self::UpdateMessage {
+                    ..
+                } => 7,
+                Self::Autocomplete {
+                    ..
+                } => 8,
+                Self::Modal {
+                    ..
+                } => 9,
```
mkrasnitski, line 91:
```suggestion
                Self::Pong => 1,
                Self::Message(_) => 4,
                Self::Defer(_) => 5,
                Self::Acknowledge => 6,
                Self::UpdateMessage(_) => 7,
                Self::Autocomplete(_) => 8,
                Self::Modal(_) => 9,
```

Weird match patterns here that got exposed by rustfmt once it wasn't wrapped in a macro call.

mkrasnitski (changes requested):
Looks good to me, sans some minor formatting.

mkrasnitski:
Whoops, one more thing. There's been a new builder added since this PR was opened: `CreateForumPost`.

kangalio:
We can add `serde::Serialize + serde::Deserialize` bounds to the `Builder` trait to enforce this. Those bounds currently only fail on `CreateForumPost` (expected) and `CreateQuickModal`.

`CreateQuickModal` doesn't really fit in with the other builders anyways... I'm thinking it should probably not implement Builder and also be moved to the utils module?

mkrasnitski:
> We can add `serde::Serialize + serde::Deserialize` bounds to the `Builder` trait to enforce this.

That would require adding a lifetime parameter to `Builder`. We could alternatively bound on `DeserializeOwned` (do we care about zero-copy deserialization at all?), but I don't think it's really that necessary to have bounds. And yeah, `CreateQuickModal` doesn't impl `Serialize` in the first place, so I'm not really sure there.

kangalio:
> That would require adding a lifetime parameter to Builder

Is that a dealbreaker? I think the return is worth it:
- clearly signals to users that they can expect any builder to be (de)serializable
- enforces this statically, including for any new PRs

The fact that CreateQuickModal doesn't implement Serialize is just a symptom of the underlying problem that CreateQuickModal is misplaced, in my opinion. In my [initial quick modal PR](https://github.com/serenity-rs/serenity/pull/2236) I was already unsure about where to put it.

All other builder types in that module belong to an HTTP request and clearly represent that request's fields as per Discord documentation. CreateQuickModal doesn't; it's an arbitrary abstraction on top of multiple HTTP requests (and gateway events, violating the pattern that the builder module doesn't have anything to do with gateway). I'm currently drafting a PR to move it to utils

AngelOnFira:
@mkrasnitski @kangalio how does this commit look? https://github.com/serenity-rs/serenity/pull/2337/commits/48073cc93e3102c777eceb187b2f5189f9785395

I haven't done a ton with lifetimes, so it could be I'm doing something a little off 😅

src/builder/create_sticker.rs
```diff
@@ -84,7 +84,10 @@ impl<'a> CreateSticker<'a> {
 
 #[cfg(feature = "http")]
 #[async_trait::async_trait]
-impl<'a> Builder for CreateSticker<'a> {
+impl<'a, 'de> Builder<'de> for CreateSticker<'a>
+where
+    'de: 'a,
+{
```
AngelOnFira: I'd wonder the most about this one, I had to add this `where` clause for it to compile.

kangalio:
Wait did I accidentally push my #2374 commits to your PR branch? 👀 Or did you manually copy them over

src/builder/create_sticker.rs
kangalio: It worked fine with just `impl<'a> Builder<'a> for CreateSticker<'a>` for me

kangalio: Oh you introduced an extra lifetime in all Builder impls? Reuse the same lifetime; adding an extra one is redundant.

kangalio:
And we should probably wait for @mkrasnitski 's response since they weren't a fan of the trait bounds

AngelOnFira:
> Wait did I accidentally push my #2374 commits to your PR branch? 👀 Or did you manually copy them over

I rebased so I would have it to use :)

re: Trait bounds, ya, I mostly made the commit to just see what it would look like. I quite like the idea of statically enforcing it, but I know very little about how it might scale, or what issues could arise. I'm happy to remove the commit if we deem it the wrong path :)

kangalio:
> We could alternatively bound on DeserializeOwned (do we care about zero-copy deserialization at all?)

After thinking about it, yeah we can do that. Serenity itself only need the serialization part anyways, deserialization is more of a convenient bonus (for users who are brave enough to rely on serenity's Discord-tailored serialization to store builders locally :woozy_face:). And it gets rid of the lifetime, i.e. not even any complexity added anymore

src/builder/create_sticker.rs
mkrasnitski: This is actually a bug - for some reason `CreateSticker::audit_log_reason` isn't marked `#[serde(skip)]`.

mkrasnitski:
Besides the stuff with `DeserializeOwned` and also `CreateSticker`, what should be done about any `files` fields on builders? They're skipped when serializing, which means users will have to re-add them if Deserializing a locally saved builder.

kangalio:
That's a side-effect of serenity's serde impls only being meant for the Discord API, not for local storing. I think our best course of action is to tell users "sorry, Discord doesn't expect files to be transmitted within the JSON so we have to adhere to that".

<sub>Actually I didn't even think of this myself and it just reaffirms my belief that trying to misuse serenity's builder struct serde implementations to store/load them locally is setting yourself up for failure, and we should not encourage this fundamentally bad idea by adding Serialize impls that only such users would need</sub>

mkrasnitski:
It might be possible to accomplish using `#[serde(serialize_with)]` to conditionally skip serializing those fields only if sending a request to Discord, but it's probably fine. Might be worth clarifying that serializing a builder currently only produces the *body* of the request, omitting any headers or form data (including files). That would be a lie in the case of `CreateSticker` (if #2376 gets merged), but it's not really a big deal.

src/builder/mod.rs
```diff
@@ -12,9 +12,16 @@ use crate::http::CacheHttp;
...
 
+/// Common trait for all HTTP request builders in this module.
+///
+/// Will have a [`serde::Serialize`]` + `[`serde::Deserialize`] bound in the future, because
+/// builders need to be serializable in order to send them to the Discord API and
+/// [`serde::Deserialize`] is useful to be able to store builders locally.
 #[cfg(feature = "http")]
 #[async_trait::async_trait]
-pub trait Builder {
+pub trait Builder<'de>: serde::Serialize + serde::Deserialize<'de> {
```
mkrasnitski:
```suggestion
pub trait Builder: serde::Serialize + serde::DeserializeOwned {
```
To avoid adding extra complexity, I think bounding on `DeserializeOwned` is better. None of the serializable fields in builders have any lifetimes, anyways.

AngelOnFira:
> Actually I didn't even think of this myself and it just reaffirms my belief that trying to misuse serenity's builder struct serde implementations to store/load them locally is setting yourself up for failure, and we should not encourage this fundamentally bad idea by adding Serialize impls that only such users would need

Maybe down the line, it could be good to document where the edges of this pattern should lie? I currently use this to be able to prepare a builder from something that just happened, so that it can be sent at an arbitrary time in the future. The alternative to this is having my own parallel way to store all the data, which seems a bit redundant :P so I've been quite invested in these changes.

src/builder/create_sticker.rs
```diff
@@ -15,13 +15,14 @@ use crate::model::prelude::*;
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
 #[must_use]
 pub struct CreateSticker<'a> {
     name: String,
     tags: String,
     description: String,
     file: CreateAttachment,
+    #[serde(skip)]
     audit_log_reason: Option<&'a str>,
 }
```
AngelOnFira: @mkrasnitski so for `CreateSticker` the `#[serde(skip)]` does seem to work. Is this the best way to go? If so, then it seems like `DeserializeOwned` should be the way to go.

src/builder/create_sticker.rs
```diff
@@ -15,13 +15,14 @@ use crate::model::prelude::*;
...
```
mkrasnitski:
```suggestion
#[derive(Clone, Debug, Serialize, Deserialize)]
#[must_use]
pub struct CreateSticker<'a> {
    name: String,
    tags: String,
    description: String,
    
    #[serde(skip)]
    file: CreateAttachment,
    #[serde(skip)]
    audit_log_reason: Option<&'a str>,
}
```
I think the `file` field should also be skipped for consistency.

mkrasnitski:
Argh, apparently `CreateAttachment` doesn't impl `Default`. What's your thoughts on this @kangalio? Should the field not be skipped after all?

AngelOnFira:
I just saw that too 😅

kangalio:
> Maybe down the line, it could be good to document where the edges of this pattern should lie? I currently use this to be able to prepare a builder from something that just happened, so that it can be sent at an arbitrary time in the future. The alternative to this is having my own parallel way to store all the data, which seems a bit redundant :P so I've been quite invested in these changes.

Do I understand correctly: some event happens, to which your bot wants to respond at some unknown time in the future, and this wait time may even persist across bot restarts?
1) I'm very interested with your bot is actually doing
2) For the purpose of "preparing a request and sending later" I think the best way would be to add Serialize+Deserialize to serenity's own [Request type](https://serenity-rs.github.io/serenity/next/serenity/http/struct.Request.html) and save that instead

AngelOnFira:
Ya, at some point I'll try to document how stuff is getting stored on the project I'm doing. I haven't touched it in a bit, but I do need to pick it up again soonish.

As for your second point, it might work? I feel like having the builder available to change later is also real nice.

kangalio:
Oh I just meant like, why does your bot need to respond later, why not immediately? What kind of bot is it, a moderation bot?

> I feel like having the builder available to change later is also real nice.

Sure, but that's messy to implement. We can't re-use the existing Serialize impls because those omit fields that are not transmitted via JSON (but via HTTP form or query param). We'd need some way to configure the Serialize impls at runtime, or put two Serialize impls on one struct, or have some alternative parallel (de)serialization scheme, or maybe something different. None of what I can think of is remotely pleasant.

Can you think of concrete use cases of having the builders themselves available to change? If not, (de)serializing the Request object is hopefully enough for all use cases

AngelOnFira:
I'll move this discussion to Discord 👍

AngelOnFira:
> Argh, apparently `CreateAttachment` doesn't impl `Default`. What's your thoughts on this @kangalio? Should the field not be skipped after all?

Thoughts on this?

kangalio:
CreateAttachment could gain a Default impl, but that would make the (De)Serialize impls of CreateSticker pretty useless, because the actual sticker image data would not get serialized. The best short-term solution would probably to not implement Deserialize and Builder for CreateSticker and just leave it as an ugly exception for now. Until we figure out what to do about the builder fields not part of the Discord API JSON

AngelOnFira:
This would mean removing the trait bound right?

kangalio:
> This would mean removing the trait bound right?

Not necessarily; the `execute()` method could be moved to `impl CreateSticker`. Although it's probably for the best to remove the trait bound, seeing as we _cannot_ actually guarantee all builders to have a functioning Deserialize impl

mkrasnitski:
IMO, having `CreateSticker` be an exception and removing its `Builder` impl doesn't really make sense, since it's a fully fledged builder through and through, as opposed to `CreateQuickModal`. I think probably the best course of action is to remove the trait bounds (plus the changes to `CreateSticker`) and add them later once the concerns we ran into here have been addressed down the line.

kangalio:
Status: we need feedback from those users who'd like to store and load builders locally, in order to devise an alternative approach. What's your use case? If you could turn builder structs into a generic `Request` struct that could be stored locally and executed later, would that work for you? @AngelOnFira @loganintech (#2338)

(Using the existing Serialize impls is problematic because they are tuned specifically for the Discord API, and hence leave out some fields of the serialized payload (bad!) or require special formatting which would have to be replicated in the Deserialize impl (annoying!))

kangalio:
Since we decided not to add Deserialize impls for now, especially because they can be always added later in a patch release but not removed, I'll close this PR

Thank you for your work @AngelOnFira

AngelOnFira:
No worries, I'll open another issue once I get back on the project I'm working on, and find if there is a specific use case that can't be satisfied in other ways :+1:

kangalio:
Thanks, that would be valuable

PR #2341, opened 2023-02-05
Unify documentation formatting across the codebase

mkrasnitski:
The main change is wrapping docs and comments at a consistent 100 characters (which is the standard value for Rust). This does not change how the docs are rendered, so it's only an internal change. In a few places I adjusted wording or fixed typos in docs.

Besides that, replaced some of the initialization code in doctests with calls to `unimplemented!()`, since those lines are hidden when rendered, and the tests are marked `no_run`, so they are only type-checked.


kangalio:
Is it possible to automate this formatting via rustfmt? (If not, I wonder if we should even attempt to unify the doc formatting at all, because it will invariably drift out of sync again and it's not like that's a big deal)

mkrasnitski:
Rustfmt will not do this automatically if lines are shorter than the linewidth, which is why most of this PR was done manually. However, it will wrap them if they are longer. The rustfmt option itself is quite unstable, but most editors (e.g. neovim, which I use) will automatically wrap comments at 100 characters while you're typing them. This can also be caught in new code as PRs are submitted.

arqunis (approved):

kangalio:
> However, it will wrap them if they are longer. The rustfmt option itself is quite unstable

Could you add it to the PR? Serenity already relies on loads of unstable rustfmt options anyways

I just feel like these huge mechanical transformation PRs should ensure that new code also follows the new pattern. Otherwise it's an uphill battle

Edit: woops was merged already

mkrasnitski:
The relevant options are `wrap_comments` and `comment_width`. Apparently the default value is 80.

PR #2347, opened 2023-02-25
Fix `defer_ephemeral` for modals

mkrasnitski:
Deferring an interaction ephemerally should be done using `DeferredChannelMessageWithSource`. Apparently modals support also using `DeferredUpdateMessage`, but this is undocumented.

### DIFF
--- a/src/model/application/interaction/modal.rs
+++ b/src/model/application/interaction/modal.rs
@@ -269,7 +269,7 @@ impl ModalSubmitInteraction {
     /// API response.
     pub async fn defer_ephemeral(&self, http: impl AsRef<Http>) -> Result<()> {
         self.create_interaction_response(http, |f| {
-            f.kind(InteractionResponseType::DeferredUpdateMessage)
+            f.kind(InteractionResponseType::DeferredChannelMessageWithSource)
                 .interaction_response_data(|f| f.ephemeral(true))
         })
         .await


arqunis (approved):

matveitolochko:
@mkrasnitski broken commit.
[MSRV](https://github.com/serenity-rs/serenity/actions/runs/4295026659/jobs/7484872395) Fails.
You need to update the version of `time`.
From:
```toml
...
62. package = "time"
63. version = "0.3.6"
...
```
To:
```toml
...
62. package = "time"
63. version = "0.3.20"
...
```
Due to the workflow code here:
```yaml
...
# time 0.3.10+ MSRV is 1.57
- name: Pin time to 0.3.9
  run: |
    # We have `time` 0.1 & 0.3 in the tree
    latest_version=$(cargo search --limit 1 time | head -1 | cut -d'"' -f2)
    cargo update -p time:$latest_version --precise 0.3.9
...
```

mkrasnitski:
I have doubts that changing the version in `Cargo.toml` will fix it - the version of `time` on CI is 0.3.15 (for some reason). Besides, `cargo` is aware of semantic versioning and should pick the latest minor release compatible with 0.3.6, which should be 0.3.20. So, I don't know the reason why CI is failing.

matveitolochko:
`time:0.3.15` is being downloaded as a sub dependency of something, which explains where it appears from.
```log
error: package ID specification `time:0.3.20` did not match any packages
Did you mean one of these?

  time:0.1.45
  time:0.3.15
```
I am quite sure this only retrieves a strict version.
`latest_version` will in this case be `0.3.20` which is the latest
```
latest_version=$(cargo search --limit 1 time | head -1 | cut -d'"' -f2)
cargo update -p time:$latest_version --precise 0.3.9
```

mkrasnitski:
Looking in `Cargo.lock`, it doesn't seem like `time 0.3` is a sub-dependency anywhere. That "latest version" check assumes that the correct version is being picked up by cargo, which should match (at least until a 0.4 release happens).

PR #2349, opened 2023-03-04
Unwrap .spawn() to support updated tokio task builder API

zackradisic:
When compiling serenity with the `tokio_task_builder` feature enabled, the following error occurs:
```
error[E0308]: mismatched types
 --> /Users/zackradisic/.cargo/registry/src/github.com-1ecc6299db9ec823/serenity-0.11.5/src/internal/tokio.rs:9:5
  |
4 | pub fn spawn_named<F, T>(name: &str, future: F) -> tokio::task::JoinHandle<T>
  |                                                    -------------------------- expected `tokio::task::JoinHandle<T>` because of return type
...
9 |     tokio::task::Builder::new().name(&*format!("serenity::{}", name)).spawn(future)
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `tokio::task::JoinHandle`, found enum `std::result::Result`
  |
  = note: expected struct `tokio::task::JoinHandle<_>`
               found enum `std::result::Result<tokio::task::JoinHandle<_>, std::io::Error>`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `serenity` due to previous error
```

In [this commit](https://github.com/tokio-rs/tokio/commit/3b6c74a40ad87837ab53c105a323c2807a4e1b38)  tokio's `task::Builder::spawn*` methods were changed to return an `std::io::Result<_>` to allowing opt-in to gracefully handling errors when spawning tasks ([PR](https://github.com/tokio-rs/tokio/pull/4823)). According to git it has been present since tokio version 1.21.0

The PR says the previous behaviour was to simply panic on failure, so a simple fix is to do the same and `.unwrap()` the value returned from spawning

### DIFF
--- a/src/internal/tokio.rs
+++ b/src/internal/tokio.rs
@@ -6,7 +6,10 @@ where
     F: Future<Output = T> + Send + 'static,
     T: Send + 'static,
 {
-    tokio::task::Builder::new().name(&*format!("serenity::{}", name)).spawn(future).unwrap()
+    tokio::task::Builder::new()
+        .name(&*format!("serenity::{}", name))
+        .spawn(future)
+        .expect("called outside tokio runtime")
 }
 
 #[cfg(not(all(tokio_unstable, feature = "tokio_task_builder")))]
--- a/src/internal/tokio.rs
+++ b/src/internal/tokio.rs
@@ -6,7 +6,7 @@ where
     F: Future<Output = T> + Send + 'static,
     T: Send + 'static,
 {
-    tokio::task::Builder::new().name(&*format!("serenity::{}", name)).spawn(future)
+    tokio::task::Builder::new().name(&*format!("serenity::{}", name)).spawn(future).unwrap()
 }
 
 #[cfg(not(all(tokio_unstable, feature = "tokio_task_builder")))]


mkrasnitski:
This is alright to do on current (since the build is otherwise broken), but on `next` it might be better to change the function signature to return a `Result<JoinHandle<T>>`.

src/internal/tokio.rs
```diff
@@ -6,7 +6,7 @@ where
     F: Future<Output = T> + Send + 'static,
     T: Send + 'static,
 {
-    tokio::task::Builder::new().name(&*format!("serenity::{}", name)).spawn(future)
+    tokio::task::Builder::new().name(&*format!("serenity::{}", name)).spawn(future).unwrap()
```
mkrasnitski: You should either decorate this with `#[allow(clippy::unwrap_used)]`, or use `.expect()`.

kangalio: Probably `.expect("called outside tokio runtime")`. That way, readers of the code know it's very reasonable to panic here

kangalio (approved):

arqunis (approved):

PR #2360, opened 2023-03-10
Remove redundant `ChannelId` in `CreateStageInstance` builder

mkrasnitski:
Change the `channel_id` parameter to be required only at execution time. However, since the payload must also include a `channel_id` field, make the field an `Option<ChannelId>` and fill it in at execution time.

Also, adds support for the `send_start_notification` field.

Breaking changes: 
 \- Remove `CreateStageInstance::channel_id`.
 \- Remove the `channel_id` parameter of `CreateStageInstance::new`.

### DIFF
--- a/src/builder/create_stage_instance.rs
+++ b/src/builder/create_stage_instance.rs
@@ -8,8 +8,10 @@ use crate::model::prelude::*;
 #[derive(Clone, Debug, Serialize)]
 #[must_use]
 pub struct CreateStageInstance<'a> {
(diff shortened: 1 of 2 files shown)


mkrasnitski:
I just realized that using a private struct isn't the best way to do things. Making `channel_id` an `Option` actually has the exact same tradeoffs, turns out. Also, `EditVoiceState` already does things this way.

arqunis (approved):

PR #2367, opened 2023-03-19
Add GuildAuditLogEntryCreate

cheesycod:
Adds GuildAuditLogEntryCreate and the GUILD_MODERATION gateway intent. Untested but *should* (TM) work

### DIFF
--- a/src/client/event_handler.rs
+++ b/src/client/event_handler.rs
@@ -6,6 +6,7 @@ use super::context::Context;
 use crate::client::bridge::gateway::event::*;
 use crate::http::RatelimitInfo;
 use crate::model::application::{CommandPermission, Interaction};
+use crate::model::guild::audit_log::AuditLogEntry;
 use crate::model::guild::automod::{ActionExecution, Rule};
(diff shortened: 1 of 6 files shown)


cheesycod:
This now works and is ready for review

arqunis (approved):

src/client/event_handler.rs
```diff
@@ -139,6 +140,10 @@ event_handler! {
     /// The old channel data is only provided when the cache feature is enabled.
     async fn channel_update(&self, ChannelUpdate { ctx: Context, old: Option<Channel>, new: Channel });
 
+    /// Disptahced when a new audit log entry is created
+    ///
```
mkrasnitski, line 144: Typo that was missed before merge. Also best to put a period at the end. The extra line is also unnecessary.

arqunis: [Manually fixed](https://github.com/serenity-rs/serenity/commit/b792c981d965644a6920ae3d47d9889c1da99520).

PR #2372, opened 2023-04-04
Remove ShardManagerMonitor and other gateway cleanup

kangalio:
The bulk of this PR is removing ShardManagerMonitor. It was just a background task that received ShardManagerMessage's and called the respective ShardManager function. Now, you can just call the respective ShardManager function directly.

The commits about removing ShardManagerMonitor are marked with `[SMM]`. The other commits do other miscellaneous related clean-up. See commit titles and descriptions for more details.

Largely untested


mkrasnitski:
Will review this later today. Just quickly noticed that the docs CI job is failing.

src/client/bridge/gateway/shard_manager.rs
```diff
@@ -117,30 +111,43 @@ pub struct ShardManager {
...
+
         let mut shard_queuer = ShardQueuer {
             data: opt.data,
             event_handlers: opt.event_handlers,
             raw_event_handlers: opt.raw_event_handlers,
             #[cfg(feature = "framework")]
             framework: opt.framework,
             last_start: None,
-            manager_tx: thread_tx.clone(),
+            manager: manager.clone(),
```
mkrasnitski:
```suggestion
            manager: Arc::clone(&manager),
```

To indicate that this is just a refcount increment.

src/client/bridge/gateway/mod.rs
```diff
@@ -43,7 +43,7 @@
 pub mod event;
 
 mod shard_manager;
-mod shard_manager_monitor;
+// mod shard_manager_monitor;
```
mkrasnitski:
```suggestion
```

src/client/bridge/gateway/shard_queuer.rs
```diff
@@ -187,7 +185,7 @@ impl ShardQueuer {
             raw_event_handlers: self.raw_event_handlers.clone(),
             #[cfg(feature = "framework")]
             framework: self.framework.get().map(Arc::clone),
-            manager_tx: self.manager_tx.clone(),
+            manager: self.manager.clone(),
```
mkrasnitski:
```suggestion
            manager: Arc::clone(&self.manager),
```
Ditto.

mkrasnitski (changes requested):
I'd say this looks good, aside from a few tiny nits. Untangling the mess that is the gateway code is definitely a good thing (I say this as someone relatively unfamiliar with gateway internals).

Could you provide a sentence or two describing the breaking changes in this PR, for changelog purposes?

kangalio:
```
- Removed `InterMessage`, `ShardClientMessage`, `ShardManagerMessage`, `ShardManagerMonitor`, `ShardManagerError`
- Changed `ShardManager::new` return type from `(_, ShardManagerMonitor)` to `(_, UnboundedReceiver<Result<(), GatewayError>)`
  - Instead of calling `ShardManagerMonitor.run()`, you just wait for the return value to arrive via the `UnboundedReceiver`
- Changed `ShardMessenger::new` parameter type from `Sender<InterMessage>` to `Sender<ShardRunnerMessage>`
  - All shard-relevant actions have been centralized into `ShardRunnerMessage`
- Changed `ShardMessenger::send_to_shard` return type from `Result<_, _>` to `()`
  - Errors are now logged in the function directly
- Changed `ShardQueuer`'s and `ShardRunnerOptions`'s field `manage_tx: Sender<ShardManagerMessage>` field to `manager: Arc<Mutex<ShardManager>>`
  - Instead of invoking actions on `ShardManager` by sending messages through a channel, you now invoke methods on `ShardManager` directly
```

Since you said "a sentence or two":
```
`ShardManagerMonitor` has been removed in favor of interacting with `ShardManager` directly.
```

arqunis (approved):

kangalio:
Here is some additional context about the removed types:

Removed:
- `ShardClientMessage`
    - used to send commands to a ShardRunner. But we already had a type for that: `ShardRunnerMessage`. `ShardClientMessage` existed solely because some shard runner commands ended up in the `ShardManagerMessage` enum for mysterious reasons, so `ShardClientMessage` was an enum over either `ShardRunnerMessage` or `ShardManagerMessage`. This PR moves all shard runner commands into `ShardRunnerMessage`, so you should use that type instead now.
- `InterMessage`
    - an enum that only ever contained `ShardClientMessage`. Use that type directly; or rather it's new replacement `ShardRunnerMessage`.
- `ShardManagerMonitor`
    - received `ShardManagerMessage`'s and propagated them to the actual shard manager. This redundancy has been removed and you can instead call methods on `ShardManager` directly now. All places that previously stored a `Sender<ShardManagerMessage`> now have a `Arc<Mutex<ShardManager>>`.
- `ShardManagerMessage`
    - see above
- `ShardManagerError`
    - duplicated parts of `GatewayError` and was removed in favor of `GatewayError`.

Thanks to @tazz4843 for pointing out the lacking explanation

PR #2374, opened 2023-04-07
Move CreateQuickModal to utils

kangalio:
All other builder types in that module belong to an HTTP request and clearly represent that request's fields as per Discord documentation. CreateQuickModal doesn't; it's an arbitrary abstraction on top of multiple HTTP requests (and gateway events, violating the pattern that the builder module doesn't have anything to do with gateway). This kind of feature fits better in the utils module

This also opens the possibility to add `serde::Serialize + serde::Deserialize` bounds to the `Builder` trait (cc #2337)


src/builder/mod.rs
```diff
@@ -12,9 +12,16 @@ use crate::http::CacheHttp;
 #[cfg(feature = "http")]
 use crate::internal::prelude::*;
 
+/// Common trait for all HTTP request builders in this module.
+///
+/// Will have a [`serde::Serialize`]` + `[`serde::Deserialize`] bound in the future, because
+/// builders need to be serializable in order to send them to the Discord API and
+/// [`serde::Deserialize`] is useful to be able to store builders locally.
```
mkrasnitski:
```suggestion
/// Common trait for all HTTP request builders in this module.
```

I don't think the clarification here is necessary.

src/builder/mod.rs
```diff
@@ -12,9 +12,16 @@ use crate::http::CacheHttp;
...
 #[cfg(feature = "http")]
 #[async_trait::async_trait]
 pub trait Builder {
+    /// Further context that's not stored in the builder itself, like which channel to execute the
+    /// request in.
     type Context<'ctx>;
     type Built;
```
mkrasnitski:
```suggestion
Serializes a builder's fields and sends the request off the API, returning the response.
```

src/builder/mod.rs
```diff
@@ -12,9 +12,16 @@ use crate::http::CacheHttp;
...
+/// Common trait for all HTTP request builders in this module.
+///
+/// Will have a [`serde::Serialize`]` + `[`serde::Deserialize`] bound in the future, because
+/// builders need to be serializable in order to send them to the Discord API and
+/// [`serde::Deserialize`] is useful to be able to store builders locally.
 #[cfg(feature = "http")]
 #[async_trait::async_trait]
 pub trait Builder {
+    /// Further context that's not stored in the builder itself, like which channel to execute the
+    /// request in.
```
mkrasnitski:
```suggestion
    /// Additional data that's only required when sending a request off to the API.
```

mkrasnitski (changes requested):
I think this makes sense to do. Got some minor docs changes for you though.

arqunis (approved):

PR #2376, opened 2023-04-07
Derive `Serialize` for `CreateSticker`

mkrasnitski:
`CreateSticker::audit_log_reason` accidentally wasn't marked to be skipped in #2105.


mkrasnitski:
Whoops, apparently `CreateSticker` doesn't even impl `Serialize`. I think this might require writing a custom serializer.

mkrasnitski:
Writing a whole serializer just to serialize 3 fields in a struct doesn't make sense.

mkrasnitski:
Addressed by #2337.

PR #2380, opened 2023-04-08
Move client::bridge into gateway

kangalio:
The `serenity::client::bridge::gateway::` paths were always an eye-sore. And the types within (ShardManager, ShardRunner, ShardId, StageUpdateEvent...) are integral parts of connecting to the Discord gateway, so they should rather be moved to `serenity::gateway`, were the rest of the types already are (Shard, WsClient, ConnectionStage...).

`serenity::gateway` now:
Screenshot_20230408_231759


mkrasnitski:
This also changes the feature flag on these types from `client` to `gateway`, right? Just making sure, is that intentional? I don't know the use-cases when one feature but not the other is enabled, but maybe you might.

kangalio:
`client` requires `gateway` (logically) and `gateway` is de-facto dependent on `client` too. For example in 0.11.5:

Screenshot_20230409_003546

I've never seen this spelled out so clearly. We could actually just merge the two flags :thinking:

arqunis (approved):

Issue #2333, opened 2023-01-08
Race condition when `SUPPRESS_EMBEDS` is set when editing a message


oSumAtrIX:
Discord does not cancel their async task when they receive a request to edit a message with the flag `SUPPRESS_EMBEDS`. So when a message is sent and the flag is set when sending the message edit request right after that, Discord will still display the embed. 

For that reason, Serenity should queue a task to edit the message after `MESSAGE_UPDATE` event is fired to fix this race condition. A certain timeout should be considered as well which Discord does not document.

If this is too abstract for Serenity, this race condition should be documented in the inline docs.

kangalio:
Can't reproduce. Code:

```rust
let mut msg = channel_id.say(&ctx, "https://github.com/serenity-rs/serenity/issues/2333").await?;
msg.edit(&ctx, EditMessage::new().suppress_embeds(true)).await?;
```

Can you provide a minimal reproducible example?

oSumAtrIX:
The race condition is a little rare, but most of the time, it works, though I was able to reproduce the race condition with the following code:

https://github.com/revanced/revanced-discord-bot/blob/main/src/utils/code_embed/utils.rs#L96-L98

kangalio:
> The race condition is a little rare, but most of the time, it works

I just tried 63 times and the embed was successfully suppressed everytime except the very first try - and that one resolved after refreshing, so it was just a client-side display error

I am testing on the next branch, although I'd be surprised if that was the reason for our different results - the sent API requests should have stayed same

oSumAtrIX:
Just to be sure, how does Serenity handle suppressing embeds? I am imagining following scenario:

1. User sends a link
2. Discord server starts parsing the link
3. Serenity suppresses an embed
4. Discord server has not parsed the link yet and therefore dismisses the suppression
5. Embed is displayed on client once Discord finishes parsing

In this scenario we would have a race condition. I can also imagine Discord queuing suppressing the embed before finishing parsing the link. In that case this would not explain, why the first time, the embed was not suppressed in your test case.

As a new test case, a slow loading page can be used to check if there's a race condition. Discord probably only parses until `</head>`, so if the site sends the content slowly the window to suppress the embed to cause a race condition could be higher.

kangalio:
> how does Serenity handle suppressing embeds?

First sends an HTTP request for sending the message, then sends an HTTP request to suppress embeds.

[...]

PR #2388, opened 2023-04-17
Register collectors via shared Mutex instead of channel

kangalio:
That means, collector registration sidesteps the 500ms "game loop" of ShardRunner, which previously caused issues with missed events (https://discord.com/channels/381880193251409931/673965002805477386/1097264574828396646, https://github.com/serenity-rs/serenity/issues/2333#issuecomment-1510872780).

(This fix makes the workaround code in #2389 actually work)

In the future, more ShardRunnerMessage variants can be expressed as simple method calls on a Mutex-guarded resource. This would make function calls much more traceable, it's also more performant, and eliminates an extra type and a background loop

### DIFF
--- a/src/gateway/bridge/shard_runner.rs
+++ b/src/gateway/bridge/shard_runner.rs
@@ -46,7 +46,7 @@ pub struct ShardRunner {
     pub cache: Arc<Cache>,
     pub http: Arc<Http>,
     #[cfg(feature = "collector")]
-    collectors: Vec<CollectorCallback>,
+    pub(crate) collectors: std::sync::Arc<std::sync::Mutex<Vec<CollectorCallback>>>,
 }
 
 impl ShardRunner {
@@ -70,7 +70,7 @@ impl ShardRunner {
             cache: opt.cache,
             http: opt.http,
             #[cfg(feature = "collector")]
-            collectors: vec![],
+            collectors: std::sync::Arc::new(std::sync::Mutex::new(vec![])),
         }
     }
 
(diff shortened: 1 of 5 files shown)


kangalio:
Breaking changes:

- `ShardMessenger::new` is not `const` anymore
    - the required argument wasn't obtainable in a const context anyway
- `ShardMessenger::new` takes a `&ShardRunner` instead of a `Sender<ShardRunnerMessage>` now
- `ShardRunnerMessage::AddCollector` was removed
    - to add a collector manually, use `ShardMessenger::add_collector` instead

arqunis (approved):

src/gateway/bridge/shard_runner.rs
```diff
@@ -165,7 +165,7 @@ impl ShardRunner {
 
             if let Some(event) = event {
                 #[cfg(feature = "collector")]
-                self.collectors.retain_mut(|callback| (callback.0)(&event));
+                self.collectors.lock().expect("poison").retain_mut(|callback| (callback.0)(&event));
```
mkrasnitski, line 168: How come you chose to use `std::sync::Mutex` instead of `tokio::sync::Mutex` here? Since this function is `async` doing a synchronous block isn't a great idea. But, this isn't really likely to block, because users aren't going to be registering collectors very often. And, it probably doesn't make sense to make `Shard::add_collector` async.

kangalio: I initially used tokio::sync::Mutex, actually, until I realized it would make the `collector` function (and Shard::add_collector too) async. And I thought about it and none of the code that interacts with collectors ever block - so a non-async Mutex makes more sense here

PR #2389, opened 2023-04-17
Note suppress_embeds race condition in docs

### LINKED ISSUE #2333: Race condition when `SUPPRESS_EMBEDS` is set when editing a message

kangalio:
Fixes #2333


kangalio:
No breaking changes. This is a docs-only PR with little risk to merge

arqunis (approved):

PR #2399, opened 2023-04-23
Overhaul command permission builders

kangalio:
- :abc: Renames `CreateCommandPermissionsData` builder to `EditCommandPermissions`
  - Name is more fitting because the builder calls the `edit_guild_command_permissions` endpoint
- :abc: Renames `CreateCommandPermissionData` builder to `CreateCommandPermission`
- :abc: Renames `CommandPermission` model type to `CommandPermissions`
  - Name is more fitting because it actually stores a Vec of all permissions for a single command
- :abc: Renames `CommandPermissionData` to `CommandPermission`
- :abc: Renames `[GuildId, Guild, PartialGuild]::create_command_permission` to `edit_command_permission`
  - Less misleading - this method overwrites all permissions of the given command
  - Matches the HTTP endpoint name
- :sparkles: Replaces `EditCommandPermissions` methods `new()`, `add_permission(CreateCommandPermission)`, `set_permissions(Vec<CreateCommandPermission>)` with just `new(Vec<CreateCommandPermission>)`
  - Not providing any `CreateCommandPermission` instances makes no sense, and `new(vec![a, b, c])` is easier for both serenity and the user than `new().add_permission(a).add_permission(b).add_permission(c)`
- :sparkles: Replaces `CreateCommandPermissionData` methods `new()`, `kind(CommandPermissionType)`, `id(CommandPermissionId)`, `permission(bool)` with `role(RoleId, bool)`, `user(UserId, bool)`, `channel(ChannelId, bool)`
  - Much shorter to invoke
  - Cannot be misused (previously your configured `kind` could mismatch the type of ID you passed in)
  - Shows everything you can do with a command permission on one glance
- :sparkles: Adds `CreateCommandPermission::everyone(GuildId, bool)` and `CreateCommandPermission::all_channels(GuildId, bool)`
  - These were previously missing
- :x: Removes `edit_guild_commands_permissions` endpoint
  - This endpoint has been disabled with Permissions V2 ([Discord docs](https://discord.com/developers/docs/interactions/application-commands#batch-edit-application-command-permissions))
  - Don't confuse with `edit_guild_command_permissions` (note the plural) - that endpoint is still active!
- :wrench: Replaces `CreateCommandPermission`'s inner private fields by making `CreateCommandPermission` a newtype around `CommandPermission`
  - They share the same schema in the Discord docs so we can also share the same type struct definition

This PR has not been runtime tested

### DIFF
--- a/src/builder/create_command_permission.rs
+++ b/src/builder/create_command_permission.rs
@@ -4,46 +4,31 @@ use super::Builder;
 use crate::http::CacheHttp;
 #[cfg(feature = "http")]
 use crate::internal::prelude::*;
-#[cfg(feature = "http")]
-use crate::model::application::CommandPermission;
-use crate::model::application::CommandPermissionType;
-use crate::model::id::CommandPermissionId;
-#[cfg(feature = "http")]
-use crate::model::id::{CommandId, GuildId};
+use crate::model::prelude::*;
 
-/// A builder for creating several [`CommandPermissionData`].
-///
-/// [`CommandPermissionData`]: crate::model::application::CommandPermissionData
+/// A builder for creating several [`CommandPermission`].
+// Cannot be replaced by a simple Vec<CreateCommandPermission> because we need the schema with
+// the `permissions` field, and also to be forward compatible if a new field beyond just
+// `permissions` is added to the HTTP endpoint
 #[derive(Clone, Debug, Default, Serialize)]
 #[must_use]
-pub struct CreateCommandPermissionsData {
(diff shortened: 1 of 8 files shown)


kangalio:
Noticing the delay of merging: I made these changes because it makes more sense to do things this way imho - however, as always, I'm not sure if this gain for new code outweighs the churn for existing code. If others believe the breaking change of this PR is not worth it, I'm ok with not having this PR merged.

arqunis (approved):

PR #2406, opened 2023-04-27
Update checkout and cache action

jontze:
This PR updates the checkout action to v3 As github started to deprecate nodev12 actions this change was done to update the checkout action to v3. There are no other breaking changes with the version bump beside of the drop of the support for node v12.

Further, the rust cache action was updated to v2. With v2 the action added support for multiple workspaces and different`target` directories. But as this repo doesn't used any inputs of the action, we're not affected by the breaking changes and can update the achtion to v2 without any adjustments.

https://github.com/Swatinem/rust-cache/blob/master/CHANGELOG.md#200

EDIT: I recognized that the `Pin time to 0.3.9` step of the  `MSRV` job is failing. But this seems to happen as well on `current` and not related to this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
 
     steps:
       - name: Checkout sources
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Install toolchain
         run: |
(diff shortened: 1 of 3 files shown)


arqunis (approved):
Thanks!

PR #2415, opened 2023-05-04
Make discriminator optional and non-zero in preparation for new usernames

Xaeroxe:
This PR requires a major version bump!

Per https://discord.com/blog/usernames the discriminator field is about to become optional. This will proceed in two phases. It's not clear when phase 1 will start. It may have started already.

Phase 0 (Prior behavior): All users have discriminators. 

Phase 1: Users with discriminators co-exist among users without discriminators. Users without discriminators have `"discriminator": 0` in their JSON during this phase.

Phase 2: Once all users are migrated, discriminators are eliminated entirely and won't be returned in the JSON.

This PR is designed such that it can continue working correctly regardless of which phase we are currently in.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -365,8 +365,10 @@ impl Message {
             let mut at_distinct = String::with_capacity(38);
             at_distinct.push('@');
             at_distinct.push_str(&u.name);
-            at_distinct.push('#');
-            write!(at_distinct, "{:04}", u.discriminator).unwrap();
+            if let Some(discriminator) = u.discriminator {
+                at_distinct.push('#');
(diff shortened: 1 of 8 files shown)


tazz4843:
Breaking changes should target `next`, not current.

Xaeroxe:
Alright, let me get these conflicts sorted

Xaeroxe:
This is ready for review now. I was winging it and I didn't do much in the way of consulting people about API design here. I recognize this is not good practice and I apologize.

Xaeroxe:
Formatting wasn't correct because I didn't invoke it properly before. It's fixed now.

arqunis (approved):

arqunis:
Thanks!

PR #2393, opened 2023-04-21
Sync model structs/enums

kangalio:
I made a Python script that parses serenity's model structs (via rustdoc JSON), extracts the Discord docs link contained in most of them, parses the Discord docs, and prints out mismatches. Going through it is a long process. It's 8 hours since I started writing the script and like 5 hours since I started working through its output (37 commits now!). It's 4:30 AM now and imma take a break lul.

The commits are atomic and should be easy to review.

Closes #2056 

Todo:
- Some of the Discord docs link are outdated (the script has link rewriting and retry logic built in to workaround this, but the links should be updated within serenity itself
- Test this PR

<details>
<summary>Python script used (I'm building up the large list of false positives as I go)</summary>

```py
from typing import *
import json, re, bs4, requests, marko, marko.ext.gfm as gfm, joblib, glob

memory = joblib.Memory("cachedir", verbose=0)
@memory.cache
def http_get(url: str) -> str:
    return requests.get(url).text

def convert_discord_to_github_link(url: str) -> str:
    query = "developers/docs/"
    path = url[url.find(query) + len(query):]
    filename = path[path.rfind("/")+1:]
    new_filename = "_".join(word[0].upper() + word[1:] for word in filename.split("-"))
    path = path.replace(filename, new_filename)
    return f"https://raw.githubusercontent.com/discord/discord-api-docs/main/docs/{path}.md"

def find_markdown_after(url: str, words: [str]) -> Optional[str]:
    markdown = http_get(convert_discord_to_github_link(url))
    for i in range(len(words) - 1):
        query = "# " + " ".join(words[i:]) + "\n"
        pos = markdown.find(query)
        if pos != -1:
            return markdown[markdown.find(query) + len(query):]
    return None

def red(text: str) -> str:
    ANSI_RED_BOLD = "\033[31;1m"
    ANSI_RESET = "\033[0m"
    return ANSI_RED_BOLD + text + ANSI_RESET

renames = {}
for path in glob.iglob("../src/model/**/*.rs", recursive = True):
    regex = """#\[serde\(rename = "([a-z_]+)"\)\]\n    pub ([a-z_]+):"""
    renames_in_this_file = {match.group(2): match.group(1) for match in re.finditer(regex, open(path).read())}
    renames[path[path.find("src/model"):]] = renames_in_this_file
def rename_field(file_path: str, field: str) -> str:
    if file_path in renames:
        if field in renames[file_path]:
            return renames[file_path][field]
    return field

[...]

PR #2418, opened 2023-05-06
Change `CurrentUser` to a newtype around `User`.

mkrasnitski:
Changes `CurrentUser` from a type-alias to `User`, to a newtype around `User`, and implements `Deref` and `DerefMut` so that all methods on `User` are still accessible. This change is done to remove `User::edit`, which was accidentally exposed in #2393.  Calling the method would edit the current user's profile, regardless of the id of the `User` it was called on, which was confusing. Now, the `edit` function can only be called on `CurrentUser`, which upholds the requirement that a user can only edit their own profile.

### DIFF
--- a/src/model/channel/reaction.rs
+++ b/src/model/channel/reaction.rs
@@ -186,11 +186,11 @@ impl Reaction {
             #[cfg(feature = "cache")]
             {
                 if let Some(cache) = cache_http.cache() {
-                    return Ok(cache.current_user().clone());
+                    return Ok(cache.current_user().clone().into());
                 }
             }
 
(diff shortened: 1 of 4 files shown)


kangalio:
Cache.get_current_user() should be changed to return CurrentUser

mkrasnitski:
Are you referring to `Cache::current_user`? If so, it already does.

kangalio:
Oh lmao I saw this and assumed the function returned CurrentUser and the `.into()` calls were needed to go User -> CurrentUser
Screenshot_20230506_193034

kangalio (approved):

arqunis (approved):

PR #2419, opened 2023-05-06
Make model items accessible via `serenity::model::ITEM` instead of `serenity::model::MODULE::ITEM`

### LINKED ISSUE #2394: Flatten serenity::model
It would be better if all model types were accessible via just `serenity::model::ModelType`. I don't think anyone can remember which serenity::model submodule a model type belongs to.

This would also make the job of `serenity::model::prelude` redundant.

A few types should be renamed to be less ambiguous, like `Rule` and `Options`.

There should ideally be redirects from the old paths to the new paths.

This needs to wait until at least after #2393

kangalio:
Fixes #2394

This PR is backwards-compatible, except for renaming model::Error to ModelError

This PR consists of two commits:
1. moves the re-exports from model::prelude to model itself
2. changes all paths within serenity to the new style

The first commit is more important. The second commit serves to clean up long paths within serenity and express that using the new short path is recommended for new code.

The second commit is very huge, though, and manually reviewing all its changes is infeasible. If wanted, I can re-do that commit and record my screen. The review work would be reduced to checking the regexes I used. I would also be okay with cutting out the second commit entirely, if that is preferred. If we do keep the second commit, we should add it to .git-blame-ignore-revs


mkrasnitski:
Per https://github.com/serenity-rs/serenity/issues/2394#issuecomment-1517196509, I'd say this should go in after the release of 0.12. That would give us the opportunity to either deprecate the old paths in some way, or just remove them (by making all the submodules of `model` private).

I think one way you could test that the second commit caught every use is by doing the above and seeing if there are errors.

kangalio:
Why do we necessarily need to deprecate or remove the old paths?

mkrasnitski:
> Why do we necessarily need to deprecate or remove the old paths?

Because imo having types be importable both from `model::*` and from `model::submodule::*` doesn't sit right with me. It's best to have one canonical place to import types from. Deprecating at minimum is probably enough to signal to users that they should instead be using the new path.

However, while working on #2429, I realized that I don't think that `model` should be entirely flattened to be 1-deep, with every possible model type accessible at that level. Instead, it would be better to flatten it to 2 levels deep, such that all types are accessible at the `model::submodule::*` level. Then, the `model::prelude` could fill mostly the same purpose as a 1-level flattening.

From what I can tell, the only submodules that are not already flat are `model::colour` and `model::guild`.

PR #2422, opened 2023-05-08
do not feature gate `MessageUpdateEvent::apply_to_message`

fee1-dead:
This is used in `poise` and it should be useful even when `cache` is disabled.


kangalio (approved):

kangalio:
Context:

https://github.com/serenity-rs/poise/pull/166

Screenshot_20230508_155001

https://discord.com/channels/381880193251409931/919310428344029265/1105116848590164048

PR #2429, opened 2023-05-11
Clean up `serenity::model::prelude`

mkrasnitski:
Cleans up the model prelude by removing some re-exports, namely:

- Don't re-export all the submodules of `model`
- Don't re-export 3rd party types that are only used internally
- Define the `Color` type alias in `model::colour`
- Clean up collisions around `automod::EventType` by not glob-exporting from `guild::automod::*`

Note that `guild::audit_log` and `guild::automod` still end up getting re-exported via `guild::*`, but I'm not sure how to address that without manually re-exporting each type from `guild`, which would be annoying.


mkrasnitski:
Weird CI failure, I suspect some weird import path in the old prelude was responsible

arqunis (approved):

PR #2431, opened 2023-05-14
Add parameter `with_localizations` to `get_global_application_commands`

proteida:


kangalio:
This is a breaking change. Ideally the PR should be redesigned to not need a breaking change, but if impossible, it should be retargeted at the `next` branch

kangalio:
Thank you!

mkrasnitski:
I actually prefer the breaking changes - parameterizing the endpoint using two separate methods seems kinda weird. However, if that's what we're sticking with, you should add wrapper methods to `Guild` and `PartialGuild` that wrap the new method on `GuildId`.

arqunis (approved):

arqunis:
[Merged manually to fix formatting](https://github.com/serenity-rs/serenity/commit/94ed67bfeb1821e3f212a9057c50c4e3b95916f2).

kangalio:
By the way @acdenisSK, you can also push commits onto PR branches. Would be a bit cleaner and possibly easier than manually merging

PR #2433, opened 2023-05-16
Don't claim that the default text channel's ID equals the guild ID

mkrasnitski:
Follow up to #2148, which missed this line.


mkrasnitski:
CI failures unrelated

arqunis (approved):

PR #2436, opened 2023-05-17
Wrap `RatelimitingKind` in `Option` and remove its `None` variant

mkrasnitski:
Purely internal change to remove the codesmell of having a `None` variant on an enum.

### DIFF
--- a/src/http/routing.rs
+++ b/src/http/routing.rs
@@ -20,8 +20,6 @@ enum RatelimitingKind {
     PathAndId(NonZeroU64),
     /// Requests with the same path should be ratelimited together.
     Path,
-    /// No ratelimiting should be performed.
-    None,
 }
 
 /// A macro for defining routes as well as the type of ratelimiting they perform. Takes as input a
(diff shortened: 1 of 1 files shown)


arqunis (approved):

kangalio:
FWIW I deliberately made RatelimitingKind::None its own enum variant. Option::None represents the absence of a value, which RatelimitingKind::None was not. Every route has _a_ ratelimiting policy, you can't have no ratelimiting policy. I merely used the word "none" for its English definition, not because it was semantically similar to Option::None in this case.

It could've been `RatelimitingKind::DontRatelimit` or `RatelimitingKind::Unlimited` but I chose the term "none" in particular because "RatelimitingKind::None" reads nicely in source code, almost as if it was plain English: "ratelimiting (kind)? none"

mkrasnitski:
I think the semantics of having `Option::None` signify that an endpoint doesn't ratelimit requests is totally fine. It doesn't always have to signify the complete absence of value. In other words, it signifies the "null" or trivial ratelimiting policy, which is to not ratelimit.

PR #2438, opened 2023-05-19
Make AutocompleteChoice constructable again

kangalio:
Noticed this because poise's serenity-next branch failed to build

### DIFF
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -278,8 +278,8 @@ impl CreateInteractionResponseMessage {
 // Same as CommandOptionChoice according to Discord, see
 // [Autocomplete docs](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-autocomplete).
 #[must_use]
-#[serde(transparent)]
 #[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(transparent)]
 pub struct AutocompleteChoice(CommandOptionChoice);
 impl AutocompleteChoice {
     pub fn new(name: impl Into<String>, value: impl Into<Value>) -> Self {
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -278,6 +278,26 @@ impl CreateInteractionResponseMessage {
 /// They're same according to Discord, see
 /// [Autocomplete docs](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-autocomplete).
 pub type AutocompleteChoice = CommandOptionChoice;
+impl AutocompleteChoice {
+    pub fn new(name: impl Into<String>, value: impl Into<serde_json::Value>) -> Self {
+        Self {
+            name: name.into(),
+            name_localizations: None,
+            value: value.into(),
+        }
+    }
+
+    pub fn add_localized_name(
+        mut self,
+        locale: impl Into<String>,
+        localized_name: impl Into<String>,
+    ) -> Self {
+        self.name_localizations
+            .get_or_insert_with(Default::default)
+            .insert(locale.into(), localized_name.into());
+        self
+    }
+}
 
 /// [Discord docs](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-autocomplete)
 #[derive(Clone, Debug, Default, Serialize)]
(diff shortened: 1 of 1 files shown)


mkrasnitski:
Since `AutocompleteChoice` is a type alias, this also adds those impls to `CommandOptionChoice`. Is this something we want?

kangalio:
In my opinion, yes.

Advantages of having CommandOptionChoice and AutocompleteChoice be the same type:
- matches Discord docs
  - in particularly, since those types are the same in Discord's official model, mirroring this logic may cause less friction with upstream API changes in the future
- We have to keep just one struct definition up-to-date instead of two duplicated definitions

Advantages of separating the two:
- The possibility of confused users due to constructor methods on CommandOptionChoice, a type that's typically only returned from the Discord API and not created yourself
  - Can probably be remedied by noting the purpose of the constructor methods in their docs

I feel like merging the types is better altogether.

mkrasnitski:
Could we not just make `AutocompleteChoice` a newtype? The fields therefore won't get out of sync, and the constructors will be only on the one. However, there's a tradeoff - right now all the fields of `CommandOptionChoice` are public, which means it and also `AutocompleteChoice` can be instantiated normally. But, making `AutocompleteChoice` a newtype will probably necessitate making the wrapped `CommandOptionChoice` private, therefore the only instantiation method will be via the new constructors.

src/builder/create_interaction_response.rs
```diff
@@ -278,6 +278,26 @@ impl CreateInteractionResponseMessage {
 /// They're same according to Discord, see
 /// [Autocomplete docs](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-autocomplete).
 pub type AutocompleteChoice = CommandOptionChoice;
+impl AutocompleteChoice {
+    pub fn new(name: impl Into<String>, value: impl Into<serde_json::Value>) -> Self {
```
mkrasnitski:
```suggestion
    pub fn new(name: impl Into<String>, value: impl Into<Value>) -> Self {
```

It's already imported via `internal::prelude::*`

src/builder/create_interaction_response.rs
```diff
@@ -278,6 +278,26 @@ impl CreateInteractionResponseMessage {
 /// They're same according to Discord, see
 /// [Autocomplete docs](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-autocomplete).
 pub type AutocompleteChoice = CommandOptionChoice;
```
mkrasnitski:
```suggestion
#[must_use]
pub type AutocompleteChoice = CommandOptionChoice;
```

We should probably make this `#[must_use]` since it gets sent off to the API.

mkrasnitski (changes requested):
With regards to the newtype discussion - would having `AutocompleteChoice` not having public fields be a problem? Since it's meant to be constructed and then sent to the API, I don't think it'll be too much of an issue, unless someone wanted to construct it and then mutate it in some way.

Also, you could probably now impl `CreateAutocompleteResponse::add_{int,string,number}_choice` in terms of `AutocompleteChoice::new`.

src/builder/create_interaction_response.rs
```diff
@@ -275,9 +275,32 @@ impl CreateInteractionResponseMessage {
     super::button_and_select_menu_convenience_methods!(self.components);
 }
 
-/// They're same according to Discord, see
-/// [Autocomplete docs](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-autocomplete).
-pub type AutocompleteChoice = CommandOptionChoice;
+// Same as CommandOptionChoice according to Discord, see
+// [Autocomplete docs](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-autocomplete).
+#[must_use]
+#[derive(Clone, Debug, Deserialize, Serialize)]
```
mkrasnitski, line 281:
```suggestion
#[must_use]
#[serde(transparent)]
#[derive(Clone, Debug, Deserialize, Serialize)]
```

To be safe.

kangalio: If it makes you feel better :p

mkrasnitski: I guess for json it doesn't make a difference, see: https://serde.rs/json.html. Feel free to revert this if you feel like it. I doubt Discord is changing their API's data format anytime soon :P

arqunis (approved):

PR #2449, opened 2023-05-28
Use files in `Interaction::{create,edit}_followup_message`

proteida:
`MessageComponentInteraction::create_followup_message`
`MessageComponentInteraction::edit_followup_message`

`ModalSubmitInteraction::create_followup_message`
`ModalSubmitInteraction::edit_followup_message`

### DIFF
--- a/src/model/application/interaction/message_component.rs
+++ b/src/model/application/interaction/message_component.rs
@@ -188,13 +188,29 @@ impl MessageComponentInteraction {
     {
         let mut interaction_response = CreateInteractionResponseFollowup::default();
         f(&mut interaction_response);
+        self._create_followup_message(http.as_ref(), interaction_response).await
+    }
 
+    async fn _create_followup_message<'a>(
+        &self,
+        http: &Http,
+        interaction_response: CreateInteractionResponseFollowup<'a>,
+    ) -> Result<Message> {
         let map = json::hashmap_to_json_map(interaction_response.0);
 
-        Message::check_content_length(&map)?;
-        Message::check_embed_length(&map)?;
+        Message::check_lengths(&map)?;
 
-        http.as_ref().create_followup_message(&self.token, &Value::from(map)).await
+        if interaction_response.1.is_empty() {
+            http.as_ref().create_followup_message(&self.token, &Value::from(map)).await
+        } else {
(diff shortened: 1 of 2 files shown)


mkrasnitski:
For context, this brings the behavior of these functions in line with `ApplicationCommandInteraction::{create,edit}_followup_message` (by not accidentally throwing away attachments). The behavior is already fixed on `next`.

src/model/application/interaction/message_component.rs
```diff
@@ -219,15 +235,32 @@ impl MessageComponentInteraction {
...
+        let message_id = message_id.into().into();
+        self._edit_followup_message(http.as_ref(), message_id, interaction_response).await
+    }
 
+    async fn _edit_followup_message<'a>(
+        &self,
+        http: &Http,
+        message_id: u64,
+        interaction_response: CreateInteractionResponseFollowup<'a>,
+    ) -> Result<Message> {
```
mkrasnitski: Since `ApplicationCommandInteraction::_edit_followup_message` doesn't exist, I think that should be duplicated here, regardless of any extra monomorphization costs that might incur by calling everything inside the main `edit_followup_message` function. Besides, having `message_id: u64` isn't ideal (on `next` we've shifted to passing IDs around everywhere instead).

src/model/application/interaction/modal.rs
```diff
@@ -222,15 +238,32 @@ impl ModalSubmitInteraction {
...
+        self._edit_followup_message(http.as_ref(), message_id.into().into(), interaction_response)
+            .await
+    }
 
+    async fn _edit_followup_message<'a>(
+        &self,
+        http: &Http,
+        message_id: u64,
+        interaction_response: CreateInteractionResponseFollowup<'a>,
+    ) -> Result<Message> {
```
mkrasnitski: Ditto.

mkrasnitski (changes requested):
I think the impl's of `edit_followup_message` should be made identical to `ApplicationCommandInteraction::edit_followup_message`, so that rebasing `next` on top of this work will be simpler. Other than that, looks good.

mkrasnitski (approved):

arqunis (approved):

Issue #2450, opened 2023-05-29
Select Menu does not include User Select, Role Select, Mentionable Select, or Mentionable Select


ghost:
The select menu builder does not include User Select, Role Select, Mentionable Select, or Channel Select. It only seems to have the String select.  

The full list from [Discord's API](https://discord.com/developers/docs/interactions/message-components#component-object-component-types) can be seen in the image below. 

![image](https://github.com/serenity-rs/serenity/assets/19809270/e29630f6-87e4-4f92-bc84-0b2ecbbc59fa)

I was creating a menu like the following 

![image](https://github.com/serenity-rs/serenity/assets/19809270/1bacdf49-0fab-4c94-bee1-ab508073c097)

When I noticed Discord has a menu for it in their API.

![image](https://github.com/serenity-rs/serenity/assets/19809270/75d5d562-17db-4609-bd9d-6d241ddec547)

ghost:
I created a quickly modified version in [this commit](https://github.com/world177/serenity/commit/5151f901b80506e70f07bba4eeaf43b3d0fce7d7). I think this is only half of what is required for it. I think it still needs the "[resolved object](https://discord.com/developers/docs/interactions/message-components#select-menu-object-select-menu-resolved-object)."

![image](https://github.com/serenity-rs/serenity/assets/19809270/189a6efe-d15b-4bf8-a8cd-74d5fd80e092)

![image](https://github.com/serenity-rs/serenity/assets/19809270/18c4dee1-1650-4416-af8a-54e48248f70a)

It has a lot of duplicated code, but Discord also seems to be treating the different select options as their own component types. So, I felt like this approach was probably more future proof to changes. I didn't really test it yet, but they do seem to all have a bot respond with the correct select menu.

GnomedDev:
As mentioned in the closed PR above, this was added in #2224.

PR #2453, opened 2023-05-30
User Select, Role Select, Mentionable Select, and Channel Select 

ghost:
This adds functionality for the other types of select menus. It also adds to the component example demonstrating their usage. 

I previously opened an [issue stating that there was not currently support for them](https://github.com/serenity-rs/serenity/issues/2450).

### DIFF
--- a/examples/e17_message_components/src/main.rs
+++ b/examples/e17_message_components/src/main.rs
@@ -3,7 +3,7 @@ use std::time::Duration;
 
 use dotenv::dotenv;
 use serenity::async_trait;
-use serenity::builder::CreateButton;
+use serenity::builder::{CreateButton, CreateChannelSelect, SelectChannelType};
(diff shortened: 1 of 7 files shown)


kangalio:
Note that support already exists in the `next` branch, i.e. the next major version, which is hopefully released soon

ghost:
I just [missed that it was added](https://github.com/serenity-rs/serenity/pull/2224).

The version I submitted was less concise, with lots of repeated code. It did not feel like clean code.  

I was worried about writing it the way that it is written in the ```next``` branch because some menus have different requirements already. Only ```String Select``` has options. Only ```Channel Select``` has channel. These result in more API results when using an IDE, that I felt like was confusing. I kind of expected Discord to continue this trend with future updates.

Anyway, I highly appreciate the fix, as I now I can just use the next version instead. Thanks

PR #2455, opened 2023-06-02
Bump MSRV to Rust 1.70

mkrasnitski:


mkrasnitski:
Hmm, clippy is failing, but I think it's actually a false positive, because the variables are used inside of `quote!`. Removing the clones causes compilation to fail. Is this something that should be reported? @acdenisSK @kangalio

kangalio:
Sure, doesn't hurt to report

mkrasnitski:
Already reported in https://github.com/rust-lang/rust-clippy/issues/10545

arqunis (approved):

PR #2457, opened 2023-06-02
Replace `once_cell::sync::OnceCell` with `std::sync::OnceLock`

mkrasnitski:
Newly stabilized in 1.70. Allows us to remove `once_cell` as a dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,6 @@ mime_guess = { version = "2.0", optional = true }
 dashmap = { version = "5.1.0", features = ["serde"], optional = true }
 parking_lot = { version = "0.12", optional = true }
 ed25519-dalek = { version = "1.0.1", optional = true }
-once_cell = { version = "1.16.0", optional = true }
 # Serenity workspace crates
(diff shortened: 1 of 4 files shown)


mkrasnitski:
`OnceLock::with_value` doesn't exist (neither does `std::cell::OnceCell::with_value`, even though the implementation would be identical to that found in the `once_cell` crate). Is there some other way we could achieve partial initialization of `ShardManagerOptions::framework`? Or, could it somehow be separated out such that `Framework::init` doesn't have circular dependency problems?

mkrasnitski:
`OnceLock::with_value` doesn't exist (neither does `std::cell::OnceCell::with_value`, even though the implementation would be identical to that found in the `once_cell` crate). Is there some other way we could achieve partial initialization of `ShardManagerOptions::framework`? Or, could it somehow be separated out such that `Framework::init` doesn't have circular dependency problems?

cc @kangalio

mkrasnitski:
Whoops, turns out both types impl `From<T>` and I somehow missed that.

arqunis (approved):

PR #2462, opened 2023-06-11
Add dedicated type for storing image hashes

GnomedDev:
This converts the hex representation that is used across the API into a much smaller (56 vs 17) type which improves type safety, saves space, and reduces heap allocations (only during a clone, but deserialization could be made work without an allocation if we pull in an ArrayString dependency).

### DIFF
--- a/src/model/guild/mod.rs
+++ b/src/model/guild/mod.rs
@@ -626,7 +626,7 @@ impl Guild {
     pub async fn create(
         http: impl AsRef<Http>,
         name: &str,
-        icon: ImageHash,
+        icon: Option<ImageHash>,
     ) -> Result<PartialGuild> {
         let map = json!({
             "icon": icon,
--- a/src/model/guild/mod.rs
+++ b/src/model/guild/mod.rs
@@ -96,18 +96,18 @@ pub struct Guild {
     /// The hash of the icon used by the guild.
     ///
     /// In the client, this appears on the guild list on the left-hand side.
-    pub icon: Option<String>,
+    pub icon: Option<ImageHash>,
     /// Icon hash, returned when in the template object
-    pub icon_hash: Option<String>,
+    pub icon_hash: Option<ImageHash>,
     /// An identifying hash of the guild's splash icon.
     ///
     /// If the `InviteSplash` feature is enabled, this can be used to generate a URL to a splash
     /// image.
-    pub splash: Option<String>,
+    pub splash: Option<ImageHash>,
     /// An identifying hash of the guild discovery's splash icon.
     ///
     /// **Note**: Only present for guilds with the `DISCOVERABLE` feature.
-    pub discovery_splash: Option<String>,
+    pub discovery_splash: Option<ImageHash>,
     // Omitted `owner` field because only Http::get_guilds uses it, which returns GuildInfo
     /// The Id of the [`User`] who owns the guild.
     pub owner_id: UserId,
@@ -626,7 +626,7 @@ impl Guild {
     pub async fn create(
         http: impl AsRef<Http>,
         name: &str,
-        icon: Option<&str>,
+        icon: ImageHash,
     ) -> Result<PartialGuild> {
         let map = json!({
             "icon": icon,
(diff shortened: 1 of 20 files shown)


src/model/guild/mod.rs
```diff
@@ -626,7 +626,7 @@ impl Guild {
     pub async fn create(
         http: impl AsRef<Http>,
         name: &str,
-        icon: Option<&str>,
+        icon: ImageHash,
```
arqunis: Why was this changed to a plain `ImageHash`? Guilds can be created without an icon, so this should stay as `Option`al.

```suggestion
        icon: Option<ImageHash>,
```

src/model/misc.rs
```diff
@@ -53,6 +51,108 @@ impl_from_str! {
     ChannelId, ChannelIdParseError, parse_channel;
 }
 
+/// An image hash returned from the Discord API.
+///
+/// Note: This is parsed into a compact form when constructed, then turned back
+/// into the cannonical hex representation used by the API when needed.
```
arqunis: Speaking of my previous comment, it is not immediately clear how this type should be constructed by the user if it is going to be passed down as an argument to functions like `Guild::create`. We should mention the `FromStr` implementation here with an example.

src/model/misc.rs
```diff
@@ -53,6 +51,108 @@ impl_from_str! {
...
+        for byte in self.hash {
+            write!(f, "{byte:02x}")?;
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum ImageHashParseError {
```
arqunis, line 140: If `FromStr` will be the go-to method for constructing an `ImageHash`, the error type should be documented as well.

src/model/misc.rs
```diff
@@ -53,6 +51,108 @@ impl_from_str! {
...
+}
+
+impl std::fmt::Display for ImageHashParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::InvalidLength(length) => {
+                write!(f, "Invalid length {length}, expected 32 or 34 characters")
+            },
+            Self::MissingAnimatedMark => f.write_str("Input is 34 characters long, but missing a_"),
+            Self::UnparsableBytes(bytes) => write!(f, "Could not parse to hex: {bytes}"),
```
arqunis: Types that implement `std::error::Error` should not be embedded within `Display`'d messages. Instead, they should be returned as [the `source` of the error](https://doc.rust-lang.org/nightly/std/error/trait.Error.html#method.source), so that error crates [like `color-eyre`](https://github.com/yaahc/color-eyre) can integrate with these types nicely. While we're on the topic, `ImageHashParseError` should implement `std::error::Error`.
```suggestion
            Self::UnparsableBytes(_) => f.write_str("Could not parse to hex"),
```

src/model/misc.rs
```diff
@@ -53,6 +51,108 @@ impl_from_str! {
...
+impl<'de> serde::Deserialize<'de> for ImageHash {
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        // TODO: Replace this with ArrayString<34>?
+        let helper = String::deserialize(deserializer)?;
+        Self::from_str(&helper).map_err(serde::de::Error::custom)
+    }
+}
+
+impl std::fmt::Display for ImageHash {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
arqunis, line 121: Additionally, if `Display` will the the go-to way for 'peeking' into an `ImageHash`, it should be documented how the output should look like.

src/model/misc.rs
```diff
@@ -53,6 +51,108 @@ impl_from_str! {
...
+        let (hex, is_animated) = if s.len() == 34 {
+            if &s.as_bytes()[0..2] != b"a_" {
+                return Err(ImageHashParseError::MissingAnimatedMark);
+            }
+
+            (&s[2..], true)
+        } else if s.len() == 32 {
+            (s, false)
+        } else {
+            return Err(Self::Err::InvalidLength(s.len()));
```
arqunis, line 179: Pedantic, but for consistency.
```suggestion
            return Err(ImageHashParseError::InvalidLength(s.len()));
```

arqunis (changes requested):

src/model/guild/mod.rs
GnomedDev: Ah, guess this was a mistake

arqunis (approved):

PR #2461, opened 2023-06-10
Add global username

oSumAtrIX:
## About

This PR is an attempt to initiate adding global usernames according to the recently merged PR https://github.com/discord/discord-api-docs/pull/6218.

## TODO

~~Before merging this pull request, the added TODOs should be solved.~~ Additionally, the changes should be checked for completeness and tested roughly because I am unsure if I have changed everything necessary to support global usernames and it may be possible that a `get_global_username` function is missing or similar for example.

PR #2468, opened 2023-06-13
Add global_name property and update Member::display_name to handle

GnomedDev:


GnomedDev:
Whoops, didn't see #2461

PR #2470, opened 2023-06-22
Support creating standalone public threads

fenhl:
Discord's API has two endpoints for creating threads: one for [creating a thread as a reply to a message](https://discord.com/developers/docs/resources/channel#start-thread-from-message) and one for [creating a thread that's not connected to any existing message](https://discord.com/developers/docs/resources/channel#start-thread-without-message). The former only supports creating public threads, but the latter supports both public and private threads. However, serenity did not allow creating public standalone threads. This PR fills that gap, and renames the methods for calling these endpoints to more accurately describe what they do.

When I brought up the missing functionality in the #serenity channel, @arqunis claimed that the “start thread without message” endpoint was only for private threads, pointing to this part of the documentation:

> \* `type` currently defaults to `PRIVATE_THREAD` in order to match the behavior when thread documentation was first published. In a future API version this will be changed to be a required field, with no default.

I believe that this is a misunderstanding of the docs. They say “defaults to `PRIVATE_THREAD`”, not “must be `PRIVATE_THREAD`”. I also tested calling the endpoint manually with `type` set to `PUBLIC_THREAD` and it works as expected.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -393,7 +393,7 @@ impl Http {
     }
 
     /// Creates a thread channel not attached to a message in the [`GuildChannel`] given its Id.
-    pub async fn create_standalone_thread(
+    pub async fn create_thread(
         &self,
         channel_id: ChannelId,
         map: &impl serde::Serialize,
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -368,9 +368,8 @@ impl Http {
         .await
     }
 
-    /// Creates a public thread channel in the [`GuildChannel`] given its Id, with a base message
-    /// Id.
-    pub async fn create_public_thread(
+    /// Creates a thread channel in the [`GuildChannel`] given its Id, with a base message Id.
+    pub async fn create_thread_from_message(
         &self,
         channel_id: ChannelId,
         message_id: MessageId,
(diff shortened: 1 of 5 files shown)


fenhl:
CI failure seems to be unrelated.

src/http/client.rs
```diff
@@ -393,8 +392,8 @@ impl Http {
         .await
     }
 
-    /// Creates a private thread channel in the [`GuildChannel`] given its Id.
-    pub async fn create_private_thread(
+    /// Creates a thread channel not attached to a message in the [`GuildChannel`] given its Id.
+    pub async fn create_standalone_thread(
```
mkrasnitski:
```suggestion
    pub async fn create_thread(
```

Given that `create_thread_from_message` is self-explanatory, no need to include `standalone` in the method name. Same goes for `{ChannelId,GuildChannel}::create_standalone_thread`.

mkrasnitski (changes requested):
Just a few minor things. I think mentioning that threads linked to messages will always be public is something we should do, so the docs change to the `create_thread_from_message` methods should be reverted. It might also be worth changing the docs on `CreateThread::kind` to mention how the field is ignored for message threads (as well as how it defaults to private for standalone threads).

arqunis (approved):

PR #2471, opened 2023-06-22
Fix rustfmt CI

mkrasnitski:
Newest nightly has some formatting differences.


mkrasnitski:
Seems command_attr doesn't like when things are decorated with `#[rustfmt::skip]`.

cc @arqunis

arqunis:
> Seems command_attr doesn't like when things are decorated with `#[rustfmt::skip]`.
> 
> cc @arqunis

[Opened a PR to fix that](https://github.com/serenity-rs/serenity/pull/2473).

arqunis (approved):

PR #2485, opened 2023-07-06
Implement Get Current User Guild Member endpoint

skreborn:
This change adds the missing [Get Current User Guild Member](https://discord.com/developers/docs/resources/user#get-current-user-guild-member) endpoint.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -3741,6 +3741,33 @@ impl Http {
         .await
     }
 
+    /// Returns a guild member object for the current user.
+    ///
+    /// This method only works for user tokens with the [`GuildsMembersRead`]
+    /// OAuth2 scope.
+    ///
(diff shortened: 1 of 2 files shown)


mkrasnitski:
Thanks for the PR! We're currently preparing for a new major release, having just released the final minor release for the v0.11 series, v0.11.6. That being said, do you think you could retarget your PR at the `next` branch instead?

Adding new endpoints is slightly different on `next`, because you only do so in a [single place](https://github.com/serenity-rs/serenity/blob/next/src/http/routing.rs#L84). Also, the `Request` struct has new `method` and `params` fields to store the `LightMethod` and any url query params, respectively.

skreborn:
I'll get right on that, sure.

skreborn:
@mkrasnitski I've opened #2487 - that one targets the `next` branch.

PR #2486, opened 2023-07-06
Revert #2400

mkrasnitski:
It was determined that #2400 broke shard restarts and caused them to not be initiated when a network error occurred, leading to crashes because of uncaught errors. As a hotfix, this reverts that PR.


arqunis:
[Manually merged to fix the merge conflict and get this over with](https://github.com/serenity-rs/serenity/commit/c6fe6f60a3267452ba4ac6e18876f4ef1192c62a).

PR #2493, opened 2023-07-13
Add `edit_original_message()` for `MessageComponentInteraction`

B-2U:
in the discord api [Interaction Callback Type](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-interaction-callback-type)
with value`7` we can update the message the components belong to as a interaction response, but currently under the [MessageComponentInteraction](https://docs.rs/serenity/latest/serenity/model/prelude/prelude/interaction/message_component/struct.MessageComponentInteraction.html#) there's only methods related to defer, reponse and follow-up

So i tried to add it myself, basically the same but specify the `.kind(InteractionResponseType::UpdateMessage)` 
it does work but I'm not sure should it using the same `CreateInteractionResponse` struct, 
although it's editing a message but for discord server it should be the same.

thanks for reading

### DIFF
--- a/src/model/application/interaction/message_component.rs
+++ b/src/model/application/interaction/message_component.rs
@@ -3,9 +3,7 @@ use serde::Serialize;
 
 #[cfg(feature = "http")]
 use crate::builder::{
-    CreateInteractionResponse,
-    CreateInteractionResponseFollowup,
-    EditInteractionResponse,
+    CreateInteractionResponse, CreateInteractionResponseFollowup, EditInteractionResponse,
 };
 #[cfg(feature = "http")]
 use crate::http::Http;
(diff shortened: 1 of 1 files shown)


B-2U:
I'm not sure if I need to deal with these CI errors?

arqunis (approved):
The CI failures must be due to the old version of the toolchain. I built it locally, which worked fine.

mkrasnitski:
Just now taking a close look at this as we're approaching release. I'm not sure I understand why this isn't covered by `create_interaction_response`. Couldn't you just call `.kind(InteractionResponseType::UpdateMessage)` within the closure that you pass in as argument?

B-2U:
> Just now taking a close look at this as we're approaching release. I'm not sure I understand why this isn't covered by `create_interaction_response`. Couldn't you just call `.kind(InteractionResponseType::UpdateMessage)` within the closure that you pass in as argument?

You are right, it's much better to do so, I was new to rust at the moment so that happened...

mkrasnitski:
That's ok. I submitted a PR which simplifies the implementation. However, it's likely that on the `next` branch the function will have to be removed, since in #2228, `CreateInteractionResponse` was turned into an enum with variants representing each response type. Therefore, since you can't mutate one enum variant into another, it would be impossible to take an arbitrary `CreateInteractionResponse` and make it of kind `UpdateMessage`.

PR #2502, opened 2023-08-08
Make owner field optional in CurrentApplicationInfo

beyazit:


mkrasnitski:
Hi, and welcome! Note that this is already fixed on the `next` [branch](https://github.com/serenity-rs/serenity/blob/next/src/model/application/mod.rs#L55), which is where breaking changes go. Since this is a change to the type of a public field, it can't go on `current`.

PR #2504, opened 2023-08-09
Implement FromStr for ids

onlycs:
Useful for databasing, i.e. in Postgres, where ID's must be stored as strings because `u64::MAX` is larger than the largest BigInt type.

```rs
// before
Id::from(some_string.parse::<u64>().unwrap());

// after
Id::from_str(&some_string).unwrap();
```

### DIFF
--- a/src/model/id.rs
+++ b/src/model/id.rs
@@ -73,6 +73,14 @@ macro_rules! id_u64 {
                     id.0 as i64
                 }
             }
+
+            impl std::str::FromStr for $name {
+                type Err = <u64 as std::str::FromStr>::Err;
+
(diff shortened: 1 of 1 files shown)


tazz4843:
Unrelated to this PR, but for Postgres, you can simply cast to a i64 and it'll be fine even once the 64th bit is used as long as you cast back as well.

scottbot95:
>  you can simply cast to a i64 and it'll be fine even once the 64th bit is used as long as you cast back as well.

And here I was foolishly using `std::mem::transmute` to "re-interpret" the u64 as an i64 cuz I just assumed the cast would fail or otherwise behave in an undesirable way....

arqunis (approved):

PR #2511, opened 2023-08-15
Fix being unable to create webhooks on forum channels

tazz4843:
Forum channels are not text based, but they can have webhooks, which can be used in posts of a forum channel. This appears to require special casing, with no obvious way around I can see.


tazz4843:
I did some quick testing on this and it seems to work fine in beta (webhooks function as normal when created on a forum through this method).

arqunis (approved):

PR #2514, opened 2023-08-16
SQLx encode/decode support

tazz4843:
Currently only supports IDs, might expand to guilds/users/etc as a shorthand for calling ID. Opened this PR looking for feedback mainly.


tazz4843:
Currently running into issues where the compiler seems like it just refuses to use the implementations, and continues suggesting casting to i64. Not sure what the heck is going on, any insight from someone else would be appreciated.

tazz4843:
Accidentally deleted the branch trying to clean up my repo, but yeah I couldn't figure this out. Doesn't seem possible from discussion in the serenity server either, unfortunately.

PR #2520, opened 2023-08-26
Added static_face as variation of face

bleuthoot-sven:
This copies the `face()` function in `user.rs`, but uses `static_avatar_url` instead of `avatar_url`.

I am not sure if the `#[must_use]` is needed, since I don't exactly know what it does. If not, I will remove it in a follow up commit.

### DIFF
--- a/src/model/user.rs
+++ b/src/model/user.rs
@@ -516,6 +516,16 @@ impl User {
         self.avatar_url().unwrap_or_else(|| self.default_avatar_url())
     }
 
+    /// Retrieves the URL to the static version of the user's avatar, falling back to the default
+    /// avatar if needed.
+    ///
+    /// This will call [`Self::static_avatar_url`] first, and if that returns [`None`], it then
+    /// falls back to [`Self::default_avatar_url`].
+    #[must_use]
+    pub fn static_face(&self) -> String {
+        self.static_avatar_url().unwrap_or_else(|| self.default_avatar_url())
+    }
+
     /// Check if a user has a [`Role`]. This will retrieve the [`Guild`] from the [`Cache`] if it
     /// is available, and then check if that guild has the given [`Role`].
     ///


bleuthoot-sven:
Something of note, the `model/guild/members.rs` does not have a `static_avatar_url` function, so I did not add the `static_face` function to there.

arqunis:
> I am not sure if the `#[must_use]` is needed, since I don't exactly know what it does. If not, I will remove it in a follow up commit.

`#[must_use]` is an attribute you can use on functions and structs. If on functions, it will prompt the compiler to issue a diagnostic if the caller of the function forgets to use the return value in any way, e. g.:
```rust
#[must_use]
fn rand() -> i32 {
    4
}

fn main() {
    rand(); // warning: unused return value of `rand` that must be used
}
```
On structs, the effect of the attribute will apply anywhere the type is used (that is, in any function where it is used as return type).

It is very useful if the whole point of calling a function is to perform further calculations on its return value, as the function does not have a side-effect (like writing to the filesystem); or if it does, there could be reasons why ignoring the return value might be bad (for instance, `Result`s, because they might be `Err`s, or `Iterator`s and `Future`s, which perform no work until `next()`/`poll()` is called).

bleuthoot-sven:
> > I am not sure if the `#[must_use]` is needed, since I don't exactly know what it does. If not, I will remove it in a follow up commit.
> 
> `#[must_use]` is an attribute you can use on functions and structs. If on functions, it will prompt the compiler to issue a diagnostic if the caller of the function forgets to use the return value in any way, e. g.:
> 
> ```rust
> #[must_use]
> fn rand() -> i32 {
>     4
> }
> 
> fn main() {
>     rand(); // warning: unused return value of `rand` that must be used
> }
> ```
> 
> On structs, the effect of the attribute will apply anywhere the type is used (that is, in any function where it is used as return type).
> 
> It is very useful if the whole point of calling a function is to perform further calculations on its return value, as the function does not have a side-effect (like writing to the filesystem); or if it does, there could be reasons why ignoring the return value might be bad (for instance, `Result`s, because they might be `Err`s, or `Iterator`s and `Future`s, which perform no work until `next()`/`poll()` is called).

Ah, that's quite useful to know. Thanks for the information.

arqunis (approved):

PR #2523, opened 2023-08-27
Change Arc<Mutex<ShardManager>> to Arc<ShardManager>

### LINKED ISSUE #2507: Shutdown is broken on `next` branch
Since the revert of #2400 in #2486, it is impossible to shutdown a serenity bot without killing the process. Calling `ShardManager.shutdown_all()` will hang forever and results in the following warning being logged:
```
WARN shutdown_all:shutdown{shard_id=ShardId(0) code=1000}: serenity::gateway::bridge::shard_manager: 287: Failed to cleanly shutdown shard 0, reached timeout: Elapsed(())
```

tazz4843:
Has the side effect of (theoretically)
fixing shutdown by removing the major deadlock

Required converting all `&mut self` methods on ShardManager to &self

Closes #2507

### DIFF
--- a/src/gateway/bridge/shard_manager.rs
+++ b/src/gateway/bridge/shard_manager.rs
@@ -330,6 +330,10 @@ impl ShardManager {
         }
 
         drop(self.shard_queuer.unbounded_send(ShardQueuerMessage::Shutdown));
+
+        // this message is received by Client::start_connection, which lets the main thread know
+        // and finally return from Client::start
+        drop(self.return_value_tx.lock().await.unbounded_send(Ok(())));
     }
 
     #[instrument(skip(self))]
--- a/src/gateway/bridge/shard_manager.rs
+++ b/src/gateway/bridge/shard_manager.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, VecDeque};
+use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::Arc;
 #[cfg(feature = "framework")]
 use std::sync::OnceLock;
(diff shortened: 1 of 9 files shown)


tazz4843:
Accidentally opened as ready for merge, converted to draft as I don't want a round 2 of last time we tried fixing shutdown. I'm going to fire up a local bot as a test and leave it running for a few days.

src/gateway/bridge/shard_manager.rs
```diff
@@ -96,20 +97,25 @@ use crate::model::gateway::GatewayIntents;
...
     /// The shard runners currently managed.
     ///
     /// **Note**: It is highly unrecommended to mutate this yourself unless you need to. Instead
     /// prefer to use methods on this struct that are provided where possible.
     pub runners: Arc<Mutex<HashMap<ShardId, ShardRunnerInfo>>>,
     /// The index of the first shard to initialize, 0-indexed.
-    shard_index: u32,
+    ///
+    /// Atomic integers are used here that way we don't need any mutable references to Self to
+    /// update any of these values.
```
mkrasnitski:
```suggestion
    // Atomics are used here to allow for mutation without requiring a mutable reference to self.
```
This should be made a regular comment IMO.

src/gateway/bridge/shard_manager.rs
```diff
@@ -96,20 +97,25 @@ use crate::model::gateway::GatewayIntents;
...
+    shard_index: AtomicU32,
     /// The number of shards to initialize.
-    shard_init: u32,
+    shard_init: AtomicU32,
     /// The total shards in use, 1-indexed.
-    shard_total: u32,
+    shard_total: AtomicU32,
     shard_queuer: Sender<ShardQueuerMessage>,
-    shard_shutdown: Receiver<ShardId>,
+    // we can safely use a Mutex for this field, as it is only ever used in one single place
```
mkrasnitski:
```suggestion
    // We can safely use a Mutex for this field, as it is only ever used in one single place
```
Nit

mkrasnitski (commented):
Looks good to me - let's hope this works. Your doctests are failing, though.

tazz4843:
Turns out this breaks poise as well, I'll make a PR to that as a companion.

src/gateway/bridge/shard_manager.rs
tazz4843: I thought I did that, might've accidentally copied and pasted. Thanks!

GnomedDev:
Is this a good idea? Have you tested running shutdown concurrently to see if that breaks anything? It worries me if this was written with the assumption that it's `&mut self` that a new class of bugs will be introduced (that cannot be caught by rustc, as we are mutating the state of the program via IO to discord).

mkrasnitski:
The only fields of `ShardManager` that end up being mutated are `shard_shutdown` as well as the integers `shard_index`, `shard_init`, and `shard_total`. The former is now behind a Mutex that only gets locked in one place, so no opportunity for deadlocks on the same thread.

The other three fields all get written to only in `.set_shards`. There *could* feasibly be a race condition where a context switch happens between two of the calls to `.store`, and then an invariant of the manager might get broken? I don't know. Maybe we should synchronize the three atomic writes to make sure they can't get interrupted.

tazz4843:
> Have you tested running shutdown concurrently to see if that breaks anything?

As far as I can tell, `shutdown` is a completely idempotent function, even with these new changes. The worst that could happen is an extra irrelevant "timed out" warning after a successful shutdown, which would have happened previously anyways. `shutdown` simply fires a message and then waits for a response, and if it doesn't get it, times out with an error.

> The other three fields all get written to only in `.set_shards`. There _could_ feasibly be a race condition where a context switch happens between two of the calls to `.store`, and then an invariant of the manager might get broken? I don't know. Maybe we should synchronize the three atomic writes to make sure they can't get interrupted.

`.set_shards` is only ever called in one place, `Client::start_connection`, which takes `&mut self` on the `Client`, preventing concurrent calls. Regardless, we could synchronize the calls, however I'm not sure how we could go about that.

tazz4843:
As with the tokio-tungstenite upgrade, failing test seems to be unrelated: https://github.com/dtolnay/proc-macro2/issues/398

tazz4843:
Test bot has been online for three days straight now without any heartbeat failures (through network loss etc etc). I think this has finally solved the issue.

arqunis (approved):

PR #2525, opened 2023-08-28
Add methods for deleting message reactions

Collin-Swish:


mkrasnitski:
Per https://github.com/serenity-rs/serenity/pull/2526#issuecomment-1701584462, methods for doing this already exist.

PR #2526, opened 2023-08-28
Add methods for deleting message reactions

Collin-Swish:


kangalio:
Looks good to me (apart from the failing CI tests). Has someone tested the two new methods?

Collin-Swish:
> Looks good to me (apart from the failing CI tests). Has someone tested the two new methods?

I've had a bot which makes frequent use of these methods running for two days now without incident.

mkrasnitski:
Seems like methods/endpoints for this already exist. See `Http::delete_reaction`, as well as `{ChannelId,GuildChannel,PrivateChannel}::delete_reaction`, and also `Message::delete_reaction_emoji`.

Collin-Swish:
> Seems like methods/endpoints for this already exist. See `Http::delete_reaction`, as well as `{ChannelId,GuildChannel,PrivateChannel}::delete_reaction`, and also `Message::delete_reaction_emoji`.

Damn I must not have seen that. I would argue that delete_own_reaction still makes sense as a separate method since it requires different permissions. ADD_REACTIONS for for deleting the bots own reaction vs MANAGE_MESSAGES for deleting another users reaction. 
But I do agree that delete_user_reaction is redundant.

mkrasnitski:
Actually, the discord docs don't specify any required permissions for _deleting_ your own reaction. Also, permissions checks in the cache don't bring a whole lot of benefit in general, even in terms of number of network calls. Most endpoints don't need them.

mkrasnitski:
IMO this PR as well as #2525 can be closed.

PR #2530, opened 2023-09-05
Bump dependency versions

mkrasnitski:
* simd-json 0.7 -> 0.10
* tokio-tungstenite 0.19 -> 0.20
* ed25519-dalek 1.0 -> 2.0
* criterion 0.4 -> 0.5
* notify 5.0.0-pre.14 -> 6.1

In tungstenite 0.20, the `WebSocketConfig::max_send_queue` is deprecated for `max_write_buffer_length`. The default value corresponds to an unlimited length, same as before. However, there's a new `write_buffer_length` field which controls how much buffering is done before a flush occurs and the buffer gets written to the underlying system. It defaults to 128KiB (so 128 * 1024) - is this a fine value for us to use?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,14 +32,14 @@ secrecy = { version = "0.8.0", features = ["serde"] }
 arrayvec = { version = "0.7.3", features = ["serde"] }
 # Optional dependencies
 fxhash = { version = "0.2.1", optional = true }
-simd-json = { version = "0.7", optional = true }
+simd-json = { version = "0.10", optional = true }
 uwl = { version = "0.6.0", optional = true }
 levenshtein = { version = "1.0.5", optional = true }
 chrono = { version = "0.4.22", default-features = false, features = ["clock", "serde"], optional = true }
 flate2 = { version = "1.0.13", optional = true }
 reqwest = { version = "0.11.7", default-features = false, features = ["multipart", "stream"], optional = true }
 static_assertions = { version = "1.1", optional = true }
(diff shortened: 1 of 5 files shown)


mkrasnitski:
Is there anything in this PR that would be considering a breaking change?

arqunis:
> Is there anything in this PR that would be considering a breaking change?

In the code itself? Nothing. It was decided a while ago that any version changes to the dependencies constitutes a breaking change, so I'm only honouring that decision. Though I suppose we could reconsider whether that has any real merit and only apply the "breaking change" label if a dependency update actually alters the public interface.

mkrasnitski:
> It was decided a while ago that any version changes to the dependencies constitutes a breaking change

This is unnecessarily conservative IMO. As long as a dependency bump doesn't impact the public interface or MSRV, then I don't think it should be considered breaking.

arqunis (approved):
> It defaults to 128KiB (so 128 * 1024) - is this a fine value for us to use?

As long as it keeps the connection and performance steady, I don't see why not.

arqunis:
re CI failure on the examples: It appears locking the version of `notify` to `5.0.0-pre.14` is still necessary, unless we're content with just removing example 15 altogether? [`rillrate`'s repository has been archived for almost a year now](https://github.com/rillrate-fossil/rillrate).

mkrasnitski:
Ah, I guess probably best to leave it until we can reimplement e15 using [`prometheus`](https://github.com/tikv/rust-prometheus).

PR #2531, opened 2023-09-05
Bump bitflags to 2.x

mkrasnitski:
In bitflags 2.x, bitflag structs changed to tuple structs and can't be instantiated the same way as before. Unfortunately, performing bitops on them isn't possible in `const` contexts, so a workaround is to perform them on the raw `.bits()` and pass the result to `from_bits_truncate`.

Also, bitflag structs no longer automatically derive the following traits: `Clone`, `Copy`, `Debug`, `Default`, `Hash`, `PartialEq`, `Eq`, `PartialOrd`, and `Ord`. To avoid breaking backcompat, I added them back, but serenity does not rely on all of them. Nothing relies on `PartialOrd`/`Ord`, and only the `Permissions` and `SpeakingState` types rely on `Hash`/`PartialEq`/`Eq`. The common required traits are just `Copy`/`Clone` and `Debug`. It's worth considering if we should remove extra impls for types which we expect won't ever make use of them in user code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ members = ["examples/*"]
 
 [dependencies]
 # Required dependencies
-bitflags = "1.3"
+bitflags = "2.4"
 serde_json = "1.0.75"
 async-trait = "0.1.54"
 tracing = { version = "0.1.23", features = ["log"] }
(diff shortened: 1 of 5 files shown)


arqunis:
> Nothing relies on `PartialOrd`/`Ord`

Then to shave off compilation times a bit, they should be removed. I doubt users are sorting `Permissions`, or utilising a data structure like `BTreeMap` on these types.

arqunis (approved):

Issue #2532, opened 2023-09-06
delete_reaction not implemented on message struct


Collin-Swish:
I noticed that the channel_id struct implements methods which can delete individual reactions [channel_id::delete_reaction] while the message struct can only perform bulk deletes of reactions [message::delete_reactions]

I think it makes sense to have both methods implemented on both structs since the endpoints for both methods already exist.

PR #2533, opened 2023-09-06
Added additional reaction controls to channel_id and message structs

Collin-Swish:
#2532 In reference to this issue.

### DIFF
--- a/src/model/channel/channel_id.rs
+++ b/src/model/channel/channel_id.rs
@@ -280,8 +280,14 @@ impl ChannelId {
             .await
     }
     /// Deletes all of the [`Reaction`]s associated with the provided message id.
-    /// 
+    ///
     /// **Note**: Requires the [Manage Messages] permission.
+    ///
+    /// [Manage Messages]: Permissions::MANAGE_MESSAGES
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::Http`] if the current user lacks permission
     ///
     /// [Manage Messages]: Permissions::MANAGE_MESSAGES
     #[inline]
@@ -290,10 +296,7 @@ impl ChannelId {
         http: impl AsRef<Http>,
         message_id: impl Into<MessageId>,
     ) -> Result<()> {
-        http.as_ref().delete_message_reactions(
-            self.0,
-            message_id.into().0,
-        ).await
+        http.as_ref().delete_message_reactions(self.0, message_id.into().0).await
     }
 
     /// Deletes all [`Reaction`]s of the given emoji to a message within the channel.
--- a/src/model/channel/channel_id.rs
+++ b/src/model/channel/channel_id.rs
@@ -279,6 +279,22 @@ impl ChannelId {
             )
             .await
     }
+    /// Deletes all of the [`Reaction`]s associated with the provided message id.
+    /// 
+    /// **Note**: Requires the [Manage Messages] permission.
+    ///
+    /// [Manage Messages]: Permissions::MANAGE_MESSAGES
+    #[inline]
+    pub async fn delete_reactions(
+        &self,
+        http: impl AsRef<Http>,
+        message_id: impl Into<MessageId>,
+    ) -> Result<()> {
+        http.as_ref().delete_message_reactions(
+            self.0,
+            message_id.into().0,
+        ).await
+    }
 
     /// Deletes all [`Reaction`]s of the given emoji to a message within the channel.
     ///
--- a/src/model/channel/guild_channel.rs
+++ b/src/model/channel/guild_channel.rs
@@ -439,6 +439,26 @@ impl GuildChannel {
         self.id.delete_reaction(&http, message_id, user_id, reaction_type).await
     }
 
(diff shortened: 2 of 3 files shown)


src/model/channel/channel_id.rs
```diff
@@ -279,6 +279,22 @@ impl ChannelId {
             )
             .await
     }
+    /// Deletes all of the [`Reaction`]s associated with the provided message id.
+    /// 
+    /// **Note**: Requires the [Manage Messages] permission.
+    ///
+    /// [Manage Messages]: Permissions::MANAGE_MESSAGES
```
mkrasnitski: Should include an `# Errors` section.

src/model/channel/message.rs
```diff
@@ -264,6 +264,36 @@ impl Message {
...
+    /// # Errors
+    /// 
+    /// Returns [`Error::Http`] if the currend user did not perform the reaction,
+    /// and lacks permission
+    /// 
+    /// [Manage Messages]: Permissions::MANAGE_MESSAGES
+    #[inline]
+    pub async fn delete_reaction(
+        self,
+        cache_http: impl CacheHttp,
```
mkrasnitski: If a permissions check isn't done on the `CacheHttp`, better to change this to `http: impl AsRef<Http>`.

src/model/channel/message.rs
```diff
@@ -264,6 +264,36 @@ impl Message {
...
+        cache_http
+            .http()
+            .as_ref()
+            .delete_reaction(
+                self.channel_id.0,
+                self.id.0,
+                user_id.map(|uid| uid.0),
+                &reaction_type.into(),
+            )
+            .await
```
mkrasnitski: This should be reworked to just a wrapper around `ChannelId::delete_reaction`. The same goes for `Message::delete_reactions`, now that `ChannelId::delete_reactions` exists.

src/model/channel/message.rs
```diff
@@ -264,6 +264,36 @@ impl Message {
         cache_http.http().as_ref().delete_message_reactions(self.channel_id.0, self.id.0).await
     }
 
+    /// Remove reaction from the message given an [`Emoji`] or unicode character
+    /// 
+    /// **Note**: Requires the [Manage Messages] permission, _if_ the current
+    /// user did not perform the reaction
```
```suggestion
    /// **Note**: Requires the [Manage Messages] permission, _if_ the current user did not perform
    /// the reaction.
```
formatting

src/model/channel/message.rs
```diff
@@ -264,6 +264,36 @@ impl Message {
...
+    /// 
+    /// # Errors
+    /// 
+    /// Returns [`Error::Http`] if the currend user did not perform the reaction,
+    /// and lacks permission
```
```suggestion
    /// Returns [`Error::Http`] if the current user did not perform the reaction, or lacks
    /// permission.
```
formatting

mkrasnitski (changes requested):
Could you add a `GuildChannel::delete_reactions` wrapper method as well? (A `PrivateChannel::delete_reactions` method doesn't make sense, because only reactions by the current user can be removed in a private channel.)

Also, make sure to run `cargo +nightly fmt`.

src/model/channel/message.rs
Collin-Swish: Good point. The change has been made.

src/model/channel/guild_channel.rs
```diff
@@ -439,6 +439,26 @@ impl GuildChannel {
...
+    /// Returns [`Error::Http`] if the current user lacks permission
+    ///
+    /// [Manage Messages]: Permissions::MANAGE_MESSAGES
+    #[inline]
+    pub async fn delete_reactions(
+        &self,
+        http: impl AsRef<Http>,
+        message_id: impl Into<MessageId>,
+    ) -> Result<()> {
+        http.as_ref().delete_message_reactions(self.id.0, message_id.into().0).await
```
mkrasnitski:
```suggestion
        self.id.delete_reactions(http, message_id).await
```

src/model/channel/message.rs
```diff
@@ -264,6 +264,32 @@ impl Message {
         cache_http.http().as_ref().delete_message_reactions(self.channel_id.0, self.id.0).await
     }
 
+    /// Remove reaction from the message given an [`Emoji`] or unicode character
```
```suggestion
    /// Deletes the given [`Reaction`] from the message.
```
Adjusted wording to mirror docs for `ChannelId::delete_reaction`.

mkrasnitski (changes requested):
Looks pretty good. Could you also change the implementation of `Message::delete_reactions` to be:
```rust
self.channel_id.delete_reactions(cache_http.http(), self.id).await
```

mkrasnitski (approved):

arqunis (approved):

mkrasnitski:
This slipped past in review, but `Message::delete_reaction` should take `&self` instead of `self` as it doesn't make sense to take ownership. Conversely, `ChannelId::delete_reactions` should in fact take `self` instead of `&self`, because `ChannelId` is `Copy`.

Issue #2474, opened 2023-06-23
Feature Flag "simd_json" causes failure to compile


Moocow9m:
Serenity Version: Latest Release(11.5) and Latest Git Commit(9fc3f118)
Log: `https://pastebin.com/fzKUYSKh`

Temporary fix: Cloned to local path and updated simd_json to 0.9. Then set `serenity\src\json.rs:77` to `Ok(unsafe { simd_json::from_str(s)? })`. This works for my use case, but I would rather not maintain this. I plan to make it not use unsafe later.

mkrasnitski:
This is unfortunately not possible to fix (in a satisfactory way) without introducing breaking changes. Note that this is in fact fixed on the `next` branch, so once 0.12 releases it should not be an issue. See #2353 for details, but in essence, `simd_json::from_str` can potentially break an invariant of its string argument by mutating it such that it no longer contains valid UTF-8. Therefore, passing in an `&mut str`, and then using the string later on, can lead to UB. This is the case even in versions of simd_json prior to 0.7, when the function was marked unsafe (prior to that it was basically just unsound).

Because `serenity::json::from_str` is public, patching this soundness hole requires changing the function to take ownership of its argument, to prevent the user from causing UB by using the string after it's been parsed. This is what #2353 did. I don't know what we should do on `current`. As it stands, simd_json support is entirely broken, but the soundness hole will still be exposed if we upgrade to a newer version without also changing the signature of `serenity::json::from_str`.

Moocow9m:
I see. I'll just run a modified local copy for now then.

kangalio:
Attempted fix in #2536

kangalio:
Fixed in https://github.com/serenity-rs/serenity/pull/2610

PR #2536, opened 2023-09-12
Update simd-json to make serenity compile again

kangalio:
And fix an unsoundness (lol)

I found out that serenity doesn't compile with the simd_json feature in poise https://github.com/serenity-rs/poise/actions/runs/6158123854/job/16710276804

### DIFF
--- a/src/json.rs
+++ b/src/json.rs
@@ -1,5 +1,5 @@
 //! This module exports different types for JSON interactions.
-//! It encapsulates the differences between serde_json and simd-json to allow
+//! It encapsulates the differences between serde_json and simd_json to allow
 //! ignoring those in the rest of the codebase.
 
 use std::collections::HashMap;
@@ -10,28 +10,28 @@ use serde::ser::Serialize;
 
 use crate::Result;
 
-#[cfg(not(feature = "simd-json"))]
+#[cfg(not(feature = "simd_json"))]
 pub type Value = serde_json::Value;
-#[cfg(feature = "simd-json")]
+#[cfg(feature = "simd_json")]
 pub type Value = simd_json::OwnedValue;
 
-#[cfg(not(feature = "simd-json"))]
+#[cfg(not(feature = "simd_json"))]
 pub use serde_json::json;
-#[cfg(not(feature = "simd-json"))]
+#[cfg(not(feature = "simd_json"))]
 pub use serde_json::Error as JsonError;
-#[cfg(feature = "simd-json")]
+#[cfg(feature = "simd_json")]
 pub use simd_json::json;
-#[cfg(feature = "simd-json")]
+#[cfg(feature = "simd_json")]
 pub use simd_json::Error as JsonError;
 
-#[cfg(not(feature = "simd-json"))]
+#[cfg(not(feature = "simd_json"))]
 pub type JsonMap = serde_json::Map<String, Value>;
-#[cfg(feature = "simd-json")]
+#[cfg(feature = "simd_json")]
 pub type JsonMap = simd_json::owned::Object;
 
-#[cfg(not(feature = "simd-json"))]
+#[cfg(not(feature = "simd_json"))]
 pub const NULL: Value = Value::Null;
-#[cfg(feature = "simd-json")]
+#[cfg(feature = "simd_json")]
 pub const NULL: Value = Value::Static(simd_json::StaticNode::Null);
 
 /// Converts a HashMap into a final [`JsonMap`] representation.
@@ -43,7 +43,7 @@ where
     map.into_iter().map(|(k, v)| (k.to_string(), v)).collect()
 }
 
-#[cfg(not(feature = "simd-json"))]
+#[cfg(not(feature = "simd_json"))]
 pub(crate) fn to_string<T>(v: &T) -> Result<String>
 where
     T: Serialize,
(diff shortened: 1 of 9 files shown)


kangalio:
After reading https://github.com/serenity-rs/serenity/issues/2474#issuecomment-1637242395, I realize serenity actually re-exports the raw serialization and deserialization functions from the underlying JSON library in serenity::json::prelude. As such, any upgrade to serde-json or simd-json is a breaking change >:(

We could hack around this unfortunate decision by replacing the re-exported simd_json::from_str with a custom impl that keeps the old signature (at the cost of a soundness hole, that probably no one ever will encounter because this is such a niche serenity API). We should also maybe deprecate the re-exports from the underlying JSON libraries?

mkrasnitski:
Is changing the trait bound to `DeserializeOwned` a breaking change? I'm not sure - since we're dealing with JSON, are there any types that don't implement DeserializeOwned?

kangalio:
serenity::json::from_str isn't public API! Only serenity::json::prelude::from_str is. The two differ in that the public one is directly re-exported from serde_json or simd_json, whereas the private one is a wrapper function around serde_json or simd_json

---

I have another hack idea to fix the semver issue: depend on both simd_json 0.4 and 0.10 - use 0.10 internally, re-export 0.4. It's silly but it's a silly solution to a silly problem (namely, serenity publicly re-exporting implementation details)

src/json.rs
```diff
@@ -145,7 +146,7 @@ pub mod prelude {
         to_vec_pretty,
     };
     #[cfg(feature = "simd-json")]
```
arqunis: I believe the feature flag name here should be the one with the underscore:
```suggestion
    #[cfg(feature = "simd_json")]
```

Otherwise, only `simd-json` gets pulled in.

arqunis (changes requested):

src/json.rs
kangalio: Ah good catch. That's actually fine though (it was in the code before this PR, even) - the `simd-json` feature flag is an automatic feature from Cargo that cannot be turned off that's always set if simd-json was pulled in. The `simd_json` is our own feature flag that aliases to `simd-json` for naming consistency. So gating by `simd-json` or `simd_json` does the same..

It is confusing though, I give you that.

Should I change all places in the codebase that gate by `simd-json` to say `simd_json`?

arqunis: > Should I change all places in the codebase that gate by `simd-json` to say `simd_json`?

We should be using our feature flag, i.e. `simd_json`. So yes, please. Not just in code, but also [in the test CI job](https://github.com/serenity-rs/serenity/blob/next/.github/workflows/ci.yml#L43), which is why it currently fails.

src/json.rs
```diff
@@ -62,19 +60,22 @@ where
...
+pub(crate) fn from_str<T>(s: &str) -> Result<T>
 where
-    T: Deserialize<'a>,
+    T: DeserializeOwned,
 {
-    Ok(simd_json::from_str(s)?)
+    // Have to clone the argument because simd_json mutates the argument
+    // And we can't just take &mut str either because simd_json mutates the string such that it may
+    // not be UTF-8 afterwards, which is unsound
+    Ok(simd_json::from_slice(&mut s.as_bytes().to_vec())?)
```
mkrasnitski, line 78: Does `from_slice` work with multi-byte characters? I think that passing in a `&mut s.to_string()` would still be fine here, since even if the string becomes invalid UTF-8, it gets dropped immediately afterward.

kangalio: > Does from_slice work with multi-byte characters?

Yes:
Screenshot_20230919_033925

> I think that passing in a &mut s.to_string() would still be fine here, since even if the string becomes invalid UTF-8, it gets dropped immediately afterward.

I originally thought the mere existence of a non UTF-8 str is instant UB but I looked it up and it seems you're right

Still, that would need an unsafe block. Using `from_slice` needs no unsafe block and has no performance penalty (I expect?

src/json.rs
kangalio: Done

mkrasnitski:
Ok, turns out this approach doesn't entirely work because having things like `OwnedValue` and `Object` imported from simd-json 0.10 is not compatible with methods like `to_string` which are imported from 0.4. What would need to happen is for two separate preludes to be created, one which is `pub` that re-exports everything from 0.4, and one which is internal containing everything from 0.10. Additionally, the pub types like `Value`, `JsonMap`, `JsonError`, etc. would need to be set to their versions from `simd_json_old`, which means that their new versions should therefore also be exported only in the "internal-only" prelude.

Considering that the `json` macro is broken on old simd-json versions, I'm really not sure if it's worth fixing this just to re-export a broken API, because the later rebases of `next` on top of that work  would be pretty painful. Maybe we just bite the bullet and re-export the new versions of the functions, except for `from_str` for which the wrapper is already `pub` anyways? I don't think any of the other function signatures have been broken by simd-json.

The other fixes in this PR (like the feature flag typo fix) are fine and can be left in.

mkrasnitski:
Any updates on this?

kangalio:
I'm currently not motivated to implement that separate prelude thing, and pushing breaking changes into `current` feels wrong to me. Maybe it's ok if I just close this PR, let simd-json continue to be broken on current, and wait for `next` instead ™ ?

mkrasnitski:
If you'd like, we can close this and split it into separate PRs.

kangalio:
Split into what PRs? What would they do?

mkrasnitski:
Replacing the uses of the incorrect `simd-json` feature with `simd_json` is relatively simple and worth incorporating.

kangalio:
More trouble than its worth tbh. That replacing is a "pick up a piece of litter in your way" level of change

PR #2537, opened 2023-09-16
Implement ID as a generic type with a marker

GnomedDev:
Implements all the common ID behavior on an `Id` type with a `Marker` for each model. This shouldn't be breaking on it's own, due to type aliases being left in as the "intended" way to use for beginners.

This also makes the internal representation of `Id` as a `NonZeroU64` a lot more private, which allows for using `repr(packed)` without confusing beginners, which makes the models a tiny bit smaller (although a lot of the enums for arguments have probably gotten smaller, they just aren't covered under the tests).

This also implements `FromStr` for all IDs, simply forwarding to the ID's `FromStr` implementation. This was the assumed behavior by a lot of serenity's internals and allows users to opt-in to having mention parsing with the specific `utils` methods.

This also creates the `cast` method, although this has been left as `pub(crate)` to be made public with a future PR after discussion.

Someone should probably check the compile time impact of this.


src/model/id.rs
```diff
@@ -1,338 +1,315 @@
...
+#[derivative(
+    Clone(bound = ""),
+    Copy(bound = ""),
+    Debug(bound = ""),
+    Eq(bound = ""),
+    Hash(bound = ""),
+    Ord(bound = ""),
+    PartialEq(bound = ""),
+    PartialOrd(bound = "")
+)]
```
mkrasnitski, line 25: What's the reason for using the derivative crate?

GnomedDev: If it didn't, the Marker types would have to be all of these traits for any of the derives to work. derivative allows providing the bounds (as none) to avoid having to write all the implementations out manually.

kangalio:
Idk man this is yet another big breaking change on top of the pile, for imo negligible benefit. You don't have to agree, I just wanna put this concern out there

tazz4843:
Ran a quick benchmark on the compile time. 

## Debug mode
Codegen time is essentially unchanged, 2.9s for both this branch and serenity-next, but total time is within error margins.

### This branch
image

### serenity-next upstream
image

Command used was `taskset -c 0-15 cargo +nightly build --timings --features full -j16` (used taskset to lock to P-cores only on 13700K)

## Release mode
`serenity-next` is actually faster? Not a once-off, happens every single time.

### This branch
image

### serenity-next
image

`taskset -c 0-15 cargo +nightly build --timings --features full -j16 --release`

## Overall
Given that I'm already on a super powerful system the compile time impact would be magnified for most people who are on sometimes much weaker systems. Overall given the tiny amounts of memory savings, the hit to compile time, and what kanga said above about this being yet another big breaking change, I don't think this is worth it.

GnomedDev:
This isn't a breaking change for those who are still on the recommended and published version of serenity. If I cannot make breaking changes to make incremental improvements, there isn't a point for serenity to not be on 1.0 yet, let alone have the maintained version be a git branch specifically for breaking changes.

GnomedDev:
Cast is an improvement, and so is the model size decreases, and the future library API improvements from being able to be generic over IDs. For users who were using Id::new and get, this isn't even a breaking change, and it's a bad idea to stablise the internals as non-zero in case it needs changing.

src/model/id.rs
mkrasnitski: I think this could be achieved by just manually implementing the traits for `Id`. Additionally, I think it would be useful to have a `Marker` trait that acts as a marker trait (heh) for `Id` to bound its parameter over (so just `Id<M: Marker>`), as having `Id` be completely generic doesn't really make sense.

mkrasnitski: Tried my hand at it and this is what I got. The extra variable assignments are due to the fact that making a reference to a packed field can potentially cause undefined behavior if the reference is unaligned. Therefore, copying the value onto the stack and passing a reference to the copy avoids this problem.

```rust
impl<Marker> Copy for Id<Marker> {}
impl<Marker> Clone for Id<Marker> {
    fn clone(&self) -> Self {
        *self
    }
}

impl<Marker> fmt::Debug for Id<Marker> {
        let inner = self.inner;
        write!(f, "{:?}", inner)
    }
}

impl<Marker> PartialEq for Id<Marker> {
    fn eq(&self, other: &Self) -> bool {
        let inner = self.inner;
        let other_inner = other.inner;
        inner == other_inner
    }
}

impl<Marker> Eq for Id<Marker> {}

impl<Marker> Hash for Id<Marker> {
    fn hash<H: Hasher>(&self, state: &mut H) {
        let inner = self.inner;
        inner.hash(state)
    }
}

impl<Marker> PartialOrd for Id<Marker> {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        let inner = self.inner;
        let other_inner = other.inner;
        inner.partial_cmp(&other_inner)
    }
}

impl<Marker> Ord for Id<Marker> {
    fn cmp(&self, other: &Self) -> Ordering {
        let inner = self.inner;
        let other_inner = other.inner;
        inner.cmp(&other_inner)
    }
}
```

GnomedDev: I never said they couldn't be implemented, I just used derivative because it's easier and I didn't feel like writing them all out as you did. If this is a blocker to merge though, I'll be fine to copy paste that into this PR.

mkrasnitski: This approach is lighter weight and avoids the compile-time hit of adding `derivative`, which wasn't anywhere in the dependency tree before.

mkrasnitski:
This actually causes a big hit to documentation, in that the documentation for `CommandPermissionId`, `TargetId`, `ChannelId`, `MessageId`, `GuildId`, `RoleId`, `StickerId`, `UserId`, and `WebhookId` have all been duplicated together onto the doc page for `Id`. The documentation for the generic `impl<Marker> Id<Marker>` also appears in the middle of this list (since it's sorted in alphabetical order). As it stands, on this basis alone, I'm opposed to merging this PR.

Some cursory testing of build times on my end seems to indicate that there isn't really an impact on build time. Maybe a slight increase, but run-to-run variance makes me doubt that.

I like some parts of this PR - for one, I do think aligning the `FromStr` implementations to be the same for all IDs is a good move, as from one perspective, having `UserId`, `ChannelId`, and `RoleId` be special in that regard can be a bit surprising. So, making mention parsing for those ids more opt-in is good in theory, but we just need to make sure that the change is advertised well so that it doesn't silently break underneath users' feet.

Also, `cast` is a nice method, but its use is relatively niche in practice (e.g., the role id for `@everyone` is the same as the guild id). If there were a way to have this method just between the various Id types as they are currently on `next`, that would be nice.

With that in mind, IMO it would be best to split this PR into multiple parts so that concerns are separated and review is easier.

PR #2548, opened 2023-09-27
Add support for `default_values` on select menus

mkrasnitski:
For select menus not of type `String`, the `default_values` json field allows for autofilled values to be populated. Therefore, fields have been added to the corresponding `CreateSelectMenuKind` enum variants to support this.

I'd like feedback, specifically for `Mentionable`. Should there be separate arrays for default UserIds and RoleIds, or should there just be a single array of GenericIds? This would allow for more controlled ordering e.g. interleaving users and roles in the list.

I've tested that this works, but for some reason if `default_values` is populated, the client has a 5 second delay to render the select menu. Can anyone else confirm this? Unsure what could be causing it, as the time to response from the API remains normal with the field populated.

### DIFF
--- a/examples/testing/src/main.rs
+++ b/examples/testing/src/main.rs
@@ -328,11 +328,22 @@ async fn interaction(
                                 CreateSelectMenuOption::new("bar", "bar"),
                             ],
                         }))
-                        .select_menu(CreateSelectMenu::new("1", CreateSelectMenuKind::Mentionable))
-                        .select_menu(CreateSelectMenu::new("2", CreateSelectMenuKind::Role))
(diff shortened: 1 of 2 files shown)


arqunis (approved):
> I'd like feedback, specifically for Mentionable. Should there be separate arrays for default UserIds and RoleIds, or should there just be a single array of GenericIds? This would allow for more controlled ordering e.g. interleaving users and roles in the list.

While the latter would obviously offer more control over the order of users and roles together, I believe the current users->roles ordering should suffice. Users will likely play a bigger role in bots' select menus than roles, so showing them first will let them be emphasised.

PR #2551, opened 2023-10-05
Return the full data about a thread when it was deleted

arqunis:
This adds a new field/parameter to the `ThreadDelete`/`thread_delete` event.

The partial thread data from Discord is lacking a lot of details about the thread, like its name. This simply passes on the `GuildChannel` instance for the deleted thread from the cache to the user.


jamesbt365 (approved):
image

PR #2552, opened 2023-10-07
Convert owner_id on guild to optional

tazz4843:
Discord appears to not be sending `owner_id` at times, causing deserialization errors quite often for me. This PR fixes that.


tazz4843:
Unrelated issue, discussion ongoing in serenity server

PR #2557, opened 2023-10-08
Use repr(packed) for ID types

GnomedDev:
Relies on #2556. Simple change, makes no difference to user usage, just shrinks models.

### DIFF
--- a/src/model/id.rs
+++ b/src/model/id.rs
@@ -83,7 +83,8 @@ macro_rules! id_u64 {
 
             impl fmt::Display for $name {
                 fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-                    fmt::Display::fmt(&self.0, f)
+                    let inner = self.0;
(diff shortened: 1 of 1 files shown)


mkrasnitski:
How does this affect taking references to id fields inside of model structs? Are they always guaranteed to be aligned?

PS: also, you may want to rebase now that #2556 has been merged.

GnomedDev:
Makes no difference, from simple testing and the fact that serenity still compiles with the small set of changes.

mkrasnitski:
Oh, I guess it's references to the packed _fields_ that matter here. And since the inner id is private, this isn't an issue.

Small example:
```rust
#[repr(packed)]
struct Wrapper(NonZeroU64);

struct Wrapper2(Wrapper);

fn main() {
    let x = Wrapper2(Wrapper(NonZeroU64::new(1).unwrap()));
    
    let y = &x.0; // fine
    let z = &x.0.0; // not allowed
}
```

arqunis (approved):

PR #2560, opened 2023-10-11
Add support for new monetization APIs

mkrasnitski:
This PR adds support for the new monetization support added to Discord. Namely, the following endpoints:
* List SKUs
* List Entitlements
* Create Test Entitlement
* Delete Test Entitlement

As well, added via the `PREMIUM_REQUIRED` interaction response type and the `entitlements` field in interaction data. Also, added the `ENTITLEMENT_CREATE`, `ENTITLEMENT_UPDATE`, and `ENTITLEMENT_DELETE` gateway events.

Note: This PR is untested and there are some things I'd like verified before this gets merged. In particular:
- [x] The `PREMIUM_REQUIRED` response type is not valid for `PING` and `APPLICATION_COMMAND_AUTOCOMPLETE` interactions. I'm not sure if this means that the payloads received from those interactions don't contain an `entitlements` field or not. I've elected to keep the `entitlements` field for all interaction types, but I'd like this verified as well.
    - **ANSWER**: Turns out the field is present for `AUTOCOMPLETE` and not for `PING`.
- [x] Are the `before` and `after` parameters to the [List Entitlements](https://github.com/discord/discord-api-docs/pull/6477) endpoint mutually exclusive in any way, like they are for some other endpoints?
    - **ANSWER**: No, they aren't.

### DIFF
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -57,6 +57,13 @@ pub enum CreateInteractionResponse {
     ///
     /// Corresponds to Discord's `MODAL`.
     Modal(CreateModal),
+    /// Not valid for autocomplete and Ping interactions. Only available for applications with
+    /// monetization enabled.
+    ///
+    /// Responds to the interaction with an upgrade button.
+    ///
+    /// Corresponds to Discord's `PREMIUM_REQUIRED'.
+    PremiumRequired,
 }
 
 impl serde::Serialize for CreateInteractionResponse {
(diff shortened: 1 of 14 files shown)


aalhitennf:
Hopefully this gets added and released soon, many thanks for your work :+1:

mkrasnitski:
If anyone has access to a monetized app and wants to test the last remaining question, that would be greatly appreciated. These are the steps you should take:
1. Checkout my branch on this PR
2. Create multiple test entitlements (at least 3, preferably 5)
3. Query the `List Entitlements` endpoint with combinations of the `before` and `after` parameters (just `before`, just `after`, and both) to see what gets returned
4. Report whether or not having both `before` and `after` works as expected or not

I take it @aalhitennf you have access to a monetized app and would like to use these features in mainline serenity?

aalhitennf:
> I take it @aalhitennf you have access to a monetized app and would like to use these features in mainline serenity?

Yes. I'm using poise as serenity framework but i can probably test it out later.

aalhitennf:
I created 5 test entitlements. Here's some results with before and after parameters when querying the list endpoint.

I reduced the ids to single digit numbers for the sake of sanity. (order is apparently reversed because api returns them in that order, results are still valid)  


```
Before = None, After = None

[1,2,3,4,5]

Before = 5
[]

Before = 4
[5]

Before = 1
[2,3,4,5]

After = 5
[4,3,2,1]

After = 4
[3,2,1]

After = 1
[]

Before = 5, After = 4
[3,2,1]

Before = 4, After = 5
[4,3,2,1]

Before = 1, After = 1
[]

Before = 1, After = 3
[2,1]

Before = 2, After = 4
[3,2,1]

Before = 4, After = 2
[1]
```

Doing same queries with http requests using Insomnia/Postman/whatever returns same results. So, looks good to me :+1:

mkrasnitski:
Alright, seems like the two parameters play nice with each other. So, we don't need to make them mutually exclusive. This PR should be ready for merge now.

arqunis (approved):

PR #2564, opened 2023-10-15
Rename `utils::parse_{username,role,channel}` methods

mkrasnitski:
Renames the methods to `parse_{user,role,channel}_mention` while leaving the old names as deprecated aliases.


mkrasnitski:
Weird that clippy triggers `must_use_candidate` only on the functions taking `&str` as parameter and not the ones taking `impl AsRef<str>`. I wonder why that is :thinking:

arqunis (approved):

PR #2565, opened 2023-10-16
Implement VOICE_CHANNEL_STATUS_UPDATE

jamesbt365:
I will cleanup the code, format this right later and document whatever I can later. I thought  thought it would be in a decent enough state to be opened as a draft.

Not sure if you even want this yet given that the documentation for this isn't even merged yet and is just a draft pull request on discord/discord-api-docs

I had no idea what I was doing with the CacheUpdate, so the code is probably not even close to what it could be.

### DIFF
--- a/src/builder/edit_channel.rs
+++ b/src/builder/edit_channel.rs
@@ -72,6 +72,8 @@ pub struct EditChannel<'a> {
     default_sort_order: Option<SortOrder>,
     #[serde(skip_serializing_if = "Option::is_none")]
     default_forum_layout: Option<ForumLayoutType>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    status: Option<String>,
 
     #[serde(skip)]
     audit_log_reason: Option<&'a str>,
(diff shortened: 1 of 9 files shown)


jamesbt365:
I have gone ahead and pushed a new function to EditChannel to change the status, not sure if the comments I've put on it are 100% correct because I am yet to fully test.

Also included the permission to do so, not sure if I need to change anything else alongside that.

I am pulling everything off discord/discord-api-docs/pull/6398 but its not merged yet so things might change or I might be wrong in some places.

jamesbt365:
If anything would be better off split into a different pull request I'll do so, just ask.

src/builder/edit_channel.rs
```diff
@@ -287,6 +289,14 @@ impl<'a> EditChannel<'a> {
...
         self
     }
+    /// The status of a voice channel.
+    /// The user must be within the channel to change the status but if the
+    /// user has Manage Channels the user can reset it without being in the channel.
+    /// Can be None, an empty string or up to 500 characters.
+    pub fn status(mut self, status: impl Into<Option<String>>) -> Self {
+        self.status = status.into();
+        self
+    }
```
arqunis: Shouldn't this be done for the special [Set Voice Channel Status](https://github.com/discord/discord-api-docs/pull/6400/files) endpoint instead of the Edit Channel endpoint?

jamesbt365: I guess you are probably right, yeah setting it to None *does* work (haven't tested trying to set it to an actual status because I haven't set up joining vcs on my bot yet) but it fires CHANNEL_UPDATE instead of VOICE_CHANNEL_STATUS_UPDATE so yeah I'm guessing this needs to be moved.

Any pointers on how I could implement the set voice channel status endpoint?

AngelOnFira (approved):
Doing this for a workshop sorry

src/builder/edit_channel.rs
jamesbt365: I have figured out how to do this but it currently has an experiment lock so it won't actually do anything yet. I think its best if we remove the changing status code from this pull request and split it off into a different pull request until it no longer has this limitation.

arqunis (approved):

PR #2569, opened 2023-10-18
Replace `serenity::json::prelude` with public wrapper functions

mkrasnitski:
To avoid future breakage, we shouldn't re-export functions from dependency libs in case a function signature changes, which would pass on breaking changes to serenity's users. Providing wrapper functions solves this issue.

Note: changed `from_str`to take `&str` instead of `String`, and simply clone it for the `simd_json` case. This means the more common `serde_json` use case no longer requires ownership (or an extra clone before passing the string in).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ secrecy = { version = "0.8.0", features = ["serde"] }
 arrayvec = { version = "0.7.3", features = ["serde"] }
 # Optional dependencies
 fxhash = { version = "0.2.1", optional = true }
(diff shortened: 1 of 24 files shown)


mkrasnitski:
I have no idea why the doctest on `RichInvite::url` is failing only if the `simd_json` feature is enabled :pensive:

GnomedDev:
Taking Cow and documenting the clone for simd_json would be the best path.

arqunis (approved):

PR #2570, opened 2023-10-19
Add optional `description` field to `CreateAttachment`

### LINKED ISSUE #2563: optional `description` field missing in `CreateAttachment`
Specifying a `description` for attachments when uploading or editing messages is currently not possible, because `CreateAttachment` does not support this optional field. 

The content of field is displayed as the image description/alt text when uploading image attachments. 

### Further References
- [API Attachment Object](https://discord.com/developers/docs/resources/channel#attachment-object)
- [API Example Payload](https://discord.com/developers/docs/reference#editing-message-attachments-example-request-bodies-multipartformdata)

Kneemund:
Closes #2563.

## Description
Currently, uploading files as message attachments is done by simply appending them as multipart form parts, without creating an entry in the `attachments` field of the JSON payload.
Because the `description` is specified in the entries of that field, this PR updates all builders that are used for uploading files: during the build/execute process, entries for each file of the `files` array are added to the `attachments` field of the payload.

The PR also makes changes to the `Multipart` struct.
Currently, the part name of the 0th file is set to `file`, all other file parts are called `fileN` because they must have unique names. This is done to work around the fact that Discord expects the part name for endpoints that take in single file uploads to be `file`, e.g. when creating stickers.
I believe that using `fileN` only works because the format of part names is not actively enforced unless you use the `attachments` field like described above.

When creating entries in the `attachments` field, Discord expects their ID to be set to `N` (starting at 0), and the name of the part containing the data to be set to `files[N]` (see [example payload](https://discord.com/developers/docs/reference#editing-message-attachments-example-request-bodies-multipartformdata)). Due to these strict naming rules, this PR replaces the `files` field of the `Multipart` struct with two new fields: 
- `upload_file: Option<CreateAttachment>` for uploading a single file to endpoints like creating stickers - the part name is set to `file`.
- `attachment_files: Option<Vec<CreateAttachment>>` for uploading multiple files as message attachments - the part name is set to `files[N]`.

## Breaking changes
- Field `description: Option<String>` added to struct `CreateAttachment`.
- Field `files: Vec<CreateAttachment>` removed, fields `upload_file: Option<CreateAttachment>` and `attachment_files: Option<Vec<CreateAttachment>>` added to struct `Multipart`.

## Things to do, change and discuss
- [ ] Testing.
- [ ] Adding `attachments` entries for new `files` in the execute function of the builders is a bit weird and should probably be done in the functions that also modify the `files` array (like `new_attachment`). I'm not sure how exactly that should be done (in a clean way).
- [ ] Structs and fields might need to be renamed (e.g. `upload_file`/`attachment_files`).
- [x] `ExistingAttachment` cannot continue to use the type `AttachmentId` for its ID field ATM, because the inner type is `NonZeroU64` and the ID of new files is an index starting at zero (see above). I'm not sure how to properly handle this. Maybe there should be a new struct for this and `ExistingAttachment` should be left as-is.
- ~~If the description and/or file name of existing attachments can be edited, this should (and easily could) be exposed when editing messages.~~ This doesn't seem to be possible.
- [ ] As far as I can tell, the `filename` field of the `NewAttachment` struct is redundant and we could remove it, because the filename is already included directly in the multipart request.

Help, critique and any other comments on some of these points and the code would be much appreciated.

### DIFF
--- a/src/builder/create_attachment.rs
+++ b/src/builder/create_attachment.rs
@@ -10,7 +10,6 @@ use crate::error::Error;
 use crate::error::Result;
 #[cfg(feature = "http")]
 use crate::http::Http;
-use crate::model::id::AttachmentId;
 
 /// [Discord docs] with the caveat at the top "For the attachments array in Message Create/Edit
 /// requests, only the id is required."
@@ -18,9 +17,29 @@ use crate::model::id::AttachmentId;
 /// [Discord docs]: https://discord.com/developers/docs/resources/channel#attachment-object-attachment-structure
 #[derive(Clone, Debug, Serialize)]
 pub(crate) struct ExistingAttachment {
-    pub id: AttachmentId,
+    // TODO: should be AttachmentId, but the ID of entries for uploaded files is the file index
+    // which starts at zero (thus not fitting into NonZeroU64).
+    pub id: u64,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub filename: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
     // TODO: add the other non-required attachment fields? Like content_type, description,
     // ephemeral (ephemeral in particular seems pretty interesting)
+}
+
+impl ExistingAttachment {
+    pub(crate) fn from_files(files: &[CreateAttachment]) -> Vec<Self> {
+        files
+            .iter()
+            .enumerate()
+            .map(|(i, file)| Self {
+                id: i as u64,
+                filename: Some(file.filename.clone()),
+                description: file.description.clone(),
+            })
+            .collect()
+    }
 }
 
 /// Enum that allows a user to pass a [`Path`] or a [`File`] type to [`send_files`]
@@ -32,6 +51,7 @@ pub(crate) struct ExistingAttachment {
 pub struct CreateAttachment {
     pub data: Vec<u8>,
     pub filename: String,
+    pub description: Option<String>,
 }
 
 impl CreateAttachment {
@@ -40,6 +60,7 @@ impl CreateAttachment {
         CreateAttachment {
             data: data.into(),
             filename: filename.into(),
+            description: None,
         }
     }
 
@@ -63,6 +84,7 @@ impl CreateAttachment {
         Ok(CreateAttachment {
             data,
             filename: filename.to_string_lossy().to_string(),
+            description: None,
         })
     }
 
@@ -78,6 +100,7 @@ impl CreateAttachment {
         Ok(CreateAttachment {
             data,
             filename: filename.into(),
+            description: None,
         })
     }
 
@@ -101,6 +124,7 @@ impl CreateAttachment {
         Ok(CreateAttachment {
             data,
             filename: filename.to_string(),
+            description: None,
         })
     }
 
@@ -116,5 +140,10 @@ impl CreateAttachment {
         };
         encoded.insert_str(0, "data:image/png;base64,");
         encoded
+    }
+
+    pub fn description(mut self, description: Option<String>) -> Self {
+        self.description = description;
+        self
     }
 }
--- a/src/builder/create_attachment.rs
+++ b/src/builder/create_attachment.rs
@@ -10,6 +10,15 @@ use crate::error::Error;
 use crate::error::Result;
 #[cfg(feature = "http")]
 use crate::http::Http;
+use crate::model::id::AttachmentId;
+
+/// Enum that allows to add existing attachments and new attachments to the payload.
+#[derive(Clone, Debug, Serialize)]
+#[serde(untagged)]
+pub(crate) enum MessageAttachment {
+    Existing(ExistingAttachment),
+    New(NewAttachment),
+}
 
 /// [Discord docs] with the caveat at the top "For the attachments array in Message Create/Edit
 /// requests, only the id is required."
@@ -17,26 +26,32 @@ use crate::http::Http;
 /// [Discord docs]: https://discord.com/developers/docs/resources/channel#attachment-object-attachment-structure
 #[derive(Clone, Debug, Serialize)]
 pub(crate) struct ExistingAttachment {
-    // TODO: should be AttachmentId, but the ID of entries for uploaded files is the file index
-    // which starts at zero (thus not fitting into NonZeroU64).
-    pub id: u64,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub filename: Option<String>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub description: Option<String>,
+    pub id: AttachmentId,
     // TODO: add the other non-required attachment fields? Like content_type, description,
     // ephemeral (ephemeral in particular seems pretty interesting)
 }
 
-impl ExistingAttachment {
+/// Represents a new attachment in the payload, and allows for passing a description.
+#[derive(Clone, Debug, Serialize)]
+pub(crate) struct NewAttachment {
+    #[serde(rename = "id")]
+    pub index: u64,
+    pub filename: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
+}
+
+impl MessageAttachment {
     pub(crate) fn from_files(files: &[CreateAttachment]) -> Vec<Self> {
         files
             .iter()
             .enumerate()
-            .map(|(i, file)| Self {
-                id: i as u64,
-                filename: Some(file.filename.clone()),
-                description: file.description.clone(),
+            .map(|(i, file)| {
+                Self::New(NewAttachment {
+                    index: i as u64,
+                    filename: file.filename.clone(),
+                    description: file.description.clone(),
+                })
             })
             .collect()
     }
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -1,6 +1,12 @@
 #[cfg(feature = "http")]
 use super::{check_overflow, Builder};
-use super::{CreateActionRow, CreateAllowedMentions, CreateAttachment, CreateEmbed};
+use super::{
+    CreateActionRow,
+    CreateAllowedMentions,
+    CreateAttachment,
+    CreateEmbed,
+    ExistingAttachment,
+};
 #[cfg(feature = "http")]
 use crate::constants;
 #[cfg(feature = "http")]
@@ -132,7 +138,11 @@ impl Builder for CreateInteractionResponse {
         let files = match &mut self {
             CreateInteractionResponse::Message(msg)
             | CreateInteractionResponse::Defer(msg)
-            | CreateInteractionResponse::UpdateMessage(msg) => std::mem::take(&mut msg.files),
+            | CreateInteractionResponse::UpdateMessage(msg) => {
+                let files = std::mem::take(&mut msg.files);
+                msg.attachments.extend(ExistingAttachment::from_files(&files));
+                files
+            },
             _ => Vec::new(),
         };
 
@@ -156,6 +166,8 @@ pub struct CreateInteractionResponseMessage {
     flags: Option<MessageFlags>,
     #[serde(skip_serializing_if = "Option::is_none")]
     components: Option<Vec<CreateActionRow>>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    attachments: Vec<ExistingAttachment>,
 
     #[serde(skip)]
     files: Vec<CreateAttachment>,
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -5,7 +5,7 @@ use super::{
     CreateAllowedMentions,
     CreateAttachment,
     CreateEmbed,
-    ExistingAttachment,
+    MessageAttachment,
 };
 #[cfg(feature = "http")]
 use crate::constants;
@@ -140,7 +140,7 @@ impl Builder for CreateInteractionResponse {
             | CreateInteractionResponse::Defer(msg)
             | CreateInteractionResponse::UpdateMessage(msg) => {
                 let files = std::mem::take(&mut msg.files);
-                msg.attachments.extend(ExistingAttachment::from_files(&files));
+                msg.attachments.extend(MessageAttachment::from_files(&files));
                 files
             },
             _ => Vec::new(),
@@ -167,7 +167,7 @@ pub struct CreateInteractionResponseMessage {
     #[serde(skip_serializing_if = "Option::is_none")]
     components: Option<Vec<CreateActionRow>>,
     #[serde(skip_serializing_if = "Vec::is_empty")]
-    attachments: Vec<ExistingAttachment>,
+    attachments: Vec<MessageAttachment>,
 
     #[serde(skip)]
     files: Vec<CreateAttachment>,
--- a/src/builder/create_interaction_response_followup.rs
+++ b/src/builder/create_interaction_response_followup.rs
@@ -5,7 +5,7 @@ use super::{
     CreateAllowedMentions,
     CreateAttachment,
     CreateEmbed,
-    ExistingAttachment,
+    MessageAttachment,
 };
 #[cfg(feature = "http")]
 use crate::constants;
@@ -33,7 +33,7 @@ pub struct CreateInteractionResponseFollowup {
     #[serde(skip_serializing_if = "Option::is_none")]
     flags: Option<MessageFlags>,
     #[serde(skip_serializing_if = "Vec::is_empty")]
-    attachments: Vec<ExistingAttachment>,
+    attachments: Vec<MessageAttachment>,
 
     #[serde(skip)]
     files: Vec<CreateAttachment>,
@@ -188,7 +188,7 @@ impl Builder for CreateInteractionResponseFollowup {
         self.check_length()?;
 
         let files = std::mem::take(&mut self.files);
-        self.attachments.extend(ExistingAttachment::from_files(&files));
+        self.attachments.extend(MessageAttachment::from_files(&files));
 
         let http = cache_http.http();
         match ctx.0 {
--- a/src/builder/create_interaction_response_followup.rs
+++ b/src/builder/create_interaction_response_followup.rs
@@ -1,6 +1,12 @@
 #[cfg(feature = "http")]
 use super::{check_overflow, Builder};
-use super::{CreateActionRow, CreateAllowedMentions, CreateAttachment, CreateEmbed};
+use super::{
+    CreateActionRow,
+    CreateAllowedMentions,
+    CreateAttachment,
+    CreateEmbed,
+    ExistingAttachment,
+};
 #[cfg(feature = "http")]
 use crate::constants;
 #[cfg(feature = "http")]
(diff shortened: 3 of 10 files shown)


mkrasnitski:
IMO, `ExistingAttachment` should be left as-is because having it perform double duty by representing both *existing* attachments whose IDs are snowflakes, and also *new* attachments whose IDs are indices, makes no sense.

Kneemund:
I agree, I added a new struct and enum for splitting the logic.

mkrasnitski:
Could you rebase this on top of `next` please? You may need to cherry-pick each of the two commits.

kangalio:
I rebased, I hope you didn't mind @Kneemund

mkrasnitski:
Here's an idea: because `upload_file` and `attachment_files` are mutually exclusive (I think), why not change the `files` field into an enum like this:
```rust
enum Files {
    Upload(CreateAttachment)
    Attachments(Vec<CreateAttachments>)
}
```
The names could use some work, but this would make it a lot cleaner IMO.

src/builder/create_attachment.rs
```diff
@@ -117,4 +156,9 @@ impl CreateAttachment {
         encoded.insert_str(0, "data:image/png;base64,");
         encoded
     }
+
+    pub fn description(mut self, description: Option<String>) -> Self {
```
kangalio, line 160: Any reason against `description: impl Into<String>`? Seems more useful and less boilerplatey for users

src/builder/create_interaction_response.rs
```diff
@@ -156,6 +166,8 @@ pub struct CreateInteractionResponseMessage {
     flags: Option<MessageFlags>,
     #[serde(skip_serializing_if = "Option::is_none")]
     components: Option<Vec<CreateActionRow>>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    attachments: Vec<MessageAttachment>,
```
kangalio, line 170: Pretty sure this should be NewAttachment instead https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-response-object-messages

src/builder/create_interaction_response_followup.rs
```diff
@@ -26,6 +32,8 @@ pub struct CreateInteractionResponseFollowup {
     components: Option<Vec<CreateActionRow>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     flags: Option<MessageFlags>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    attachments: Vec<MessageAttachment>,
```
kangalio, line 36: NewAttachment, see above

src/builder/create_interaction_response_followup.rs
```diff
@@ -178,7 +186,9 @@ impl Builder for CreateInteractionResponseFollowup {
         ctx: Self::Context<'_>,
     ) -> Result<Self::Built> {
         self.check_length()?;
+
         let files = std::mem::take(&mut self.files);
+        self.attachments.extend(MessageAttachment::from_files(&files));
```
kangalio, line 191: This seems weird.. there are now two places for the same thing (attachments)?

src/builder/create_message.rs
```diff
@@ -61,6 +67,8 @@ pub struct CreateMessage {
     sticker_ids: Vec<StickerId>,
     #[serde(skip_serializing_if = "Option::is_none")]
     flags: Option<MessageFlags>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    attachments: Vec<MessageAttachment>,
```
kangalio, line 71: NewAttachment, see above

src/builder/edit_interaction_response.rs
```diff
@@ -126,7 +127,15 @@ impl Builder for EditInteractionResponse {
         ctx: Self::Context<'_>,
     ) -> Result<Self::Built> {
         self.0.check_length()?;
+
         let files = std::mem::take(&mut self.0.files);
+        if !files.is_empty() {
+            self.0
+                .attachments
+                .get_or_insert(Vec::new())
+                .extend(MessageAttachment::from_files(&files));
```
kangalio, line 136: Two places for the same thing? See above

src/builder/edit_message.rs
```diff
@@ -55,7 +56,7 @@ pub struct EditMessage {
     #[serde(skip_serializing_if = "Option::is_none")]
     components: Option<Vec<CreateActionRow>>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    attachments: Option<Vec<ExistingAttachment>>,
+    attachments: Option<Vec<MessageAttachment>>,
```
kangalio, line 59: Do I understand Discord docs correctly that the 0-based index for new attachments, and the snowflake-based ID for existing attachments, are mixed here? Like, `attachments: [{"id": 0, ...}, {"id": 1234, ...}]` is valid, where `1234` is the ID an existing attachment that should be kept, and `0` is the `files[n]` index for a new attachment which is supplied in the POST form data?

This seems like hella weird design from Discord, but if true, `MessageAttachment` would indeed be the correct way to represent it in Rust, I guess

src/builder/create_attachment.rs
```diff
@@ -12,6 +12,14 @@ use crate::error::Result;
 use crate::http::Http;
 use crate::model::id::AttachmentId;
 
+/// Enum that allows to add existing attachments and new attachments to the payload.
+#[derive(Clone, Debug, Serialize)]
+#[serde(untagged)]
+pub(crate) enum MessageAttachment {
```
kangalio, line 18: I'd vouch for this type to be called `NewOrExistingAttachment`, to make clearer that it's for the special case of editing attachments where you have to juggle either case. With the name `MessageAttachment`, I would think it means what `NewAttachment` actually is

src/builder/create_attachment.rs
```diff
@@ -23,6 +31,32 @@ pub(crate) struct ExistingAttachment {
     // ephemeral (ephemeral in particular seems pretty interesting)
 }
 
+/// Represents a new attachment in the payload, and allows for passing a description.
+#[derive(Clone, Debug, Serialize)]
+pub(crate) struct NewAttachment {
```
kangalio, line 36: I'd vouch for this to be renamed to `NewAttachmentMetadata` to make it clearer what it is (I needed a very long time just now to untangle my thoughts and understand what this struct actually is)

src/builder/create_message.rs
kangalio, line 71: Also, it would be much clearer imo to rename this field to `attachments_metadata`, and rename `files` to `attachments`. That more closely reflects what they are: `attachments_metadata` is always empty, except very briefly before sending the request, when metadata is extracted out of `attachments` (new naming) into `attachments_metadata`, because that's where Discord expects the metadata to be.

Same proposal applies for all other builder structs that have both `attachments` and `files` currently

src/builder/create_interaction_response_followup.rs
kangalio: 20min later, I now understand this line of code (and the long time I needed to understand this is part of my motivation to rename `attachments` + `files` to `attachments_metadata` + `attachments`).

Anyways, I think `self.attachments = ` would be better than `.extend()` because it makes it even clearer that `attachments_metadata` (new naming) _does not_ have any items so far, and will only be filled by this line of code alone

src/builder/edit_interaction_response.rs
```diff
@@ -126,7 +127,15 @@ impl Builder for EditInteractionResponse {
         ctx: Self::Context<'_>,
     ) -> Result<Self::Built> {
         self.0.check_length()?;
+
         let files = std::mem::take(&mut self.0.files);
+        if !files.is_empty() {
```
kangalio, line 132: Why this check?

src/builder/edit_message.rs
```diff
@@ -258,7 +264,14 @@ impl Builder for EditMessage {
         ctx: Self::Context<'_>,
     ) -> Result<Self::Built> {
         self.check_length()?;
+
         let files = std::mem::take(&mut self.files);
+        if !files.is_empty() {
```
kangalio, line 269: Why this check? (see above)

src/builder/edit_webhook_message.rs
```diff
@@ -189,7 +192,14 @@ impl Builder for EditWebhookMessage {
         ctx: Self::Context<'_>,
     ) -> Result<Self::Built> {
         self.check_length()?;
+
         let files = std::mem::take(&mut self.files);
+        if !files.is_empty() {
```
kangalio, line 197: Dito

kangalio (changes requested):

kangalio:
> Here's an idea: because `upload_file` and `attachment_files` are mutually exclusive (I think), why not change the `files` field into an enum like this:
> 
> ```rust
> enum Files {
>     Upload(CreateAttachment)
>     Attachments(Vec<CreateAttachments>)
> }
> ```
> 
> The names could use some work, but this would make it a lot cleaner IMO.

I just played around with it locally.. I now think the added complexity with the nested types is not worth it for the little gain

src/builder/edit_message.rs
kangalio: I just tested, and yes, indeed. That is how Discord works.

Screenshot_20231026_113756

This `attachments` contents indeed edits the existing attachment of ID 1167034226038800415, _and_ adds a new attachment of index 0. That's wild

src/builder/edit_interaction_response.rs
Kneemund: If the payload includes the `attachments` field, all existing attachments but the ones specified there are removed. If we unconditionally create the vector, the `attachments` field is not skipped during serialization and added as an empty array, meaning that all existing attachments would always be removed.

If we skip the serialization if the vector is empty, users are unable to clear all attachments using `clear_existing_attachments`.

The fact that adding new files or calling `clear_existing_attachments` or `keep_existing_attachment` (both initialize the vector, and the latter extends it) removes all existing, not manually added attachments from the message is known and (I think) documented behavior. The check is needed to keep that behavior.

src/builder/create_attachment.rs
Kneemund: What about removing the description? Or should you have to do that manually using assignment or another function?
Depending on what's desirable, this could also be `Option<impl Into<String>>` I guess.

src/builder/edit_message.rs
Kneemund: > edits the existing attachment of ID 1167034226038800415

Wait, you can do that? I did very little amount of testing on this and couldn't get editing the description and/or filename to work. Can you also edit the ephemeral flag? Seems a bit weird

src/builder/create_interaction_response.rs
Kneemund: Yeah, I don't *think* that you can send existing attachments in new messages. I just did this for simplicity, but I guess moving `MessageAttachment::from_files` to `NewAttachment::from_files` and changing these types would be better.

src/builder/create_interaction_response_followup.rs
Kneemund: Alright. 

> Anyways, I think self.attachments = would be better than .extend() because it makes it even clearer that attachments_metadata (new naming) does not have any items so far, and will only be filled by this line of code alone

This only applies to message creation though. `attachments_metadata` can have entries when editing messages.

src/builder/create_attachment.rs
kangalio: `description` is None by default, no? So to set to None, you'd just not call this function

src/builder/create_interaction_response.rs
kangalio: > Yeah, I don't think that you can send existing attachments in new message

As in, copy attachments from other messages into a new message by just referencing their attachment ID? That sounds absolutely wild, I'd be surprised if that works haha

src/builder/create_interaction_response_followup.rs
kangalio: That is a good point 😵‍💫 This all is confusing.. Maybe there's a way to centralize all this attachment complexity somewhere instead of needing it spread out across all builders.. that'd be cool. But not sure if possible

src/builder/edit_interaction_response.rs
kangalio: I see. All of that makes perfect sense.

This situation, where existing attachments are sometimes removed, sometimes not, depending on which methods are called, is really confusing ngl. I wonder if serenity can invent a less footgunny API here. Or maybe we can get away _always_ supplying `attachments`, but pre-filling it with AttachmentId's wherever serenity has the required data, and requiring users to do it where not?

src/builder/edit_message.rs
kangalio: Editing the filename triggered a Discord error ATTACHMENT_FILENAME_NOT_EDITABLE (or something like that), from which I deduced that Discord indeed understood I meant to edit the existing attachment, and that it's principally possible.

Editing `ephemeral` did nothing. Editing `description` - not sure, I was too lazy to try image attachments, and descriptions dont seem to show in text attachments

Kneemund:
> > Here's an idea: because `upload_file` and `attachment_files` are mutually exclusive (I think), why not change the `files` field into an enum like this:
> > ```rust
> > enum Files {
> >     Upload(CreateAttachment)
> >     Attachments(Vec<CreateAttachments>)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The names could use some work, but this would make it a lot cleaner IMO.
> 
> I just played around with it locally.. I now think the added complexity with the nested types is not worth it for the little gain

Hm, I think it's a bit more readable. Especially because it highlights the mutual exclusivity.

src/builder/create_attachment.rs
Kneemund: Fair enough. I guess the cases in which you've set it and want to remove it later on are too rare to really consider.

src/builder/edit_interaction_response.rs
kangalio: I wrote up all the possible combinations of `attachments` and `files`, with existing and new attachments, and what will happen to the message:

```md
If we edit a message that currently has attachments [old], then the following possible request variants exist:

attachments None, files []
-> attachments unchanged

attachments None, files [new]
-> attachments replaced

attachments Some([]), files []
-> attachments removed

attachments Some([]), files [new]
-> attachments removed (new attachments NOT ADDED!)

attachments Some([old]), files []
-> attachments unchanged, existing attachments configurable

attachments Some([old]), files [new]
-> attachments unchanged, existing attachments configurable (new attachments NOT ADDED!)

attachments Some([new]), files []
-> attachments removed

attachments Some([new]), files [new]
-> attachments replaced, new attachments configurable (e.g. `description`)

attachments Some([old, new]), files []
-> attachments unchanged, existing attachments configurable

attachments Some([old, new]), files [new]
-> attachments appended, both existing and new attachments configurable
```

So these are all possible unique actions we can take with what we have:

1. attachments removed (no extra data needed)
2. attachments unchanged (no extra data needed)
3. attachments unchanged, existing attachments configurable (**old attachment IDs needed**)
4. attachments replaced (no extra data needed)
5. attachments replaced, new attachments configurable (e.g. `description`) (no extra data needed)
6. attachments appended, both existing and new attachments configurable (**old attachment IDs needed**)

Of these, 4 is redundant, because 5 is strictly more powerful and requires no extra data

So, these remaining 5 possibilities could be represented in Rust like this

```rust
pub enum EditAttachments {
    RemoveExisting, // 1
    KeepExisting, // 2
    ConfigureExisting(Vec<ExistingAttachment>), // 3
    RemoveExistingAndAddNew(Vec<NewAttachment>), // 5
    KeepOrConfigureExistingAndAddNew(Vec<ExistingAttachment>, Vec<NewAttachment>), // 6
}
```

Technically, RemoveExisting is redundant in favor of ConfigureExisting(vec![]), RemoveExistingAndAddNew(...) is redundant in favor of KeepOrConfigureExistingAndAddNew(vec![], ...), and ConfigureExisting(...) is redundant in favor of KeepOrConfigureExistingAndAddNew(..., vec![]). So this enum can be consolidated into:

```rust
enum EditAttachments {
    KeepExisting,
    Configure {
        // Set to vec![] to remove all, set to vec![<all AttachmentIds from the msg>] to keep all
        existing: Vec<ExistingAttachment>,
        // Set to vec![] if you just want to edit existing
        new: Vec<NewAttachment>,
    }
}
```

We could go one step ahead, and say that builder structs should always store EditAttachments in an Option, and None will correspond to KeepExisting. Then the code simplifies to:

```rust
pub struct EditAttachments {
    /// Set to vec![] to remove all, set to vec![<all AttachmentIds from the msg>] to keep all
    pub existing: Vec<ExistingAttachment>,
    /// Set to vec![] if you just want to edit existing
    pub new: Vec<NewAttachment>,
}
```

Now, `NewAttachment` could also start containing the `CreateAttachment` including the file contents within itself. This would simplify the API quite a lot I think, and remove footguns

kangalio:
Man Discord attachments keeps on giving.

So I mentioned how I assumed the properties of attachments were actually editable in principle, because Discord has a specialized error message for how the `filename` field in particular _cannot_ be edited:

Screenshot_20231026_162259

This implies that other fields _can_ be edited, right? However, when I try, nothing happens. When supplying a non-null value for anything but `filename` in the attachment object in the edit request, Discord throws no error, but doesn't change anything either. So.. even though Discord evidently has infrastructure for it, you _cannot_ currently edit existing attachments at all.

So.. I suppose serenity doesn't need to expose editing existing attachments in its API?

mkrasnitski:
> So.. I suppose serenity doesn't need to expose editing existing attachments in its API?

If we're talking about editing the properties of existing attachments, then it's fine to err on the safe side and not expose this functionality if Discord doesn't fully support it.

src/builder/edit_interaction_response.rs
Kneemund: > We could go one step ahead, and say that builder structs should always store EditAttachments in an Option, and None will correspond to KeepExisting. Then the code simplifies to:
>
> ```rust 
> pub struct EditAttachments {
>     /// Set to vec![] to remove all, set to vec![<all AttachmentIds from the msg>] to keep all
>     pub existing: Vec<ExistingAttachment>,
>     /// Set to vec![] if you just want to edit existing
>     pub new: Vec<NewAttachment>,
> }
> ```

Internally, that's pretty much how data is laid out right now, but separated and as a single struct (`Option<EditAttachments>` instead of `Option<Vec<MessageAttachment>>`). Or do you want to expose this to the users as a new builder, and remove all existing methods of the message builders?

kangalio: I think that would be easier to understand, yes. Easier to understand in the code for devs, and easier to understand for users

kangalio: I'm currently working on it locally as an experiment

kangalio:
I also noticed Discord wants the filename duplicated which is also kinda goofy lol
image

kangalio:
Btw, sorry that this PR erupted so much @Kneemund. It's because serenity's attachment/file uploading needed an overhaul anyways, and Discord's API for it is quite hard to understand, and this PR was just the catalyst for the mountain of pending work and complexity that has accumulated

Kneemund:
> I also noticed Discord wants the filename duplicated which is also kinda goofy lol

The example seems like it, but it could be possible to omit it from the `attachments` field. But since this part of the specification really sucks (this is an *example* payload), the only way to find out is testing.

[This statement](https://discord.com/developers/docs/resources/channel#attachment-object) however makes it seem like you *can* omit the filename from the payload:

> grafik

Kneemund:
> Btw, sorry that this PR erupted so much @Kneemund. It's because serenity's attachment/file uploading needed an overhaul anyways, and Discord's API for it is quite hard to understand, and this PR was just the catalyst for the mountain of pending work and complexity that has accumulated

Yeah, I tried to fit in the `description` without breaking existing behavior and interfaces. If you want to overhaul the system as you imagine it to be, you could also take over from here.

kangalio:
By the way, thank you for investing the initial bulk of work for this PR! This PR was exceptionally well made, well documented, well integrated into the codebase. Thanks!

PR #2573, opened 2023-10-24
update async-tungstenite

### LINKED ISSUE #2550: Vulnerability discovered in Tungstenite
A vulnerability has been discovered in the version of Tungstenite used in this crate. Unfortunately, upgrading to the newest version causes this crate to no longer compile. Not sure what the best way forward is. Creating this issue to raise awareness and hopefully find a way to be able to upgrade. 

To generate the output below requires [cargo deny](https://crates.io/crates/cargo-deny) to be installed `cargo install cargo-deny`

Output of `cargo deny check advisories`

```
2023-10-04 22:42:09 [WARN] unable to find a config path, falling back to default config
error[vulnerability]: Tungstenite allows remote attackers to cause a denial of service
    ┌─ /home/user/serenity/Cargo.lock:137:1
    │
137 │ tungstenite 0.17.3 registry+https://github.com/rust-lang/crates.io-index
    │ ------------------------------------------------------------------------ security vulnerability detected
    │
    = ID: RUSTSEC-2023-0065
    = Advisory: https://rustsec.org/advisories/RUSTSEC-2023-0065
    = The Tungstenite crate through 0.20.0 for Rust allows remote attackers to cause
      a denial of service (minutes of CPU consumption) via an excessive length of an
      HTTP header in a client handshake. The length affects both how many times a parse
      is attempted (e.g., thousands of times) and the average amount of data for each
      parse attempt (e.g., millions of bytes).
    = Announcement: https://github.com/snapview/tungstenite-rs/issues/376
    = Solution: Upgrade to >=0.20.1
    = tungstenite v0.17.3
      └── async-tungstenite v0.17.2
          └── serenity v0.11.6

advisories FAILED

```

mh84:
Closes #2550 

I'm not sure if we should accept the default values for `write_buffer_size` and `max_write_buffer_size`, see https://docs.rs/tungstenite/latest/src/tungstenite/protocol/mod.rs.html#78 , so any feedback would be appreciated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,7 +92,7 @@ version = "1.1"
 default-features = false
 features = ["tokio-runtime"]
 optional = true
-version = "0.17"
+version = "0.23.0"
 
 [dependencies.typemap_rev]
 optional = true
--- a/src/internal/ws_impl.rs
+++ b/src/internal/ws_impl.rs
@@ -82,8 +82,7 @@ pub(crate) async fn create_client(url: Url) -> Result<WsStream> {
     let config = async_tungstenite::tungstenite::protocol::WebSocketConfig {
         max_message_size: None,
         max_frame_size: None,
-        max_send_queue: None,
-        accept_unmasked_frames: false,
+        ..Default::default()
     };
     let (stream, _) =
         async_tungstenite::tokio::connect_async_with_config(url, Some(config)).await?;


arqunis:
Thanks for the PR, but we've just released 0.11.7 and decided that's the final version we'll make for 0.11.x. Could you switch the base branch to the `next` branch instead?

mkrasnitski:
This will unfortunately have to wait until `0.12` is released, as we've just had our final release in the `0.11.x` series and are under code freeze on `current`. However, we are planning a patch v0.11.8 to be released after v0.12 in order to fix bugs without needing to rebase on top of them, so this can make it into that release.

mkrasnitski:
> Thanks for the PR, but we've just released 0.11.7 and decided that's the final version we'll make for 0.11.x. Could you switch the base branch to the `next` branch instead?

jinx :sweat_smile: Note that on `next` we've switched from `async-tungstenite` to `tokio-tungstenite`, so this PR doesn't make sense in that context. But, addressing the advisory is still something we should probably do on 0.11.

arqunis:
> jinx 😅 Note that on `next` we've switched from `async-tungstenite` to `tokio-tungstenite`, so this PR doesn't make sense in that context. But, addressing the advisory is still something we should probably do on 0.11.

Ah, right. I suppose we can leave the PR open as a reminder? Then once 0.12 is released and the 0.11.x branch is made, the PR can be retargetted to 0.11.x and 0.11.8 can be released as a bugfix.

mkrasnitski (approved):

PR #2575, opened 2023-10-27
Add CreateAttachment.description and EditAttachment

kangalio:
Summary: support attachment description, replace ad-hoc attachment edit methods with designated EditAttachments builder

Based on #2570

CreateAttachment is now Serialize and can be used for the `attachments` field directly, eliminating the extra `files` fields.

Attachment handling is now centralized in EditAttachments. This removes redundancy in serenity code and makes it more maintainable. Also, by modifying attachments through EditAttachments instead of ad-hoc methods on the builder itself, the side effects of adding attachments can now be documented much more clearly, as well as mirroring those side effects in the builder syntax (also see https://github.com/serenity-rs/serenity/pull/2570#discussion_r1373236965). The old ad-hoc methods are kept for backwards-compatibility, but marked deprecated

### DIFF
--- a/examples/testing/src/main.rs
+++ b/examples/testing/src/main.rs
@@ -59,7 +59,7 @@ async fn message(ctx: &Context, msg: Message) -> Result<(), serenity::Error> {
             )
             .await?;
         // Pre-PR, this falsely triggered a MODEL_TYPE_CONVERT Discord error
-        msg.edit(&ctx, EditMessage::new().add_existing_attachment(msg.attachments[0].id)).await?;
+        msg.edit(&ctx, EditMessage::new().attachments(EditAttachments::new_keep_all(&msg))).await?;
     } else if msg.content == "unifiedattachments" {
         let mut msg = channel_id.send_message(ctx, CreateMessage::new().content("works")).await?;
         msg.edit(ctx, EditMessage::new().content("works still")).await?;
@@ -72,9 +72,10 @@ async fn message(ctx: &Context, msg: Message) -> Result<(), serenity::Error> {
             .await?;
         msg.edit(
             ctx,
-            EditMessage::new()
-                .attachment(CreateAttachment::url(ctx, IMAGE_URL_2).await?)
-                .add_existing_attachment(msg.attachments[0].id),
+            EditMessage::new().attachments(
+                EditAttachments::new_keep_all(&msg)
+                    .add(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
+            ),
         )
         .await?;
     } else if msg.content == "ranking" {
@@ -240,9 +241,10 @@ async fn interaction(
         let msg = interaction
             .edit_response(
                 &ctx,
-                EditInteractionResponse::new()
-                    .new_attachment(CreateAttachment::url(ctx, IMAGE_URL_2).await?)
-                    .keep_existing_attachment(msg.attachments[0].id),
+                EditInteractionResponse::new().attachments(
+                    EditAttachments::new_keep_all(&msg)
+                        .add(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
+                ),
             )
             .await?;
 
@@ -253,8 +255,7 @@ async fn interaction(
             .edit_response(
                 &ctx,
                 EditInteractionResponse::new()
-                    .clear_existing_attachments()
-                    .keep_existing_attachment(msg.attachments[1].id),
+                    .attachments(EditAttachments::new().keep(msg.attachments[1].id)),
             )
             .await?;
     } else if interaction.data.name == "unifiedattachments1" {
@@ -291,8 +292,9 @@ async fn interaction(
         interaction
             .edit_response(
                 ctx,
-                EditInteractionResponse::new()
-                    .new_attachment(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
+                EditInteractionResponse::new().attachments(
+                    EditAttachments::new().add(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
+                ),
             )
             .await?;
 
--- a/examples/testing/src/main.rs
+++ b/examples/testing/src/main.rs
@@ -59,7 +59,7 @@ async fn message(ctx: &Context, msg: Message) -> Result<(), serenity::Error> {
             )
             .await?;
         // Pre-PR, this falsely triggered a MODEL_TYPE_CONVERT Discord error
-        msg.edit(&ctx, EditMessage::new().attachments(EditAttachments::new_keep_all(&msg))).await?;
+        msg.edit(&ctx, EditMessage::new().attachments(EditAttachments::keep_all(&msg))).await?;
     } else if msg.content == "unifiedattachments" {
         let mut msg = channel_id.send_message(ctx, CreateMessage::new().content("works")).await?;
         msg.edit(ctx, EditMessage::new().content("works still")).await?;
@@ -73,8 +73,7 @@ async fn message(ctx: &Context, msg: Message) -> Result<(), serenity::Error> {
         msg.edit(
             ctx,
             EditMessage::new().attachments(
-                EditAttachments::new_keep_all(&msg)
-                    .add(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
+                EditAttachments::keep_all(&msg).add(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
             ),
         )
         .await?;
@@ -242,7 +241,7 @@ async fn interaction(
             .edit_response(
                 &ctx,
                 EditInteractionResponse::new().attachments(
-                    EditAttachments::new_keep_all(&msg)
+                    EditAttachments::keep_all(&msg)
                         .add(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
                 ),
             )
--- a/examples/testing/src/main.rs
+++ b/examples/testing/src/main.rs
@@ -291,9 +291,8 @@ async fn interaction(
         interaction
             .edit_response(
                 ctx,
-                EditInteractionResponse::new().attachments(
-                    EditAttachments::new().add(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
-                ),
+                EditInteractionResponse::new()
+                    .new_attachment(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
             )
             .await?;
 
--- a/src/builder/create_attachment.rs
+++ b/src/builder/create_attachment.rs
@@ -5,6 +5,7 @@ use tokio::io::AsyncReadExt;
 #[cfg(feature = "http")]
 use url::Url;
 
+use crate::all::Message;
 #[cfg(feature = "http")]
 use crate::error::Error;
 use crate::error::Result;
@@ -12,61 +13,21 @@ use crate::error::Result;
 use crate::http::Http;
 use crate::model::id::AttachmentId;
 
-/// Enum that allows to add existing attachments and new attachments to the payload.
-#[derive(Clone, Debug, Serialize)]
-#[serde(untagged)]
-pub(crate) enum MessageAttachment {
-    Existing(ExistingAttachment),
-    New(NewAttachment),
-}
-
-/// [Discord docs] with the caveat at the top "For the attachments array in Message Create/Edit
-/// requests, only the id is required."
-///
-/// [Discord docs]: https://discord.com/developers/docs/resources/channel#attachment-object-attachment-structure
-#[derive(Clone, Debug, Serialize)]
-pub(crate) struct ExistingAttachment {
-    pub id: AttachmentId,
-    // TODO: add the other non-required attachment fields? Like content_type, description,
-    // ephemeral (ephemeral in particular seems pretty interesting)
-}
-
-/// Represents a new attachment in the payload, and allows for passing a description.
-#[derive(Clone, Debug, Serialize)]
-pub(crate) struct NewAttachment {
-    #[serde(rename = "id")]
-    pub index: u64,
-    pub filename: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub description: Option<String>,
-}
-
-impl MessageAttachment {
-    pub(crate) fn from_files(files: &[CreateAttachment]) -> Vec<Self> {
-        files
-            .iter()
-            .enumerate()
-            .map(|(i, file)| {
-                Self::New(NewAttachment {
-                    index: i as u64,
-                    filename: file.filename.clone(),
-                    description: file.description.clone(),
-                })
-            })
-            .collect()
-    }
-}
-
 /// Enum that allows a user to pass a [`Path`] or a [`File`] type to [`send_files`]
 ///
+/// [Discord docs](https://discord.com/developers/docs/resources/channel#attachment-object-attachment-structure).
+///
 /// [`send_files`]: crate::model::id::ChannelId::send_files
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize)]
 #[non_exhaustive]
 #[must_use]
 pub struct CreateAttachment {
-    pub data: Vec<u8>,
+    pub(crate) id: u64, // Placeholder ID will be filled in when sending the request
     pub filename: String,
     pub description: Option<String>,
+
+    #[serde(skip)]
+    pub data: Vec<u8>,
 }
 
 impl CreateAttachment {
@@ -76,6 +37,7 @@ impl CreateAttachment {
             data: data.into(),
             filename: filename.into(),
             description: None,
+            id: 0,
         }
     }
 
@@ -96,11 +58,7 @@ impl CreateAttachment {
             )
         })?;
 
-        Ok(CreateAttachment {
-            data,
-            filename: filename.to_string_lossy().to_string(),
-            description: None,
-        })
+        Ok(CreateAttachment::bytes(data, filename.to_string_lossy().to_string()))
     }
 
     /// Builds an [`CreateAttachment`] by reading from a file handler.
@@ -112,11 +70,7 @@ impl CreateAttachment {
         let mut data = Vec::new();
         file.try_clone().await?.read_to_end(&mut data).await?;
 
-        Ok(CreateAttachment {
-            data,
-            filename: filename.into(),
-            description: None,
-        })
+        Ok(CreateAttachment::bytes(data, filename))
     }
 
     /// Builds an [`CreateAttachment`] by downloading attachment data from a URL.
@@ -136,11 +90,7 @@ impl CreateAttachment {
             .and_then(Iterator::last)
             .ok_or_else(|| Error::Url(url.to_string()))?;
 
-        Ok(CreateAttachment {
-            data,
-            filename: filename.to_string(),
-            description: None,
-        })
+        Ok(CreateAttachment::bytes(data, filename))
     }
 
     /// Converts the stored data to the base64 representation.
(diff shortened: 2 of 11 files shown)


src/builder/create_attachment.rs
```diff
@@ -117,4 +106,174 @@ impl CreateAttachment {
...
+    pub fn keep(mut self, id: AttachmentId) -> Self {
+        self.new_and_existing_attachments.push(NewOrExisting::Existing(ExistingAttachment {
+            id,
+        }));
+        self
+    }
+
+    /// This method removes an existing attachment from the list of attachments that are kept after
+    /// editing.
+    pub fn dont_keep(mut self, id: AttachmentId) -> Self {
```
mkrasnitski: I think a better name would be `remove`, or maybe `drop`.

src/http/multipart.rs
```diff
@@ -5,11 +5,23 @@ use reqwest::multipart::{Form, Part};
...
+
 /// Holder for multipart body. Contains files, multipart fields, and payload_json for creating
 /// requests with attachments.
 #[derive(Clone, Debug)]
 pub struct Multipart {
-    pub files: Vec<CreateAttachment>,
+    /// Files that are sent with the form data as individual uploads.
+    pub upload_file: Option<CreateAttachment>,
+    /// Files that are sent with the form data as message attachments.
+    pub attachment_files: Option<Vec<CreateAttachment>>,
```
mkrasnitski: Do we still want to combine these into a single enum, since they seem to be mutually exclusive? Would remove the need for `CreateAttachment::into_part` as the part could be calculated inline using a `match`.

src/builder/create_attachment.rs
```diff
@@ -117,4 +106,174 @@ impl CreateAttachment {
...
+                        id: a.id,
+                    })
+                })
+                .collect(),
+        }
+    }
+
+    /// This method adds an existing attachment to the list of attachments that are kept after
+    /// editing.
+    pub fn keep(mut self, id: AttachmentId) -> Self {
```
mkrasnitski, line 228: What do you think of the name `retain`?

src/builder/create_attachment.rs
```diff
@@ -117,4 +106,174 @@ impl CreateAttachment {
...
+    ///
+    /// Shorthand for [`Self::new()`] and calling [`Self::keep()`] for every
+    /// [`AttachmentId`] in [`Message::attachments`].
+    ///
+    /// If you only want to keep a subset of attachments from the message, either implement this
+    /// method manually, or use [`Self::dont_keep()`].
+    ///
+    /// **Note: this EditAttachments must be run on the same message as is supplied here, or else
+    /// Discord will throw an error!**
+    pub fn new_keep_all(msg: &Message) -> Self {
```
mkrasnitski: The `new_` prefix is unnecessary. This could also be named `retain_all`, similar to `keep` above. What do you think?

src/builder/create_attachment.rs
```diff
@@ -117,4 +106,174 @@ impl CreateAttachment {
...
+    #[allow(clippy::should_implement_trait)] // Clippy thinks add == std::ops::Add::add
+    pub fn add(mut self, attachment: CreateAttachment) -> Self {
+        self.new_and_existing_attachments.push(NewOrExisting::New(attachment));
+        self
+    }
+
+    /// Clones all new attachments into a new Vec, keeping only data and filename, because those
+    /// are needed for the multipart form data. The data is taken out of `self` in the process, so
+    /// this method can only be called once.
+    pub(crate) fn take_files(&mut self) -> Vec<CreateAttachment> {
```
mkrasnitski, line 258: Could be renamed to just `take`, though not that important.

src/builder/create_attachment.rs
```diff
@@ -117,4 +106,174 @@ impl CreateAttachment {
...
+        let mut id_placeholder = 0;
```
mkrasnitski, line 259: This variable would be better named `index` or something related. Ideally you could just call `.filter_map().enumerate()` on the iterator instead of manually incrementing a counter.

src/builder/create_attachment.rs
```diff
@@ -117,4 +106,174 @@ impl CreateAttachment {
...
+        self.new_and_existing_attachments.retain(|a| match a {
+            NewOrExisting::Existing(a) if a.id == id => false,
+            _ => true,
+        });
+        self
+    }
+
+    /// Adds a new attachment to the attachment list.
+    #[allow(clippy::should_implement_trait)] // Clippy thinks add == std::ops::Add::add
+    pub fn add(mut self, attachment: CreateAttachment) -> Self {
```
mkrasnitski, line 250: Can't think of an alternative name, since `push` doesn't really make sense. Willing to discuss further though.

mkrasnitski (changes requested):

src/builder/create_attachment.rs
kangalio: Not a fan tbh. In Rust I associate it with `collection.retain(<closure>)`, which this is not. Also, I'm always a fan if method names match natural language, and I'd naturally say "keep an attachment" rather than "retain an attachment" (for example in the docs of that method)

src/builder/create_attachment.rs
kangalio: I went through a few iterations on the method name and deliberately decided to keep `new`, to make clear that this is a constructor method.

src/builder/create_attachment.rs
kangalio: I prefer `take_files` over `take`, but I'll not expand on that if it's not that important

src/builder/create_attachment.rs
kangalio: I deliberately did not name it `index` because that's semantically not what it is, as intended by Discord. It really is a placeholder for the ID, because Discord hasn't assigned yet. That's the "click" moment I talked about earlier in the Discord. I haven't tested, but I'm pretty sure the ID placeholder could be any numbers, not just sequential starting from 0

src/builder/create_attachment.rs
kangalio: Why an alternative name? I like `add` very much - short, simple, self-explanatory. The fact that clippy throws a fit is not a reason to discard that name in my opinion

src/http/multipart.rs
kangalio: Sure we can, I don't particularly care. I frankly forgot that idea was a thing while creating this follow-up PR

src/builder/create_attachment.rs
mkrasnitski: Are you saying that the id doesn't have to match the `files[N]` name of the new attachment?

kangalio: No, they do have to match, but I believe it doesn't have to be 0, 1, 2; it can be 42, 69, 420, 1337. So `files[1337]` in the multipart and `{"id": "1337"}` in the payload_json

src/builder/create_attachment.rs
mkrasnitski: That's true, but the signature is pretty short and the absence of `self` I think is easy enough to notice. Moreover, I think having the name be consistent with `keep` is important.

kangalio: Alright

src/builder/create_attachment.rs
mkrasnitski: Mainly because it's slightly ambiguous what you're adding if you only look at the method name, if it's existing or new attachments.

src/builder/create_attachment.rs
```diff
@@ -117,4 +106,174 @@ impl CreateAttachment {
...
+    New(CreateAttachment),
+    Existing(ExistingAttachment),
+}
+
+/// You can add new attachments and edit existing ones using this builder.
+///
+/// When this builder is _not_ supplied in a message edit, Discord keeps the attachments intact.
+/// However, as soon as a builder is supplied, Discord removes all attachments from the message.
+/// If you want to keep old attachments, you have to specify this for each attachment using
+/// [`Self.keep()`].
```
mkrasnitski:
```suggestion
/// However, as soon as a builder is supplied, Discord removes all attachments from the message. If
/// you want to keep old attachments, you must specify this either using [`Self::keep_all`], or
/// individually for each attachment using [`Self::keep`].
```

src/builder/create_attachment.rs
```diff
@@ -117,4 +106,174 @@ impl CreateAttachment {
...
+    /// Existing attachments are not kept by default, either. See [`Self::keep_all()`] or
+    /// [`Self::keep()`].
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Creates a new attachments builder that keeps all existing attachments.
+    ///
+    /// Shorthand for [`Self::new()`] and calling [`Self::keep()`] for every
+    /// [`AttachmentId`] in [`Message::attachments`].
```
mkrasnitski:
```suggestion
    /// Shorthand for [`Self::new()`] and calling [`Self::keep()`] for every [`AttachmentId`] in
    /// [`Message::attachments`].
```

src/builder/create_attachment.rs
mkrasnitski: Any objections to changing this?

examples/testing/src/main.rs
```diff
@@ -291,8 +291,9 @@ async fn interaction(
         interaction
             .edit_response(
                 ctx,
-                EditInteractionResponse::new()
-                    .new_attachment(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
+                EditInteractionResponse::new().attachments(
+                    EditAttachments::new().add(CreateAttachment::url(ctx, IMAGE_URL_2).await?),
+                ),
```
mkrasnitski: This is a bit wordy. Do you think some of the now-deprecated builder methods could instead be turned into wrappers around `.attachments`? That way, complexity is still centralized into `EditAttachments` but convenience methods are still left. I'd nominate `new_attachment` be kept at least.

mkrasnitski (changes requested):

src/builder/create_attachment.rs
kangalio: Your envisioned mental model of EditAttachments seems to be: two seperate lists of attachments, initially empty, that each can be pushed to. This more closely matches internals than my envisioned mental model of EditAttachments, which is: one list of new attachments that is initially empty and can be pushed to; and a set of all existing attachments, initially full but with all items set "inactive", where you don't push to the set but rather activate specific attachments by their ID (think `existing_attachments.set_to_keep(id, true);` rather than `existing_attachments.add_to_existing(id);`)

My idea is that the latter mental model is more intuitive to people who don't happen to already know how the API works under the hood. And that's also why I chose the terms "keep"/"dont_keep"

src/builder/create_attachment.rs
kangalio: In another comment I explained my envisioned mental model here. The idea is that existing attachments shouldn't be thought of as a pushable list, but a set with togglable items. Hence, semantically, the only pushable list is the list of new attachments and `add` is unambiguous

examples/testing/src/main.rs
kangalio: Good point. Maybe I can actually just un-deprecate all those methods, while carefully documenting what `EditAttachments` invocation they're equivalent to, so users can read about and grasp the unintuitive side effects of those methods.

Because the existing methods were already chosen carefully a few months ago to minimize side effect confusion while not devolving into wordiness, so their design is probably exactly what we want for "less wordier shorthands for EditAttachments"

src/builder/create_attachment.rs
mkrasnitski: I think the name `remove` still fits with this mental model, as it marks the existing attachment to be explicitly "removed" or otherwise "excluded" from the message. And I'd say that I do in fact view the existing attachments as a set and not as an empty list. I just really dislike the use of `dont` as part of the method name like this. It's normally a contraction but obviously you can't put an apostrophe in a method name (I'm not sure if that would make it a worse name or better tbh lol). As it stands it reads not seriously at all.

In principle, the pair of `keep`/`dont_keep` sounds too passive for me. If it were `keep`/`remove`, that would be better. The pair `keep`/`drop` has a nice symmetry, but `drop` is not a very conventional term, so I like it less. There may be some replacement for `keep` that's also less passive, but I can't think of one at the moment.

src/http/multipart.rs
mkrasnitski: I experimented a bit with names locally, and I kind of like this definition:
```rust
pub enum MultipartUpload {
    /// A file sent with the form data as an individual upload. For example, a sticker.
    File(CreateAttachment),
    /// Files sent with the form as message attachments.
    Attachments(Vec<CreateAttachment>),
}
```
Originally, if you remember, the proposed enum name was simply `Files`, but that seemed too generic to me. And, on its own the `File` variant may also seem generic but when written as `MultipartUpload::File`, I think it's pretty understandable. It only would occur in a single place anyways.

kangalio: I like it too! Nice idea

src/http/client.rs
```diff
@@ -26,6 +26,7 @@ use super::{
     HttpError,
     LightMethod,
     MessagePagination,
+    MultipartUpload,
```
mkrasnitski: This should be imported directly along with `super::multipart::Multipart`, a few lines above this.

src/http/client.rs
```diff
@@ -489,8 +490,7 @@ impl Http {
             request.body = Some(to_vec(map)?);
         } else {
             request.multipart = Some(Multipart {
-                attachment_files: Some(files.into_iter().map(Into::into).collect()),
-                upload_file: None,
+                upload: MultipartUpload::Attachments(files.into_iter().map(Into::into).collect()),
```
mkrasnitski: The `.into_iter().map(Into::into).collect()` is actually a no-op because `files` is already a `Vec<CreateAttachment>`. I won't duplicate this note for every other instance, but in all cases they could just take `files` directly.

src/http/multipart.rs
```diff
@@ -14,14 +14,19 @@ impl CreateAttachment {
...
+
 /// Holder for multipart body. Contains files, multipart fields, and payload_json for creating
 /// requests with attachments.
 #[derive(Clone, Debug)]
 pub struct Multipart {
-    /// Files that are sent with the form data as individual uploads.
-    pub upload_file: Option<CreateAttachment>,
-    /// Files that are sent with the form data as message attachments.
-    pub attachment_files: Option<Vec<CreateAttachment>>,
+    pub upload: MultipartUpload,
```
mkrasnitski, line 29: Should this still be called `files`? If not, we might want to adjust the docstring on `Multipart` somehow.

mkrasnitski (changes requested):

src/http/client.rs
kangalio: Lol

src/http/multipart.rs
```diff
@@ -5,11 +5,28 @@ use reqwest::multipart::{Form, Part};
...
+#[derive(Clone, Debug)]
+pub enum MultipartUpload {
+    /// A file sent with the form data as an individual upload. For example, a sticker.
+    File(CreateAttachment),
+    /// Files sent with the form as message attachments.
+    Attachments(Vec<CreateAttachment>),
+}
+
+/// Holder for multipart body. Contains upload data, multipart fields, and payload_json for creating
 /// requests with attachments.
```
mkrasnitski:
```suggestion
/// Holder for multipart body. Contains upload data, multipart fields, and payload_json for
/// creating requests with attachments.
```
Sorry for the nit

mkrasnitski:
It seems you undeprecated the old methods without actually changing them to wrap `EditAttachments` internally?

kangalio:
They already used EditAttachments internally, no? I changed it in the initial commit already

mkrasnitski:
~~As far as I can tell, the implementations on `EditMessage` and `EditWebhookMessage` are unchanged.~~ Also, the methods on `EditInteractionResponse` still contain `#[allow(deprecated)]` even though the wrapped methods are no longer deprecated.

EDIT: Nevermind, I see what you mean. I hadn't realized that they were already expressed in terms of `EditAttachments`.

src/builder/edit_interaction_response.rs
```diff
@@ -75,29 +76,30 @@ impl EditInteractionResponse {
...
+    }
+
+    /// Adds a new attachment to the message.
     ///
-    /// If this is called one or more times, existing attachments will reset. To keep them, provide
-    /// their IDs to [`Self::keep_existing_attachment`].
+    /// Resets existing attachments. See [`Self::keep_existing_attachment`] or read
+    /// [`EditAttachments`] for explanation.
     pub fn new_attachment(self, attachment: CreateAttachment) -> Self {
+        #[allow(deprecated)]
```
mkrasnitski:
```suggestion
    /// Resets existing attachments. See the documentation for [`EditAttachments`] for details.
    pub fn new_attachment(self, attachment: CreateAttachment) -> Self {
```

src/builder/edit_interaction_response.rs
```diff
@@ -75,29 +76,30 @@ impl EditInteractionResponse {
...
         Self(self.0.new_attachment(attachment))
     }
 
-    /// Keeps an existing attachment by id.
-    ///
-    /// To be used after [`Self::new_attachment`] or [`Self::clear_existing_attachments`].
+    /// Shorthand for [`EditAttachments::keep`]
     pub fn keep_existing_attachment(self, id: AttachmentId) -> Self {
+        #[allow(deprecated)]
```
mkrasnitski:
```suggestion
    /// Shorthand for [`EditAttachments::keep`].
    pub fn keep_existing_attachment(self, id: AttachmentId) -> Self {
```

src/builder/edit_interaction_response.rs
```diff
@@ -75,29 +76,30 @@ impl EditInteractionResponse {
...
 
-    /// Clears existing attachments.
-    ///
-    /// In combination with [`Self::keep_existing_attachment`], this can be used to selectively
-    /// keep only some existing attachments.
-    pub fn clear_existing_attachments(self) -> Self {
-        Self(self.0.clear_existing_attachments())
+    /// Shorthand for [`Self::attachments`] with [`EditAttachments::new`]
+    pub fn clear_attachments(self) -> Self {
+        #[allow(deprecated)]
```
mkrasnitski:
```suggestion
    /// Shorthand for calling [`Self::attachments`] with [`EditAttachments::new`].
    pub fn clear_attachments(self) -> Self {
```

src/builder/edit_message.rs
```diff
@@ -190,35 +187,40 @@ impl EditMessage {
...
+        self.attachments = Some(attachments);
+        self
+    }
+
+    /// Adds a new attachment to the message.
     ///
-    /// This can be called multiple times.
+    /// Resets existing attachments. See [`Self::keep_existing_attachment`] or read
+    /// [`EditAttachments`] for explanation.
     pub fn attachment(mut self, attachment: CreateAttachment) -> Self {
```
mkrasnitski:
```suggestion
    /// Resets existing attachments. See the documentation for [`EditAttachments`] for details.
    pub fn new_attachment(mut self, attachment: CreateAttachment) -> Self {
```
This method name isn't in line with the `new_attachment` method on `EditWebhookMessage`/`EditInteractionResponse`, and in combination with the new `attachments` method, might be confusing. I think it'd be better to change the name.

src/builder/edit_message.rs
```diff
@@ -190,35 +187,40 @@ impl EditMessage {
...
-                attachments.remove(attachment_index);
-            };
+        if let Some(attachments) = self.attachments {
+            self.attachments = Some(attachments.remove(id));
         }
         self
     }
 
-    /// Remove all attachments.
+    /// Shorthand for [`Self::attachments`] with [`EditAttachments::new`]
```
mkrasnitski:
```suggestion
    /// Shorthand for calling [`Self::attachments`] with [`EditAttachments::new`].
```

src/builder/edit_webhook_message.rs
```diff
@@ -133,33 +131,32 @@ impl EditWebhookMessage {
...
+        self.attachments = Some(attachments);
+        self
+    }
+
+    /// Adds a new attachment to the message.
     ///
-    /// If this is called one or more times, existing attachments will reset. To keep them, provide
-    /// their IDs to [`Self::keep_existing_attachment`].
+    /// Resets existing attachments. See [`Self::keep_existing_attachment`] or read
+    /// [`EditAttachments`] for explanation.
```
mkrasnitski:
```suggestion
    /// Resets existing attachments. See the documentation for [`EditAttachments`] for details.
```

src/builder/edit_webhook_message.rs
```diff
@@ -133,33 +131,32 @@ impl EditWebhookMessage {
...
         self
     }
 
-    /// Clears existing attachments.
-    ///
-    /// In combination with [`Self::keep_existing_attachment`], this can be used to selectively
-    /// keep only some existing attachments.
-    pub fn clear_existing_attachments(mut self) -> Self {
-        self.attachments = Some(Vec::new());
+    /// Shorthand for [`Self::attachments`] with [`EditAttachments::new`]
```
mkrasnitski:
```suggestion
    /// Shorthand for calling [`Self::attachments`] with [`EditAttachments::new`].
```

src/builder/edit_message.rs
```diff
@@ -190,35 +187,40 @@ impl EditMessage {
...
     pub fn attachment(mut self, attachment: CreateAttachment) -> Self {
-        self.files.push(attachment);
+        let attachments = self.attachments.get_or_insert_with(Default::default);
+        self.attachments = Some(std::mem::take(attachments).add(attachment));
         self
     }
 
-    /// Add an existing attachment by id.
+    /// Shorthand for [`EditAttachments::keep`]
     pub fn add_existing_attachment(mut self, id: AttachmentId) -> Self {
```
mkrasnitski:
```suggestion
    /// Shorthand for [`EditAttachments::keep`].
    pub fn keep_existing_attachment(mut self, id: AttachmentId) -> Self {
```
Same here as for `EditMessage::attachment`. The method name should be inline with `EditWebhookMessage`.

src/builder/edit_message.rs
```diff
@@ -190,35 +187,40 @@ impl EditMessage {
...
-            id,
-        });
+        let attachments = self.attachments.get_or_insert_with(Default::default);
+        self.attachments = Some(std::mem::take(attachments).keep(id));
         self
     }
 
-    /// Remove an existing attachment by id.
+    /// Shorthand for [`EditAttachments::remove`]
     pub fn remove_existing_attachment(mut self, id: AttachmentId) -> Self {
```
mkrasnitski:
```suggestion
    /// Shorthand for [`EditAttachments::remove`].
    pub fn remove_existing_attachment(mut self, id: AttachmentId) -> Self {
```
I've noticed there is no equivalent helper method on `EditWebhookMessage`. Maybe worth adding?

src/builder/edit_webhook_message.rs
```diff
@@ -133,33 +131,32 @@ impl EditWebhookMessage {
...
-        self.files.push(attachment);
+        let attachments = self.attachments.get_or_insert_with(Default::default);
+        self.attachments = Some(std::mem::take(attachments).add(attachment));
         self
     }
 
-    /// Keeps an existing attachment by id.
-    ///
-    /// To be used after [`Self::new_attachment`] or [`Self::clear_existing_attachments`].
+    /// Shorthand for [`EditAttachments::keep`]
```
mkrasnitski:
```suggestion
    /// Shorthand for [`EditAttachments::keep`].
```

mkrasnitski (changes requested):
Last round of review (hopefully). Sorry the review is a bit out of order.

src/builder/edit_message.rs
kangalio: I suppose. Serenity current doesn't have it so I don't really care. Can always be added later

src/builder/create_attachment.rs
```diff
@@ -117,4 +106,178 @@ impl CreateAttachment {
...
+/// you want to keep old attachments, you must specify this either using [`Self::keep_all`], or
+/// individually for each attachment using [`Self::keep`].
+///
+/// # Examples
+///
+/// ## Removing all attachments
+///
+/// ```rust,no_run
+/// # use serenity::all::*;
+/// # async fn _foo(ctx: Context, mut msg: Message) -> Result<(), Error> {
```
mkrasnitski:
```suggestion
/// # async fn _foo(ctx: Http, mut msg: Message) -> Result<(), Error> {
```
This fixes the CI failure with no gateway. The other doctests should also be changed in this way.

src/builder/create_attachment.rs
```diff
@@ -117,4 +106,178 @@ impl CreateAttachment {
...
+                cloned_attachment.id = id_placeholder;
+                files.push(cloned_attachment);
+
+                id_placeholder += 1;
+            }
+        }
+        files
+    }
+
+    pub(crate) fn is_empty(&self) -> bool {
```
mkrasnitski, line 281:
```suggestion
    #[cfg(feature = "cache")]
    pub(crate) fn is_empty(&self) -> bool {
```
Get an unused warning if cache is disabled.

examples/testing/src/main.rs
mkrasnitski: This should be reverted now that `new_attachment` is no longer deprecated.

mkrasnitski (changes requested):
Ok, for real this time. Thank you for being so patient.

mkrasnitski (approved):
Thank you for sticking with this. This PR is good to go.

mkrasnitski:
To summarize the breaking changes in this PR (for changelog purposes):
* Renamed the following methods:
    * `{EditInteractionResponse, EditWebhookMessage}::clear_existing_attachments` -> `clear_attachments`
    * `EditMessage::attachment` -> `new_attachment`
    * `EditMessage::add_existing_attachment` -> `keep_existing_attachment`
 * Replaced `Multipart::files` field with an `upload` field of type `MultipartUpload`.

PR #2576, opened 2023-10-27
 Add GuildId::everyone_role and clarify how to edit default channel permissions

junekept:
The goal of this PR is to make it more apparent on how you are meant to modify the default permissions of a channel. The current way (where you create a `RoleId` from the guild's ID) is very unintuitive and borderline impossible to figure out without an explanation from someone else with more experience with serenity. This PR adds a new method, `everyone_role`, to `GuildId`, which returns a `RoleId` using the guild's ID, thus giving you an `@everyone` role. Documentation is also added to `PermissionOverwriteType` to reflect this change and make it easier to figure out this behavior.

### DIFF
--- a/src/model/application/interaction.rs
+++ b/src/model/application/interaction.rs
@@ -82,7 +82,7 @@ impl Interaction {
     #[must_use]
     pub fn guild_locale(&self) -> Option<&str> {
         match self {
-            Self::Ping(i) => i.guild_locale.as_deref(),
+            Self::Ping(_) => None,
             Self::Command(i) | Self::Autocomplete(i) => i.guild_locale.as_deref(),
             Self::Component(i) => i.guild_locale.as_deref(),
             Self::Modal(i) => i.guild_locale.as_deref(),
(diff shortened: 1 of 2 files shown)


src/model/channel/mod.rs
```diff
@@ -342,6 +342,9 @@ pub struct PermissionOverwrite {
 ///
 /// This is for use with methods such as [`GuildChannel::create_permission`].
 ///
+/// If you would like to modify the default permissions of a channel, you can get it's [`RoleId`] from
```
arqunis:
```suggestion
/// If you would like to modify the default permissions of a channel, you can get its [`RoleId`] from
```

arqunis (changes requested):

src/model/guild/guild_id.rs
```diff
@@ -868,6 +868,12 @@ impl GuildId {
         Ok(roles.into_iter().map(|r| (r.id, r)).collect())
     }
 
+    /// Gets the default permission role (@everyone) from the guild.
+    #[inline]
+    pub fn everyone_role(&self) -> RoleId {
+        RoleId::from(self.0)
```
```suggestion
        RoleId::from(self.get())
```

arqunis (changes requested):

arqunis:
I wanted to push some commits before merging (format code & fix clippy warning), but git was being stupid; I've manually merged the PR: https://github.com/serenity-rs/serenity/commit/4361c1c9eaf4c31eda9410f3defd61ac34383a2b

PR #2582, opened 2023-11-10
fix faulty checks in `suppress_embeds`

fawni:
the previous implementation would return a `NotAuthor` error after trying to use `suppress_embeds` on a message sent by a different user, even if the current user has Manage Messages perm.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -796,16 +796,18 @@ impl Message {
         #[cfg(feature = "cache")]
         {
             if let Some(cache) = cache_http.cache() {
-                utils::user_has_perms_cache(
-                    cache,
-                    self.channel_id,
-                    self.guild_id,
-                    Permissions::MANAGE_MESSAGES,
-                )?;
-
-                if self.author.id != cache.current_user_id() {
-                    return Err(Error::Model(ModelError::NotAuthor));
-                }
+                if let Err(Error::Model(ModelError::InvalidPermissions(..))) =
+                    utils::user_has_perms_cache(
+                        cache,
+                        self.channel_id,
+                        self.guild_id,
+                        Permissions::MANAGE_MESSAGES,
+                    )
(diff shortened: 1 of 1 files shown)


src/model/channel/message.rs
```diff
@@ -796,16 +796,18 @@ impl Message {
...
+                        cache,
+                        self.channel_id,
+                        self.guild_id,
+                        Permissions::MANAGE_MESSAGES,
+                    )
+                {
+                    if self.author.id != cache.current_user_id() {
+                        return Err(Error::Model(ModelError::NotAuthor));
+                    }
+                };
```
mkrasnitski: I think the conditionals should be swapped in order if you're nesting them like this. Since, we don't care about permissions if we're the author. Also, if `user_has_perms_cache` returns an error that isn't `InvalidPermissions`, we should propagate it up as normal.

mkrasnitski:
I just realized this is targeting the `current` branch, which is in code freeze right now. This is a valid bugfix, but it'll have to wait until after 0.12 releases, so that we can include it in the `v0.11.8` patch release.

kangalio:
Is this such a critical fix that it warrants a hotfix 0.11 release? Otherwise, we could also retarget this PR to `next`

fawni:
i don't think it's _critical_ but it is a little inconvenient how it's not possible to suppress embeds currently

i would like to note that the only reason i targeted `current` is that i'm unfamiliar with serenity's workflow / release cycle😅

mkrasnitski:
Actually, `Message::suppress_embeds` doesn't exist on `next`. A workaround is to just create an `EditMessage::new().suppress_embeds(true)` and pass the builder to `Message::edit` or an equivalent method on `GuildId`/`Guild`.

mkrasnitski (approved):

arqunis (approved):

PR #2583, opened 2023-11-10
Move sending shard shutdown messages into the global shutdown lock

tazz4843:
Should fix a race condition that shows something like "Failed to cleanly shutdown shard N: Shutdown channel sent incorrect ID"

@GnomedDev

### DIFF
--- a/src/gateway/bridge/shard_manager.rs
+++ b/src/gateway/bridge/shard_manager.rs
@@ -277,10 +277,12 @@ impl ShardManager {
 
         info!("Shutting down shard {}", shard_id);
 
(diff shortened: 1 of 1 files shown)


kangalio:
Why is this still a draft PR?

kangalio (approved):

tazz4843:
@GnomedDev is still testing it as she's the only one who's hit this issue so far

GnomedDev:
Ah, I missed this because it was a draft PR. I thought it had already been merged. There don't seem to be any negative effects this could possibly have so I would merge so I pick it up next restart.

arqunis (approved):

PR #2585, opened 2023-11-11
Update gateway intents with up-to-date information

cheesycod:
Updates the gateway intents with up to date information. Also removes GUILD_BANS as it doesn't even exist anymore

### DIFF
--- a/src/model/gateway.rs
+++ b/src/model/gateway.rs
@@ -421,29 +421,34 @@ bitflags! {
     ///
     /// [Discord docs](https://discord.com/developers/docs/topics/gateway#list-of-intents).
     ///
-    /// [gateway intent]: https://discord.com/developers/docs/topics/gateway#privileged-intents
+    /// [Gateway Intents]: https://discord.com/developers/docs/topics/gateway#gateway-intents
     /// [Privileged Intents]: https://discord.com/developers/docs/topics/gateway#privileged-intents
-    /// [the bot must be verified]: https://support.discord.com/hc/en-us/articles/360040720412-Bot-Verification-and-Data-Whitelisting
+    /// [Caveats regarding intents]: https://discord.com/developers/docs/topics/gateway#caveats
+    /// [Bot Verification and Intent Whitelisting]: https://support.discord.com/hc/en-us/articles/360040720412-Bot-Verification-and-Data-Whitelisting
     pub struct GatewayIntents: u64 {
         /// Enables following gateway events:
-        /// - GUILD_CREATE
-        /// - GUILD_DELETE
-        /// - GUILD_ROLE_CREATE
-        /// - GUILD_ROLE_UPDATE
-        /// - GUILD_ROLE_DELETE
-        /// - CHANNEL_CREATE
-        /// - CHANNEL_UPDATE
-        /// - CHANNEL_DELETE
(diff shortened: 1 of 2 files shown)


mkrasnitski:
Do you think it might be better to simply deprecate `GUILD_BANS`, since it's basically just been renamed?

cheesycod:
> Do you think it might be better to simply deprecate `GUILD_BANS`, since it's basically just been renamed?

It was deprecated until now (that was what I earlier did). But, tbh I think it should be removed now so it can be one breaking release instead of doing it later in my opinion.

mkrasnitski:
I'm talking about specifically marking the constant as `#[deprecated]`, which isn't currently the case. I think that since it's just a name, it's not a big deal to mark it as deprecated and wait a release cycle to remove it, to give people time to change. Serenity already has some things which will are marked deprecated in 0.12 and will be removed in 0.13.

But, if you think most people won't notice, then breaking it immediately might be fine too.

cheesycod:
> I'm talking about specifically marking the constant as `#[deprecated]`, which isn't currently the case. I think that since it's just a name, it's not a big deal to mark it as deprecated and wait a release cycle to remove it, to give people time to change. Serenity already has some things which will are marked deprecated in 0.12 and will be removed in 0.13.
> 
> But, if you think most people won't notice, then breaking it immediately might be fine too.

True, I'm conflicted by that too. I do think that its a really simple change to make anyways and that most people won't really notice. I bet most just enable all non-priv intents anyways

src/model/gateway.rs
```diff
@@ -453,93 +458,87 @@ bitflags! {
...
-        /// This intent is also necessary to even receive the events in contains.
         const GUILD_MEMBERS = 1 << 1;
-        /// Enables following gateway events:
+        /// Enables the following gateway events:
         /// - GUILD_AUDIT_LOG_ENTRY_CREATE
         /// - GUILD_BAN_ADD
         /// - GUILD_BAN_REMOVE
         const GUILD_MODERATION = 1 << 2;
-        /// Backwards compatibility with old gateway event name. Same as GUILD_MODERATION
-        const GUILD_BANS = 1 << 2;
```
mkrasnitski, line 465: After some thought, I think the best course of action is to just keep the old name and mark it with ```#[deprecated = "use `GatewayIntents::GUILD_MODERATION` instead"]```.

cheesycod: Alright, will do in a bit

mkrasnitski:
Also, turns out `MANAGE_EMOJIS_AND_STICKERS` has also been renamed to `MANAGE_GUILD_EXPRESSIONS` (See [docs](https://discord.com/developers/docs/topics/permissions#permissions-bitwise-permission-flags).) Would be good to rename and deprecate the old name for this permission as well.

EDIT: I was confusing permissions and gateway intents. Ignore what I said; this change should be made separately since there are other related permissions changes it could go with.

cheesycod:
> gateway intents. Ignore what I said; this change should be made separately since there are other related permissions changes

Alright

mkrasnitski:
Just to be clear, I still think `GUILD_BANS` should be deprecated with this PR. Sorry for the confusion.

tazz4843:
Can be closed in favor of #2593

PR #2587, opened 2023-11-12
Add missing audit log events

crivasr:
There are some events that are missing from the audit log

PR #2588, opened 2023-11-12
Add missing audit log events

crivasr:
Duplicate of #2587 but on the correct branch


tazz4843:
You need to run `rustfmt` with nightly Rust, as serenity uses some nightly only features of rustfmt.

mkrasnitski:
> You need to run `rustfmt` with nightly Rust, as serenity uses some nightly only features of rustfmt.

It's just a whitespace change, no nightly required in this case.

mkrasnitski (approved):

arqunis (approved):

PR #2589, opened 2023-11-13
Bump all dependencies

Milo123459:
This bumps all the dependencies in `Cargo.toml` to their latest versions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,43 +18,43 @@ members = ["examples/*"]
 
 [dependencies]
 # Required dependencies
-bitflags = "2.4"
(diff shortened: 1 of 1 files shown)


tazz4843:
Ran this through `cargo public-api`, seems like no changes were introduced here. Regardless, this is targeting a breaking branch anyways so should be fine to merge.

tazz4843 (approved):
Ran this through `cargo public-api`, seems like no changes were introduced here. Regardless, this is targeting a breaking branch anyways so should be fine to merge.

arqunis (approved):

PR #2591, opened 2023-11-15
Add `reply_builder` function

LuckyMika:
I added a `reply_builder` function, that acts in a similar way to [`send_message`](https://github.com/serenity-rs/serenity/blob/current/src/model/channel/guild_channel.rs#L1040).
It is meant to make replying with embeds, files & other things that you would need a builder for, easier.

Please don't mind the 3 commits! My terminal/installation of git didn't properly accept the backtick character, so I had to do some fixing.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -5,6 +5,7 @@ use std::fmt::Display;
 #[cfg(all(feature = "cache", feature = "model"))]
 use std::fmt::Write;
 
+#[cfg(feature = "model")]
 use crate::builder::CreateMessage;
 #[cfg(all(feature = "model", feature = "utils"))]
 use crate::builder::{CreateEmbed, EditMessage};
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -5,6 +5,7 @@ use std::fmt::Display;
 #[cfg(all(feature = "cache", feature = "model"))]
 use std::fmt::Write;
 
+use crate::builder::CreateMessage;
 #[cfg(all(feature = "model", feature = "utils"))]
 use crate::builder::{CreateEmbed, EditMessage};
 #[cfg(all(feature = "cache", feature = "model"))]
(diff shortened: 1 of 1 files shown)


mkrasnitski:
The `current` branch is under code freeze at the moment as we prepare for the release of serenity 0.12. The code for that lives on the `next` branch. In serenity 0.12, builders have been reworked so that you no longer pass in a closure, but instead a concrete builder object. In other words, `send_message` now takes a `CreateMessage` as argument instead of a `FnOnce(&mut CreateMessage) -> &mut CreateMessage`.

If you'd like to rebase your work and target this PR at the `next` branch, feel free. Since we're rapidly approaching release, there's no guarantee it'll make it in in time for 0.12, but since this won't be a breaking change, it can make it into a minor release (e.g. 12.1).

kangalio:
Do I see correctly that this is just a shorthand for `send_message`, with `reference_message` and `allowed_mentions` already set?

LuckyMika:
Yes it is, but it is shorter than getting the channel and calling send_message on that and there are a lot of those util functions in serenity and I this does some more stuff instead of just adding a reference message directly. I guess it would be possible to just inline this one but I do believe it is helpful

mkrasnitski:
This change could be made more concise by changing the signature of `Message::_reply` to take a builder, and then expressing all other wrappers in terms of it. Then, `reply_builder` could become a thin wrapper around `_reply`.

src/model/channel/message.rs
```diff
@@ -5,6 +5,7 @@ use std::fmt::Display;
 #[cfg(all(feature = "cache", feature = "model"))]
 use std::fmt::Write;
 
+use crate::builder::CreateMessage;
```
mkrasnitski, line 9:
```suggestion
#[cfg(feature = "model")]
use crate::builder::CreateMessage;
```

arqunis:
Closing this for no activity. You may re-open if you still wish to add this, but you'll have to retarget to 0.12 and incorporate the change suggested by @mkrasnitski.

PR #2592, opened 2023-11-16
Add experimental typesize support

GnomedDev:
Adds a simple integration with my new [`typesize`](https://docs.rs/typesize) library to help users find which caches are using up the most ram. This is highly experimental, hence the lack of documentation, but could be seriously useful.

### DIFF
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -47,6 +47,8 @@ mod event;
 mod settings;
 mod wrappers;
 
+#[cfg(feature = "temp_cache")]
+pub(crate) use wrappers::MaybeOwnedArc;
 use wrappers::{BuildHasher, MaybeMap, ReadOnlyMapRef};
 
 type MessageCache = DashMap<ChannelId, HashMap<MessageId, Message>, BuildHasher>;
@@ -75,8 +77,8 @@ impl<'a, K, V, T> CacheRef<'a, K, V, T> {
     }
 
     #[cfg(feature = "temp_cache")]
-    fn from_arc(inner: Arc<V>) -> Self {
-        Self::new(CacheRefInner::Arc(inner))
+    fn from_arc(inner: MaybeOwnedArc<V>) -> Self {
+        Self::new(CacheRefInner::Arc(inner.get_inner()))
     }
 
     fn from_ref(inner: Ref<'a, K, V, BuildHasher>) -> Self {
@@ -120,6 +122,7 @@ pub type GuildChannelsRef<'a> = MappedGuildRef<'a, HashMap<ChannelId, GuildChann
 pub type ChannelMessagesRef<'a> = CacheRef<'a, ChannelId, HashMap<MessageId, Message>>;
 pub type MessageRef<'a> = CacheRef<'a, ChannelId, Message, HashMap<MessageId, Message>>;
 
+#[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
 #[derive(Debug)]
 pub(crate) struct CachedShardData {
     pub total: u32,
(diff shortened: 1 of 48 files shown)


tazz4843 (approved):
Looks fine, besides the lack of docs but as you said experimental crate. Seems like a super useful feature esp with the cache metrics.

src/internal/macros.rs
```diff
@@ -157,10 +157,11 @@ macro_rules! bitflags {
             )*
         }
     ) => {
+        $(#[$outer])*
+        $vis struct $BitFlags($T);
+
         bitflags::bitflags! {
-            #[derive(Copy, Clone, Debug, Eq, Hash, PartialEq)]
```
mkrasnitski, line 161: Why remove this auto-derive of traits? You missed decorating `InteractionResponseFlags` with `#[derive(TypeSize)]` but also didn't add the other derived traits back to it.

GnomedDev: I removed the auto derive as it is against the documentation of the macro (it doesn't just forward through) and is more clear when implemented on each type. I'll fix that mess up with InteractionResponseFlags.

mkrasnitski: In what sense does it not forward through?

GnomedDev: It does, but it also adds the derives undocumented.

GnomedDev: I also didn't miss decorating InteractionResponseFlags, it's just not used anywhere apparently, looks like a separate PR is in order.

mkrasnitski: > It does, but it also adds the derives undocumented.

You mean in the sense that the definition site of the type inside serenity doesn't include these? I guess that makes sense.

src/internal/macros.rs
```diff
@@ -157,10 +157,11 @@ macro_rules! bitflags {
...
-            $(#[$outer])*
-            $vis struct $BitFlags: $T {
+            impl $BitFlags: $T {
```
mkrasnitski, line 164: How come this was changed to declare the struct outside the call to the macro? This also goes for `Permissions`, which doesn't use the internal macro because of conflicting serde impls.

GnomedDev: As otherwise bitflags generates weirdness inside the struct, instead of just the u64. This trips up the typesize derive and it seems like all the others are manually handled inside bitflags.

mkrasnitski: After running `cargo expand`, I see the difference is that if declaring the struct inline, you get:
```rust
pub struct MessageFlags(<MessageFlags as ::bitflags::__private::PublicFlags>::Internal)
```
whereas doing it like this just gets you `pub struct MessageFlags(u64)` as normal. However, maybe instead you could add manual impls of `TypeSize` to the structs instead? It would mean that `Permissions` also gets a separate manual impl. Or as an alternative, get the `typesize` crate to work with associated types, or whatever the actual limitation here is. The `PublicFlags` trait I think is publically exported (despite the `__private` name there), so I don't think visibility is the problem.

src/cache/mod.rs
```diff
@@ -54,6 +54,17 @@ use wrappers::{BuildHasher, MaybeMap, ReadOnlyMapRef};
 
 type MessageCache = DashMap<ChannelId, HashMap<MessageId, Message>, BuildHasher>;
 
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug)]
+pub struct CacheInfo {
+    /// The name of the cahce
```
tazz4843:
```suggestion
    /// The name of the cache
```

src/internal/macros.rs
GnomedDev: That would not be possible. I cannot blanket implementation
```rs
impl<F> TypeSize for F
where
    F: bitflags::Flags,
    F::Bits: TypeSize,
{
}
```
as this leads to orphan rule stuff where bitflags could technically add an implementation of Flags for `Option` or `Vec` and break typesize.

I cannot "fix" the derive in any way either to handle the internal type as that would require deriving TypeSize on the private bitflags internal type.

I could add a custom implementation in serenity, but that seems unnecessary when it currently works fine.

GnomedDev:
Just pushed an update to get rid of the manual `get_statistics` method. That is now automatically derived by the macro for all the structs annotated.

arqunis (approved):

PR #2594, opened 2023-11-17
[Voice Model] Get rid of enum_primitive dep

GnomedDev:
`enum_primitive` is unmaintained and causes duplicate `num-traits` dependencies, and just a simple macro crate.

### DIFF
--- a/voice-model/Cargo.toml
+++ b/voice-model/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 
 [dependencies]
 bitflags = "2.4"
-enum_primitive = "0.1.1"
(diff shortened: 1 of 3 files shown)


GnomedDev:
This is technically breaking by the way, as we are now re-exporting num-traits 0.2 instead of 0.1, but songbird isn't broken by it and I think that's the only dependency on voice-model.

kangalio (approved):
Nice, replacing an abandoned 2017 crate with the up-to-date ecosystem cornerstone that it only re-exported anyways

arqunis (approved):

PR #2598, opened 2023-11-17
Actually use InteractionResponseFlags

GnomedDev:
Noticed in #2592 that this type is never actually used.


mkrasnitski:
Should `CreateInteractionResponseFollowup::flags` also have its type changed?

GnomedDev:
I don't think so, since that's closer to a webhook message than a interaction response, but I'm not sure. This PR makes a positive impact on serenity as is so should be merged.

tazz4843 (approved):

kangalio (approved):
(Could argue about whether InteractionResponseFlags should exist, since Discord docs don't show it as a separate type from MessageFlags either, but it doesn't hurt to keep I suppose)

Issue #2480, opened 2023-07-01
feature request: add utility types/functions for creating/parsing discord timestamps


vidhanio:
```
<t:UNIX_TIMESTAMP:FLAG>
```

Available flags:

```
t: Short time (e.g 9:41 PM)
T: Long Time (e.g. 9:41:30 PM)
d: Short Date (e.g. 30/06/2021)
D: Long Date (e.g. 30 June 2021)
f (default): Short Date/Time (e.g. 30 June 2021 9:41 PM)
F: Long Date/Time (e.g. Wednesday, June, 30, 2021 9:41 PM)
R: Relative Time (e.g. 2 months ago, in an hour)
```

it would be nice to have convenience types/functions to create these.

PR #2600, opened 2023-11-17
feat:  Introduce `FormattedTimestamp` and `FormattedTimestampStyle` for improved message timestamp formatting

darkyeg:
### Overview

This PR introduces a new feature: the `FormattedTimestamp` struct and `FormattedTimestampStyle` enum. These additions provide a flexible way to represent and format timestamps in messages within the Serenity framework.

### Changes Made

- Added `FormattedTimestamp` struct to represent a combination of a timestamp and a formatting style.
- Introduced `FormattedTimestampStyle` enum to define various styles for formatting time in messages.
- Implemented relevant traits for ease of use, including `Clone`, `Copy`, `PartialEq`, `Eq`, `Debug`, `Hash`, `Display`, and parsing from strings.
- Included methods for creating instances, accessing timestamp and style, and displaying formatted time strings.

#### Creating a FormattedTimestamp instance:

```rust
use your_crate_name::{FormattedTimestamp, FormattedTimestampStyle, Timestamp};

let timestamp = Timestamp::now();
let message_time = FormattedTimestamp::new(timestamp, Some(FormattedTimestampStyle::ShortDateTime));
```

#### Displaying formatted time string:

```rust
println!("{}", message_time); // Outputs: <t:timestamp:f>
```

#### Parsing from a string:

```rust
let time_str = "<t:timestamp:f>";
let parsed_time: Result<FormattedTimestamp, FormattedTimestampParseError> = time_str.parse();
```

### Related Issue

Closes: #2480  

### Tests

Included comprehensive tests covering the functionality of `FormattedTimestamp` and `FormattedTimestampStyle`.

### Additional Notes

This feature addresses the request raised in issue #2480, providing enhanced functionality for formatting timestamps in messages.

Please review and provide feedback.

---

Thank you!


kangalio (changes requested):
There's a few rough edges (in my opinion) I'd like to remove:

- rename MessageTime -> FormattedTimestamp (and MessageTime[Style, ParseError...] accordingly)
  - nicely consistent with Timestamp
- change MessageTime from `(i64, MessageTimeStyle)` to `{ timestamp: Timestamp, style: MessageTimeStyle }`
- change MessageTimeParseError Display impl to be lowercase, to fit Rust standard rules
- inline the redundant 14 `let short_time = MessageTimeStyle::ShortTime` etc. variables

Is that ok? If so, I'd make those changes right into this PR

darkyeg:
> There's a few rough edges (in my opinion) I'd like to remove:
> 
> * rename MessageTime -> FormattedTimestamp (and MessageTime[Style, ParseError...] accordingly)
>   
>   * nicely consistent with Timestamp
> * change MessageTime from `(i64, MessageTimeStyle)` to `{ timestamp: Timestamp, style: MessageTimeStyle }`
> * change MessageTimeParseError Display impl to be lowercase, to fit Rust standard rules
> * inline the redundant 14 `let short_time = MessageTimeStyle::ShortTime` etc. variables
> 
> Is that ok? If so, I'd make those changes right into this PR

Currently, there is an issue with the Eq implementation when Timestamp is used as an instance of i64. The problem arises because two instances of Timestamp may have the same underlying i64 value but differ in their Datetime representation. This discrepancy can lead to unexpected behavior when comparing instances.

kangalio:
In general though, this PR seems great! I was skeptical about creating a newtype for formatted timestamps, but I take it back, it makes total sense, particularly when considering that it can also be parsed into. And you gave a lot of attention to detail in this PR, it looks like, thank you

kangalio:
> Currently, there is an issue with the Eq implementation when Timestamp is used as an instance of i64. The problem arises because two instances of Timestamp may have the same underlying i64 value but **differ in their Datetime representation**. This discrepancy can lead to unexpected behavior when comparing instances.

What are you referring to? Do you mean the fact that it uses either `chrono` or `time` backend? Since that is a compile time flag, all timestamps in any given serenity program will use consistently either `chrono` OR `time`, never both of them mixed, so that should be a non-issue

darkyeg:
> What are you referring to? Do you mean the fact that it uses either `chrono` or `time` backend? Since that is a compile time flag, all timestamps in any given serenity program will use consistently either `chrono` OR `time`, never both of them mixed, so that should be a non-issue

This is an example of what I mean:

```rs
  fn test() {
        let timestamp = MessageTime::now();

        let time_str = timestamp.to_string();
        println!("{}", time_str); 
        let time_parsed = time_str.parse::<MessageTime>().unwrap();

        println!("{}", time_parsed);

        assert_eq!(timestamp.timestamp(), time_parsed.timestamp());
    }
 ```
 ```log
 <t:1700235250:f>
<t:1700235250:f>
thread 'model::misc::tests::test' panicked at src\model\misc.rs:514:9:
assertion `left == right` failed
  left: Timestamp(2023-11-17T15:34:10.990466300Z)
 right: Timestamp(2023-11-17T15:34:10Z)
 ```

**Comment by kangalio** [comment]
Ah I see, the issue is that sub-second precision is compared too, which is unintuitive. That's a good point. Apart from that, are the changes I mentioned ok? (I'm asking because the usual back and forth on these PRs "can you please do that" "ok" "no i meant like that" "ah ok" has huge amount of useless overhead, so I'd like to experiment with a less bureaucratic approach)

**Comment by darkyeg** [comment]
> Ah I see, the issue is that sub-second precision is compared too, which is unintuitive. That's a good point. Apart from that, are the changes I mentioned ok? (I'm asking because the usual back and forth on these PRs "can you please do that" "ok" "no i meant like that" "ah ok" has huge amount of useless overhead, so I'd like to experiment with a less bureaucratic approach)

Yes, there are no problems. I felt that the name was inaccurate and needed to be changed.

**Comment by kangalio** [comment]
FormattedTimestampParseError is really not forwards compatible; it has these extremely specific three error cases (will any serenity user ever actually match on those?) with no possibility of adding or removing or changing anything at all, except breaking changes. Is it ok if I change FormattedTimestampParseError to not expose its variants? It can always be changed back in the future without a breaking change

**Comment by darkyeg** [comment]
> FormattedTimestampParseError is really not forwards compatible; it has these extremely specific three error cases (will any serenity user ever actually match on those?) with no possibility of adding or removing or changing anything at all, except breaking changes. Is it ok if I change FormattedTimestampParseError to not expose its variants? It can always be changed back in the future without a breaking change

I think it is unnecessary to explain in detail what the parsing error is.

**Comment by kangalio** [comment]
Oh and I'm realizing this PR doesn't support the non-styled formatted timestamp at all?
image

Am I missing something, is this intended?

**Comment by darkyeg** [comment]
> Oh and I'm realizing this PR doesn't support the non-styled formatted timestamp at all? image
> 
> Am I missing something, is this intended?

I didn't intend that

**Comment by darkyeg** [comment]
> Oh and I'm realizing this PR doesn't support the non-styled formatted timestamp at all? image
> 
> Am I missing something, is this intended?

Should I add it or you?

**Comment by kangalio** [comment]
While thinking how to avoid the annoying `#[cfg(all(feature = "model", feature = "utils"))]` spewed everywhere, I noticed: wouldn't this fit much better into `serenity::utils`?

**Comment by darkyeg** [comment]
> While thinking how to avoid the annoying `#[cfg(all(feature = "model", feature = "utils"))]` spewed everywhere, I noticed: wouldn't this fit much better into `serenity::utils`?

I think so, I don't know inside the library 100%.

**Review by kangalio — APPROVED** [review]

**Comment by kangalio** [comment]
Could you @darkyeg or someone else sign off on my changes to the PR?

**Review by jamesbt365 — APPROVED** [review]
nothing obviously wrong sticks out over a quick runthrough _the commit count though eeee_

**Comment by kangalio** [comment]
> nothing obviously wrong sticks out over a quick runthrough the commit count though eeee

Why is that bad? I thought granular small commits were a _good_ thing in PRs, seeing as they're squashed when merging anyways

File: src/utils/formatted_timestamp.rs [hunk 1]
```diff
@@ -0,0 +1,243 @@
+use std::error::Error as StdError;
+use std::fmt::{self, Display};
```
mkrasnitski:
```suggestion
use std::fmt;
```
Minor nit - elsewhere in the lib, we choose to impl `fmt::Display` rather than importing the trait directly.

mkrasnitski (changes requested):

kangalio:
Thanks for reviewing @mkrasnitski and @jamesbt365! I decided not to wait for a review from you @darkyeg because of the other reviews, I hope it was okay. And thank you for submitting this PR and doing most of the work!

jamesbt365:
> Why is that bad? I thought granular small commits were a good thing in PRs, seeing as they're squashed when merging anyways

eh nothing is really bad about high commit counts I'm just not personally a fan of having such a high commit count for something that overall doesn't look too complex, just a stylistic choice personally, nothing *actually* wrong with the extra verbosity :P

PR #2603, opened 2023-11-18
Remove PrivateChannel from caching and events

GnomedDev:
The entire `PrivateChannel` dashmap was never used, discord hasn't supplied events for it since API v8, and all the methods handling it were just unneeded. This allows for removal of `Channel` from a lot of the API surface, replacing it with the specific `GuildChannel` or specifically using `PrivateChannel`.

This was motivated by the `allow(clippy::large_enum_variant)` on `Channel` making me reconsider if it is strictly needed, which it is for a few endpoints (http get channel can return private channels), but was able to be removed from a lot of endpoints.

### DIFF
--- a/src/builder/create_invite.rs
+++ b/src/builder/create_invite.rs
@@ -28,7 +28,7 @@ use crate::model::prelude::*;
 /// impl EventHandler for Handler {
 ///     async fn message(&self, context: Context, msg: Message) {
 ///         if msg.content == "!createinvite" {
-///             let channel_opt = context.cache.guild_channel(msg.channel_id).as_deref().cloned();
+///             let channel_opt = context.cache.channel(msg.channel_id).as_deref().cloned();
 ///             let channel = match channel_opt {
 ///                 Some(channel) => channel,
 ///                 None => {
@@ -118,7 +118,7 @@ impl<'a> CreateInvite<'a> {
     /// # #[cfg(all(feature = "cache", feature = "client", feature = "framework", feature = "http"))]
     /// # #[command]
     /// # async fn example(context: &Context) -> CommandResult {
-    /// # let channel = context.cache.guild_channel(81384788765712384).unwrap().clone();
+    /// # let channel = context.cache.channel(81384788765712384).unwrap().clone();
     /// let builder = CreateInvite::new().max_age(3600);
     /// let invite = channel.create_invite(context, builder).await?;
     /// # Ok(())
@@ -150,7 +150,7 @@ impl<'a> CreateInvite<'a> {
     /// # #[cfg(all(feature = "cache", feature = "client", feature = "framework", feature = "http"))]
     /// # #[command]
     /// # async fn example(context: &Context) -> CommandResult {
-    /// # let channel = context.cache.guild_channel(81384788765712384).unwrap().clone();
+    /// # let channel = context.cache.channel(81384788765712384).unwrap().clone();
     /// let builder = CreateInvite::new().max_uses(5);
     /// let invite = channel.create_invite(context, builder).await?;
     /// # Ok(())
@@ -180,7 +180,7 @@ impl<'a> CreateInvite<'a> {
     /// # #[cfg(all(feature = "cache", feature = "client", feature = "framework", feature = "http"))]
     /// # #[command]
     /// # async fn example(context: &Context) -> CommandResult {
-    /// # let channel = context.cache.guild_channel(81384788765712384).unwrap().clone();
+    /// # let channel = context.cache.channel(81384788765712384).unwrap().clone();
     /// let builder = CreateInvite::new().temporary(true);
     /// let invite = channel.create_invite(context, builder).await?;
     /// # Ok(())
@@ -210,7 +210,7 @@ impl<'a> CreateInvite<'a> {
     /// # #[cfg(all(feature = "cache", feature = "client", feature = "framework", feature = "http"))]
     /// # #[command]
     /// # async fn example(context: &Context) -> CommandResult {
-    /// # let channel = context.cache.guild_channel(81384788765712384).unwrap().clone();
+    /// # let channel = context.cache.channel(81384788765712384).unwrap().clone();
     /// let builder = CreateInvite::new().unique(true);
     /// let invite = channel.create_invite(context, builder).await?;
     /// # Ok(())
@@ -265,10 +265,7 @@ impl<'a> CreateInvite<'a> {
 #[cfg(feature = "http")]
 #[async_trait::async_trait]
 impl<'a> Builder for CreateInvite<'a> {
-    #[cfg(feature = "cache")]
-    type Context<'ctx> = (ChannelId, Option<GuildId>);
-    #[cfg(not(feature = "cache"))]
-    type Context<'ctx> = (ChannelId,);
+    type Context<'ctx> = ChannelId;
     type Built = RichInvite;
 
     /// Creates an invite for the given channel.
@@ -289,15 +286,10 @@ impl<'a> Builder for CreateInvite<'a> {
         #[cfg(feature = "cache")]
         {
             if let Some(cache) = cache_http.cache() {
-                crate::utils::user_has_perms_cache(
-                    cache,
-                    ctx.0,
-                    ctx.1,
-                    Permissions::CREATE_INSTANT_INVITE,
-                )?;
+                crate::utils::user_has_perms_cache(cache, ctx, Permissions::CREATE_INSTANT_INVITE)?;
             }
         }
 
-        cache_http.http().create_invite(ctx.0, &self, self.audit_log_reason).await
+        cache_http.http().create_invite(ctx, &self, self.audit_log_reason).await
     }
 }
--- a/src/builder/edit_channel.rs
+++ b/src/builder/edit_channel.rs
@@ -294,10 +294,7 @@ impl<'a> EditChannel<'a> {
 #[cfg(feature = "http")]
 #[async_trait::async_trait]
 impl<'a> Builder for EditChannel<'a> {
-    #[cfg(feature = "cache")]
-    type Context<'ctx> = (ChannelId, Option<GuildId>);
-    #[cfg(not(feature = "cache"))]
-    type Context<'ctx> = (ChannelId,);
+    type Context<'ctx> = ChannelId;
     type Built = GuildChannel;
 
     /// Edits the channel's settings.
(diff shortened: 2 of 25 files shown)


mkrasnitski:
This is a rather large change to review this close to 0.12. Even though it's breaking, would you be alright if it was targeted for 0.13 instead? Or is this something you think should absolutely go into 0.12?

GnomedDev:
Knowing serenity's release times, I want this to be usable by users this decade, not next decade.

GnomedDev:
In a more serious note, it's misleading that there is a private channels cache at all, it hasn't worked since serenity bumped past v8, that's when this change should have been made.

src/builder/create_invite.rs
```diff
@@ -284,20 +281,19 @@ impl<'a> Builder for CreateInvite<'a> {
     async fn execute(
         self,
         cache_http: impl CacheHttp,
-        ctx: Self::Context<'_>,
+        channel: Self::Context<'_>,
```
mkrasnitski:
```suggestion
        ctx: Self::Context<'_>,
```
All other builders still name this param `ctx`.

src/builder/edit_channel.rs
```diff
@@ -315,28 +312,18 @@ impl<'a> Builder for EditChannel<'a> {
     async fn execute(
         self,
         cache_http: impl CacheHttp,
-        ctx: Self::Context<'_>,
+        channel: Self::Context<'_>,
```
mkrasnitski:
```suggestion
        ctx: Self::Context<'_>,
```
Ditto.

src/cache/mod.rs
```diff
@@ -151,7 +149,7 @@ pub struct Cache {
     ///
     /// The TTL for each value is configured in CacheSettings.
     #[cfg(feature = "temp_cache")]
-    pub(crate) temp_channels: MokaCache<ChannelId, GuildChannel, BuildHasher>,
+    pub(crate) temp_channels: MokaCache<ChannelId, Arc<GuildChannel>, BuildHasher>,
```
mkrasnitski, line 152: Why was this changed?

mkrasnitski:
What did you use as reference when figuring out when only one of `PrivateChannel` or `GuildChannel` is ever used?

GnomedDev:
Can you ask for a specific method? Most of them are quite obvious, as Webhooks cannot be created in private channels and private channels cannot be cached.

src/model/channel/channel_id.rs
```diff
@@ -415,7 +407,7 @@ impl ChannelId {
         {
             if let Some(cache) = cache_http.cache() {
                 if let Channel::Guild(guild_channel) = &channel {
-                    cache.temp_channels.insert(guild_channel.id, guild_channel.clone());
+                    cache.temp_channels.insert(guild_channel.id, Arc::new(guild_channel.clone()));
```
mkrasnitski, line 410: If `temp_channels` was changed to use `Arc`, why do we still need to clone when we insert into the cache? Is it because the original channel gets dropped here?

src/model/channel/guild_channel.rs
```diff
@@ -354,20 +350,16 @@ impl GuildChannel {
...
-                    Some(self.guild_id),
-                    Permissions::MANAGE_CHANNELS,
-                )?;
+                crate::utils::user_has_perms_cache(cache, self.id, Permissions::MANAGE_CHANNELS)?;
             }
         }
 
-        self.id.delete(cache_http.http()).await
+        let channel = self.id.delete(cache_http.http()).await?;
+        channel.guild().ok_or(Error::Model(ModelError::InvalidChannelType))
```
mkrasnitski, line 362: The error here is technically unreachable (we're a `GuildChannel` here), but meh, probably fine.

src/model/channel/private_channel.rs
```diff
@@ -69,8 +69,8 @@ impl PrivateChannel {
     /// closing a private channel on the client, which can be re-opened.
     #[allow(clippy::missing_errors_doc)]
     #[inline]
-    pub async fn delete(&self, http: impl AsRef<Http>) -> Result<Channel> {
-        self.id.delete(http).await
+    pub async fn delete(&self, http: impl AsRef<Http>) -> Result<PrivateChannel> {
+        self.id.delete(http).await?.private().ok_or(Error::Model(ModelError::InvalidChannelType))
```
mkrasnitski, line 73: Same here as for `GuildChannel`, we can probably assert that this won't ever produce an invalid channel.

src/model/webhook.rs
```diff
@@ -191,28 +194,32 @@ impl WebhookChannel {
...
+                            .insert(guild_channel.id, Arc::new(guild_channel.clone()));
+                    }
                 }
-            }
-        }
 
-        Ok(channel)
+                Ok(guild_channel)
+            },
+            other => Err(Error::Model(ModelError::NonGuildWebhookChannel(Box::new(other)))),
```
mkrasnitski:
```suggestion
            other => Err(Error::Model(ModelError::InvalidChannelType)),
```
Couldn't we just use `InvalidChannelType` here and call it a day? If it's not possible to make a webhook for a private channel, it's not worth creating an error variant that will never realistically be constructed at runtime.

src/cache/event.rs
```diff
@@ -38,113 +38,52 @@ use crate::model::user::{CurrentUser, OnlineStatus};
...
-            },
-        }
-
-        old_channel
+        cache
+            .guilds
+            .get_mut(&self.channel.guild_id)
+            .and_then(|mut g| g.channels.insert(self.channel.id, self.channel.clone()))
     }
 }
```
mkrasnitski, line 79: Having trouble reviewing this because the diff is quite messy. But, I'm assuming changing `.map` to `.and_then` works because `HashMap::insert` returns the old value wrapped in `Some` (and `None` if there is no old value)?

It's clever, but maybe it's worth adding a comment to that effect so that other people can better understand what's happening without much effort.

mkrasnitski:
> Can you ask for a specific method? Most of them are quite obvious, as Webhooks cannot be created in private channels and private channels cannot be cached.

My main concern was about webhooks, but that's been answered in converstion in the serenity discord. Almost done with review here.

src/cache/mod.rs
```diff
@@ -469,75 +445,6 @@ impl Cache {
...
-    /// # async fn run() -> Result<(), Box<dyn std::error::Error>> {
-    /// let mut client =
-    ///     Client::builder("token", GatewayIntents::default()).event_handler(Handler).await?;
-    ///
-    /// client.start().await?;
-    /// # Ok(())
-    /// # }
-    /// ```
-    ///
-    /// [`EventHandler::message`]: crate::client::EventHandler::message
```
mkrasnitski, line 520: Do we want to keep this example and just move it onto `Cache::channel`?

src/cache/mod.rs
```diff
@@ -387,39 +362,40 @@ impl Cache {
...
-        {
-            if let Some(channel) = self.temp_channels.get(&id) {
-                return Some(Channel::Guild(channel));
-            }
-        }
-
-        if let Some(private_channel) = self.private_channels.get(&id) {
-            return Some(Channel::Private(private_channel.clone()));
+        if let Some(channel) = self.temp_channels.get(&id) {
+            return Some(CacheRef::from_arc(channel));
```
mkrasnitski: Minor nit, but for diff cleanliness purposes, let's keep the wrapping braces around this. Serenity is inconsistent with how it chooses to feature-gate single-expression blocks.  Sometimes it wraps them in braces and sometimes not. But the behavior is equivalent so it's something to be decided later.

src/cache/mod.rs
```diff
@@ -387,39 +362,40 @@ impl Cache {
...
-    /// This will search the `channels` map, then the [`Self::private_channels`] map.
-    ///
-    /// If you know what type of channel you're looking for, you should instead manually retrieve
-    /// from one of the respective methods:
+    /// Retrieves a [`GuildChannel`] from the cache based on the given Id.
     ///
-    /// - [`GuildChannel`]: [`Self::guild_channel`] or [`Self::guild_channels`]
-    /// - [`PrivateChannel`]: [`Self::private_channel`] or [`Self::private_channels`]
+    /// This will search the guild cache for the [`Guild::channels`] map, then the
+    /// `Self::temp_channels` map if enabled and not found.
```
mkrasnitski: This should be reworded since `Cache::temp_channels` isn't a method and the field is only `pub(crate)`. Plus, it's feature-gated.

mkrasnitski (changes requested):

src/cache/mod.rs
GnomedDev: This allows `Cache::channel` to return a `CacheRef`, as `MokaCache` will clone the value when `get` is called. If not wrapped, `Cache::channel` would either have to clone guild channels in the more popular path or `CacheRef` would have to include owned values, which feels very wrong.

src/cache/mod.rs
GnomedDev: It seems largely useless as an example, as `Message` already has a `to_channel` method, but if you feel strongly about this example I'll move it to `Cache::channel`.

src/model/channel/channel_id.rs
GnomedDev: We do need to clone before inserting into the cache. Arc::new takes an owned value and we are returning back the channel we just inserted. A clever thought would be to use `CacheRef` to always refer back to either the `cache.channel` or the `temp_channels` map, but this falls apart when `temp_cache` is disabled.

src/model/channel/guild_channel.rs
GnomedDev: The alternative would be a panic, and in exceptional circumstances I still prefer an error.

src/model/channel/private_channel.rs
GnomedDev: See above.

src/model/channel/channel_id.rs
mkrasnitski: Right, the lifetime of the channel changes if we enable or disable `temp_cache`.

src/cache/event.rs
GnomedDev: I kind of agree with you here, but looking at the whole module sees this trick used quite a couple of times, so I don't think it's necessary to comment here.

mkrasnitski (approved):
LGTM

mkrasnitski:
Last thing - do we also want to rename `Cache::guild_channels` to `Cache::channels`, since we essentially did the same with the singular version `Cache::guild_channel` -> `Cache::channel`?

EDIT: It's probably fine to leave it, since `channel` takes a `ChannelId`, but `guild_channels` takes a `GuildId`, and all the other methods that do so also have a `guild_` prefix.

arqunis (approved):

PR #2604, opened 2023-11-19
Added AsRef on Context for Context

supercoolspy:
This fixes an issue with poise's `execute_modal_on_component_interaction` which takes in a impl AsRef<Context> however, this does not actually accept context so this fixes it

### DIFF
--- a/src/client/context.rs
+++ b/src/client/context.rs
@@ -426,6 +426,12 @@ impl Context {
     }
 }
 
+impl AsRef<Context> for Context {
+    fn as_ref(&self) -> &Context {
+        &self
+    }
+}
+
 impl AsRef<Http> for Context {
     fn as_ref(&self) -> &Http {
         &self.http


GnomedDev:
Can users not just pass `&Context` into `impl AsRef<Context>`?

supercoolspy:
May be doing something wrong but getting ```the trait `std::convert::AsRef<poise::serenity_prelude::Context>` is not implemented for `poise::serenity_prelude::Context``` when passing an &Context

supercoolspy:
Skill issue with shuttle

sarahkittyy:
This is happening to me too !! What the hell am I supposed to do https://github.com/serenity-rs/poise/issues/247

GnomedDev:
This is an issue usually caused by duplicate dependencies, please open a new issue or ask in the serenity server if anyone else runs into this.

PR #2610, opened 2023-11-20
Fix CI

mkrasnitski:
This revives #2536 in order to fix the `simd-json` feature on the `v0.11` series and get CI to actually work again.

Initially I attempted to keep the old simd-json version exported while using a newer version internally, but unfortunately this doesn't work as types from the internal version would leak into the public API (specifically in methods on `Http` which expect a `Value` or `JsonMap`). Therefore, because the old version of `simd-json` has a soundness bug in `from_str`, and because not that many people actually use this feature, I propose we just bite the bullet and upgrade to `simd-json v0.10` on this branch. This way, we can provide a safe wrapper around `simd_json::from_str`, and while `from_str` is still exported in the prelude, it is marked unsafe.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 env:
   rust_min: 1.53.0
-  rust_nightly: nightly-2022-08-11
+  rust_nightly: nightly-2023-01-26
 
 jobs:
   test:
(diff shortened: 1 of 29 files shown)


mkrasnitski:
I'm just gonna make this PR for passing all of CI, not just simd-json.

mkrasnitski:
I don't feel like fixing MSRV at the moment, it's too much work. However, I tested it and without pinning, the MSRV is 1.67.

arqunis (approved):

PR #2620, opened 2023-11-24
Fix README.md minor error

LaytonGB:
FIX: bad example was re-using the moved `framework` value

### Before

image

### After

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,8 +61,9 @@ impl EventHandler for Handler {}
 
 #[tokio::main]
 async fn main() {
-    let framework = StandardFramework::new().group(&GENERAL_GROUP);
-    framework.configure(|c| c.prefix("~")); // set the bot's prefix to "~"
+    let framework = StandardFramework::new()
+        .group(&GENERAL_GROUP)
+        .configure(|c| c.prefix("~")); // set the bot's prefix to "~"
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,8 +61,9 @@ impl EventHandler for Handler {}
 
 #[tokio::main]
 async fn main() {
-    let framework = StandardFramework::new().group(&GENERAL_GROUP);
-    framework.configure(|c| c.prefix("~")); // set the bot's prefix to "~"
+    let framework = StandardFramework::new()
+        .group(&GENERAL_GROUP)
+        .configure(|c| c.prefix("~")); // set the bot's prefix to "~"
```
```suggestion
        .configure(Configuration::new().prefix("~")); // set the bot's prefix to "~"
```
Now that #2615 has been merged, this should include those changes as well. Be sure to include `Configuration` in the imports as well.

mkrasnitski (changes requested):

mkrasnitski:
Actually, as it turns out, in `v0.12`, since the standard framework is configurable at runtime, the `configure` function no longer consumes `self`. Therefore, the example can mostly go unchanged (sans the change to the call to `configure`).

mkrasnitski (approved):

PR #2623, opened 2023-11-24
Use dedicated types for GuildId::audit_logs

### LINKED ISSUE #2488: `(Partial)Guild(Id)::audit_logs` takes in raw action type IDs instead of the `Action` enum
Currently, [`serenity::model::Guild::audit_logs`](https://docs.rs/serenity/0.11.6/serenity/model/guild/struct.Guild.html#method.audit_logs) (as well as the equivalent methods in `GuildId` and `PartialGuildId`) take an `Option<u8>` for the optional action type field. This is incredibly bad for multiple reasons, such as allowing the user to easily and mistakenly specify incorrect type IDs, and requiring a comment explaining the magic number in many cases.

Since Serenity already has an [`Action` enum](https://docs.rs/serenity/latest/serenity/model/prelude/enum.Action.html) that models this action type very clearly, we should definitely refactor the `audit_log` methods to use this enum in 0.12.x.

GnomedDev:
Fixes #2488

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2576,14 +2576,14 @@ impl Http {
     pub async fn get_audit_logs(
         &self,
         guild_id: GuildId,
-        action_type: Option<u8>,
+        action_type: Option<audit_log::Action>,
         user_id: Option<UserId>,
-        before: Option<u64>,
+        before: Option<AuditLogEntryId>,
         limit: Option<u8>,
     ) -> Result<AuditLogs> {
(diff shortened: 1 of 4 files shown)


mkrasnitski (approved):
Only change I'd suggest would be to import `audit_log::Action`.

GnomedDev:
We have other types called `Action` in the library, so that would be ambiguous, I'm usually quite happy to import stuff.

PR #2625, opened 2023-11-25
Remove some miscellaneous vectors

GnomedDev:


src/model/guild/mod.rs
```diff
@@ -1616,19 +1616,10 @@ impl Guild {
...
-        for (&id, member) in &self.members {
-            if let Some(presence) = self.presences.get(&id) {
-                if status == presence.status {
-                    members.push(member);
-                }
-            }
-        }
-
-        members
+    pub fn members_with_status(&self, status: OnlineStatus) -> impl Iterator<Item = &Member> {
```
mkrasnitski, line 1619: Why change the return type on this? I think probably most people would expect a `Vec`, so having to explicitly collect into one is an ergonomics hit.

GnomedDev: If they want a vector, they can `.collect()` pretty easily. This is just more flexible for performance reasons. It's the same as `CacheRef` where we take an easily fixable ergonomics hit to be more flexible for performance.

src/model/guild/mod.rs
```diff
@@ -1616,19 +1616,10 @@ impl Guild {
...
+        self.members.values().filter(move |m| {
+            self.presences.get(&m.user.id).filter(|presence| presence.status == status).is_some()
+        })
```
mkrasnitski:
```suggestion
        self.members.iter().filter_map(move |(id, member)| match self.presences.get(id) {
            Some(presence) if presence.status == status => Some(member),
            _ => None,
        })
```
I think it's cleaner to use the hashmap key directly, even if `member.user.id` is the same. You could replace the match statement with a call to `.filter(|presence| presence.status == status).map(|_| member)`, but that felt not as intuitive.

arqunis (approved):

PR #2626, opened 2023-11-25
Remove Context from FullEvent

GnomedDev:
The macro syntax looks a bit weird as otherwise I run into parsing issues with `ctx` getting parsed the same as a variant name, so I had to put the pipe there as a separator. This is a general clean up as FullEvent is really weird otherwise, and makes generic `Client::data` a bit easier (no need for manual clone impl) if we are going to pursue that.

### DIFF
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -67,7 +67,7 @@ pub(crate) fn dispatch_model(
                 let context = context.clone();
                 let handler = Arc::clone(&handler);
                 spawn_named(event.snake_case_name(), async move {
-                    event.dispatch(&*handler, Some(context)).await;
+                    event.dispatch(context, &*handler).await;
                 });
             }
         }
--- a/src/client/dispatch.rs
+++ b/src/client/dispatch.rs
@@ -6,7 +6,7 @@ use tracing::debug;
 use super::event_handler::{EventHandler, RawEventHandler};
 use super::{Context, FullEvent};
 #[cfg(feature = "cache")]
-use crate::cache::CacheUpdate;
+use crate::cache::{Cache, CacheUpdate};
 #[cfg(feature = "framework")]
 use crate::framework::Framework;
 use crate::internal::tokio::spawn_named;
@@ -17,18 +17,34 @@ use crate::model::guild::Member;
 use crate::model::id::GuildId;
 
 #[cfg(feature = "cache")]
-fn update_cache<E: CacheUpdate>(context: &Context, event: &mut E) -> Option<E::Output> {
-    context.cache.update(event)
+macro_rules! if_cache {
+    ($e:expr) => {
+        $e
+    };
 }
 
 #[cfg(not(feature = "cache"))]
-fn update_cache<E>(_: &Context, _: &mut E) -> Option<()> {
-    None
+macro_rules! if_cache {
+    ($e:expr) => {
+        None
+    };
+}
+
+#[cfg(feature = "cache")]
+macro_rules! update_cache {
+    ($cache:ident, $event:ident) => {
+        $event.update($cache)
+    };
+}
+
+#[cfg(not(feature = "cache"))]
+macro_rules! update_cache {
+    ($cache:ident, $event:ident) => {};
 }
 
 pub(crate) fn dispatch_model(
     event: Event,
-    context: Context,
+    context: &Context,
     #[cfg(feature = "framework")] framework: Option<Arc<dyn Framework>>,
     event_handlers: Vec<Arc<dyn EventHandler>>,
     raw_event_handlers: Vec<Arc<dyn RawEventHandler>>,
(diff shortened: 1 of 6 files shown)


src/client/event_handler.rs
```diff
@@ -14,17 +14,17 @@ macro_rules! event_handler {
     ( $(
         $( #[doc = $doc:literal] )*
         $( #[cfg(feature = $feature:literal)] )?
-        async fn $method_name:ident(&self, $variant_name:ident { $(
+        async fn $method_name:ident(&self, $($context:ident: Context,)? | $variant_name:ident { $(
```
mkrasnitski, line 17: Will get to full review later, but why is this `|` here? Shouldn't the matching work fine without it?

mkrasnitski: Ok, apparently not, since both `$context` and `$variant_name` are identifiers, the parsing becomes ambiguous. This could still be done without the bar by defining a `event_handler_method!` macro that generates one method at a time and is called inside `event_handler!` for each method. It could have a second rule that takes a special `@NO-CONTEXT` token at the start which then wouldn't generate a `ctx: Context` parameter.

GnomedDev: I'm not the best at macro_rules syntax, I'd prefer if this somewhat janky syntax is kept for now and a followup PR could correct it if possible.

src/client/dispatch.rs
```diff
@@ -70,357 +92,310 @@ pub(crate) fn dispatch_model(
 ///
 /// Can return `None` if an event is unknown.
 #[cfg_attr(not(feature = "cache"), allow(unused_mut))]
-fn update_cache_with_event(ctx: Context, event: Event) -> Option<(FullEvent, Option<FullEvent>)> {
+fn update_cache_with_event(
+    #[cfg(feature = "cache")] cache: &Cache,
+    event: Event,
+) -> Option<(FullEvent, Option<FullEvent>)> {
```
mkrasnitski, line 98: A lot of the macro invocations here end up using the `cache` unconditionally, which I guess somehow works, but it still feels weird to pass in an undefined identifier if the `cache` is disabled.

src/client/dispatch.rs
```diff
@@ -17,18 +17,34 @@ use crate::model::guild::Member;
...
+
+#[cfg(not(feature = "cache"))]
+macro_rules! update_cache {
+    ($cache:ident, $event:ident) => {};
 }
 
 pub(crate) fn dispatch_model(
     event: Event,
-    context: Context,
+    context: &Context,
```
mkrasnitski, line 47:
```suggestion
    context: Context,
```
No need to change this I don't think.

src/gateway/bridge/shard_runner.rs
```diff
@@ -168,7 +168,7 @@ impl ShardRunner {
 
                 dispatch_model(
                     event,
-                    self.make_context(),
+                    &self.make_context(),
```
mkrasnitski, line 171:
```suggestion
                    self.make_context(),
```

src/client/event_handler.rs
```diff
@@ -59,11 +59,17 @@ macro_rules! event_handler {
...
                     $( #[cfg(feature = $feature)] )?
                     Self::$variant_name { $( $arg_name ),* } => {
-                        handler.$method_name( $( $arg_name ),* ).await;
+                        $(let Some($context) = ctx else {
+                            panic!("Missing Context for dispatch call!");
+                        };)?
+                        handler.$method_name($($context,)? $( $arg_name ),* ).await;
                     }
                 )* }
             }
```
mkrasnitski: IMO a better impl is the following: 
```rust
pub async fn dispatch(self, ctx: Context, handler: &dyn EventHandler) {
    match self { $(
        $( #[cfg(feature = $feature)] )?
        Self::$variant_name { $( $arg_name ),* } => {
            $( let $context = ctx; )?
            handler.$method_name($($context,)? $( $arg_name ),* ).await;
        }
    )* }
}
```

Right now, only `FullEvent::Ratelimit` doesn't contain a `ctx`, and it's never dispatched normally. So, making `ctx` optional here doesn't fit as it's basically required anyway.

mkrasnitski (changes requested):

src/client/dispatch.rs
GnomedDev: There is, otherwise you get clippy::needless_pass_by_value.

src/client/event_handler.rs
GnomedDev: Done

src/client/dispatch.rs
GnomedDev: It's overall a bad situation, but I don't think passing Context is any better, because it's still an undefined field access in that case.

mkrasnitski: Couldn't you just inline the macro invocations and feature-flag them? Btw, I noticed there's calls to `if_cache!(event.update(cache))` which are almost equivalent to just `update_cache!(cache, event)`.

GnomedDev: `if_cache!(event.update(cache))` returns `None` if cache is disabled, whereas `update_cache!(cache, event)` returns `()`. This difference is needed because otherwise the `None` would need to be qualified for it's `T`.  I could of course inline the macro invocations, but that would get very repetitive very quickly and make it a lot harder to read.

mkrasnitski (approved):
Remaining things are purely internal and can be fixed in a followup PR. LGTM

arqunis (approved):

PR #2627, opened 2023-11-25
Fix Channel argument_convert with no cache

GnomedDev:
Broken in #2602, that PR moved the `#[cfg(feature = "cache")]` around the entire if else block, which meant the else statement didn't get compiled in if cache was disabled. Caught this by running `cargo check` with cache disabled and seeing that `guild_id` was unused.


mkrasnitski (approved):
Good catch

PR #2631, opened 2023-11-28
Get rid of TypeMap and replace with Data generic

GnomedDev:
This PR gets rid of the `Client::data` typemap entirely, swapping with a generic `Arc<D>` that the user can provide. This allows much more flexibility for users and much less overhead. It also gets rid of the somewhat weird concept of a typemap from the beginner's view. Using a generic data parameter has been praised in poise for a while, so it's time to bring this over.

On the user side, a couple of changes had to be made:
- Client::builder now takes a third argument for the data value
- A lot of the types are now generic over the data, which is the major downside of this approach
- The standard framework had to be somewhat modified, with each macro now using generic syntax to provide the Data type.

These changes can be seen in the changes to the examples, which look pretty good to me.


GnomedDev:
Plenty of interest from users, no interest from serenity developers. Closing.

PR #2636, opened 2023-11-29
Fix `event_handler!` macro syntax

mkrasnitski:
Disambiguate the optional `ctx: Context` argument by changing the macro syntax around so that the parsing no longer becomes ambiguous. I'm relatively happy with the new syntax, since any docs/attributes already apply both to the `EventHandler` method, as well as the `FullEvent` variant.

### DIFF
--- a/src/client/event_handler.rs
+++ b/src/client/event_handler.rs
@@ -70,7 +70,7 @@ macro_rules! event_handler {
                     $(
                         $( #[cfg(feature = $feature)] )?
                         Self::$variant_name { $( $arg_name ),* } => {
-                            $(let $context = ctx;)?
+                            $( let $context = ctx; )?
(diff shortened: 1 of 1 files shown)


GnomedDev:
This is in no way blocking merge, but if the args could be duplicated to the "signature" that would read easier. Eg:
```
AutoModRuleCreate { rule: Rule } => async fn auto_moderation_rule_create(&self, ctx: Context, rule: Rule);
```

mkrasnitski:
Ideally we should enforce duplicate tokens, but idk how to do that. At best the compiler will at least ensure that the order/types are the same.

EDIT: This would produce pretty long lines, not sure how I feel about that.

arqunis (approved):

PR #2639, opened 2023-11-30
`impl Display for EmojiIdentifier`

### LINKED ISSUE #2638: implement `Display` for `EmojiIdentifier`
is there a reason that there isnt a `<:{name}:{id}>` `Display` impl?

vidhanio:
displays `EmojiIdentifier` as `<:{name}:id>`.

closes #2638

### DIFF
--- a/src/model/misc.rs
+++ b/src/model/misc.rs
@@ -4,6 +4,8 @@
 use std::error::Error as StdError;
 use std::fmt;
 #[cfg(all(feature = "model", feature = "utils"))]
+use std::fmt::Write;
+#[cfg(all(feature = "model", feature = "utils"))]
 use std::result::Result as StdResult;
 use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


kangalio:
That code is missing the `animated` field. But you don't need to implement it manually, it's probably better to impl Display by converting `ReactionType` and calling its Display impl instead. That way, no code is duplicated

vidhanio:
that would require a `clone()` of the `name` field. should i just copy the implementation instead?

vidhanio:
another code-deduplication alternative would be to maybe create an auxillary function which borrows the fields from both and writes?

kangalio:
> that would require a `clone()` of the `name` field. should i just copy the implementation instead?

Good point

> another code-deduplication alternative would be to maybe create an auxillary function which borrows the fields from both and writes?

That's true. I don't mind the current state of the PR, your decision

kangalio (approved):

vidhanio:
> > that would require a `clone()` of the `name` field. should i just copy the implementation instead?
> 
> 
> 
> Good point
> 
> 
> 
> > another code-deduplication alternative would be to maybe create an auxillary function which borrows the fields from both and writes?
> 
> 
> 
> That's true. I don't mind the current state of the PR, your decision 

this feels fine, it's only a couple lines. not some gigantic function being duplicated.

kangalio:
Yeah. The CI has some failured, once those are fixed, this PR can be merged I think

arqunis (approved):

PR #2646, opened 2023-12-03
Replace new `Request` generic param with const parameter

mkrasnitski:
Follow-up to #2624

We can rework the new `Request` generics to avoid having to use a turbofish everywhere by using const parameters which can be inferred easier. Unfortunately, a concrete choice for `N` must still be made when calling `Request::new`.

Slight behavioral change - before, setting `params` to `Some(vec![])` would still append a question mark to the end of the url. Now, we check if the params array `is_empty` instead of `is_some`, so the question mark won't be appended if the params list is empty (which is IMO more correct).

Once this gets merged it's probably good to squash with #2624 so that future rebases of `next` don't become painful with back-to-back commits causing conflicts.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2437,7 +2437,7 @@ impl Http {
     ) -> Result<Message> {
         let mut params = ArrayVec::<_, 1>::new();
         if let Some(thread_id) = thread_id {
-            params.push(("thread_id", thread_id.to_string()))
+            params.push(("thread_id", thread_id.to_string()));
         }
 
         self.fire(Request {
@@ -2467,7 +2467,7 @@ impl Http {
     ) -> Result<Message> {
         let mut params = ArrayVec::<_, 1>::new();
         if let Some(thread_id) = thread_id {
-            params.push(("thread_id", thread_id.to_string()))
+            params.push(("thread_id", thread_id.to_string()));
         }
 
         let mut request = Request {
@@ -2506,7 +2506,7 @@ impl Http {
     ) -> Result<()> {
         let mut params = ArrayVec::<_, 1>::new();
         if let Some(thread_id) = thread_id {
-            params.push(("thread_id", thread_id.to_string()))
+            params.push(("thread_id", thread_id.to_string()));
         }
 
         self.wind(204, Request {
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -4475,7 +4475,7 @@ impl Http {
     /// let channel_id = ChannelId::new(381880193700069377);
     /// let route = Route::ChannelMessages { channel_id };
     ///
-    /// let mut request = Request::<'_, 0>::new(route, LightMethod::Post).body(Some(bytes));
+    /// let mut request = Request::new(route, LightMethod::Post, []).body(Some(bytes));
     ///
     /// let message: Message = http.fire(request).await?;
     ///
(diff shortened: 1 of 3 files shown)


src/http/request.rs
```diff
@@ -62,58 +63,13 @@ impl<'a, Params> Request<'a, Params> {
...
-// Makes sure only methods that use the params actually need to bound on them.
-impl<'a, Params> Request<'a, Params>
-where
-    Params: IntoIterator<Item = (&'static str, String)>,
-    Params: std::fmt::Debug + Clone,
-{
-    pub fn params(mut self, params: Option<Params>) -> Self {
-        self.params = params;
+    /// # Panics
+    /// Panics if the length of the slice exceeds the value of the const paramter `N`.
```
GnomedDev: Misspelling of `parameter`

GnomedDev (changes requested):
`N` is really non-descriptive and requires a comment to explain. Can it be changed to `MAX_PARAMS`?

src/http/client.rs
```diff
@@ -4475,9 +4475,9 @@ impl Http {
     /// let channel_id = ChannelId::new(381880193700069377);
     /// let route = Route::ChannelMessages { channel_id };
     ///
-    /// let mut request = Request::new(route, LightMethod::Post).body(Some(bytes));
+    /// let mut request = Request::<'_, 0>::new(route, LightMethod::Post).body(Some(bytes));
```
mkrasnitski: I don't like doing this. The only solution I can think of would be to make `params` a required argument to `new` - then the value of `N` could be inferred. Thoughts?

src/http/client.rs
```diff
@@ -4518,7 +4518,7 @@ impl Http {
...
```
mkrasnitski: Ditto.

mkrasnitski:
Unfortunately in this process of adding a `params` argument, `Request::new` had to become non-`const`, but considering that doing anything meaningful with the `Request` before sending it off (like setting a body and headers) can't be done `const`, I don't think there's any real downside here.

arqunis (approved):

src/http/request.rs
```diff
@@ -59,58 +60,13 @@ impl<'a, Params> Request<'a, Params> {
...
-where
-    Params: IntoIterator<Item = (&'static str, String)>,
-    Params: std::fmt::Debug + Clone,
-{
-    pub fn params(mut self, params: Option<Params>) -> Self {
-        self.params = params;
+    /// # Panics
+    /// Panics if the length of the slice exceeds the value of `MAX_PARAMS`.
+    pub fn params(mut self, params: &[(&'static str, String)]) -> Self {
+        self.params = params.try_into().expect("number of params exceeded capacity");
```
arqunis:
```suggestion
    pub fn params(mut self, params: [(&'static str, String); MAX_PARAMS]) -> Self {
        self.params = params;
```

arqunis (changes requested):

src/http/request.rs
mkrasnitski: The motivation here is that the original argument was of type `Option<Vec<_>>`. Converting a vec into an array requires calling `.try_into().unwrap()`. I guess the more explicit version is a better tradeoff.

arqunis (approved):

PR #2647, opened 2023-12-03
Fix a possible deadlock in user_perms

GnomedDev:
I don't think this can deadlock in the current state, as `get` only deadlocks if holding a mutable reference, but just to be safe.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -463,19 +463,23 @@ pub(crate) fn user_has_perms_cache(
 pub(crate) fn user_perms(cache: impl AsRef<Cache>, channel_id: ChannelId) -> Result<Permissions> {
     let cache = cache.as_ref();
 
-    let Some(channel) = cache.channel(channel_id) else {
+    let Some(guild_id) = cache.channels.get(&channel_id).map(|c| *c) else {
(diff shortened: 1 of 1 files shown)


arqunis (approved):

src/utils/mod.rs
```diff
@@ -463,19 +463,23 @@ pub(crate) fn user_has_perms_cache(
 pub(crate) fn user_perms(cache: impl AsRef<Cache>, channel_id: ChannelId) -> Result<Permissions> {
     let cache = cache.as_ref();
 
-    let Some(channel) = cache.channel(channel_id) else {
+    let Some(guild_id) = cache.channels.get(&channel_id).map(|c| *c) else {
```
arqunis, line 466: I think you meant to write this?:
```suggestion
    let Some(guild_id) = cache.channels.get(&channel_id).map(|c| c.guild_id) else {
```

arqunis (changes requested):

src/utils/mod.rs
GnomedDev: No, cache.channels is map of ChannelId to GuildId. See #2602

arqunis (approved):

PR #2650, opened 2023-12-05
 Implement max_concurrency support 

GnomedDev:
Implements support for the [max concurrency](https://discord.com/developers/docs/topics/gateway#sharding-max-concurrency) system. This allows bots on very large sharding to start up multiple shards in parallel to cut startup times down quite a bit. I've personally tested this using my bot with 256 shards and a max_concurrency of 16 and it worked flawlessly.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -30,7 +30,7 @@ use std::sync::OnceLock;
 use futures::channel::mpsc::UnboundedReceiver as Receiver;
 use futures::future::BoxFuture;
 use futures::StreamExt as _;
-use tokio::sync::{Mutex, RwLock};
+use tokio::sync::RwLock;
 use tracing::{debug, error, info, instrument};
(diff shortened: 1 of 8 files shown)


mkrasnitski:
Might be better to separate concerns to make sure the rework is reviewed separately.

GnomedDev:
Done, and drafted until that PR is merged.

mkrasnitski:
This was discussed in the discord, but basically: When creating a batch of shards, to ensure both maximum batch size and avoid being ratelimited, their ratelimit keys should all be unique (calculated via `shard_id % max_concurrency`). That means that if a shard fails to start, when including it in the next batch for retry, it should replace a shard with the same rate limit key, in order to make sure the batch doesn't include any shards which are ratelimited together.

GnomedDev:
@mkrasnitski is going to take over this effort.

PR #2653, opened 2023-12-05
Clean up ShardManager/Queuer/Runner

GnomedDev:
Just some basic cleanups I noticed.

### DIFF
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -25,6 +25,7 @@
 
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::hash::Hash;
+use std::num::NonZeroU16;
 #[cfg(feature = "temp_cache")]
 use std::sync::Arc;
 #[cfg(feature = "temp_cache")]
@@ -125,7 +126,7 @@ pub type MessageRef<'a> = CacheRef<'a, ChannelId, Message, HashMap<MessageId, Me
 #[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
 #[derive(Debug)]
 pub(crate) struct CachedShardData {
-    pub total: u32,
+    pub total: NonZeroU16,
     pub connected: HashSet<ShardId>,
     pub has_sent_shards_ready: bool,
 }
@@ -281,7 +282,7 @@ impl Cache {
             message_queue: DashMap::default(),
 
             shard_data: RwLock::new(CachedShardData {
-                total: 1,
+                total: NonZeroU16::MIN,
                 connected: HashSet::new(),
                 has_sent_shards_ready: false,
             }),
@@ -539,7 +540,7 @@ impl Cache {
 
     /// Returns the number of shards.
     #[inline]
-    pub fn shard_count(&self) -> u32 {
+    pub fn shard_count(&self) -> NonZeroU16 {
         self.shard_data.read().total
     }
 
--- a/src/gateway/bridge/mod.rs
+++ b/src/gateway/bridge/mod.rs
@@ -50,6 +50,7 @@ mod shard_runner_message;
 mod voice;
 
 use std::fmt;
+use std::num::NonZeroU16;
 use std::time::Duration as StdDuration;
 
 pub use self::event::ShardStageUpdateEvent;
@@ -68,9 +69,10 @@ use crate::model::id::ShardId;
 /// A message to be sent to the [`ShardQueuer`].
 #[derive(Clone, Debug)]
 pub enum ShardQueuerMessage {
-    /// Message to start a shard, where the 0-index element is the ID of the Shard to start and the
-    /// 1-index element is the total shards in use.
-    Start(ShardId, ShardId),
+    /// Message to set the shard total.
+    SetShardTotal(NonZeroU16),
+    /// Message to start a shard.
+    Start(ShardId),
     /// Message to shutdown the shard queuer.
     Shutdown,
     /// Message to dequeue/shutdown a shard.
--- a/src/gateway/bridge/shard_manager.rs
+++ b/src/gateway/bridge/shard_manager.rs
@@ -1,5 +1,5 @@
 use std::collections::{HashMap, VecDeque};
-use std::sync::atomic::{AtomicU32, Ordering};
+use std::num::NonZeroU16;
 use std::sync::Arc;
 #[cfg(feature = "framework")]
 use std::sync::OnceLock;
@@ -65,7 +65,10 @@ use crate::model::gateway::GatewayIntents;
 /// impl RawEventHandler for Handler {}
 ///
 /// # let http: Arc<Http> = unimplemented!();
-/// let ws_url = Arc::new(Mutex::new(http.get_gateway().await?.url));
+/// let gateway_info = http.get_bot_gateway().await?;
+///
+/// let shard_total = gateway_info.shards;
+/// let ws_url = Arc::from(gateway_info.url);
 /// let data = Arc::new(RwLock::new(TypeMap::new()));
 /// let event_handler = Arc::new(Handler) as Arc<dyn EventHandler>;
 /// let framework = Arc::new(StandardFramework::new()) as Arc<dyn Framework + 'static>;
@@ -75,15 +78,10 @@ use crate::model::gateway::GatewayIntents;
 ///     event_handlers: vec![event_handler],
 ///     raw_event_handlers: vec![],
 ///     framework: Arc::new(OnceLock::from(framework)),
-///     // the shard index to start initiating from
-///     shard_index: 0,
-///     // the number of shards to initiate (this initiates 0, 1, and 2)
-///     shard_init: 3,
-///     // the total number of shards in use
-///     shard_total: 5,
 ///     # #[cfg(feature = "voice")]
 ///     # voice_manager: None,
 ///     ws_url,
+///     shard_total,
 ///     # #[cfg(feature = "cache")]
 ///     # cache: unimplemented!(),
 ///     # http,
@@ -103,13 +101,6 @@ pub struct ShardManager {
     /// **Note**: It is highly unrecommended to mutate this yourself unless you need to. Instead
     /// prefer to use methods on this struct that are provided where possible.
     pub runners: Arc<Mutex<HashMap<ShardId, ShardRunnerInfo>>>,
-    /// The index of the first shard to initialize, 0-indexed.
-    // Atomics are used here to allow for mutation without requiring a mutable reference to self.
-    shard_index: AtomicU32,
-    /// The number of shards to initialize.
-    shard_init: AtomicU32,
-    /// The total shards in use, 1-indexed.
-    shard_total: AtomicU32,
     shard_queuer: Sender<ShardQueuerMessage>,
     // We can safely use a Mutex for this field, as it is only ever used in one single place
     // and only is ever used to receive a single message
@@ -131,10 +122,7 @@ impl ShardManager {
 
         let manager = Arc::new(Self {
             return_value_tx: Mutex::new(return_value_tx),
-            shard_index: AtomicU32::new(opt.shard_index),
-            shard_init: AtomicU32::new(opt.shard_init),
             shard_queuer: shard_queue_tx,
-            shard_total: AtomicU32::new(opt.shard_total),
             shard_shutdown: Mutex::new(shutdown_recv),
             shard_shutdown_send: shutdown_send,
             runners: Arc::clone(&runners),
@@ -155,6 +143,7 @@ impl ShardManager {
             #[cfg(feature = "voice")]
             voice_manager: opt.voice_manager,
             ws_url: opt.ws_url,
+            shard_total: opt.shard_total,
             #[cfg(feature = "cache")]
             cache: opt.cache,
             http: opt.http,
@@ -182,32 +171,20 @@ impl ShardManager {
     /// This will communicate shard boots with the [`ShardQueuer`] so that they are properly
     /// queued.
     #[instrument(skip(self))]
-    pub fn initialize(&self) -> Result<()> {
-        let shard_index = self.shard_index.load(Ordering::Relaxed);
-        let shard_init = self.shard_init.load(Ordering::Relaxed);
-        let shard_total = self.shard_total.load(Ordering::Relaxed);
-
+    pub fn initialize(
+        &self,
+        shard_index: u16,
+        shard_init: u16,
+        shard_total: NonZeroU16,
+    ) -> Result<()> {
         let shard_to = shard_index + shard_init;
 
+        self.set_shard_total(shard_total);
         for shard_id in shard_index..shard_to {
-            self.boot([ShardId(shard_id), ShardId(shard_total)]);
+            self.boot(ShardId(shard_id));
         }
 
         Ok(())
-    }
-
-    /// Sets the new sharding information for the manager.
-    ///
-    /// This will shutdown all existing shards.
-    ///
-    /// This will _not_ instantiate the new shards.
-    #[instrument(skip(self))]
-    pub async fn set_shards(&self, index: u32, init: u32, total: u32) {
-        self.shutdown_all().await;
-
-        self.shard_index.store(index, Ordering::Relaxed);
-        self.shard_init.store(init, Ordering::Relaxed);
-        self.shard_total.store(total, Ordering::Relaxed);
     }
 
     /// Restarts a shard runner.
@@ -232,12 +209,9 @@ impl ShardManager {
     /// [`ShardRunner`]: super::ShardRunner
     #[instrument(skip(self))]
     pub async fn restart(&self, shard_id: ShardId) {
-        info!("Restarting shard {}", shard_id);
+        info!("Restarting shard {shard_id}");
         self.shutdown(shard_id, 4000).await;
-
-        let shard_total = self.shard_total.load(Ordering::Relaxed);
-
-        self.boot([shard_id, ShardId(shard_total)]);
+        self.boot(shard_id);
     }
 
     /// Returns the [`ShardId`]s of the shards that have been instantiated and currently have a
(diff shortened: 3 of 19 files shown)


src/gateway/bridge/shard_queuer.rs
```diff
@@ -116,14 +116,19 @@ impl ShardQueuer {
...
-                Ok(Some(ShardQueuerMessage::Start(id, total))) => {
-                    debug!("[Shard Queuer] Received to start shard {} of {}.", id.0, total.0);
-                    self.checked_start(id, total.0).await;
+                Ok(Some(ShardQueuerMessage::Start(shard_info))) => {
+                    debug!(
+                        "[Shard Queuer] Received to start shard {} of {}.",
+                        shard_info.id, shard_info.total
+                    );
+
+                    self.check_last_start().await;
```
mkrasnitski:
```suggestion
```
This already gets called inside `checked_start`.

src/gateway/bridge/shard_messenger.rs
```diff
@@ -58,24 +58,17 @@ impl ShardMessenger {
...
-    /// # async fn run() -> Result<(), Box<dyn std::error::Error>> {
-    /// #     let mutex = Arc::new(Mutex::new("".to_string()));
-    /// #
-    /// #     let shard_info = ShardInfo {
-    /// #         id: ShardId(0),
-    /// #         total: 1,
-    /// #     };
-    /// #     let mut shard = Shard::new(mutex.clone(), "", shard_info, GatewayIntents::all(), None).await?;
-    /// #
+    /// # async fn run(mut shard: Shard) -> Result<(), Box<dyn std::error::Error>> {
```
mkrasnitski, line 62:
```suggestion
    /// # async fn run() -> Result<(), Box<dyn std::error::Error>> {
```
Hard to create an auto-applicable suggestion here, but the best way to do this would be to add a line as follows:
```rust
let mut shard: Shard = unimplemented!();
```
The other places you defined `async fn run(mut shard: Shard)` should also be changed as such.

src/gateway/bridge/mod.rs
```diff
@@ -61,16 +61,16 @@ pub use self::shard_runner_message::ShardRunnerMessage;
 #[cfg(feature = "voice")]
 pub use self::voice::VoiceGatewayManager;
 use super::ChunkGuildFilter;
+use crate::all::ShardInfo;
```
mkrasnitski:
```suggestion
use crate::modell::gateway::ShardInfo;
```
We should avoid importing from `crate::all` internally.

src/gateway/bridge/shard_manager.rs
```diff
@@ -15,6 +14,7 @@ use typemap_rev::TypeMap;
 #[cfg(feature = "voice")]
 use super::VoiceGatewayManager;
 use super::{ShardId, ShardQueuer, ShardQueuerMessage, ShardRunnerInfo};
+use crate::all::ShardInfo;
```
mkrasnitski:
```suggestion
use crate::model::gateway::ShardInfo;
```
Ditto.

src/gateway/bridge/shard_manager.rs
```diff
@@ -182,34 +166,19 @@ impl ShardManager {
...
 
-    /// Sets the new sharding information for the manager.
-    ///
-    /// This will shutdown all existing shards.
-    ///
-    /// This will _not_ instantiate the new shards.
-    #[instrument(skip(self))]
-    pub async fn set_shards(&self, index: u32, init: u32, total: u32) {
-        self.shutdown_all().await;
-
```
mkrasnitski, line 207: Strange, I wonder why this line was even included if the only time this gets called is when the `ShardManager` is first instantiated. Is there some sort of control flow related to the channels/message passing that might reach this function through some other means?

src/model/gateway.rs
```diff
@@ -375,7 +375,7 @@ pub struct ShardInfo {
 
 impl ShardInfo {
     #[must_use]
-    pub(crate) fn new(id: ShardId, total: u32) -> Self {
+    pub fn new(id: ShardId, total: u32) -> Self {
```
mkrasnitski: If `ShardInfo` no longer becomes part of the public API, this should be reverted too.

mkrasnitski (changes requested):
Since `ShardManager::{restart, restart_shard}` are public functions, relying on the user to make sure that `ShardInfo::total` is correct seems brittle. Maybe `ShardManager::shard_total` could be kept and changed to type `Arc<u32>`? It's only set on initialization and never changed, and therefore could be kept immutable this way.

src/gateway/bridge/shard_messenger.rs
GnomedDev: Why? A lot of other examples take Client as an argument and I don't see the benefit at all of doing `unimplemented!()`.

mkrasnitski: yeah, you're right - this can stay, it was just a minor nit

src/gateway/bridge/shard_manager.rs
GnomedDev: Seeing that I removed no call of this function other than the initial startup, it must have been removed at some point.

src/gateway/bridge/shard_queuer.rs
GnomedDev: Fixed

src/gateway/bridge/mod.rs
GnomedDev: Fixed

src/gateway/bridge/shard_manager.rs
GnomedDev: Fixed

src/model/gateway.rs
GnomedDev: Eh, I don't see why. The fields are already public.

src/gateway/bridge/shard_manager.rs
```diff
@@ -103,13 +96,6 @@ pub struct ShardManager {
     /// **Note**: It is highly unrecommended to mutate this yourself unless you need to. Instead
     /// prefer to use methods on this struct that are provided where possible.
     pub runners: Arc<Mutex<HashMap<ShardId, ShardRunnerInfo>>>,
-    /// The index of the first shard to initialize, 0-indexed.
-    // Atomics are used here to allow for mutation without requiring a mutable reference to self.
-    shard_index: AtomicU32,
-    /// The number of shards to initialize.
-    shard_init: AtomicU32,
-    /// The total shards in use, 1-indexed.
-    shard_total: AtomicU32,
```
mkrasnitski, line 112: Add `shard_total` back as a `OnceLock<u32>`.

src/gateway/bridge/shard_manager.rs
```diff
@@ -131,10 +117,7 @@ impl ShardManager {
 
         let manager = Arc::new(Self {
             return_value_tx: Mutex::new(return_value_tx),
-            shard_index: AtomicU32::new(opt.shard_index),
-            shard_init: AtomicU32::new(opt.shard_init),
             shard_queuer: shard_queue_tx,
-            shard_total: AtomicU32::new(opt.shard_total),
```
mkrasnitski, line 137: Should be added back, initialized to `OnceLock::new()`.

src/gateway/bridge/shard_manager.rs
```diff
@@ -182,34 +165,19 @@ impl ShardManager {
     /// This will communicate shard boots with the [`ShardQueuer`] so that they are properly
     /// queued.
     #[instrument(skip(self))]
-    pub fn initialize(&self) -> Result<()> {
-        let shard_index = self.shard_index.load(Ordering::Relaxed);
-        let shard_init = self.shard_init.load(Ordering::Relaxed);
-        let shard_total = self.shard_total.load(Ordering::Relaxed);
-
+    pub fn initialize(&self, shard_index: u32, shard_init: u32, shard_total: u32) -> Result<()> {
```
mkrasnitski:
```suggestion
    pub fn initialize(&self, shard_index: u32, shard_init: u32, shard_total: u32) -> Result<()> {
        self.shard_total.set(shard_total).expect("initializing shard_total failed");
```

src/gateway/bridge/shard_manager.rs
```diff
@@ -182,34 +165,19 @@ impl ShardManager {
...
         let shard_to = shard_index + shard_init;
 
         for shard_id in shard_index..shard_to {
-            self.boot([ShardId(shard_id), ShardId(shard_total)]);
+            self.boot(ShardInfo {
+                id: ShardId(shard_id),
+                total: shard_total,
+            });
```
mkrasnitski:
```suggestion
            self.boot(ShardId(shard_id));
```

src/gateway/bridge/shard_manager.rs
```diff
@@ -220,24 +188,23 @@ impl ShardManager {
...
     /// [`ShardRunner`]: super::ShardRunner
     #[instrument(skip(self))]
-    pub async fn restart(&self, shard_id: ShardId) {
-        info!("Restarting shard {}", shard_id);
-        self.shutdown(shard_id, 4000).await;
-
-        let shard_total = self.shard_total.load(Ordering::Relaxed);
+    pub async fn restart(&self, shard: ShardInfo) {
+        info!("Restarting shard {}", shard.id);
+        self.shutdown(shard.id, 4000).await;
```
mkrasnitski:
```suggestion
    pub async fn restart(&self, shard_id: ShardId) {
        info!("Restarting shard {}", shard_id);
        self.shutdown(shard_id, 4000).await;
```

src/gateway/bridge/shard_manager.rs
```diff
@@ -220,24 +188,23 @@ impl ShardManager {
...
 
-        self.boot([shard_id, ShardId(shard_total)]);
+        self.boot(shard);
```
mkrasnitski:
```suggestion
        self.boot(shard_id);
```

src/gateway/bridge/shard_manager.rs
```diff
@@ -351,9 +317,9 @@ impl ShardManager {
         }
     }
 
-    pub async fn restart_shard(&self, id: ShardId) {
-        self.restart(id).await;
-        if let Err(e) = self.shard_shutdown_send.unbounded_send(id) {
+    pub async fn restart_shard(&self, shard: ShardInfo) {
+        self.restart(shard).await;
+        if let Err(e) = self.shard_shutdown_send.unbounded_send(shard.id) {
```
mkrasnitski:
```suggestion
    pub async fn restart_shard(&self, shard_id: ShardId) {
        self.restart(shard_id).await;
        if let Err(e) = self.shard_shutdown_send.unbounded_send(shard_id) {
```

src/gateway/bridge/shard_manager.rs
```diff
@@ -325,11 +292,10 @@ impl ShardManager {
...
 
     #[instrument(skip(self))]
-    fn boot(&self, shard_info: [ShardId; 2]) {
-        info!("Telling shard queuer to start shard {}", shard_info[0]);
-
-        let msg = ShardQueuerMessage::Start(shard_info[0], shard_info[1]);
+    fn boot(&self, shard: ShardInfo) {
+        info!("Telling shard queuer to start shard {}", shard.id);
 
+        let msg = ShardQueuerMessage::Start(shard);
```
mkrasnitski:
```suggestion
    fn boot(&self, shard_id: ShardId) {
        info!("Telling shard queuer to start shard {}", shard_id);

        let shard_total = self.shard_total.get().expect("shard_total uninitialized");
        let msg = ShardQueuerMessage::Start(ShardInfo::new(shard_id, *shard_total));
```

src/gateway/bridge/shard_runner.rs
```diff
@@ -458,12 +458,12 @@ impl ShardRunner {
 
         self.update_manager().await;
 
-        let shard_id = self.shard.shard_info().id;
-        self.manager.restart_shard(shard_id).await;
+        let shard_info = self.shard.shard_info();
+        self.manager.restart_shard(shard_info).await;
```
mkrasnitski:
```suggestion
        self.manager.restart_shard(shard_info.id).await;
```

mkrasnitski (changes requested):
Ok, so rather than an `Arc<u32>`, I figured the best way to avoid breaking the signatures of `restart` and `restart_shard` is to have `shard_total` be a `OnceLock<u32>` that gets filled inside `ShardManager::initialize` and is fetched inside `ShardManager::boot`.

GnomedDev:
Right then, fixed your concerns by making `ShardQueuer` just store the total shard count and stop using `ShardInfo` everywhere redundantly.

mkrasnitski:
Changing to `u16` is fine, but making `shard_total` a `NonZeroU16` I think is a bit awkward, since it infects the public API for `Client`.

GnomedDev:
It infecting the API for Client is somewhat intentional, it is an invariant of the API that was previously not encoded and is now in the type system.

mkrasnitski:
> It infecting the API for Client is somewhat intentional, it is an invariant of the API that was previously not encoded and is now in the type system.

That's true, but do you think the panic could be moved inside the function by converting from a `u16`? The affected functions are called early on in the life of the program, and aren't called as often as `start_autosharded` (I'm just assuming here).

GnomedDev:
If I wanted mistakes in my code to become problems at runtime, instead of compile time, I would be writing in python. I can make it panic if 0 inside the serenity code but I very much recommend against. Id::new panicking is fine enough because it's done so often and isn't likely to have a person give it `0`, but this is a one time thing for each program, it just looks like more of a migration because we have 50 bots in the repo in the form of doc examples and actual examples.

src/cache/mod.rs
```diff
@@ -125,7 +126,7 @@ pub type MessageRef<'a> = CacheRef<'a, ChannelId, Message, HashMap<MessageId, Me
 #[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
 #[derive(Debug)]
 pub(crate) struct CachedShardData {
-    pub total: u32,
+    pub total: std::num::NonZeroU16,
```
mkrasnitski:
```suggestion
    pub total: NonZeroU16,
```

src/gateway/bridge/shard_manager.rs
```diff
@@ -182,34 +171,21 @@ impl ShardManager {
...
-    pub fn initialize(&self) -> Result<()> {
-        let shard_index = self.shard_index.load(Ordering::Relaxed);
-        let shard_init = self.shard_init.load(Ordering::Relaxed);
-        let shard_total = self.shard_total.load(Ordering::Relaxed);
-
+    pub fn initialize(
+        &self,
+        shard_index: u16,
+        shard_init: u16,
+        shard_total: NonZeroU16,
```
mkrasnitski, line 178: This parameter gets ignored and doesn't actually make it to the `ShardQueuer`. Which means that the manual shard totals specified in `Client::start_shard` and `start_shards` get ignored, since you initialize `ShardQueuer::shard_total` with the value received from the gateway.

src/gateway/bridge/shard_queuer.rs
```diff
@@ -148,24 +151,22 @@ impl ShardQueuer {
     }
 
     #[instrument(skip(self))]
-    async fn checked_start(&mut self, id: ShardId, total: u32) {
-        debug!("[Shard Queuer] Checked start for shard {} out of {}", id, total);
-        self.check_last_start().await;
+    async fn checked_start(&mut self, shard_id: ShardId, shard_total: NonZeroU16) {
```
mkrasnitski: Could just use `self.shard_total` internally here.

src/gateway/bridge/shard_queuer.rs
```diff
@@ -148,24 +151,22 @@ impl ShardQueuer {
...
         }
 
         self.last_start = Some(Instant::now());
     }
 
     #[instrument(skip(self))]
-    async fn start(&mut self, id: ShardId, total: u32) -> Result<()> {
-        let shard_info = ShardInfo::new(id, total);
-
+    async fn start(&mut self, shard_info: ShardInfo) -> Result<()> {
```
mkrasnitski: Same here as for `checked_start` - we can construct the `ShardInfo` inside the function before giving it to `Shard::new`.

mkrasnitski (changes requested):
Initial brief review of the new changes. The shard total is being incorrectly initialized IMO. The intention is that it should be either specified manually, or default to the value provided by the gateway. Right now it can't be manually overridden.

One way to fix this is to introduce a new `ShardQueuerMessage` variant that can override the value of `ShardQueuer::shard_total`. Then, `Client::start_autosharded` won't have to fire a duplicate request to the gateway.

I don't like that `ShardManagerOptions::shard_total` exists solely to get the value into the `ShardQueuer`, but I guess changing to using a `OnceLock` would mean that `Client::start_autosharded` would be the place that initializes it, meaning a duplicate gateway request.

As for the `NonZeroU16` thing, all I'll say is that in the string `NonZeroU16::new(1).expect("1 != 0")`, out of the 35 total characters, only a *single one* is relevant. And, I'm sure there are plenty of other values serenity receives from the API that can never be 0 which aren't tracked as non-zero. Ultimately, the tradeoff of noisiness here is pretty stark to me, and not worth it.

GnomedDev:
Followed through with all your suggestions.

src/utils/mod.rs
```diff
@@ -485,8 +486,16 @@ pub(crate) fn user_perms(cache: impl AsRef<Cache>, channel_id: ChannelId) -> Res
...
 #[inline]
 #[must_use]
-pub fn shard_id(guild_id: GuildId, shard_count: u32) -> u32 {
-    ((guild_id.get() >> 22) % (shard_count as u64)) as u32
+pub fn shard_id(guild_id: GuildId, shard_count: u16) -> u16 {
+    let shard_count = check_shard_total(shard_count);
+    ((guild_id.get() >> 22) % (shard_count.get() as u64)) as u16
+}
+
+pub(crate) fn check_shard_total(total_shards: u16) -> NonZeroU16 {
```
mkrasnitski, line 494: nit: Do you think `shard_total_checked` is a better name?

mkrasnitski (approved):
Thank you for all your patience and work on this PR.

src/utils/mod.rs
GnomedDev: Not personally

mkrasnitski: I'm pulling from methods like `u16::checked_add` for the naming here. When I first saw the method in review (before I saw the body), I thought it was somehow querying the value of the shard total, because of the method name wording (using the imperative). But tbh it's not a big deal.

arqunis (approved):

PR #2654, opened 2023-12-05
Remove `_arc` methods

GnomedDev:
These are unnecessary, taking `impl Into<Arc<T>>` allows passing `T` or `Arc<T>`.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -203,7 +203,7 @@ impl ClientBuilder {
     /// Sets the voice gateway handler to be used. It will receive voice events sent over the
     /// gateway and then consider - based on its settings - whether to dispatch a command.
     #[cfg(feature = "voice")]
-    pub fn voice_manager_arc<V>(mut self, voice_manager: impl Into<Arc<V>>) -> Self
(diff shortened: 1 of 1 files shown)


src/client/mod.rs
```diff
@@ -202,31 +202,12 @@ impl ClientBuilder {
 
     /// Sets the voice gateway handler to be used. It will receive voice events sent over the
     /// gateway and then consider - based on its settings - whether to dispatch a command.
-    ///
-    /// *Info*: If a reference to the voice_manager is required for manual dispatch, use the
-    /// [`Self::voice_manager_arc`]-method instead.
     #[cfg(feature = "voice")]
-    pub fn voice_manager<V>(mut self, voice_manager: V) -> Self
+    pub fn voice_manager_arc<V>(mut self, voice_manager: impl Into<Arc<V>>) -> Self
```
arqunis:
```suggestion
    pub fn voice_manager<V>(mut self, voice_manager: impl Into<Arc<V>>) -> Self
```

arqunis (changes requested):

arqunis (approved):

PR #2656, opened 2023-12-06
Replace Vec and String with FixedArray and FixedString for all models

GnomedDev:
This shrinks type sizes by [a lot](https://www.diffchecker.com/ASYD3mq3/) however makes the user experience slightly different:

- `FixedString` must be converted to String with `.into()` or `.into_string()` before it can be pushed to, but derefs to `&str` as is.
- `FixedArray` must be converted to `Vec` with `.into()` or `.into_vec()` before it can be pushed to, but derefs to `&[T]` as is.

Currently this is a git dependency, but I hope that is fine for `next`.  I want to get some basic testing in serenity before I push to crates.io.


mkrasnitski:
Without taking a look at the changes, what's the benefit of these types over `ArrayVec` and `ArrayString` from the `arrayvec` crate?

GnomedDev:
ArrayVec and ArrayString are both inline allocated and have a compile time max capacity
FixedArray and FixedString are both heap allocated, have a max length of u32::MAX, and don't store capacity

mkrasnitski:
Looking at the lib, it seems like there are no mutable APIs for either of the types? If that's the case, why not then use boxed arrays? At the very least, at a cursory glance I'm not convinced the `FixedString` API is all that sound - I think it would be best to replace that with `Box<str>`, which is guaranteed sound.

GnomedDev:
Can you read the module level documentation of the lib?
> These can be thought of as `Box<[T]>` and `Box<str>`, except the length is denoted as a [`u32`].

As for the soundness, there is very little `unsafe` in the library, and you are free to audit it, but it's pretty simple and I can't see many failure points.

ultrabear:
theres a bit of UB (that i can see) that is amendable by erroring on platforms where pointer width is not 32 or 64 bit

[UB in question](https://github.com/GnomedDev/small-fixed-array/blob/101610c061830127811ed3414de1794996a20eb0/src/non_empty_array.rs#L15C34-L15C34)
```rust
    fn len(&self) -> usize {
        self.small_len() as usize
    }
```

GnomedDev:
Can you move that to an issue on the repo, instead of here, although thanks for letting me know, I'll see if I can restriction lint out `as` entirely.

mkrasnitski:
> Can you read the module level documentation of the lib?

Apologies about that, I wasn't being very careful with my skimming. 

With regards to the size of the struct, we're already saving a `usize` for capacity, which we can assume is 8 bytes. However, an extra 4 bytes of size is not worth the extra maintenance burden IMO, considering this type would be included in basically every model type, and in some cases be user facing. The deref impl definitely helps, but IMO it would still be too exotic.

GnomedDev:
I know I've floated the idea before, but how about an unstable-optimisations feature flag to swap implementations for this stuff around. It could be Box<str> for most users and FixedString for crazy people.

mkrasnitski:
> I know I've floated the idea before, but how about an unstable-optimisations feature flag to swap implementations for this stuff around. It could be Box for most users and FixedString for crazy people.

This could work and would save you the headache of maintaining a fork. That work could be made in a separate PR, as the sizable savings already incurred from switching to `Box` are worth merging on their own merit.

mkrasnitski:
Relying on macros to specialize on the size of the length field feels like a bit too much pixie dust to me. I definitely see how your goal is to cut down on feature-gates, but I feel like there must be a better way.

That aside, imo for the sake of separation of concerns, this PR should be split into two, the first of which converts everything it can to `Box<str>`/`Box<[T]>`, with this one on top of it. If you'd like, I can handle that to save you the duplicate effort.

GnomedDev:
Using macros is basically the only way, unless I'm feature gating all times in serenity a String is used... which no. Splitting the PR will cause conflict hell for rebasing next on current, and I don't see the point.

mkrasnitski:
> Splitting the PR will cause conflict hell for rebasing next on current.

That's true, I guess the underlying base of changing to `Box<str>` felt ready, but since it's not an urgent change, it doesn't actually need to be separated out.

mkrasnitski:
Ok, one idea that can alleviate the macro-hell is to specialize the size on a per-module basis rather than per-use. That way, you can feature-gate a type-alias definition at the top of the module. Doing a grep through the changes, most of the uses of `FixedString` within a given module use the same size for the length, with only a few outliers that will incur a few extra bytes.

GnomedDev:
I don't see how it is a macro hell? It's just a basic macro that spits out different types and ignores an argument on one featureset. If macros can be avoided, sure, but this seems much weirder? Like, oh "FixedString" in embed.rs means a u16 length string? That's kind of weird and repeats the feature cfgs multiple times.

mkrasnitski:
Having a macro invocation instead of a concrete type parameter is really awkward and confusing, way more-so than an extra type alias. Part of the reason to have an `advanced-optimizations` feature is to not incur extra maintenance burden for the more common use-cases, and using macros means that you can't really ignore it. I don't really see what's wrong with:

```rust
#[cfg(feature = "advanced-optimizations")]
type FixedString = small_fixed_array::FixedString<u16>;
```
at the top of model files (the `Box<str>` alias is already exported from the internal prelude).

GnomedDev:
That creates an invisible maintainance burden, and hurts readablility a ton, as for each model file you now have to look to the top whenever you want to see the maximum value, and a new maintainer who does not know about this can just blindly use the wrong length and cause issues. The macro is a slight syntax change which needs to be understood once and can be understood imo quite easily on that first look. To the user, it appears as the concrete type in rustdoc or intellisense or even error messages.

examples/e02_transparent_guild_sharding/src/main.rs
```diff
@@ -25,7 +26,7 @@ struct Handler;
 #[async_trait]
 impl EventHandler for Handler {
     async fn message(&self, ctx: Context, msg: Message) {
-        if msg.content == "!ping" {
+        if &*msg.content == "!ping" {
```
mkrasnitski: Agh, you're right that `Box<str>` doesn't have everything we want and doesn't always auto-deref. This sucks, I don't know what to do. Maybe a followup PR can introduce a newtype or something that implements what we want. Or we could revert some of the fields and keep them as `String` for the sake of ergonomics. I'd say `Message::content` is a good candidate for that.

arqunis (approved):

Issue #2644, opened 2023-12-03
Suppressing embeds on messages by non-self users is disabled


ryanccn:
This check prevents edits on messages that are not from the bot itself:

https://github.com/serenity-rs/serenity/blob/6f5b610213c251129f4b011bf9d339548a31a559/src/model/channel/message.rs#L352-L354

However, you can actually edit messages from other users when your edit is for the sole purpose of suppressing the embeds on them (i.e. `EditMessage::new().suppress_embeds(true)`). This check prevents that functionality at the client level.

ryanccn:
The check could be amended to take this specific use case into account, or a more laissez-faire approach would be to just remove the client check altogether and let the server do its thing.

PR #2659, opened 2023-12-11
Fix embed suppression on non-self users when cache is enabled

iulianalexa:
This pull request proposes a fix for issue #2644.

### DIFF
--- a/src/builder/edit_message.rs
+++ b/src/builder/edit_message.rs
@@ -35,7 +35,7 @@ use crate::model::prelude::*;
 /// ```
 ///
 /// [Discord docs](https://discord.com/developers/docs/resources/channel#edit-message)
-#[derive(Clone, Debug, Default, Serialize)]
+#[derive(Clone, Debug, Default, Serialize, PartialEq)]
 #[must_use]
 pub struct EditMessage {
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -221,7 +221,7 @@ impl EditMessage {
 #[cfg(feature = "http")]
 #[async_trait::async_trait]
 impl Builder for EditMessage {
-    type Context<'ctx> = (ChannelId, MessageId);
+    type Context<'ctx> = (ChannelId, MessageId, Option<UserId>);
     type Built = Message;
 
     /// Edits a message in the channel.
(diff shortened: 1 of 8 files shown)


GnomedDev (changes requested):
This is a bad solution, a proposed solution would be to move the check into `EditMessage::execute`, pass along the `author_id` in the Context tuple, and check there while having access to the message flags to check if suppress embeds has been set.

iulianalexa:
> This is a bad solution, a proposed solution would be to move the check into `EditMessage::execute`, pass along the `author_id` in the Context tuple, and check there while having access to the message flags to check if suppress embeds has been set.

I agree that a more organised solution involves doing the check in execute, however my solution checks the other fields of the struct too in addition to the suppress embeds flag, the thought process being that a request with suppress embeds and a content change would still be invalid.

I would like to know if I should keep using that approach (create a reference struct, check the serialised equality) or if I should just check the flag and let the Discord API handle it.

GnomedDev:
Why not just compare them directly? Deriving PartialEq might be needed, but it would be a ton more efficent than passing through serde.

iulianalexa:
If I compare them directly, then is it not fine to do it in `edit` like in the initial proposal, but without the serialisation? Otherwise, I would have to do the cache check inside `execute`, which to me does not make sense considering the original author check was done inside of `edit`. Furthermore, there is one more function calling `execute` that does not use a context (`ChannelId::edit_message`), so I would have to wrap it inside an `Option`, further complicating things and making them look not so nice.

To me it's fine either way, but I thought I'd let you know before I push it.

iulianalexa:
I have implemented the proposed solution, please let me know if anything should be changed.

src/builder/edit_message.rs
```diff
@@ -254,6 +254,19 @@ impl Builder for EditMessage {
     ) -> Result<Self::Built> {
         self.check_length()?;
 
+        if let Some(user_id) = ctx.2 {
+            #[cfg(feature = "cache")]
```
alexandruradovici:
```suggestion
        #[cfg(feature = "cache")]
        if let Some(user_id) = ctx.2 {
```

I think check should not be done at all if the cache feature is not enabled.

src/builder/edit_message.rs
```diff
@@ -254,6 +254,19 @@ impl Builder for EditMessage {
...
+        if let Some(user_id) = ctx.2 {
+            {
+                if let Some(cache) = cache_http.cache() {
+                    let reference_builder = EditMessage::new().suppress_embeds(true);
+
+                    if user_id != cache.current_user().id && self != reference_builder {
+                        return Err(Error::Model(ModelError::InvalidUser));
+                    }
+                }
+            }
```
alexandruradovici:
```suggestion
                if let Some(cache) = cache_http.cache() {
                    let reference_builder = EditMessage::new().suppress_embeds(true);

                    if user_id != cache.current_user().id && self != reference_builder {
                        return Err(Error::Model(ModelError::InvalidUser));
                    }
                }
```

You can delete the inner scope.

iulianalexa: Thank you for the suggestion

alexandruradovici:
Please fix the formatting issue.

arqunis (approved):

PR #2661, opened 2023-12-12
Implement `max_concurrency` support when starting shards

mkrasnitski:
Supercedes #2650.

When a session is first kicked off and we need to start many shards, we can use the value of `max_concurrency` to kick off multiple shards in parallel, and do so in batches every 5 seconds. However, the contents of the batches need to be managed so that we don't accidentally get ratelimited.

Simply put, shards are ratelimited together if their IDs are equivalent modulo `max_concurrency`. For example, if `max_concurrency` is 16, then shards 0 and 16 get ratelimited together, and so do 1 and 17, etc. So, we need to make sure that a given batch doesn't include any conflicting shards that might cause the ratelimit to get hit. If we want a batch of maximal size, we need to wait until the queue contains enough shards to create a full batch.

However, we don't always want to wait for a batch to fill up if we know it's not possible for that to happen. This happens when we restart a single shard - in order to avoid waiting the 5 second loop timeout, we need to add a `bool` field to `ShardQueuerMessage::Shard` to signal whether we want to add the shard to the queue or not.

### DIFF
--- a/src/gateway/bridge/mod.rs
+++ b/src/gateway/bridge/mod.rs
@@ -72,11 +72,11 @@ pub enum ShardQueuerMessage {
     /// Message to set the shard total.
     SetShardTotal(NonZeroU16),
     /// Message to start a shard.
-    Start(ShardId, bool),
+    Start { shard_id: ShardId, concurrent: bool },
     /// Message to shutdown the shard queuer.
     Shutdown,
     /// Message to dequeue/shutdown a shard.
-    ShutdownShard(ShardId, u16),
+    ShutdownShard { shard_id: ShardId, code: u16 },
 }
 
 /// Information about a [`ShardRunner`].
--- a/src/gateway/bridge/mod.rs
+++ b/src/gateway/bridge/mod.rs
@@ -56,7 +56,7 @@ use std::time::Duration as StdDuration;
 pub use self::event::ShardStageUpdateEvent;
 pub use self::shard_manager::{ShardManager, ShardManagerOptions};
 pub use self::shard_messenger::ShardMessenger;
-pub use self::shard_queuer::ShardQueuer;
+pub use self::shard_queuer::{ShardQueue, ShardQueuer};
 pub use self::shard_runner::{ShardRunner, ShardRunnerOptions};
 pub use self::shard_runner_message::ShardRunnerMessage;
 #[cfg(feature = "voice")]
@@ -72,7 +72,7 @@ pub enum ShardQueuerMessage {
     /// Message to set the shard total.
     SetShardTotal(NonZeroU16),
     /// Message to start a shard.
-    Start(ShardId),
+    Start(ShardId, bool),
     /// Message to shutdown the shard queuer.
     Shutdown,
     /// Message to dequeue/shutdown a shard.
--- a/src/gateway/bridge/shard_queuer.rs
+++ b/src/gateway/bridge/shard_queuer.rs
@@ -185,7 +185,7 @@ impl ShardQueuer {
         self.last_start = Some(Instant::now());
     }
 
-    #[instrument(skip(self))]
+    #[cfg_attr(feature = "tracing_instrument", instrument(skip(self)))]
     async fn checked_start_batch(&mut self, shard_ids: Vec<ShardId>) {
         if shard_ids.is_empty() {
             return;
@@ -200,7 +200,7 @@ impl ShardQueuer {
         self.last_start = Some(Instant::now());
     }
 
-    #[instrument(skip(self))]
+    #[cfg_attr(feature = "tracing_instrument", instrument(skip(self)))]
     async fn try_start(&mut self, shard_id: ShardId) {
         if let Err(why) = self.start(shard_id).await {
             warn!("[Shard Queuer] Err starting shard {shard_id}: {why:?}");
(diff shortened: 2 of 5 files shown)


src/gateway/bridge/mod.rs
```diff
@@ -72,7 +72,7 @@ pub enum ShardQueuerMessage {
     /// Message to set the shard total.
     SetShardTotal(NonZeroU16),
     /// Message to start a shard.
-    Start(ShardId),
+    Start(ShardId, bool),
```
GnomedDev: This should be converted to a struct style variant, so the bool can be named.

src/gateway/bridge/shard_queuer.rs
```diff
@@ -102,35 +100,54 @@ impl ShardQueuer {
...
-                Ok(Some(ShardQueuerMessage::SetShardTotal(shard_total))) => {
-                    self.shard_total = shard_total;
-                },
-                Ok(None) => break,
-                Err(_) => {
-                    if let Some(shard) = self.queue.pop_front() {
-                        self.checked_start(shard).await;
-                    }
-                },
+            if let Ok(msg) = timeout(TIMEOUT, self.rx.next()).await {
```
GnomedDev, line 110: Was this destructuring of the Result needed? The extra indentation makes a harder to review diff.

src/gateway/bridge/shard_queuer.rs
```diff
@@ -102,35 +100,54 @@ impl ShardQueuer {
...
+                match msg {
+                    Some(ShardQueuerMessage::SetShardTotal(shard_total)) => {
+                        self.shard_total = shard_total;
+                    },
+                    Some(ShardQueuerMessage::Start(shard_id, concurrent)) => {
+                        if concurrent {
```
GnomedDev, line 119: Non blocker: Can this concurrent bool be part of the pattern, to flatten indentation a bit.

src/gateway/bridge/shard_queuer.rs
```diff
@@ -253,3 +291,51 @@ impl ShardQueuer {
...
+    pub fn push_front(&mut self, shard_id: ShardId) {
+        let bucket = shard_id.0 % self.max_concurrency.get();
+        self.buckets.entry(bucket).or_default().push_front(shard_id);
+    }
+
+    /// Pops a `ShardId` from every bucket containing at least one and returns them all as a `Vec`.
+    pub fn pop_batch(&mut self) -> Vec<ShardId> {
+        (0..self.max_concurrency.get())
+            .filter_map(|i| self.buckets.get_mut(&i).and_then(|bucket| bucket.pop_front()))
+            .collect()
```
GnomedDev, line 333: Non blocker: Using a HashMap here seems kind of inefficient, is there a better data structure like BTreeMap that would be automatically sorted you can use?

GnomedDev (changes requested):
I'm yet to test this out on my bot, but here's my first pass review.

src/gateway/bridge/shard_queuer.rs
mkrasnitski: I felt the nested patterns such as `Ok(Some(...))` and `Ok(None)` didn't read very well. And having the `Err(_)` case along with them didn't read well as the fall-through case for the timeout. The first commit in this PR just destructures the Result and changes indentation.

src/gateway/bridge/shard_queuer.rs
mkrasnitski: Do you mean splitting the match into two separate arms with guards, one for `if concurrent` and one for `if !concurrent`? That seems less readable, and also doesn't satisfy the compiler for exhaustive matching.

EDIT: Using `true` and `false` literals satisfies the compiler for exhaustive matching.

src/gateway/bridge/shard_queuer.rs
mkrasnitski: In what sense would a `BTreeMap` be more efficient? Insertions and lookups are constant for HashMaps, not so for `BTreeMap` - and I don't really care about the order of buckets here.

src/gateway/bridge/shard_queuer.rs
GnomedDev: Fair.

src/gateway/bridge/shard_queuer.rs
GnomedDev: I do mean true or false literals.

src/gateway/bridge/shard_queuer.rs
GnomedDev: Yeah, fair enough, I guess I saw iteration and hashmap and thought inefficient..

src/gateway/bridge/shard_queuer.rs
mkrasnitski: Unfortunately this won't play that well with struct-style variants, because rustfmt wants to indent the fields in the match arm, so you end up with something like:
```rust
Some(ShardQueuerMessage::Start {
    shard_id,
    concurrent: true,
}) => {
    /// ...
},
Some(ShardQueuerMessage::Start {
    shard_id,
    concurrent: false,
}) => {
    /// ...
},
```
As opposed to:
```rust
Some(ShardQueuerMessage::Start {
    shard_id,
    concurrent,
}) => {
    if concurrent {
        /// ...
    } else {
        /// ...
    }
},

src/gateway/bridge/mod.rs
mkrasnitski: Should I change `ShutdownShard` into a struct variant as well, since it has more than one parameter?

GnomedDev: Yes

src/gateway/bridge/shard_queuer.rs
GnomedDev: Oh that sucks, pick your poison.

arqunis (approved):

PR #2665, opened 2023-12-17
Encode timestamps in RFC3339

### LINKED ISSUE #2664: The `get_channel_archived_public_threads` method doesn't encode the timestamp well
While using the `Http::get_channel_archived_public_threads` method I got an error due to a miss-encoded `before` argument. The method asks for an `Option<u64>` but should rather ask for an `Option<Timestamp>` and encode it in ISO8601 as [described in the documentation](https://discord.com/developers/docs/resources/channel#list-public-archived-threads).

```
Error: Invalid Form Body (before: Could not parse 1700059153. Should be ISO8601.)

Caused by:
    Invalid Form Body (before: Could not parse 1700059153. Should be ISO8601.)
```

Kerollmops:
This PR fixes #2664, but I am unsure if it should be unwrapped. It would be better to output an error if we cannot encode them. Also, please tell me if [it is ok to encode in RFC3339 instead of ISO 8601](https://stackoverflow.com/a/522281/1941280).

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2799,7 +2799,7 @@ impl Http {
     pub async fn get_channel_archived_public_threads(
         &self,
         channel_id: ChannelId,
-        before: Option<u64>,
+        before: Option<Timestamp>,
         limit: Option<u64>,
     ) -> Result<ThreadsData> {
         let mut params = ArrayVec::<_, 2>::new();
@@ -2827,7 +2827,7 @@ impl Http {
     pub async fn get_channel_archived_private_threads(
         &self,
         channel_id: ChannelId,
-        before: Option<u64>,
+        before: Option<Timestamp>,
         limit: Option<u64>,
     ) -> Result<ThreadsData> {
         let mut params = ArrayVec::<_, 2>::new();
@@ -2855,7 +2855,7 @@ impl Http {
     pub async fn get_channel_joined_archived_private_threads(
         &self,
         channel_id: ChannelId,
-        before: Option<u64>,
+        before: Option<ChannelId>,
         limit: Option<u64>,
     ) -> Result<ThreadsData> {
         let mut params = ArrayVec::<_, 2>::new();
--- a/src/model/channel/channel_id.rs
+++ b/src/model/channel/channel_id.rs
@@ -1022,7 +1022,7 @@ impl ChannelId {
     pub async fn get_archived_private_threads(
         self,
         http: impl AsRef<Http>,
-        before: Option<u64>,
+        before: Option<Timestamp>,
         limit: Option<u64>,
     ) -> Result<ThreadsData> {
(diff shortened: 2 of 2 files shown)


src/http/client.rs
```diff
@@ -2828,12 +2828,12 @@ impl Http {
...
         &self,
         channel_id: ChannelId,
-        before: Option<u64>,
+        before: Option<Timestamp>,
         limit: Option<u64>,
     ) -> Result<ThreadsData> {
         let mut params = vec![];
         if let Some(before) = before {
-            params.push(("before", before.to_string()));
+            params.push(("before", before.to_rfc3339().unwrap()));
```
mkrasnitski: The implementation of `Display` for `Timestamp` already calls `to_rfc3339`, so we don't actually need to change anything here.

src/http/client.rs
```diff
@@ -2884,12 +2884,12 @@ impl Http {
     pub async fn get_channel_joined_archived_private_threads(
         &self,
         channel_id: ChannelId,
-        before: Option<u64>,
+        before: Option<Timestamp>,
```
mkrasnitski: This should actually be `Option<ChannelId>`.

src/model/channel/channel_id.rs
```diff
@@ -1049,7 +1049,7 @@ impl ChannelId {
     pub async fn get_joined_archived_private_threads(
         self,
         http: impl AsRef<Http>,
-        before: Option<u64>,
+        before: Option<Timestamp>,
```
mkrasnitski: Same here - should be `Option<ChannelId>`.

mkrasnitski (changes requested):
Thanks for the PR! A few review concerns:

1. This PR changes the signatures of public-facing functions and is therefore breaking. As such, it should target the `next` branch rather than `current`.
2. According to Discord's docs, the [`List Joined Private Archived Threads`](https://discord.com/developers/docs/resources/channel#list-joined-private-archived-threads) endpoint expects the `before` parameter to be a snowflake rather than a timestamp. Since we're dealing with threads here, I think `ChannelId` would be the right type to use.
3. You've missed an endpoint here, in particular the [`List Private Archived Threads`](https://discord.com/developers/docs/resources/channel#list-private-archived-threads) endpoint, which actually _does_ expect a formatted timestamp.

Kerollmops:
Thank you for your review 🙏

I rebased on the `next` branch and used `ChannelId`s instead of `Timestamp`s or `u64`s.

mkrasnitski:
Ah, sorry for not being clear enough. It's only the `List Joined Private Archived Threads` endpoint that takes a snowflake. The other two endpoints do indeed take timestamps.

Kerollmops:
Oups, sorry! I just fixed this an squashed the commits.

mkrasnitski (approved):

arqunis (approved):

PR #2677, opened 2023-12-31
Optimize snowflake parsing

bend-n:
snow

### DIFF
--- a/src/model/id.rs
+++ b/src/model/id.rs
@@ -5,6 +5,22 @@ use std::num::{NonZeroI64, NonZeroU64};
 
 use super::Timestamp;
 
+pub struct IDFromStrError;
+
+impl fmt::Debug for IDFromStrError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "IDFromStrError(id of zero, or too long)")
+    }
+}
+
+impl fmt::Display for IDFromStrError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Attempted to construct a id of zero, or id too long.")
+    }
+}
+
+impl std::error::Error for IDFromStrError {}
+
 macro_rules! id_u64 {
     ($($name:ident;)*) => {
         $(
@@ -107,10 +123,10 @@ macro_rules! id_u64 {
             }
 
             impl std::str::FromStr for $name {
-                type Err = ();
+                type Err = IDFromStrError;
 
                 fn from_str(s: &str) -> Result<Self, Self::Err> {
-                    Ok(Self(snowflake::parse(s).ok_or(())?))
+                    Ok(Self(snowflake::parse(s).ok_or(IDFromStrError)?))
                 }
             }
 
@@ -375,7 +391,7 @@ pub(crate) mod snowflake {
             // being UB is faster, but unsounder
             return None;
         }
-        Some(match as_chunks(x.as_ref()) {
+        NonZeroU64::new(match as_chunks(x.as_ref()) {
             // 16, ..4 (most flakes are length 18, so this is on top)
             (&[a, b], rest) => simple_parse(parse_eight(a) * 100000000 + parse_eight(b), rest),
             // 8, 4..8
@@ -387,7 +403,6 @@ pub(crate) mod snowflake {
             // 24
             _ => unsafe { std::hint::unreachable_unchecked() },
         })
-        .and_then(NonZeroU64::new)
     }
 
     pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<NonZeroU64, D::Error> {
--- a/src/model/id.rs
+++ b/src/model/id.rs
@@ -9,13 +9,13 @@ pub struct IDFromStrError;
 
 impl fmt::Debug for IDFromStrError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "IDFromStrError(id of zero, or too long)")
+        write!(f, "IDFromStrError(id of zero)")
     }
 }
 
 impl fmt::Display for IDFromStrError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "Attempted to construct a id of zero, or id too long.")
+        write!(f, "Attempted to construct a id of zero.")
     }
 }
 
@@ -386,23 +386,13 @@ pub(crate) mod snowflake {
 
     // this parse is 4x faster than [`str::parse`], see <https://github.com/serenity-rs/serenity/pull/2677#issue-2060912973>
     pub fn parse(x: &(impl AsRef<[u8]> + ?Sized)) -> Option<NonZeroU64> {
-        if x.as_ref().len() > 19 {
-            // SAFETY: max snowflake is 9223372036854775807
-            // unsafe { std::hint::unreachable_unchecked() }
-            // being UB is faster, but unsounder
-            return None;
-        }
         NonZeroU64::new(match as_chunks(x.as_ref()) {
             // 16, ..4 (most flakes are length 18, so this is on top)
             (&[a, b], rest) => simple_parse(parse_eight(a) * 100000000 + parse_eight(b), rest),
             // 8, 4..8
             (&[a], rest) => simple_parse(parse_eight(a), rest),
-            // omitting this shaves a ns
-            (&[], rest) => simple_parse(0, rest),
-            // SAFETY: as the max snowflake length is 19, it can either be split into [a, b](16),
-            // rest, [a](8), rest or [], rest. [a, b, c], rest would require a snowflake of length
-            // 24
-            _ => unsafe { std::hint::unreachable_unchecked() },
+            // largest number is 20 digits, if this is a problem you have bigger problems
+            (_, rest) => simple_parse(0, rest),
         })
     }
 
--- a/src/model/id.rs
+++ b/src/model/id.rs
@@ -107,10 +107,10 @@ macro_rules! id_u64 {
             }
 
             impl std::str::FromStr for $name {
-                type Err = <u64 as std::str::FromStr>::Err;
+                type Err = ();
 
                 fn from_str(s: &str) -> Result<Self, Self::Err> {
-                    Ok(Self(s.parse()?))
+                    Ok(Self(snowflake::parse(s).ok_or(())?))
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/model/id.rs
```diff
@@ -331,6 +331,54 @@ pub(crate) mod snowflake {
...
+            // SAFETY: max snowflake is 9223372036854775807
+            // unsafe { std::hint::unreachable_unchecked() }
+            // being UB is faster, but unsounder
+            return None;
+        }
+        fn parse(s: [u8; 8]) -> u64 {
+            let n = u64::from_ne_bytes(s);
+            let n = ((n & 0x0f000f000f000f00) >> 8) + ((n & 0x000f000f000f000f) * 10);
+            let n = ((n & 0x00ff000000ff0000) >> 16) + ((n & 0x000000ff000000ff) * 100);
+            ((n & 0x0000ffff00000000) >> 32) + ((n & 0x000000000000ffff) * 10000)
```
arqunis: Could this be given a comment on the bit wizardry happening here? As in, what does it rely upon to make this work and what it produces? Otherwise, without deciphering the meaning of the numbers, a person just sees magic numbers.

src/model/id.rs
```diff
@@ -331,6 +331,61 @@ pub(crate) mod snowflake {
...
+                n = n * 10 + (b - b'0') as u64;
+            }
+            n
+        }
+
+        if x.as_ref().len() > 19 {
+            #[cfg(test)]
+            {
+                return NonZeroU64::new(generic(0, x.as_ref()));
+            }
```
arqunis: This is a band-aid for a test that was broken by this change. The test itself should be fixed to account for the parsing changes (i.e. pick a number that can fit in 19 decimal digits).

src/model/id.rs
```diff
@@ -331,6 +331,61 @@ pub(crate) mod snowflake {
...
+            // 8, 4..8
+            (&[a], rest) => generic(parse(a), rest),
+            // omitting this shaves a ns
+            (&[], rest) => generic(0, rest),
+            // SAFETY: as the max snowflake length is 19, it can either be split into [a, b](16),
+            // rest, [a](8), rest or [], rest. [a, b, c], rest would require a snowflake of length
+            // 24
+            _ => unsafe { std::hint::unreachable_unchecked() },
+        })
+        .and_then(NonZeroU64::new)
```
arqunis: We're not in a functional language limited to one expression in a function.
```suggestion
        let n = match as_chunks(x.as_ref()) {
            // 16, ..4 (most flakes are length 18, so this is on top)
            (&[a, b], rest) => generic(parse(a) * 100000000 + parse(b), rest),
            // 8, 4..8
            (&[a], rest) => generic(parse(a), rest),
            // omitting this shaves a ns
            (&[], rest) => generic(0, rest),
            // SAFETY: as the max snowflake length is 19, it can either be split into [a, b](16),
            // rest, [a](8), rest or [], rest. [a, b, c], rest would require a snowflake of length
            // 24
            _ => unsafe { std::hint::unreachable_unchecked() },
        };
        
        NonZeroU64::new(n)
```

src/model/id.rs
```diff
@@ -331,6 +331,61 @@ pub(crate) mod snowflake {
...
+    }
+
+    pub fn parse(x: &(impl AsRef<[u8]> + ?Sized)) -> Option<NonZeroU64> {
+        #[allow(clippy::items_after_statements)]
+        fn generic(mut n: u64, s: &[u8]) -> u64 {
+            for &b in s {
+                n = n * 10 + (b - b'0') as u64;
+            }
+            n
+        }
```
arqunis: I'd hoist `generic` and the innermost `parse` as private functions of the `snowflake` module to reduce noise within the outermost `parse` function (thus avoiding the warning from clippy).

src/model/id.rs
```diff
@@ -331,6 +331,61 @@ pub(crate) mod snowflake {
...
+        assert!(N != 0, "chunk size must be non-zero");
+        let len = slice.len() / N;
+        let (multiple_of_n, remainder) = slice.split_at(len * N);
+        let new_len = multiple_of_n.len() / N;
+        // SAFETY: We cast a slice of `new_len * N` elements into
+        // a slice of `new_len` many `N` elements chunks.
+        (unsafe { std::slice::from_raw_parts(multiple_of_n.as_ptr().cast(), new_len) }, remainder)
+    }
+
+    pub fn parse(x: &(impl AsRef<[u8]> + ?Sized)) -> Option<NonZeroU64> {
```
arqunis: This should have a comment explaining that compared to the solution offered by the standard library (`str::parse`), this is 4x as fast (ns-wise), to justify the complex code for parsing snowflakes.

src/model/id.rs
```diff
@@ -112,10 +112,10 @@ macro_rules! id_u64 {
             }
 
             impl std::str::FromStr for $name {
-                type Err = <u64 as std::str::FromStr>::Err;
+                type Err = ();
```
arqunis: This should refer to a type implementing the `std::error::Error` trait; `()` isn't one. You can use a simple unit struct.

arqunis (changes requested):

src/model/id.rs
bend-n: But.. I like my functional code :(

arqunis: If the `match` were a function that returned `Option<u64>`, then sure, I would've allowed it. But, no, don't just abuse `Option`'s combinators.

bend-n: what if i wrap the whole thing in NonZeroU64::new (the and_then thing is a bit of a mistake)

arqunis: That's okay, then. The `match` isn't too big.

src/model/id.rs
```diff
@@ -5,6 +5,22 @@ use std::num::{NonZeroI64, NonZeroU64};
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "IDFromStrError(id of zero)")
+    }
+}
+
+impl fmt::Display for IDFromStrError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Attempted to construct a id of zero.")
+    }
+}
```
arqunis:
```suggestion
#[derive(Debug, Clone)]
pub struct IDFromStrError;

impl fmt::Display for IDFromStrError {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.write_str("attempted to construct an ID of zero")
    }
}
```

Otherwise, this is okay to merge.

arqunis (changes requested):

src/model/id.rs
```diff
@@ -331,6 +347,61 @@ pub(crate) mod snowflake {
...
+
+    // this parse is 4x faster than [`str::parse`], see <https://github.com/serenity-rs/serenity/pull/2677#issue-2060912973>
+    pub fn parse(x: &(impl AsRef<[u8]> + ?Sized)) -> Option<NonZeroU64> {
+        NonZeroU64::new(match as_chunks(x.as_ref()) {
+            // 16, ..4 (most flakes are length 18, so this is on top)
+            (&[a, b], rest) => simple_parse(parse_eight(a) * 100000000 + parse_eight(b), rest),
+            // 8, 4..8
+            (&[a], rest) => simple_parse(parse_eight(a), rest),
+            // largest number is 20 digits, if this is a problem you have bigger problems
+            (_, rest) => simple_parse(0, rest),
```
arqunis, line 379: > Otherwise, this is okay to merge.

Actually, could you fall back to `str::parse` here or handle the integer overflow yourself for this case? The test has a dummy snowflake that's too big to parse into a `u64` (or just change the snowflake to be smaller).

arqunis (changes requested):

src/model/id.rs
bend-n: the test has a digit of length 24?

bend-n: oh, `123123123123123123123`.

hmm, with `str::parse` it throws `Err(ParseIntError { kind: PosOverflow })`, but i really cant afford that

bend-n: So, im looking at some tests:

https://github.com/serenity-rs/serenity/blob/c5e6b3d923043e0028b8ef3c2315becbafd0670d/src/utils/mod.rs#L130-L131

I think we have a problem.

src/model/id.rs
```diff
@@ -112,10 +123,13 @@ macro_rules! id_u64 {
...
             impl std::str::FromStr for $name {
-                type Err = <u64 as std::str::FromStr>::Err;
+                type Err = IDFromStrError;
 
                 fn from_str(s: &str) -> Result<Self, Self::Err> {
-                    Ok(Self(s.parse()?))
+                    #[cfg(debug_assertions)]
+                    { return Ok(Self(s.parse::<u64>().ok().and_then(NonZeroU64::new).ok_or(IDFromStrError)?)) }
+                    #[cfg(not(debug_assertions))]
+                    { return Ok(Self(snowflake::parse(s).ok_or(IDFromStrError)?)) }
```
GnomedDev: I dislike this, we should always rely on the dangerous path when there are debug assertions as otherwise that path will hide bugs.

src/model/id.rs
```diff
@@ -331,6 +345,61 @@ pub(crate) mod snowflake {
...
 
+    /// impl from std ([T]::as_chunks)
+    const fn as_chunks<T, const N: usize>(slice: &[T]) -> (&[[T; N]], &[T]) {
+        assert!(N != 0, "chunk size must be non-zero");
+        let len = slice.len() / N;
+        let (multiple_of_n, remainder) = slice.split_at(len * N);
+        let new_len = multiple_of_n.len() / N;
+        // SAFETY: We cast a slice of `new_len * N` elements into
+        // a slice of `new_len` many `N` elements chunks.
+        (unsafe { std::slice::from_raw_parts(multiple_of_n.as_ptr().cast(), new_len) }, remainder)
```
GnomedDev: Is there no way to get rid of this unsafe?

GnomedDev (changes requested):

src/model/id.rs
bend-n: This comes from std-- as far as im concerned, this isnt any more unsafe than using a Vec.

src/model/id.rs
bend-n: I, uh, cant.

src/model/id.rs
bend-n: i can add a `// FIXME replace when [<[T]>::as_chunks](https://github.com/rust-lang/rust/issues/74985)` is stabilized

bend-n: Or i can add a dependency to bytemuck

GnomedDev: I personally recommend `bytemuck`, tbh

bend-n: Really?

GnomedDev: Yes, a discord bot library should not be using unsafe.

bend-n: Its the same thing though...

src/model/id.rs
GnomedDev: What do you mean? You can't remove the `#[cfg]`?

bend-n: It would break this test

https://github.com/serenity-rs/serenity/blob/c5e6b3d923043e0028b8ef3c2315becbafd0670d/src/utils/mod.rs#L130-L131

GnomedDev: So the solution to a change breaking a test is to obscure the bug behind a `#[cfg(not(test)]`?

bend-n: You have a good point.

src/model/id.rs
0xpr03: > same thing though

except there are more eyes on the bytemuck unsafe part, and thus if bytemuck gets something fixed, so does serenity - which isn't true if everyone just copy-pastes unsafe stuff that could be a crate

bend-n: i turned it into a crate, [`pieced`](https://docs.rs/pieced)

GnomedDev:
Closing after discussion on Discord. This PR reduces correctness for speed in an unacceptable way, as invalid IDs are considered valid but mangled data.

PR #2678, opened 2023-12-31
Add unusual_dm_activity_until to Member and PartialMember

jamesbt365:
I'm assuming this change doesn't count as a breaking change in any way that you consider so, but if it needs to be moved to next the base can be changed cleanly.

There is currently no docs for this feature so this is done from what I can test. I can show some sample member objects if that makes it easier to prove this is fine.

I'm not the most familiar with the serenity codebase so I hope I didn't miss anything immediately obvious.

### DIFF
--- a/src/model/guild/member.rs
+++ b/src/model/guild/member.rs
@@ -61,7 +61,7 @@ pub struct Member {
     /// The unique Id of the guild that the member is a part of.
     #[serde(default)]
     pub guild_id: GuildId,
-    /// If the member is current flagged for sending excessive DMs to non-friend server members
+    /// If the member is currently flagged for sending excessive DMs to non-friend server members
     /// in the last 24 hours.
     ///
     /// Will be None if the user is not currently flagged.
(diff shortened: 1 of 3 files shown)


src/model/guild/member.rs
```diff
@@ -61,6 +61,11 @@ pub struct Member {
     /// The unique Id of the guild that the member is a part of.
     #[serde(default)]
     pub guild_id: GuildId,
+    /// If the member is current flagged for sending excessive DMs to non-friend server members
```
GnomedDev:
```suggestion
    /// If the member is currently flagged for sending excessive DMs to non-friend server members
```

src/model/guild/member.rs
```diff
@@ -592,6 +597,11 @@ pub struct PartialMember {
     ///
     /// [`Interaction`]: crate::model::application::Interaction
     pub permissions: Option<Permissions>,
+    /// If the member is current flagged for sending excessive DMs to non-friend server members
```
GnomedDev: Ditto

arqunis (approved):

GnomedDev (changes requested):

src/model/guild/member.rs
jamesbt365: crap

PR #2679, opened 2024-01-01
Add UserPublicFlags::SPAMMER

jamesbt365:
For the most part this is stealing off discord.py.

https://github.com/Rapptz/discord.py/blob/078b500657c651f72a6ca78ada23332422e23adf/discord/enums.py#L492-L512

The only one I can actually verify what it means/its the right value is spammer, which i have tested and it lines up correctly. If you don't want potentially wrong flags, I can remove all but the spammer flag which I have personally tested.

I'm only actually bothered about the spammer flag anyway.

Wasn't sure how to match up with the current documentation, "flag as" instead of "flag for" in these cases looked a bit weird, so I'm open to suggestions there.

If this should be moved to `next` it should be a clean change without any additional messing around.

### DIFF
--- a/src/model/user.rs
+++ b/src/model/user.rs
@@ -349,6 +349,9 @@ bitflags! {
         const DISCORD_CERTIFIED_MODERATOR = 1 << 18;
         /// Bot's running with HTTP interactions
         const BOT_HTTP_INTERACTIONS = 1 << 19;
+        /// User's flag for suspected spam activity.
+        #[cfg(feature = "unstable_discord_api")]
+        const SPAMMER = 1 << 20;
         /// User's flag as active developer
         const ACTIVE_DEVELOPER = 1 << 22;
     }


jamesbt365:
image
when looking into why the spammer flag isn't documented, I guess discord devs are kinda unreasonable thats why.

image
but this wouldn't explain why it still happens now AND has been added to the member tab for mods to view. Also in some cases I've seen people get it and *never* get scheduled for deletion, never lost access to the account ever.

mkrasnitski:
Shouldn't we just adhere to what the Discord docs say? If they don't see a use for undocumented flags to anyone but themselves, should we care?

jamesbt365:
I'm not bothered about support for anything but the spammer flag here, I don't understand why Discord doesn't seem the utility here, and what they've said about active users having it being a bug doesn't really line up to the fact that they recently allowed you to filter by it in the new members tab, you could only filter by timeout and dm activity a couple weeks ago and now you can filter by "unusual account activity" too.
image


If what they said was true and they were "fixing it in a couple weeks" and it still happens now either they did fix this, which wouldn't line up to the fact that I saw a user talking with this flag yesterday, or the feature was repurposed into what it is now.
image



Why would discord allow you to filter by something that is strictly attached to disabled accounts? Nothing really lines up in regard to that.

Adhering to what the Discord docs say would mean removing a feature that is actually used and has utility behind it.

I'm happy to remove everything but the spammer flag, as I only added the other flags while I was looking into how other libraries handle the flag. I don't want the spammer flag to not be included simply because Discord devs for *some* unclear reason deny documentation for a feature that they are supporting now more than ever.

mkrasnitski:
Are you sure it's not filtering according to the `unusual_dm_activity_until` member flag? Also, I just realized that field isn't documented anywhere either, so I'm worried if we got the name right.

EDIT: I'm slightly blind.

GnomedDev (changes requested):
I am happy to merge if these are locked behind the `unstable_discord_api` feature.

jamesbt365:
I'm not too confident on the other flags I stole from discord.py, so its probably best I remove those too, I'll get around to this hopefully tomorrow.

jamesbt365:
Decided to only do the SPAMMER flag because its the only one I see utility in and can actually prove the documentation of/what it actually does.

PR #2681, opened 2024-01-02
Use nonmax::NonMax* types in Option where possible

GnomedDev:
Swaps fields that store `Option<Int>` for `Option<NonMaxInt>` where the maximum value would be ludicrous. Since `nonmax` uses `NonZero` internally, this gives us niche opts, so model sizes can drop some more. <https://www.diffchecker.com/Hr8ZqANM/>

I've had to include a workaround for https://github.com/LPGhatguy/nonmax/issues/17 in `optional_string` by making my own `TryFrom<u64>`, so that should be removable once that issue is fixed.


GnomedDev:
Thanks to @TestingPlant on the Rust Community Server for the idea to use nonmax, by the way.

arqunis (approved):

PR #2688, opened 2024-01-05
Rewrite builders to take Cows

GnomedDev:
`String` -> `Cow<'a, str>`
`Vec<T>` -> `Cow<'a, [T]>`
`impl IntoIterator<...>` -> `Cow<'a, [T]>`

### DIFF
--- a/src/builder/create_allowed_mentions.rs
+++ b/src/builder/create_allowed_mentions.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::HashSet;
 
 use serde::{Deserialize, Serialize};
@@ -20,17 +21,20 @@ enum ParseValue {
 /// ```rust,no_run
 /// # use serenity::builder::CreateMessage;
 /// # use serenity::model::channel::Message;
+/// # use serenity::model::id::*;
 /// #
 /// # async fn run() -> Result<(), Box<dyn std::error::Error>> {
 /// use serenity::builder::CreateAllowedMentions as Am;
 ///
 /// // Mention only the user 110372470472613888
 /// # let m = CreateMessage::new();
-/// m.allowed_mentions(Am::new().users(vec![110372470472613888]));
+/// m.allowed_mentions(Am::new().users([UserId::new(110372470472613888)].as_slice()));
 ///
 /// // Mention all users and the role 182894738100322304
 /// # let m = CreateMessage::new();
-/// m.allowed_mentions(Am::new().all_users(true).roles(vec![182894738100322304]));
+/// m.allowed_mentions(
+///     Am::new().all_users(true).roles([RoleId::new(182894738100322304)].as_slice()),
+/// );
 ///
 /// // Mention all roles and nothing else
 /// # let m = CreateMessage::new();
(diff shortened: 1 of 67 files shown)


arqunis (approved):

src/builder/create_allowed_mentions.rs
```diff
@@ -103,29 +109,29 @@ impl CreateAllowedMentions {
 
     /// Sets the *specific* users that will be allowed mentionable.
     #[inline]
-    pub fn users(mut self, users: impl IntoIterator<Item = impl Into<UserId>>) -> Self {
-        self.users = users.into_iter().map(Into::into).collect();
+    pub fn users(mut self, users: impl Into<Cow<'a, [UserId]>>) -> Self {
+        self.users = users.into();
```
mkrasnitski, line 113: Sorry for the late review on this. I feel like this is an ergonomics hit, since the `Into<UserId>` is gone.

GnomedDev: I dislike the Into Id signatures personally, as it ruins the strict typing of the ID types. Since this is a performance hit for a signature style I don't like, I removed it.

mkrasnitski: There's only a `From<u64>` impl on IDs. Since there's no way to implicitly cast an Id of one kind to a different one, I don't see how the strict typing is hurt. Being able to just pass in literals is a boon to ergonomics, and still guarantees that a list of multiple will all be converted into the same kind of Id.

GnomedDev: I'm not saying it allows for converting one type of ID to another, I believe that `Id::new` should be a boundary that needs explicitly typing to allow for the second of "wait, is this actually a user id".

mkrasnitski: As opposed to what? Not many constants will look like IDs, unless maybe we're talking about UNIX timestamps. Plus, the builder methods are named well enough that they clue you in already.

What looks better to you?
```rust
m.allowed_mentions(Am::new().users([UserId::new(110372470472613888), UserId::new(1103724704727656385), UserId::new(110372470472918472)].as_slice()));
```
vs
```rust
m.allowed_mentions(Am::new().users(vec![110372470472613888, 1103724704727656385, 110372470472918472]));
```

Another thing: calling `.into` on each ID is equivalent to the first example but is no longer explicit about the Id kind:
```rust
m.allowed_mentions(Am::new().users([110372470472613888.into(), 1103724704727656385.into(), 110372470472918472.into()].as_slice()));
```

Mapping over the entire array is technically possible, but is still longer than passing in just the array:
```rust
m.allowed_mentions(Am::new().users([110372470472613888, 1103724704727656385 110372470472918472].map(UserId::from).as_slice()));
```

So, both sides of the `Into<Cow>` are resultantly longer/noisier. Besides, aren't builders relatively short-lived objects? They aren't cached in any way, so the only real performance hit is the cost of creating a few allocations. These aren't large objects either, so those allocator calls are not going to be expensive.

GnomedDev: Small allocations still cause fragmentation and are just absolutely not needed. I agree this is a hit for passing u64s in, but in real bots these Ids should be in `const`s or config files anyway, I just believe this example shows the worst case scenario, which isn't even too bad.

PR #2689, opened 2024-01-05
Swap Id from NonZero to NonMax

GnomedDev:
Discord seems to internally default Ids to 0, which is a bug whenever exposed, but this makes ID parsing more resilient. I also took the liberty to remove the NonZero From implementations, to prevent future headaches, as it was impossible to not break public API as we exposed NonZero in Id::parse.

### DIFF
--- a/src/model/id.rs
+++ b/src/model/id.rs
@@ -1,9 +1,20 @@
 //! A collection of newtypes defining type-strong IDs.
 
 use std::fmt;
-use std::num::{NonZeroI64, NonZeroU64};
+
+use nonmax::NonMaxU64;
 
 use super::Timestamp;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ParseIdError(nonmax::ParseIntError);
+
+impl std::error::Error for ParseIdError {}
+impl std::fmt::Display for ParseIdError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
 
 macro_rules! id_u64 {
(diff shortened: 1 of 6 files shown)


src/model/id.rs
```diff
@@ -1,32 +1,43 @@
...
+                        None => panic!(concat!("Attempted to call ", stringify!($name), "::new with invalid (u64::MAX) value"))
                     }
                 }
 
                 /// Retrieves the inner `id` as a [`u64`].
                 #[inline]
                 #[must_use]
                 pub const fn get(self) -> u64 {
-                    self.0.get()
+                    {self.0}.get()
```
arqunis, line 43: Are the braces for optimisation? If so, add a comment explaining their purpose to prevent other contributors from removing them.

src/model/id.rs
```diff
@@ -77,20 +76,7 @@ macro_rules! id_u64 {
...
-            impl From<$name> for NonZeroU64 {
-                fn from(id: $name) -> NonZeroU64 {
-                    id.0
-                }
-            }
-
-            impl From<$name> for NonZeroI64 {
-                fn from(id: $name) -> NonZeroI64 {
-                    NonZeroI64::new(id.get() as i64).unwrap()
+                    fmt::Display::fmt(&{self.0}, f)
```
arqunis, line 83: Ditto.

arqunis (changes requested):

src/model/id.rs
GnomedDev: I've added a comment explaining it.

arqunis (approved):

arqunis (approved):

PR #2695, opened 2024-01-07
Revert `Request::params` back to `Option`

mkrasnitski:
Follow-up to #2694.

When `Request::params` was made into an ArrayVec, the `Option` around it was removed in order to avoid having to add a turbofish on `None` to specify the value of `MAX_PARAMS`. Also, `Request::new` also needed to be changed so that the value of `MAX_PARAMS` could be inferred. Now that the field is a slice again, we can wrap it in `Option` again (at no cost to size, thanks to niche opts).

We ensure we never store the redundant `Some(&[])` by checking for an empty slice and storing `None` instead. This way, we ensure we never read an empty slice out of the `Some` variant.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -240,7 +240,7 @@ impl Http {
                     guild_id,
                     user_id,
                 },
-                params: &[],
+                params: None,
             })
             .await?;
 
@@ -273,7 +273,7 @@ impl Http {
                 role_id,
                 user_id,
             },
-            params: &[],
+            params: None,
         })
         .await
     }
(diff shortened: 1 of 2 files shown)


mkrasnitski:
Unsure if using `<[_; 1]>::as_slice` is preferable to just mapping `AsRef::as_ref`.

GnomedDev (approved):
Looks good, wasn't too sure if to implement as it leads to the none vs Some(&[]) confusion, but that's a fine fix. Part of me wants a NonEmptySlice that uses NonZero for the length, but that would be entirely a waste of time for this case.

mkrasnitski:
Yeah, it's not like the type system disallows `Some(&[])`, but we can easily still prevent it at runtime. Also, optional slices are already niche optimized, so not really any benefit to doing it.

arqunis (approved):

PR #2703, opened 2024-01-09
Use NonMax to signal builder range constraints

GnomedDev:


mkrasnitski:
Are we sure we want to expose a third-party type to the user like this? What's it like to convert from primitive ints to `NonMax` versions?

GnomedDev:
We have already exposed it to users in many model types, constructing them is the same as NonZero.

mkrasnitski:
At least for those it's easy to just call `.into()` on them. However, constructing them requires importing the type from `nonmax`. Also, we're not even limiting the value ranges (except for `MAX`), only signalling that Discord has a narrower range than all of `u16` for instance. IMO in terms of using "ranged ints", this feels like a half-measure.

GnomedDev:
It is a half measure, and signalling that discord has a narrower range than u16 is exactly what is required to get someone to read the docs. It isn't easy to just call .into() for non zero types, that isn't an impl.

mkrasnitski:
What do you mean? Every `NonMax*` provides an implementation of `impl From<NonMax*> for *`

GnomedDev:
Ah, you meant getting the value out of a NonMax.

arqunis (approved):

Issue #2479, opened 2023-06-29
Could not establish a websocket connection through a proxy


forsummer:
Unable to establish a websocket connection through a proxy, the code is as follows:
```rust
let intents = GatewayIntents::MESSAGE_CONTENT |
    GatewayIntents::GUILD_MESSAGES |
    GatewayIntents::DIRECT_MESSAGES;

let http = HttpBuilder::new(bot_token)
    .application_id(app_id)
    .proxy("http://localhost:7890")
    .map_or_else(|e| panic!("{}", e), |h| h.build());

let mut bot = ClientBuilder::new_with_http(http, intents)
    .event_handler(Handler)
    .await
    .expect("Build bot client failed");

bot.start()
    .await
    .expect("Bot connect success");
```
The error message is as follows:
```
Build bot client failed: Http(Request(reqwest::Error { kind: Request, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("discord.com")), port: None, path: "/api/v10/gateway", query: None, fragment: None }, source: hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 110, kind: TimedOut, message: "Connection timed out" })) }))
```

**PS: Strangely enough, websocket connections can be established normally using `wscat` (via http proxy)**

PR #2709, opened 2024-01-12
Add ws_proxy option for connecting gateway through a proxy.

zzzzRuby:
#2479


cheesycod:
This would help me as well. I forked nirn to hackily replace the ``ws`` url.

GnomedDev:
I'm kinda wondering why this is required at all, can't the gateway url just be exposed (if not already) to point to the proxy, instead of having this proxy option, and why is the manual http-over-tcp required? None of this is commented or explained.

zzzzRuby:
> I'm kinda wondering why this is required at all, can't the gateway url just be exposed (if not already) to point to the proxy, instead of having this proxy option, and why is the manual http-over-tcp required? None of this is commented or explained.

Sorry I'm new to opensource.
In some network environments like Chinese networks, there must be a proxy to connect with discord gateway. There're two options to do this, the first one is open a tunnel to the proxy likes this pr does, the second one is creating a custom gateway and redirect all message through a tunnel to the proxy. The second one seems too complex.
As to the manual http-over-tcp, I'm not sure if I can add a new dependency in Cargo.toml.

GnomedDev:
Can you not perform this proxying system wide, instead of coding support for it into every application? I'm honestly not familar with this.

zzzzRuby:
> Can you not perform this proxying system wide, instead of coding support for it into every application? I'm honestly not familar with this.

Normally, libraries like libcurl or reqwest get a default proxy from environment or system. But tungstenite doesn't support this and they seem have no interest in this (snapview/tungstenite-rs#177), the only choice to perform a proxy system wide is proxychains-like-tools. But I don't think proxychains is a good idea, it depends on linking libc dynamically, so on some distro like alpine it's hard to use with rust program and sometimes not work.

src/client/mod.rs
```diff
@@ -293,6 +295,23 @@ impl ClientBuilder {
...
+    pub fn ws_proxy<T: Into<String>>(mut self, proxy: T) -> Self {
+        self.ws_proxy = Some(proxy.into());
+        self
+    }
+
+    /// Remove websocket proxy.
+    pub fn no_ws_proxy(mut self) -> Self {
+        self.ws_proxy = None;
+        self
+    }
```
GnomedDev: Can this method be removed, I don't see a useful point to it.

src/client/mod.rs
```diff
@@ -351,6 +371,7 @@ impl IntoFuture for ClientBuilder {
                 #[cfg(feature = "voice")]
                 voice_manager: voice_manager.as_ref().map(Arc::clone),
                 ws_url: Arc::clone(&ws_url),
+                ws_proxy,
```
GnomedDev, line 374: Can this be wrapped in `Arc` like `ws_url`, and preprocessed into `Url` to fail here if the user provides an invalid URL.

src/gateway/ws.rs
```diff
@@ -101,13 +108,72 @@ const TIMEOUT: Duration = Duration::from_millis(500);
 const DECOMPRESSION_MULTIPLIER: usize = 3;
 
 impl WsClient {
-    pub(crate) async fn connect(url: Url) -> Result<Self> {
+    async fn connect_with_proxy_async(
```
GnomedDev, line 109: This whole method should be replaced with something using `reqwest`, which is already in the library.

GnomedDev (commented):
Okay, I understand now, the situation sucks (this should be on tungstenite) but if we can make a positive impact lets go for this

src/gateway/ws.rs
Itsusinn: > This whole method should be replaced with something using `reqwest`, which is already in the library.

Reqwest doesn't expose API that can be used in websocket proxy. But there is an implementation that borrows from reqwest's design.
https://github.com/snapview/tungstenite-rs/issues/177#issuecomment-1352808621

src/client/mod.rs
zzzzRuby: Removed it.

src/gateway/ws.rs
zzzzRuby: Just like [Itsusinn](https://github.com/Itsusinn) said. I replaced the http connect code with async-http-proxy[https://crates.io/crates/async-http-proxy] which is mentioned in that issue.

src/client/mod.rs
zzzzRuby: Yes it can. Finished

mkrasnitski: `ws_url` is only parsed into a `Url` inside of the `gateway::shard::connect` helper function. I think for consistency's sake the same should be done with `ws_proxy` - a followup PR could move the parsing logic for both to be earlier in the initialization stage.

mlesin:
Are there any plans to merge this?

GnomedDev:
As the author seems to be MIA, no. If anyone wants to rebase and solve outstanding reviews, feel free to open a new PR.

qthree:
fyi, I cherry-picked commit from this PR on top of `next` branch and added socks4/5 proxy support, feel free to use it https://github.com/qthree/serenity/tree/ws-proxy

Morb0:
a critically important feature

PR #2735, opened 2024-01-23
Remove simd-json support

GnomedDev:
simd-json is not a performance improvement for serenity, has a ton of sketchy unsafe, and increases the maintenance burden and leads to serenity containing a json compatibility layer(!). This PR just removes all that, although a faster json decoding implementation is on the table in the future if the compatibility layer is kept out of tree and it is measurably faster for serenity's usecase (if benchmark data is wanted, I can probably provide anonymized payloads for a 260k+ guild startup).


arqunis (approved):

lucascompython:
Could serenity just use [sonic-rs](https://github.com/cloudwego/sonic-rs) instead of serde_json, that way serenity doesn't need to support two json crates nor a compatibility layer?
It's also serde compatible.

I'd be willing to submit a PR if so.

PR #2739, opened 2024-01-24
Fix massive code size of client::dispatch

GnomedDev:
This replaces fancy iter stuff with just writing the code, and therefore reduces the code size significantly and spawns less tasks.


GnomedDev:
This is ready as-is.

arqunis (approved):

PR #2741, opened 2024-01-25
Remove the builder trait and only take required arguments

GnomedDev:
Removes unnecessary trait, which only limited the signatures to always take `CacheHttp` and required the async_trait boxing and compile time cost. This also makes a lot less methods generic, improving compile times.

### DIFF
--- a/src/builder/edit_channel.rs
+++ b/src/builder/edit_channel.rs
@@ -306,18 +306,26 @@ impl<'a> EditChannel<'a> {
     pub async fn execute(
         self,
         cache_http: impl CacheHttp,
-        chan_id: ChannelId,
+        channel_id: ChannelId,
     ) -> Result<GuildChannel> {
         #[cfg(feature = "cache")]
         {
             if let Some(cache) = cache_http.cache() {
-                crate::utils::user_has_perms_cache(cache, chan_id, Permissions::MANAGE_CHANNELS)?;
+                crate::utils::user_has_perms_cache(
+                    cache,
+                    channel_id,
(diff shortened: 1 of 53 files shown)


src/builder/edit_channel.rs
```diff
@@ -311,21 +302,22 @@ impl<'a> Builder for EditChannel<'a> {
...
     /// [Manage Channels]: Permissions::MANAGE_CHANNELS
     /// [Manage Roles]: Permissions::MANAGE_ROLES
-    async fn execute(
+    #[cfg(feature = "http")]
+    pub async fn execute(
         self,
         cache_http: impl CacheHttp,
-        ctx: Self::Context<'_>,
-    ) -> Result<Self::Built> {
+        chan_id: ChannelId,
```
mkrasnitski:
```suggestion
        channel_id: ChannelId,
```

mkrasnitski (changes requested):

src/builder/edit_channel.rs
GnomedDev: I did this because it makes the below usage of it blow the line limit and expand a ton, although I guess it doesn't matter really. Would you prefer the consistent naming over caring about lines?

mkrasnitski: Consistent naming is better. There's plenty of other places where calls to `user_has_perms_cache` end up wrapping.

GnomedDev: Fair enough, good to merge after fixed?

arqunis (approved):

PR #2750, opened 2024-01-31
Avoid HashMap in ShardQueue

GnomedDev:
This avoids hashing costs, is probably more cache efficient, and avoids storing duplicate information (length of buckets and max concurrency are equal).

### DIFF
--- a/src/gateway/bridge/shard_queuer.rs
+++ b/src/gateway/bridge/shard_queuer.rs
@@ -301,46 +301,43 @@ impl ShardQueuer {
 /// [`max_concurrency`](crate::model::gateway::SessionStartLimit::max_concurrency).
 #[must_use]
 pub struct ShardQueue {
-    buckets: HashMap<u16, VecDeque<ShardId>>,
-    max_concurrency: NonZeroU16,
+    buckets: FixedArray<VecDeque<ShardId>, u16>,
 }
 
 impl ShardQueue {
     pub fn new(max_concurrency: NonZeroU16) -> Self {
+        let buckets = vec![VecDeque::new(); max_concurrency.get() as usize].into_boxed_slice();
+        let buckets = FixedArray::try_from(buckets).expect("should fit without truncation");
+
         Self {
-            buckets: HashMap::with_capacity(max_concurrency.get() as usize),
-            max_concurrency,
+            buckets,
         }
+    }
+
+    fn calculate_bucket(&self, shard_id: ShardId) -> u16 {
+        shard_id.0 % self.buckets.len()
     }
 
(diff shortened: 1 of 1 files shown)


mkrasnitski:
Actually, we could even use a regular `Vec` to avoid pulling in extra machinery. Under the hood, VecMap uses a `Vec<Option<V>>`, but storing `VecDeque::new()` is pretty much the same semantically as storing `None`, so just making it a `Vec<VecDeque<ShardId>>` would work.

GnomedDev:
Sure, good idea.

src/gateway/bridge/shard_queuer.rs
```diff
@@ -301,46 +301,44 @@ impl ShardQueuer {
...
         Self {
-            buckets: HashMap::with_capacity(max_concurrency.get() as usize),
-            max_concurrency,
+            buckets,
         }
     }
 
+    fn calculate_bucket(&self, shard_id: ShardId) -> usize {
+        let max_concurrency = u16::try_from(self.buckets.len()).expect("length is stored as u16");
+        usize::from(shard_id.0 % max_concurrency)
```
mkrasnitski:
```suggestion
        usize::from(shard_id.0) % self.buckets.len()
```
We don't need to be so verbose.

mkrasnitski (changes requested):

src/gateway/bridge/shard_queuer.rs
GnomedDev: This doesn't work, as buckets.len returns u32 because the max length type of a FixedArray is u32... although this is kind of weird and I'm going to go fix that for small-fixed-array 0.3

GnomedDev: Done.

mkrasnitski:
Looking at your changes in `small-fixed-array 0.3`,  it seems the type inference for indices is a problem. I don't think it makes sense to provide more than one implementation of `Index`. Sticking with `usize` is probably best for now.

GnomedDev:
I definitely see that, but I believe that the type inference problems are worth it to make indexing using the length type provided work.

mkrasnitski:
> I definitely see that, but I believe that the type inference problems are worth it to make indexing using the length type provided work.

I don't quite understand the benefit of indexing with the length type. What does it provide?

GnomedDev:
If the type for indexing does not match the type returned from `len`, a lot of basic patterns become a pain. Further discussion on this should be done on the small-fixed-array repo, as this is not the right place for it.

arqunis (approved):

PR #2754, opened 2024-02-05
Use FxHash everywhere

GnomedDev:
Swaps the HashMap uses throughout serenity to use FxHash, this should be checked to avoid hashdos first.


GnomedDev:
This is not a good idea, at least to do in a find and replace manor like this.

PR #2757, opened 2024-02-08
Simplify the message cache

GnomedDev:
This PR:
- Reduces complexity by only storing messages in one field
- Allows the user to get a sorted view of the messages
- Speeds up the MessageCreate path
- but may slow down the MessageUpdate path, as it has to iterate over the messages in the channel's cache.

### DIFF
--- a/src/cache/event.rs
+++ b/src/cache/event.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::{HashSet, VecDeque};
 use std::num::NonZeroU16;
 
 use super::{Cache, CacheUpdate};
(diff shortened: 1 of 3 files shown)


src/cache/event.rs
```diff
@@ -337,18 +337,15 @@ impl CacheUpdate for MessageCreateEvent {
...
         if messages.len() == max {
-            if let Some(id) = queue.pop_front() {
-                removed_msg = messages.remove(&id);
-            }
+            removed_msg = messages.pop_front();
         }
 
-        queue.push_back(self.message.id);
-        messages.insert(self.message.id, self.message.clone());
+        if !messages.iter().any(|m| m.id == self.message.id) {
```
arqunis, line 346: Is this uniqueness check necessary? Each message that is received through the `MESSAGE_CREATE` event should have a different ID every time.

arqunis (approved):

src/cache/event.rs
GnomedDev: Yes, both to make the tests pass, and because I'm pretty sure discord can deliver duplicate events (although this might be solved by the sequence number).

arqunis: Alright, it can stay. The PR is otherwise fine.

PR #2764, opened 2024-02-11
Defer deserialisation of Dispatch to handle errors

### LINKED ISSUE #2763: Sequence number does not get updated for invalid events
Currently on the next branch, although this may occur on the current branch, if an unknown event is received a warn is emitted in the style of 
```
[serenity::gateway::shard]: [ShardInfo { id: ShardId(..), total: .. }] Sequence off; them: 266560, us: 266558
```

This is because the sequence number is part of the GatewayEvent payload and therefore when deserialisation fails it cannot update the Shard's sequence to fix this. Possible fixes would be to make `GatewayEvent` lazy, deserialising to `JsonMap` before deserialising again to `Event` after fetching the sequence number, or maybe some other hack is possible.

GnomedDev:
Fixes #2763, takes advantage of the already existing deserialisation to JsonMap to defer event deserialisation errors until after the sequence number has been handled.

### DIFF
--- a/src/gateway/bridge/shard_runner.rs
+++ b/src/gateway/bridge/shard_runner.rs
@@ -421,42 +421,36 @@ impl ShardRunner {
             Err(why) => Err(why),
         };
 
-        let action = match self.shard.handle_event(&gateway_event) {
-            Ok(Some(action)) => Some(action),
-            Ok(None) => None,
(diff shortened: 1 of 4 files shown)


GnomedDev:
Ready for review, I've cleaned up and tested that this seems to work on my prod token for a bot with no event handler.

arqunis (approved):

PR #2766, opened 2024-02-13
Fix VoiceServerUpdate::guild_id being wrapped in Option

GnomedDev:
Noticed by @vicky5124, this does not need to be optional.
https://discord.com/developers/docs/topics/gateway-events#voice-server-update

### DIFF
--- a/src/gateway/bridge/shard_runner.rs
+++ b/src/gateway/bridge/shard_runner.rs
@@ -341,11 +341,9 @@ impl ShardRunner {
                         .await;
                 },
                 Event::VoiceServerUpdate(event) => {
-                    if let Some(guild_id) = event.guild_id {
-                        voice_manager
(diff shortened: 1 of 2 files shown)


vicky5124:
Apparently it is optional because back in the day bots could join group voice chats, but that's no longer possible.
The same thing happens with `VoiceState::guild_id`, but this one is slightly more useful for really old bots that are still in group chats, as the voice state update event may still be received without a guild_id if a voice call is started in those group chats, even if the bot can't join it, it can still detect who joined and left that voice call.

Other libraries like hikari have [both values directly without optional](https://docs.hikari-py.dev/en/latest/reference/hikari/voices/#hikari.voices.VoiceState.guild_id), cause bots that are in DMs are extremely rare, and it's not worth the hassle to support them in such a niche area.

arqunis (approved):

PR #2768, opened 2024-02-15
Fill message cache with ChannelId::messages and ChannelId::pins

GnomedDev:
Fills the (non temp) message cache with messages from `ChannelId::messages`/`pins`. This doesn't need to be temporary as they will stay consistent with events and will be thrown out when new messages are received anyway.

### DIFF
--- a/src/builder/get_messages.rs
+++ b/src/builder/get_messages.rs
@@ -89,6 +89,9 @@ impl GetMessages {
 
     /// Gets messages from the channel.
     ///
+    /// If the cache is enabled, this method will fill up the message cache for the channel, if the
+    /// messages returned are newer than the existing cached messages or the cache is not full yet.
+    ///
     /// **Note**: If the user does not have the [Read Message History] permission, returns an empty
     /// [`Vec`].
     ///
--- a/src/builder/get_messages.rs
+++ b/src/builder/get_messages.rs
@@ -1,5 +1,5 @@
 #[cfg(feature = "http")]
-use crate::http::{Http, MessagePagination};
+use crate::http::{CacheHttp, MessagePagination};
 #[cfg(feature = "http")]
 use crate::internal::prelude::*;
 use crate::model::prelude::*;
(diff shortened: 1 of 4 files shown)


src/builder/get_messages.rs
```diff
@@ -98,8 +98,21 @@ impl GetMessages {
     ///
```
arqunis, line 101: There should be a mention about this change in the documentation that by using this method, the message cache will be populated with messages generated by it.

src/model/channel/channel_id.rs
```diff
@@ -543,8 +543,15 @@ impl ChannelId {
     /// Returns [`Error::Http`] if the current user lacks permission to view the channel.
     ///
```
arqunis, line 554: Ditto.

arqunis (approved):
I only have a suggestion for documentation, otherwise it's good to merge.

src/cache/mod.rs
```diff
@@ -616,6 +616,31 @@ impl Cache {
...
+        }
+
+        let mut channel_messages = self.messages.entry(channel_id).or_default();
+
+        // Fill up the existing cache
+        channel_messages.extend(new_messages.take(max_messages));
+        // Make sure the cache stays sorted to messages
+        channel_messages.make_contiguous().sort_unstable_by_key(|m| m.id);
+        // Get rid of the overflow at the front of the queue.
+        channel_messages.drain(..max_messages);
```
arqunis:
```suggestion
        channel_messages.truncate(max_messages);
```

Unlike `Vec::drain`, `Vec::truncate` won't panic if the end point is bigger than `Vec::len()`; it will just do nothing.

arqunis (changes requested):
Actually, as pointed out on Discord, there is a possibility of a panic. That should be resolved before merging.

src/cache/mod.rs
GnomedDev: We can't use truncate here, since we need to truncate from the front of the deque, not the back. I'll fix this panic though.

GnomedDev:
Sorted out the review comments.

arqunis (approved):

PR #2770, opened 2024-02-16
Remove duplicate `external_emojis` accessor

### LINKED ISSUE #2769: Use External Emojis shows up twice in formatted Permissions string
Reason:

There's actually just one permission, called USE_EXTERNAL_EMOJIS, but there's two shorthand accessor methods, called use_external_emojis() and external_emojis() (both using the same USE_EXTERNAL_EMOJIS internally). Every accessor method seems to be assigned an associated display name, which will be iterated when formatting permissions, and both use_external_emojis() and external_emojis() got the display name of "Use External Emojis"

Found in: https://discord.com/channels/381880193251409931/381912587505500160/1207880394725400626

wackbyte:
Closes #2769.

### DIFF
--- a/src/model/permissions.rs
+++ b/src/model/permissions.rs
@@ -379,7 +379,6 @@ generate_get_permission_names! {
     create_public_threads: "Create Public Threads",
     deafen_members: "Deafen Members",
     embed_links: "Embed Links",
-    external_emojis: "Use External Emojis",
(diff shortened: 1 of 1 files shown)


kangalio (approved):
LGTM

arqunis (approved):

PR #2771, opened 2024-02-16
Remove `CreateAttachment::id` field

mkrasnitski:
For new attachments, the value of the `id` field corresponds to their order in the payload. Because these ids are sequential, we can simply calculate them when forming the multipart payload/serializing.

cc @kangalio

### DIFF
--- a/src/builder/create_attachment.rs
+++ b/src/builder/create_attachment.rs
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 use std::path::Path;
 
+use serde::ser::{Serialize, SerializeSeq, Serializer};
 use tokio::fs::File;
 use tokio::io::AsyncReadExt;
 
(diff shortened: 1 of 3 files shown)


src/builder/create_attachment.rs
```diff
@@ -284,3 +272,37 @@ impl<'a> EditAttachments<'a> {
...
+                        description: &new_attachment.description,
+                    };
+                    id += 1;
+                    seq.serialize_element(&attachment)?;
+                },
+                NewOrExisting::Existing(existing_attachment) => {
+                    seq.serialize_element(existing_attachment)?;
+                },
+            }
+        }
```
GnomedDev, line 305: Can this be rewritten to use `Serializer::collect_seq`?

GnomedDev (changes requested):
Looks good, just a small improvement before I can approve

src/builder/create_attachment.rs
mkrasnitski: I don't think so, considering the types we're serializing are different between the two match arms.

GnomedDev: Ah yes, I don't know how I missed that, this would be possible if we had Either but probably not. I'll merge and look at it later if I get round to it.

GnomedDev (approved):

Issue #2690, opened 2024-01-05
feat(builder): CreateAttachment stream support


ghost:
Currently the CreateAttachment has a data field with static attachment data, therefore all attachments are fully loaded to memory before being sent, this causes huge memory use when dealing with a lot of files.

Perhaps there is a way to provide streams instead which would solve the issue.

mkrasnitski:
Implemented in #3322; attachments can now be streamed directly from files on disk. Because of how we implement Discord ratelimiting, our request data structure needs to be clone-able. Therefore, support for arbitrary byte streams and streaming from network request bodies (both of which are not `Clone`) isn't possible at the moment. However, for most use-cases, streaming from disk should suffice.

PR #2774, opened 2024-02-19
Add `CreateAttachmentStream` type

mkrasnitski:
Addresses #2690 by adding a type which allows lazily creating attachments which only get resolved at request time. The `Clone` requirement means that we can't store the `Reader` types directly, which is unfortunate but not too bad.

Draft for now because I don't know how to concisely add support to the `create_*` builders. Adding support to the `edit_*` builders was easy by just adding an `EditAttachments::add_stream` method. However, the `create_*` builders have three separate methods for adding attachments: `add_file`, `add_files`, and `files`. Duplicating all three to accommodate the streaming variant seems quite noisy.

### DIFF
--- a/src/builder/create_attachment.rs
+++ b/src/builder/create_attachment.rs
@@ -4,6 +4,7 @@ use std::path::Path;
 use serde::ser::{Serialize, SerializeSeq, Serializer};
 use tokio::fs::File;
 use tokio::io::AsyncReadExt;
+use url::Url;
 
 #[allow(unused)] // Error is used in docs
 use crate::error::{Error, Result};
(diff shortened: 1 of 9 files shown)


GnomedDev:
I don't see how this solves the linked issue. These types still require loading all the files into memory before sending them, instead of streaming them, it just defers the load to the builder execution instead of the builder creation. Implementing proper streaming would probably require dropping down to hyper, but I'm not sure.

mkrasnitski:
I guess I was thinking about long-lived builders which keep their data around in memory. If we want to use readers directly, we'll have to drop the `Clone` requirement which would require separating out the attachment methods for builders so that the builders themselves can still remain `Clone`.

mkrasnitski:
Closing this for now as it isn't the solution we really want.

PR #2779, opened 2024-02-28
Remove comment about bots not being able to send stickers

jamesbt365:
The builders support this already, but for whatever reason this comment still is here saying that bots cannot send them.

image


arqunis (approved):

ComputerVirusUD:
omg is that me

PR #2786, opened 2024-03-09
Use ToArrayString for int -> str 

GnomedDev:
This replaces the slow heap allocations with stack allocations at the cost of complicating the code because of lifetime shenanigans.

It also makes ToArrayString a public trait implementation for Ids, which shouldn't be a problem, but I can remove if wanted.


GnomedDev:
Undrafted

arqunis (approved):

PR #2788, opened 2024-03-10
switch to strum for fullevent and event

cheesycod:
Add Strum for getting variant names and list of all variants for large enums such as FullEvent and Event. Also removes a hack (``Event::name``) where serde is used to get the variant name

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ arrayvec = { version = "0.7.4", features = ["serde"] }
 small-fixed-array = { version = "0.4", features = ["serde"] }
 bool_to_bitflags = { version = "0.1.0" }
 nonmax = { version = "0.5.5", features = ["serde"] }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -34,6 +34,8 @@ arrayvec = { version = "0.7.4", features = ["serde"] }
 small-fixed-array = { version = "0.4", features = ["serde"] }
 bool_to_bitflags = { version = "0.1.0" }
 nonmax = { version = "0.5.5", features = ["serde"] }
+strum = "0.26"
```
GnomedDev: Why do you need `strum` as well as `strum_macros`, I don't see it being used.

GnomedDev (changes requested):
Just a small nitpick

Cargo.toml
cheesycod: Will fix, sec

GnomedDev (approved):

PR #2797, opened 2024-03-13
Use ExtractMap instead of HashMap when possible

GnomedDev:
ExtractMap is HashSet with a Map interface to prevent duplicating the key if it needs to be stored in the value. This swaps serenity to using it which improves memory efficiency and maybe runtime performance while deserialising as only the value needs to be deserialised.


GnomedDev:
Just published extract_map and a typesize compatible version, so ready for review.

arqunis (approved):

PR #2801, opened 2024-03-17
Add banner editing to EditProfile

GnomedDev:


src/builder/edit_profile.rs
```diff
@@ -64,6 +66,20 @@ impl EditProfile {
         self.username = Some(username.into());
         self
     }
+
+    /// Sets the banner of the current user.
+    ///
+    /// This takes the raw image data as base64 encoding is handled internally.
+    pub fn banner(mut self, banner: &[u8]) -> Self {
```
arqunis: Shouldn't this use `CreateAttachment` just like `EditProfile::avatar`? If not, the documentation should mention that `banner` should be PNG data.

GnomedDev: I don't know how the current CreateAttachment stuff works, I'll take a look.

GnomedDev: This seems the wrong way to use it, since it doesn't require a filename, but meh that's not this PR's job to fix.

GnomedDev: Sorted this out.

arqunis (approved):

PR #2806, opened 2024-03-18
Remove Value from Http

GnomedDev:
This avoids allocating serde_json::Value

### DIFF
--- a/src/builder/edit_role.rs
+++ b/src/builder/edit_role.rs
@@ -177,7 +177,9 @@ impl<'a> EditRole<'a> {
         };
 
         if let Some(position) = self.position {
-            http.edit_role_position(guild_id, role.id, position, self.audit_log_reason).await?;
+            guild_id
+                .edit_role_position_with_reason(http, role.id, position, self.audit_log_reason)
+                .await?;
         }
         Ok(role)
     }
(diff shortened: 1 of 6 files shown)


arqunis (approved):

src/builder/edit_role.rs
```diff
@@ -177,7 +177,9 @@ impl<'a> EditRole<'a> {
         };
 
         if let Some(position) = self.position {
-            http.edit_role_position(guild_id, role.id, position, self.audit_log_reason).await?;
+            guild_id
+                .edit_role_position_with_reason(http, role.id, position, self.audit_log_reason)
+                .await?;
```
mkrasnitski, line 182: This introduced a dependency on the `model` feature inside code that is normally just gated behind `http` plus `builder`. Is it possible to fix this?

GnomedDev: I'm going to make a PR soon that moves `builder` inside of `model` and gets rid of the `builder` feature, since `builder` shouldn't be used without `model` anyway.

mkrasnitski: Can you elaborate? I'm thinking of the builder-no-http usecase as a specific counter-example to that idea.

EDIT: It would be feasible to gate the `execute` methods behind `model` instead of just `http`, but the structs themselves I don't think should be gated behind `model`.

GnomedDev: Let's move this to the lib-development chat as this is going quite off-topic for this PR.

PR #2807, opened 2024-03-18
Implement User Apps

GnomedDev:
Implements all the changes documented in the [User Apps Change Log](https://discord.com/developers/docs/change-log#march-18-2024-api-changes). This includes fun new models and fun new deprecation warnings.

### DIFF
--- a/src/model/application/command.rs
+++ b/src/model/application/command.rs
@@ -2,6 +2,8 @@ use std::collections::HashMap;
 
 use serde::Serialize;
 
+#[cfg(feature = "unstable_discord_api")]
+use super::{InstallationContext, InteractionContext};
 #[cfg(feature = "model")]
 use crate::builder::{Builder, CreateCommand};
 #[cfg(feature = "model")]
@@ -73,11 +75,23 @@ pub struct Command {
     /// Indicates whether the command is available in DMs with the app, only for globally-scoped
     /// commands. By default, commands are visible.
     #[serde(default)]
+    #[deprecated = "Use Command::contexts"]
     pub dm_permission: Option<bool>,
     /// Indicates whether the command is [age-restricted](https://discord.com/developers/docs/interactions/application-commands#agerestricted-commands),
     /// defaults to false.
     #[serde(default)]
     pub nsfw: bool,
+    /// Installation context(s) where the command is available, only for globally-scoped commands.
+    ///
+    /// Defaults to [`InstallationContext::Guild`]
+    #[cfg(feature = "unstable_discord_api")]
+    #[serde(default)]
+    pub integration_types: Vec<InstallationContext>,
+    /// Interaction context(s) where the command can be used, only for globally-scoped commands.
+    ///
+    /// By default, all interaction context types are included.
+    #[cfg(feature = "unstable_discord_api")]
+    pub contexts: Option<Vec<InteractionContext>>,
     /// An autoincremented version identifier updated during substantial record changes.
     pub version: CommandVersionId,
 }
--- a/src/model/application/command_interaction.rs
+++ b/src/model/application/command_interaction.rs
@@ -4,6 +4,8 @@ use serde::de::{Deserializer, Error as DeError};
 use serde::ser::{Error as _, Serializer};
 use serde::{Deserialize, Serialize};
 
+#[cfg(feature = "unstable_discord_api")]
+use super::{AuthorizingIntegrationOwners, InteractionContext};
 #[cfg(feature = "model")]
 use crate::builder::{
     Builder,
@@ -74,6 +76,7 @@ pub struct CommandInteraction {
     /// Always `1`.
     pub version: u8,
     /// Permissions the app or bot has within the channel the interaction was sent from.
+    // TODO(next): This is now always serialized.
     pub app_permissions: Option<Permissions>,
     /// The selected language of the invoking user.
     pub locale: String,
@@ -81,6 +84,13 @@ pub struct CommandInteraction {
     pub guild_locale: Option<String>,
     /// For monetized applications, any entitlements of the invoking user.
     pub entitlements: Vec<Entitlement>,
+    /// The owners of the applications that authorized the interaction, such as a guild or user.
+    #[serde(default)]
+    #[cfg(feature = "unstable_discord_api")]
+    pub authorizing_integration_owners: AuthorizingIntegrationOwners,
+    /// The context where the interaction was triggered from.
+    #[cfg(feature = "unstable_discord_api")]
+    pub context: Option<InteractionContext>,
 }
 
 #[cfg(feature = "model")]
--- a/src/model/application/component_interaction.rs
+++ b/src/model/application/component_interaction.rs
@@ -62,6 +62,13 @@ pub struct ComponentInteraction {
     pub guild_locale: Option<String>,
     /// For monetized applications, any entitlements of the invoking user.
     pub entitlements: Vec<Entitlement>,
+    /// The owners of the applications that authorized the interaction, such as a guild or user.
+    #[serde(default)]
+    #[cfg(feature = "unstable_discord_api")]
+    pub authorizing_integration_owners: AuthorizingIntegrationOwners,
+    /// The context where the interaction was triggered from.
+    #[cfg(feature = "unstable_discord_api")]
+    pub context: Option<InteractionContext>,
 }
 
 #[cfg(feature = "model")]
--- a/src/model/application/interaction.rs
+++ b/src/model/application/interaction.rs
@@ -1,14 +1,20 @@
 use serde::de::{Deserialize, Deserializer, Error as DeError};
 use serde::ser::{Serialize, Serializer};
 
+#[cfg(feature = "unstable_discord_api")]
+use super::InstallationContext;
 use super::{CommandInteraction, ComponentInteraction, ModalInteraction, PingInteraction};
 use crate::internal::prelude::*;
 use crate::json::from_value;
 use crate::model::guild::PartialMember;
 use crate::model::id::{ApplicationId, InteractionId};
+#[cfg(feature = "unstable_discord_api")]
+use crate::model::id::{GuildId, MessageId, UserId};
 use crate::model::monetization::Entitlement;
 use crate::model::user::User;
 use crate::model::utils::deserialize_val;
+#[cfg(feature = "unstable_discord_api")]
+use crate::model::utils::StrOrInt;
 use crate::model::Permissions;
 
 /// [Discord docs](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object)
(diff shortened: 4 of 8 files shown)


GnomedDev:
This is now ready for review, I've not tested it yet as I don't have the time so that would be lovely if someone else could.

src/model/application/mod.rs
```diff
@@ -76,6 +77,55 @@ pub struct CurrentApplicationInfo {
...
+    #[non_exhaustive]
+    pub enum InteractionContext {
+        /// Interaction can be used within servers
+        Guild = 0,
+        /// Interaction can be used within DMs with the app's bot user
+        BotDm = 2,
+        /// Interaction can be used within Group DMs and DMs other than the app's bot user
+        PrivateChannel = 3,
+        _ => Unknown(u8),
+    }
```
fgardt: I believe this should be `0`, `1`, `2`.
https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-interaction-context-types
```suggestion
    pub enum InteractionContext {
        /// Interaction can be used within servers
        Guild = 0,
        /// Interaction can be used within DMs with the app's bot user
        BotDm = 1,
        /// Interaction can be used within Group DMs and DMs other than the app's bot user
        PrivateChannel = 2,
        _ => Unknown(u8),
    }
```

src/model/application/command.rs
```diff
@@ -73,11 +75,22 @@ pub struct Command {
...
+    /// Installation context(s) where the command is available, only for globally-scoped commands.
+    ///
+    /// Defaults to [`InstallationContext::Guild`]
+    #[cfg(feature = "unstable_discord_api")]
+    pub integration_types: Vec<InstallationContext>,
+    /// Interaction context(s) where the command can be used, only for globally-scoped commands.
+    ///
+    /// By default, all interaction context types are included.
+    #[cfg(feature = "unstable_discord_api")]
+    pub contexts: Vec<InteractionContext>,
```
fgardt: from my testing it appears that this can also be `null` so this would either need to be wrapped as an `Option` or `#[serde(default)]`

src/model/application/command_interaction.rs
```diff
@@ -74,13 +76,21 @@ pub struct CommandInteraction {
...
     /// The selected language of the invoking user.
     pub locale: String,
     /// The guild's preferred locale.
     pub guild_locale: Option<String>,
     /// For monetized applications, any entitlements of the invoking user.
     pub entitlements: Vec<Entitlement>,
+    #[cfg(feature = "unstable_discord_api")]
+    #[serde(rename = "authorizing_integration_owners")]
+    /// The owner of the application that authorized the interaction, such as a guild or user.
+    pub authorizing_integration_owner: AuthorizingIntegrationOwner,
```
fgardt:
```suggestion
    pub authorizing_integration_owner: Option<AuthorizingIntegrationOwner>,
```

> A key will **only** be present if the following are true...

https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-authorizing-integration-owners-object

src/model/application/component_interaction.rs
```diff
@@ -62,6 +62,13 @@ pub struct ComponentInteraction {
     pub guild_locale: Option<String>,
     /// For monetized applications, any entitlements of the invoking user.
     pub entitlements: Vec<Entitlement>,
+    #[cfg(feature = "unstable_discord_api")]
+    #[serde(rename = "authorizing_integration_owners")]
+    /// The owner of the application that authorized the interaction, such as a guild or user.
+    pub authorizing_integration_owner: AuthorizingIntegrationOwner,
```
fgardt:
```suggestion
    pub authorizing_integration_owner: Option<AuthorizingIntegrationOwner>,
```
_same deal_

src/model/application/interaction.rs
```diff
@@ -313,3 +390,29 @@ pub struct MessageInteraction {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub member: Option<PartialMember>,
 }
+
+/// Metadata about the interaction, including the source of the interaction relevant server and
+/// user IDs.
+#[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[cfg(feature = "unstable_discord_api")]
+pub struct MessageInteractionMetadata {
```
fgardt, line 424: during testing I realized that this also includes a `name` property which seems to behave identical to `MessageInteraction::name`
_this is currently not documented by discord tho_

src/model/application/interaction.rs
```diff
@@ -289,11 +293,84 @@ bitflags! {
...
+                match map.next_key()? {
+                    Some(InstallationContext::Guild) => {
+                        map.next_value().map(Self::Value::GuildInstall)
+                    },
+                    Some(InstallationContext::User) => {
+                        map.next_value().map(Self::Value::UserInstall)
+                    },
+                    Some(key @ InstallationContext::Unknown(_)) => Ok(Self::Value::Unknown(key)),
+                    None => Err(serde::de::Error::missing_field("key")),
+                }
```
fgardt: this does not work since `key` is a string but the `InstallationContext` enum is `u8`.

src/model/event.rs
```diff
@@ -525,8 +525,12 @@ pub struct MessageUpdateEvent {
     pub flags: Option<Option<MessageFlags>>,
     #[serde(default, deserialize_with = "deserialize_some")]
     pub referenced_message: Option<Option<Box<Message>>>,
+    #[cfg_attr(not(ignore_serenity_deprecated), deprecated = "Use interaction_metadata")]
     #[serde(default, deserialize_with = "deserialize_some")]
+    #[allow(deprecated)]
     pub interaction: Option<Option<Box<MessageInteraction>>>,
+    #[cfg(feature = "unstable_discord_api")]
+    pub interaction_metadata: Option<Option<MessageInteractionMetadata>>,
```
fgardt: double `Option`? seems matched to `interaction` but why?

src/model/application/mod.rs
```diff
@@ -1,6 +1,7 @@
 //! Models about OAuth2 applications.
 
 mod command;
+
```
fgardt: intentional?

src/model/application/interaction.rs
```diff
@@ -313,3 +390,29 @@ pub struct MessageInteraction {
...
+    /// The ID of the interaction
+    pub id: InteractionId,
+    /// The type of interaction
+    #[serde(rename = "type")]
+    pub kind: InteractionType,
+    /// The ID of the user who triggered the interaction
+    pub user_id: UserId,
+    /// The IDs for installation context(s) related to an interaction.
+    #[serde(rename = "authorizing_integration_owners")]
+    pub authorizing_integration_owner: AuthorizingIntegrationOwner,
```
fgardt: not sure if this would also need to be wrapped as an `Option` since the conditions for the field to be present should be given when interaction metadata is also present?

fgardt (changes requested):

src/model/application/interaction.rs
GnomedDev: I'm not going to add undocumented features.

src/model/event.rs
GnomedDev: Option for being updated, Option for the real field

src/model/application/command_interaction.rs
GnomedDev: From the way we are deserialising this, I don't think this is wrong? In what situation is an interaction going to be fired for an unsupported installation/interaction context.

fgardt: well in my testing I did not receive this field ever. Probably because I also did not set the `integration_types` or `contexts` when registering the commands (?).
Without marking it optional no interaction would work for me.

src/model/application/command.rs
```diff
@@ -73,11 +75,23 @@ pub struct Command {
...
+    ///
+    /// Defaults to [`InstallationContext::Guild`]
+    #[cfg(feature = "unstable_discord_api")]
+    pub integration_types: Vec<InstallationContext>,
+    /// Interaction context(s) where the command can be used, only for globally-scoped commands.
+    ///
+    /// By default, all interaction context types are included.
+    #[cfg(feature = "unstable_discord_api")]
+    #[serde(default)]
+    pub contexts: Vec<InteractionContext>,
```
fgardt: tested this change just now, I was wrong :/
unless there is another nice way to deserialize `null` as an empty `Vec<T>` this should be an `Option`

```suggestion
    pub contexts: Option<Vec<InteractionContext>>,
```

src/model/application/mod.rs
```diff
@@ -76,6 +76,55 @@ pub struct CurrentApplicationInfo {
...
+        User = 1,
+        _ => Unknown(u8),
+    }
+}
+
+#[cfg(feature = "unstable_discord_api")]
+enum_number! {
+    /// An enum representing the different [interaction contexts].
+    ///
+    /// [interaction contexts](https://discord.com/developers/docs/monetization/skus#sku-object-sku-flags).
```
fgardt:
```suggestion
    /// [interaction contexts](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-interaction-context-types).
```

fgardt (changes requested):

src/model/application/interaction.rs
```diff
@@ -289,11 +293,86 @@ bitflags! {
...
+            },
+            Self::UserInstall(inner) => {
+                serializer.serialize_entry(&InstallationContext::User, &inner)
+            },
+            Self::Unknown(inner) => serializer.serialize_entry(&inner, &()),
+        }?;
+
+        serializer.end()
+    }
+}
```
fgardt: A response can contain multiple (2) authorizing integration owners. I'm not sure what the best way to go about representing it would be. I'm currently experimenting with having something like this
```rs
#[derive(Clone, Debug)]
pub struct AuthorizingIntegrationOwners(Vec<AuthorizingIntegrationOwner>);
```
and then implementing the (de)serialization for that.

src/model/application/command_interaction.rs
fgardt: did this change get reverted due to force pushing?
```suggestion
    pub authorizing_integration_owner: Option<AuthorizingIntegrationOwner>,
```

fgardt (changes requested):

src/model/application/interaction.rs
GnomedDev: What situation causes multiple authorizing integration owners?

fgardt: Having the app installed as a user app aswell as on a guild you're on.

GnomedDev: Okay, done.

src/model/application/interaction.rs
```diff
@@ -289,11 +293,101 @@ bitflags! {
...
+            fn visit_map<A>(self, mut map: A) -> StdResult<Self::Value, A::Error>
+            where
+                A: serde::de::MapAccess<'de>,
+            {
+                let mut out = Vec::new();
+                while let Some(key_str) = map.next_key::<serde_cow::CowStr<'_>>()? {
+                    let key_int = key_str.0.parse::<u8>().map_err(serde::de::Error::custom)?;
+                    let value = match InstallationContext::from(key_int) {
+                        InstallationContext::Guild => {
+                            AuthorizingIntegrationOwner::GuildInstall(map.next_value()?)
```
fgardt: This deserialization fails when the interaction is triggered from a bot DM.
Key is `GUILD_INSTALL` but the value is `"0"` which is not a valid `GuildId`.

Maybe it would make sense to have a separate `AuthorizingIntegrationOwner` just for bot DMs?
Would also mean it could be `GuildInstall(GuildId)` instead of `GuildInstall(Option<GuildId>)`.

fgardt (commented):
its working fine now apart from the one issue :)

src/model/application/interaction.rs
fgardt: alternatively of course
```suggestion
                            AuthorizingIntegrationOwner::GuildInstall(map.next_value().ok())
```

GnomedDev: Fixed this issue, this is dumb though.

GnomedDev:
Merging this in as it isn't doing any good as a PR and has been tested by fgart, thanks!.

PR #2817, opened 2024-03-25
Remove deprecated Member and Guild methods

GnomedDev:
Followup to #2816

### DIFF
--- a/src/model/guild/mod.rs
+++ b/src/model/guild/mod.rs
@@ -1269,7 +1269,6 @@ impl Guild {
     ///
     /// [`position`]: Role::position
     #[must_use]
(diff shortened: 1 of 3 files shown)


src/model/guild/mod.rs
```diff
@@ -1268,13 +1268,9 @@ impl Guild {
     /// owner, their ID is returned.
     ///
     /// [`position`]: Role::position
+    #[must_use]
     #[cfg(feature = "cache")]
```
arqunis: This shouldn't need to be feature-gated anymore.
```suggestion
```

arqunis (changes requested):

arqunis (approved):

PR #2819, opened 2024-03-28
added from<User> to CreateEmbedAuthor

RegenJacob:
This adds `From<User>` to `CreateEmbedAuthor`

### DIFF
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -271,7 +271,8 @@ impl From<EmbedAuthor> for CreateEmbedAuthor {
 
 impl From<User> for CreateEmbedAuthor {
     fn from(user: User) -> Self {
-        Self::new(user.name).icon_url(user.face())
+        let avatar_icon = user.face();
+        Self::new(user.name).icon_url(avatar_icon)
     }
 }
 
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -269,6 +269,12 @@ impl From<EmbedAuthor> for CreateEmbedAuthor {
     }
 }
 
+impl From<User> for CreateEmbedAuthor {
+    fn from(user: User) -> Self {
+        Self::new(&user.name).icon_url(user.face())
+    }
+}
+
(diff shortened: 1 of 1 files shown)


src/builder/create_embed.rs
```diff
@@ -269,6 +269,12 @@ impl From<EmbedAuthor> for CreateEmbedAuthor {
     }
 }
 
+impl From<User> for CreateEmbedAuthor {
+    fn from(user: User) -> Self {
+        Self::new(&user.name).icon_url(user.face())
```
GnomedDev: You should pass user.name is as-is, to avoid cloning.
```suggestion
        Self::new(user.name).icon_url(user.face())
```

GnomedDev (commented):
LGTM, just a small nitpick.

src/builder/create_embed.rs
RegenJacob: actually no this doesn't work because
```
error[E0382]: borrow of partially moved value: `user`
   --> src/builder/create_embed.rs:274:39
    |
274 |         Self::new(user.name).icon_url(user.face())
    |                   ---------           ^^^^ value borrowed here after partial move
    |                   |
    |                   value partially moved here
    |
    = note: partial move occurs because `user.name` has type `std::string::String`, which does not implement the `Copy` trait
   ```

GnomedDev: Call user.face before passing in the name then, I should have spotted that when making the suggestion lol.

RegenJacob: Yeah also bad call from me to just commit that. But `CreateEmbedAuthor::new()` uses user.name as argument so I guess I should just make it without the builder function via `From<EmbedAuthor>` but wouldn't that also clone?

GnomedDev: No. You can call user.face() and store the resulting String in a variable, then pass user.name into the builder, then pass the String into the builder.

RegenJacob: oh right

arqunis (approved):

PR #2824, opened 2024-03-30
refactor!: remove `fs` feature for `tokio` by default.

UserIsntAvailable:
I currently working with wasm on one of my personal projects. I'm would like to use `serenity` to get access to the discord API response models, but currently the library can not the compiled to wasm, because the `tokio/fs` feature is not supported in that target.

I went ahead and removed the need to include that feature by default. By the looks of it, it only was used with the `builder` and `model` features, so I conditionally added `tokio/fs` back when those features are being used (the `model` feature already depends on `builder` so that doesn't need to be changed).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ async-trait = "0.1.74"
 tracing = { version = "0.1.40", features = ["log"] }
 serde = { version = "1.0.192", features = ["derive"] }
 url = { version = "2.4.1", features = ["serde"] }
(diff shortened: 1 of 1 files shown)


GnomedDev (changes requested):
This is a breaking change as this changes the set of features a dependent would have access to due to feature unification. Please re-target this to the next branch.

UserIsntAvailable:
I wasn't aware this was considered breaking. I updated the branch and the title to reflect that :+1:.

GnomedDev (approved):
No problem, looks good to me!

PR #2825, opened 2024-03-30
Deprecate Guild::is_large and Message::is_own/private

GnomedDev:
These methods are all wrong and/or easily performed without serenity doing it.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -203,6 +203,7 @@ impl Message {
 
     /// A util function for determining whether this message was sent by someone else, or the bot.
     #[cfg(feature = "cache")]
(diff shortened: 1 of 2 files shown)


src/model/channel/message.rs
```diff
@@ -453,6 +454,7 @@ impl Message {
     /// never set for those, which this method relies on.
     #[inline]
     #[must_use]
+    #[deprecated = "Check if guild_id is None if the message is recieved from the gateway."]
```
```suggestion
    #[deprecated = "Check if guild_id is None if the message is received from the gateway."]
```

arqunis (changes requested):

arqunis (approved):

PR #2830, opened 2024-04-01
 Add support for the bulk ban endpoint 

GnomedDev:
This also fixes up the usage of the deprecated delete_message_days by replacing it with delete_message_seconds internally.


GnomedDev:
This currently links to the PR that documented bulk ban, not the actual docs, as it seems to not be on the hosted docs yet.

arqunis (approved):

PR #2836, opened 2024-04-10
Add support for Message Polls

GnomedDev:
This PR adds support for Polls attached to messages. This has been tested for deserialising, creating, getting answer voters, and ending a poll.

The builder is designed differently, as there are many required fields, so I used the typestate pattern to prevent the user from using the builder until it is ready.


cheesycod:
@GnomedDev Can this be merged ASAP since it's already out and I want to add stuff to my bot related to polls

GnomedDev:
No? It doesn't work as-is and I haven't implemented much anyway. If you want to take over this PR, feel free, but I'm not paid to work on serenity so I will work when I feel like it.

cheesycod:
> No? It doesn't work as-is and I haven't implemented much anyway. If you want to take over this PR, feel free, but I'm not paid to work on serenity so I will work when I feel like it.

Oh alright, sorry for the comment btw

GnomedDev:
This is ready for review, I've tested the happy paths but it may need a bit more testing, this can be done after merge though as it's not like we are pushing a release right after.

arqunis (approved):

PR #2839, opened 2024-04-11
No longer update topic on VoiceChannelStatusUpdate

jamesbt365:
In my original PR I was updating topic because the event was changing the topic of the VC because it was changing server side but without firing `ChannelUpdate`.

This doesn't seem to be the case anymore, I guess they are still stabilizing things considering the documentation for this feature isn't even merged yet.

This change targets next because it messes with stuff users could access, if this doesn't count as breaking I'll change the base target.

### DIFF
--- a/src/cache/event.rs
+++ b/src/cache/event.rs
@@ -609,8 +609,6 @@ impl CacheUpdate for VoiceChannelStatusUpdateEvent {
 
         let old = channel.status.clone();
         channel.status.clone_from(&self.status);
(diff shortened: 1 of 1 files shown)


GnomedDev:
This should be targeted at current.

jamesbt365:
Okay its just not gonna do CI?

Edit: commenting made it work

GnomedDev (approved):
LGTM, I would have liked the whole feature to be flagged under unstable if it's undocumented, but as-is this is the best we can do.

PR #2841, opened 2024-04-16
system flags in audit log changes

bend-n:
fixes bug where audit log entries of welcome messages being disabled etc would deserialize as `Unknown`

### DIFF
--- a/src/model/guild/audit_log/change.rs
+++ b/src/model/guild/audit_log/change.rs
@@ -6,6 +6,7 @@ use crate::model::guild::{
     DefaultMessageNotificationLevel,
     ExplicitContentFilter,
     MfaLevel,
+    SystemChannelFlags,
     VerificationLevel,
 };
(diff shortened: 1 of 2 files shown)


GnomedDev:
What's the point of the macro? All I can see it being used for is generating the `name` method which isn't used in this PR and isn't documented in the PR description.

bend-n:
The point is the `name` method which i put, as originally I didnt see that there was already a `SystemFlags` so i modelled it after `Permissions`, so as to make them have more similar apis.

I can remove it if that would be required

GnomedDev:
Please do remove the changes that are not explicitly related to fixing the bug, I see this as an unnecessary complexity.

bend-n:
image
is this normal

GnomedDev:
I think you ran into the emoji in branches bug, that causes github's infra to fall apart if you have an emoji in a branch name. Reopen the PR with a different branch name lol

PR #2843, opened 2024-04-18
Fix for media types in CreateAttachment to base64 conversion

NotNorom:


GnomedDev (commented):
Why is this trying to guess instead of using an enum argument? The enum could have a `guess` method on that returned `Option<Self>` to get rid of the hidden failure of defaulting to image/png.

NotNorom:
Would you rather have this media type enum present in the CreateAttachment constructor or in the to_base64 function?

GnomedDev:
CreateAttachment constructor, ig it will be useless when passed into methods that don't call to_base64 but that's fine.

NotNorom:
How should we handle the case where to_base64 is called, but the media_type is neither jpg, png or gif? Should the signature be changed to `to_base64(&self) -> Option<String>` and return None?

GnomedDev:
That won't be a problem if MediaType is an enum of `jpg`, `png`, `gif` and the guess method returns Option<Self>.

NotNorom:
So I got your suggestion to work but I just realized this might fuck over users who use CreateAttachment for anything else besides uploading images. The MediaType restriction makes only sense when needing to upload an image to e.g. banner, or icon endpoints but when uploading an mp3 file for example as an attachment to a message this would no longer work I think.

GnomedDev:
Okay, this is dumb, this is an abuse of CreateAttachment and we should have a type that means base64 encoded data with a mime type. Can you implement that, or do you want to revert back to the original PR and I'll merge that *for now*.

NotNorom:
Yeah I agree, this is not what CreateAttachment should be.
I think for now I'd like to split this PR so it's possible to merge [1e7ba1e](https://github.com/serenity-rs/serenity/pull/2843/commits/1e7ba1ebb2a45d3e31ce61bb2061bd677ef2e705)  so we can make the EditGuild API a little more consistent without having to worry about media type issues.

I think for a base64 encoded type with a mime type we could workshop that in discord maybe?

Another idea that I had:
What's your opinion on converting CreateAttachment to a trait? Then there could be structs for Icons, Banners, regular files, etc implementing their own restrictions. Maybe have a ToBase64 trait too?

src/builder/create_attachment.rs
```diff
@@ -12,6 +13,49 @@ use crate::http::Http;
...
+            _ => None,
+        }
+    }
+
+    pub fn guess_from_data<'a>(data: impl Into<Cow<'a, [u8]>>) -> Option<Self> {
+        // magic byte values from https://en.m.wikipedia.org/wiki/List_of_file_signatures
+        match &*data.into() {
+            &[0x89, b'P', b'N', b'G', 0x0D, 0x0A, 0x1A, 0x0A, ..] => Some(Self::Png),
+            &[0xFF, 0xD8, 0xFF, 0xDB | 0xEE | 0xE0, ..] => Some(Self::Jpg),
+            &[b'G', b'I', b'F', 0x38, 0x37 | 0x39, 0x61, ..] => Some(Self::Gif),
```
bend-n, line 53:
```suggestion
            &[0x89, b'P', b'N', b'G', ..] => Some(Self::Png),
            &[0xFF, 0xD8, 0xFF, ..] => Some(Self::Jpg),
            &[b'G', b'I', b'F', ..] => Some(Self::Gif),
```

PR #2278, opened 2022-11-09
Next try at properly merging "Rewrite collectors #2240"

kangalio:
#2240 had to be reverted because, in the process of rebasing, I accidentally reverted a few already-merged commits. ~~I don't know exactly what went wrong so I just did the same thing again, carefully.~~ I used the wrong PR base commit. I first used a01ee32, then cf38735d1b92363e6b391e21bed6855da06df9bf, but the correct one was eee1c85005

PR #2847, opened 2024-04-22
Restore match arm for `Error::Gateway` variant

mkrasnitski:
This was mistakenly removed in #2278 and wasn't caught because of the wildcard pattern in the match.


GnomedDev:
Can you spell out the variants being caught by the wildcard, so this doesn't happen again?

mkrasnitski:
I removed the wildcard, so there are no extra variants being caught by it anymore.

GnomedDev (approved):
I'm silly!

PR #2849, opened 2024-04-24
Remove multiple event handler support

GnomedDev:
As voted for on the lib dev channel: this removes multiple event handler support and restores the owned values passed to event handlers.


GnomedDev:
Ready for review.

arqunis (approved):

PR #2850, opened 2024-04-24
Split ReactionType into FixedReactionType and ReactionType

GnomedDev:
Draft as WIP, solves an exposure of FixedString in the user constructable types.


GnomedDev:
Splitting this isn't a good idea, and it's probably the best idea to just deal with the exposure or swap to String.

PR #2851, opened 2024-04-25
Use RwLock for active collectors collection 

TheCataliasTNT2k:
One of the main functions used a Mutex to guard a shared variable.
This may not impact small bots, but can significantly slow down bots, which have a huge volume of GatewayEvents.
This Pullrequest replaces these Mutexes with a RwLock, which allows multiple simultaneous readers.
This should not break any existing code.


GnomedDev:
This is a breaking change, please re-target to the next branch.

TheCataliasTNT2k:
I do not know, what this should break, I tested it with current and it worked without any other additions, but I changed it.

GnomedDev:
```rs
let mutex = Arc::new(tokio::sync::Mutex::new("".to_string()));
Shard::new(mutex, "", shard_info, GatewayIntents::all(), None).await?;
```

This is an example of user code that would break if this PR was merged. Breaking changes means *any* public API change that stops compilation, with few exceptions. This PR now needs rebasing to drop the extra commits from current and to solve the conflicts.

TheCataliasTNT2k:
Ah, I forgot about the ws_url, I see. I will hust rebase it onto next, this should be enough.

TheCataliasTNT2k:
I now got a lot of "Unresolved reference: `instrument`", but I ignored them, because they already existed on the branch.

GnomedDev:
Disable the tracing_instrument feature in your builds, it is currently broken, that will get those errors to go away.

TheCataliasTNT2k:
I did not get build errors, `cargo test --features full` runs successfully, jut my IDE is very unhappy.

GnomedDev:
Yes, but you have probably enabled `--all-features` in your IDE, so it enables the broken feature.

src/gateway/bridge/shard_runner.rs
```diff
@@ -42,8 +42,9 @@ pub struct ShardRunner {
     #[cfg(feature = "cache")]
     pub cache: Arc<Cache>,
     pub http: Arc<Http>,
+    // do not use mutex! this is used for almost every gateway event!
```
GnomedDev: Remove this comment, if there is a reason someone PRs this back to Mutex it will have to justify the performance hit anyway.

TheCataliasTNT2k:
I will let rustfmt fix the linter issue.
What to do about the other linter issue? Should I really create a new type alias for that?

src/gateway/bridge/shard_runner.rs
```diff
@@ -171,7 +172,14 @@ impl ShardRunner {
 
             if let Some(event) = event {
                 #[cfg(feature = "collector")]
-                self.collectors.lock().expect("poison").retain_mut(|callback| (callback.0)(&event));
+                {
+                    let read_lock = self.collectors.read();
+                    let to_remove: Vec<_> = read_lock.iter().filter(|callback| !callback.0(&event)).cloned().collect();
+                    drop(read_lock);
+                    if !to_remove.is_empty() {
+                        self.collectors.write().retain(|f| !to_remove.contains(f));
```
GnomedDev, line 183: What happens if a collector gets pushed into this Vec between the read lock being dropped and the write lock being acquired?

src/gateway/bridge/mod.rs
```diff
@@ -97,9 +98,16 @@ pub struct ShardRunnerInfo {
...
+pub struct CollectorCallback(pub Arc<Box<dyn Fn(&Event) -> bool + Send + Sync>>);
 impl std::fmt::Debug for CollectorCallback {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_tuple("CollectorCallback").finish()
     }
 }
+
+impl PartialEq for CollectorCallback {
+    fn eq(&self, other: &Self) -> bool {
+        Arc::as_ptr(&self.0) == Arc::as_ptr(&other.0)
```
GnomedDev: Arc has a method for this, `Arc::ptr_eq`.

src/gateway/bridge/mod.rs
```diff
@@ -97,9 +98,16 @@ pub struct ShardRunnerInfo {
 /// Newtype around a callback that will be called on every incoming request. As long as this
 /// collector should still receive events, it should return `true`. Once it returns `false`, it is
 /// removed.
-pub struct CollectorCallback(pub Box<dyn Fn(&Event) -> bool + Send + Sync>);
+#[derive(Clone)]
+pub struct CollectorCallback(pub Arc<Box<dyn Fn(&Event) -> bool + Send + Sync>>);
```
GnomedDev: Arc does not require T to be sized, so you can store `Arc<dyn Fn(&Event) + ...>`

GnomedDev (changes requested):

src/gateway/bridge/mod.rs
GnomedDev: This should incidentally solve your type complexity lint issue.

src/gateway/bridge/shard_runner.rs
TheCataliasTNT2k: Nothing, the collector will not be run. But the code should not delete "wrong" collectors.

src/gateway/bridge/mod.rs
TheCataliasTNT2k: It has, but the conversion does not work as expected. This method failed my tests.

src/gateway/bridge/shard_runner.rs
GnomedDev: Okay, might be worth adding a comment here as it looks mighty suspicious without looking into it.

src/gateway/bridge/mod.rs
TheCataliasTNT2k: Will update that.

src/gateway/bridge/shard_runner.rs
TheCataliasTNT2k: Ok, I will remove it.

src/collector.rs
```diff
@@ -1,3 +1,8 @@
+// Or we'll get deprecation warnings from our own deprecated type (seriously Rust?)
+#![allow(deprecated)]
```
GnomedDev: Why is this here?

src/gateway/bridge/mod.rs
GnomedDev: I very much doubt that... https://doc.rust-lang.org/stable/src/alloc/sync.rs.html#1786

src/gateway/bridge/shard_runner.rs
TheCataliasTNT2k: I did that

src/collector.rs
TheCataliasTNT2k: This was not added by me, I thought, it was in the Git history somewhere, but I can not find the commit. I will remove it then...

src/gateway/bridge/mod.rs
TheCataliasTNT2k: I did not realise, this fas fixed, because it was a problem some time ago: https://github.com/rust-lang/rust/issues/103763
I will change it.

src/collector.rs
TheCataliasTNT2k: I guess, it was added during rebase onto next yesterday by some of the commits.

TheCataliasTNT2k:
Too stupid to see errors sometimes...

GnomedDev (approved):
Looks good to me, thanks for the PR!

TheCataliasTNT2k:
Thanks for the help!

PR #2852, opened 2024-04-26
Provide old presence data on update

jamesbt365:
Gaming


jamesbt365:
Hopefully this isn't considered "too expensive", though if you are running with presence data you probably are expecting the additional overhead.

Also, anything against collapsing getting the guild into a `let mut guild = cache.guilds.get_mut(guild_id)?;` instead of `if let Some(x)`?

jamesbt365:
Actually ready for review now, I stopped being stupid.

arqunis (approved):

PR #2853, opened 2024-04-26
Adds a way to ignore specific events

TheCataliasTNT2k:
I need a possibility to ignore specific events.
This adds a way to do exactly that.

### DIFF
--- a/src/client/event_handler.rs
+++ b/src/client/event_handler.rs
@@ -30,11 +30,18 @@ macro_rules! event_handler {
                 }
             )*
 
-            /// Check if `event` should be dispatched (`true`) or ignored (`false`) by the bot.
-            /// This affects [`crate::collector::collect`], [`crate::framework::Framework::dispatch`] and this EventHandler trait.
+            /// Checks if the `event` should be dispatched (`true`) or ignored (`false`).
             ///
-            /// WARNING: This method will be run synchronously for every received event in the main dispatch loop of the receiving shard!
-            /// It should be as fast as possible to not block the bot from handling events!
+            /// This affects [`crate::collector::collect`], [`crate::framework::Framework::dispatch`] and this `EventHandler` trait.
+            ///
+            /// ## Warning
+            ///
+            /// This will run synchronously on every event in the dispatch loop
+            /// of the shard that is receiving the event. If your filter code
+            /// takes too long, it may delay other events from being dispatched
+            /// in a timely manner. It is recommended to keep the runtime
+            /// complexity of the filter code low to avoid unnecessarily blocking
+            /// your bot.
             fn filter_event(&self, context: &Context, event: &Event) -> bool {
                     // Suppress unused argument warnings
                     #[allow(dropping_references, dropping_copy_types)]
--- a/src/client/event_handler.rs
+++ b/src/client/event_handler.rs
@@ -504,13 +504,18 @@ pub trait RawEventHandler: Send + Sync {
     /// Dispatched when any event occurs
     async fn raw_event(&self, _ctx: Context, _ev: Event) {}
 
-    /// Check if `event` should be dispatched (`true`) or ignored (`false`) by the bot.
-    /// This affects [`crate::collector::collect`], [`crate::framework::Framework::dispatch`] and
-    /// this EventHandler trait.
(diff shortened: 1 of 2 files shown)


src/client/event_handler.rs
```diff
@@ -29,6 +29,18 @@ macro_rules! event_handler {
                     drop(( $($context,)? $($arg_name),* ))
                 }
             )*
+
+            /// Check if `event` should be dispatched (`true`) or ignored (`false`) by the bot.
+            /// This affects [`crate::collector::collect`], [`crate::framework::Framework::dispatch`] and this EventHandler trait.
+            ///
+            /// WARNING: This method will be run synchronously for every received event in the main dispatch loop of the receiving shard!
+            /// It should be as fast as possible to not block the bot from handling events!
```
arqunis:
```suggestion
            /// Checks if the `event` should be dispatched (`true`) or ignored (`false`).
            ///
            /// This affects [`crate::collector::collect`], [`crate::framework::Framework::dispatch`] and this `EventHandler` trait.
            ///
            /// ## Warning
            ///
            /// This will run synchronously on every event in the dispatch loop
            /// of the shard that is receiving the event. If your filter code
            /// takes too long, it may delay other events from being dispatched
            /// in a timely manner. It is recommended to keep the runtime
            /// complexity of the filter code low to avoid unnecessarily blocking
            /// your bot.
```

src/client/event_handler.rs
```diff
@@ -484,6 +496,20 @@ event_handler! {
...
     /// Dispatched when any event occurs
     async fn raw_event(&self, _ctx: Context, _ev: Event) {}
+
+    /// Check if `event` should be dispatched (`true`) or ignored (`false`) by the bot.
+    /// This affects [`crate::collector::collect`], [`crate::framework::Framework::dispatch`] and
+    /// this EventHandler trait.
+    ///
+    /// WARNING: This method will be run synchronously for every received event in the main dispatch
+    /// loop of the receiving shard! It should be as fast as possible to not block the bot from
+    /// handling events!
```
arqunis:
```suggestion
    /// Checks if the `event` should be dispatched (`true`) or ignored (`false`).
    ///
    /// This affects [`crate::collector::collect`], [`crate::framework::Framework::dispatch`] and this `EventHandler` trait.
    ///
    /// ## Warning
    ///
    /// This will run synchronously on every event in the dispatch loop
    /// of the shard that is receiving the event. If your filter code
    /// takes too long, it may delay other events from being dispatched
    /// in a timely manner. It is recommended to keep the runtime
    /// complexity of the filter code low to avoid unnecessarily blocking
    /// your bot.
```

src/gateway/bridge/shard_runner.rs
```diff
@@ -170,29 +170,43 @@ impl ShardRunner {
...
+                let context = self.make_context();
+                if match &self.event_handler {
+                    Some(InternalEventHandler::Normal(handler)) => {
+                        handler.filter_event(&context, &event)
+                    },
+                    Some(InternalEventHandler::Raw(handler)) => {
+                        handler.filter_event(&context, &event)
+                    },
+                    None => true,
+                } {
```
arqunis:
```suggestion
                let can_dispatch = match &self.event_handler {
                    Some(InternalEventHandler::Normal(handler)) => {
                        handler.filter_event(&context, &event)
                    },
                    Some(InternalEventHandler::Raw(handler)) => {
                        handler.filter_event(&context, &event)
                    },
                    None => true,
                };

                if can_dispatch {
```

arqunis (changes requested):

src/gateway/bridge/shard_runner.rs
```diff
@@ -170,29 +170,45 @@ impl ShardRunner {
...
+                        // this compares the inner pointer of the Arc
+                        if !to_remove.is_empty() {
+                            self.collectors.write().retain(|f| !to_remove.contains(f));
+                        }
                     }
+                    spawn_named(
+                        "shard_runner::dispatch",
+                        dispatch_model(
+                            event,
+                            self.make_context(),
```
arqunis:
```suggestion
                            context,
```
So the context created for `filter_event` is reused.

arqunis (changes requested):

arqunis (approved):

PR #2854, opened 2024-04-28
Implements changes to allow developers more control about event handling

TheCataliasTNT2k:
Adds Framework::dispatch_automatically
Makes client::dispatch::update_cache_with_event public, so it can be called manually by developers


GnomedDev (changes requested):
I dislike `update_cache_with_event` becoming public, as it's quite internal. If you could make a nicer API for this that doesn't require making internal modules public, that would resolve this.

TheCataliasTNT2k:
The only thing in this method I really care about, is to get the returned values.
I do not really want to copy this function to get the FullEvent objects to use them in my library. If this can be achieved in any other way (while not losing the cache functionality) I am happy.

cheesycod:
This would be useful for me as well

TheCataliasTNT2k:
I can not really create a better API, because I need to update the cache.
Do you have any suggestions?
(I will resolve the conflicts when all changes are approved.)

> I dislike `update_cache_with_event` becoming public, as it's quite internal. If you could make a nicer API for this that doesn't require making internal modules public, that would resolve this.

Already did the rebase (or rather just moved my commits to the new HEAD).

cheesycod:
Any updates on this?

TheCataliasTNT2k:
I don't think so 🤷
I am just using a fork which has everything I need, which is rejected by the maintainers.

GnomedDev:
Closing as you have said you are not interested in getting this to a mergable state.

TheCataliasTNT2k:
> Closing as you have said you are not interested in getting this to a mergable state.

I did not say, that I am not interested.
I proposed changes, you were unhappy, and I asked you, which solution you would prefer to achieve the same functionality [here](https://github.com/serenity-rs/serenity/pull/2854#issuecomment-2086969687) and [here](https://github.com/serenity-rs/serenity/pull/2854#issuecomment-2159149613).
You never replied to any of these comments.

Give me a solution, which you would accept, I will implement it and rebase my branch :woman_shrugging: .

TheCataliasTNT2k:
If you need a reference what I am doing with this change:
I created my own framework based off of poise and serenity.
I want to be able to control event handling how I need it to to be.
So I use the "raw_event_handler" method provided by poise / serenity and handle the events myself.
But I need a way to convert the raw events into FullEvents and update the cache.
This is done by this function exactly and therefore it needs to be public.
There is no "nicer API" I can think of.

PR #2878, opened 2024-05-28
Replace Framework/VoiceGatewayManager with PassthroughEventHandler

GnomedDev:
WIP, may require an `init` method to be added to `EventHandler` before this is a good idea.


GnomedDev:
I'm going to revisit this after my EventHandler redesign, as it turns out that initialization is probably a good idea.

PR #2888, opened 2024-06-05
Add into_inner for CreateEmbed's wrapped Embed field

calum4:
Admittedly, it's an incredibly niche change, but it annoyed me so here we are.

### DIFF
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -198,6 +198,13 @@ impl CreateEmbed {
         super::check_overflow(length, crate::constants::EMBED_MAX_LENGTH)
             .map_err(|overflow| Error::Model(ModelError::EmbedTooLarge(overflow)))
     }
+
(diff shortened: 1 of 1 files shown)


src/builder/create_embed.rs
```diff
@@ -198,6 +198,13 @@ impl CreateEmbed {
         super::check_overflow(length, crate::constants::EMBED_MAX_LENGTH)
             .map_err(|overflow| Error::Model(ModelError::EmbedTooLarge(overflow)))
     }
+
+    /// Retrieves the inner [`Embed`].
+    #[inline]
+    #[must_use]
+    pub fn get(self) -> Embed {
```
arqunis: Methods that consume and convert a type to another should follow the `into_*` convention:
```suggestion
    pub fn into_inner(self) -> Embed {
```

arqunis (changes requested):

arqunis (approved):

PR #2887, opened 2024-06-03
Add serde rename for `burst_colors`

thmasq:

PR #2893, opened 2024-06-07
Fix deserialization of burst_colours on MessageReaction

### LINKED ISSUE #2892: Invalid Type Error
Iterating over all the messages in a server yields me an invalid type error every time. The error is:
`invalid type: string "#d0334d", expected u32 at line 1 column 123537`
I don't really know what to do in order to help debug this but I assume this is a deserialization error.
Branch: `current`

jamesbt365:
Fixes #2892.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -22,7 +22,7 @@ use crate::gateway::ShardMessenger;
 #[cfg(feature = "model")]
 use crate::http::{CacheHttp, Http};
 use crate::model::prelude::*;
-use crate::model::utils::StrOrInt;
+use crate::model::utils::{discord_colours, StrOrInt};
 #[cfg(all(feature = "model", feature = "cache"))]
 use crate::utils;
 
@@ -947,7 +947,7 @@ pub struct MessageReaction {
     #[serde(rename = "emoji")]
     pub reaction_type: ReactionType,
     // The colours used for super reactions.
-    #[serde(rename = "burst_colors")]
+    #[serde(rename = "burst_colors", deserialize_with = "discord_colours")]
     pub burst_colours: Vec<Colour>,
 }
 
--- a/src/model/channel/reaction.rs
+++ b/src/model/channel/reaction.rs
@@ -8,7 +8,6 @@ use std::str::FromStr;
 use percent_encoding::{utf8_percent_encode, NON_ALPHANUMERIC};
 use serde::de::Error as DeError;
 use serde::ser::{Serialize, SerializeMap, Serializer};
-use serde_cow::CowStr;
 #[cfg(feature = "model")]
 use tracing::warn;
 
@@ -16,6 +15,7 @@ use tracing::warn;
 use crate::http::{CacheHttp, Http};
 use crate::internal::prelude::*;
 use crate::model::prelude::*;
+use crate::model::utils::discord_colours_opt;
 
 /// An emoji reaction to a message.
 ///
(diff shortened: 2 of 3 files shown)


GnomedDev (changes requested):
I'm not merging this as-is, we cannot duplicate logic like this. There must be a better way to do this.

jamesbt365:
I couldn't think of a nicer way at the time, It would probably involve the creation of another module (so that it doesn't reexport the function in serenity::all, or changing it carefully so it reexports *everything* but this function) alongside somehow handling the difference between the optional and non optional version.

Do you have any suggestions in mind?

GnomedDev:
Can this not go in `/src/model/utils.rs`?

jamesbt365:
I didn't think about moving it there to be fair, <sub>I don't think much</sub>, that only leaves one issue with one being an option and the other not.

jamesbt365:
Would this pass?

```rs
fn discord_colours_opt<'de, D>(deserializer: D) -> Result<Option<Vec<Colour>>, D::Error>
where
    D: Deserializer<'de>,
{
    let vec_str: Option<Vec<CowStr<'_>>> = Deserialize::deserialize(deserializer)?;

    let Some(vec_str) = vec_str else { return Ok(None) };

    if vec_str.is_empty() {
        return Ok(None);
    }

    deserialize_colours::<D>(vec_str).map(Some)
}

fn discord_colours<'de, D>(deserializer: D) -> Result<Vec<Colour>, D::Error>
where
    D: Deserializer<'de>,
{
    let vec_str: Vec<CowStr<'_>> = Deserialize::deserialize(deserializer)?;

    deserialize_colours::<D>(vec_str)
}

fn deserialize_colours<'de, D>(vec_str: Vec<CowStr<'_>>) -> Result<Vec<Colour>, D::Error>
where
    D: Deserializer<'de>,
{
    vec_str
        .into_iter()
        .map(|s| {
            let s = s.0.strip_prefix('#').ok_or_else(|| DeError::custom("Invalid colour data"))?;

            if s.len() != 6 {
                return Err(DeError::custom("Invalid colour data length"));
            }

            u32::from_str_radix(s, 16)
                .map(Colour::new)
                .map_err(|_| DeError::custom("Invalid colour data"))
        })
        .collect()
}
```

GnomedDev:
Sure

AlexCzar:
Do I understand correctly, that 0.12.* won't work with reactions until this is merged?
I'm trying to update my bot to latest version and see following error:
```none
missing field `burst_colours` at line 1 column 816
```
at
```rust
    ctx.http
        .get_message(reaction.channel_id.into(), message_id.into())
        .await
```

jamesbt365:
This is a 2 part fix, the 1st part was #2887.

AlexCzar:
Can't wait :) I tried applying it myself and specifying git dependency, but some magic happens, and my shuttle app refuses to recognize `serenity::Client` 🤷🏽

GnomedDev (approved):

FichteFoll:
Hi, I'd just like to mention that a release with this fix would be much appreciated, also considering the number of duplicate issues of #2892 since people keep running into it. For now, I'm staying on 0.10 which still works but there have been some breaking API changes since then, so I presume not everyone would be happy to revert to that.

gkuznik:
@FichteFoll you can use a newer version, this issue was introduced in 0.12.2. `serenity = { version = "=0.12.1" }` in Cargo,.toml also works if you use Poise

PR #2901, opened 2024-06-12
Adds better abilities to check, what exactly was rate limited

TheCataliasTNT2k:
Add "OwnedRoute" to "RatelimitInfo" which allows to check the cause of the rate limit and which identifiers were affected (e.g. which channel).


TheCataliasTNT2k:
The failing check seems broken, at least I do not get any error when looking at the details...

jamesbt365:
> The failing check seems broken, at least I do not get any error when looking at the details...

`cargo +nightly fmt --all`

src/http/routing.rs
```diff
@@ -80,412 +145,534 @@ macro_rules! routes {
                 }))
             }
 
+            /// This returns the common identifiers, which are used in rate limit buckets.
+            /// If you are interested in less common identifiers,
+            /// please match the variants you are interested in.
```
GnomedDev: This documentation just duplicates the `RatelimitCause` docs, remove it.

src/http/routing.rs
```diff
@@ -80,412 +145,534 @@ macro_rules! routes {
...
+            #[must_use]
+            #[allow(unused_variables)] // prevent compiler from complaining about unused variables
```
GnomedDev: This comment is completely superfluous, it's kinda obvious.

src/http/routing.rs
```diff
@@ -27,23 +27,24 @@ enum RatelimitingKind {
...
+            $name:ident $({ $($field_name_route:ident: $field_type:ty | $field_type_owned:ty),* })?,
+            { $($field_name:ident),* },
             $path:expr,
             $ratelimiting_kind:expr;
         )+
     }) => {
+        #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+        enum RouteKind {
+            $($name,)+
+        }
```
GnomedDev, line 39: Why was this moved?

src/http/routing.rs
```diff
@@ -80,412 +145,534 @@ macro_rules! routes {
                 }))
             }
 
+            /// This returns the common identifiers, which are used in rate limit buckets.
+            /// If you are interested in less common identifiers,
+            /// please match the variants you are interested in.
+            #[must_use]
+            #[allow(unused_variables)] // prevent compiler from complaining about unused variables
+            pub fn get_common_identifiers(&self) -> RatelimitCause {
```
GnomedDev, line 148:
```suggestion
            pub fn cause(&self) -> RatelimitCause {
```

src/http/routing.rs
```diff
@@ -27,23 +27,24 @@ enum RatelimitingKind {
 macro_rules! routes {
     ($lt:lifetime, {
         $(
-            $name:ident $({ $($field_name:ident: $field_type:ty),* })?,
+            $name:ident $({ $($field_name_route:ident: $field_type:ty | $field_type_owned:ty),* })?,
```
GnomedDev, line 30: Instead of making the macro syntax messier, can you make this use traits instead? I'm pretty sure `ToOwned` should handle this, but if it doesn't you can just write a simple private trait.

src/http/routing.rs
```diff
@@ -67,7 +68,71 @@ macro_rules! routes {
...
+                    let id = match r {
+                        RatelimitingKind::PathAndId(id) => Some(id),
+                        RatelimitingKind::Path => None,
+                    };
+                    (self.kind(), id)
+                }))
+            }
+
+            #[must_use]
+            pub fn get_owned_route(&self) -> OwnedRoute {
```
GnomedDev: Rust convention is to name getters without the prefix, but since this is a `&self` -> `Owned` I think the convention would be `to_owned_route`... however that's just ToOwned, so implement `ToOwned` instead.

GnomedDev (changes requested):

src/http/routing.rs
TheCataliasTNT2k: How? I somehow need to know the type, which the OwnedRoute should use. I do not know, how a trait would be able to do that.
The problem is the "&str" type. Can you please provide an example for how to do it?

src/http/routing.rs
TheCataliasTNT2k: I did not use `toOwned`, because this is not really a clone and should not be treated as one, which is expected by the trait however.

src/http/routing.rs
TheCataliasTNT2k: This does not really return the "cause" of the rate limit but the parts of the path, which was rate limited.
The "cause" would be the deletion of a message or something like that.

src/http/routing.rs
TheCataliasTNT2k: To have two similar blocks next to another (`Route` and `OwnedRoute`).
I could move it back, but in my opinion it is more readable like this.

src/http/routing.rs
TheCataliasTNT2k: Removed

src/http/routing.rs
TheCataliasTNT2k: Removed

src/http/routing.rs
GnomedDev: With that logic `ToOwned` should not be a thing, but this is exactly what it's for. This is internally just converting `str -> String` which is exactly the point of `ToOwned`.

src/http/routing.rs
GnomedDev: So the type should not be caused `RatelimitCause`, it should be called `RatelimitPathParts` or something else.

src/http/routing.rs
GnomedDev: The owned type is simply `<$field_type as ToOwned>::Owned`, and if that doesn't return the correct types you can duplicate the definition of `ToOwned` and implement it for the types needed.

src/http/routing.rs
TheCataliasTNT2k: That is better, yes.
Can the method name stay then?

src/http/routing.rs
TheCataliasTNT2k: Will do that.

TheCataliasTNT2k: The `field_type_owned` is actually needed, I did not find a way to get compiling code without it.

Reason:
The macro would expand into something like this (irrelevant code removed):
```rust
#[derive(Clone, Debug)]
pub enum OwnedRoute {
    ChannelMessageReaction { channel_id: <ChannelId as ToOwned>::Owned, message_id: <MessageId as ToOwned>::Owned, user_id: <UserId as ToOwned>::Owned, reaction: <&'a str as ToOwned>::Owned},
}
```

`<&'a str as ToOwned>::Owned` is not valid, because `'a` does not exist. I could add the lifetime to the enum, but then we would need to carry this useless lifetime though the whole code. This can not be fixed by replacing the `<&'a str as ToOwned>::Owned` with a custom macro, I already tried:
```rust
macro_rules! get_owned_type {
    (&'a str) => {String};
    ($type:ty) => {<$type as ToOwned>::Owned};
}

#[derive(Clone, Debug)]
pub enum OwnedRoute {
    ChannelMessageReaction { channel_id: get_owned_type!(ChannelId), message_id: get_owned_type!(MessageId), user_id: get_owned_type!(UserId), reaction: get_owned_type!(&'a str)},
}
```

will not compile for the same reason.

The compiler is not smart enough to ignore the unused lifetime at the moment.

src/http/routing.rs
TheCataliasTNT2k: This can not be done using `ToOwned`, because this would require the "copied" object to carry the lifetime with it.

I already tried that sometime, and it makes the whole code a mess.

I renamed the method though (not pushed yet because of the other change requests).

GnomedDev (changes requested):
I cannot review this PR without it being rebased correctly on-top of next.

TheCataliasTNT2k:
> I cannot review this PR without it being rebased correctly on-top of next.

Was doing that right as you wrote that, lol xD

PR #2905, opened 2024-06-18
Add support for premium buttons

### LINKED ISSUE #2904: Update to the monetization API
The [Discord API documentation](https://github.com/discord/discord-api-docs/commit/4853fbc9ba8df174fc8bf1abbe8a19bc8f14cd1c) updated with a few changes:
- Added a new sku_id field to the interaction buttons model.
- Added a new button style.
- Updated the conflicts between button fields.
- Deprecated the PREMIUM_REQUIRED interaction response type.
- Added 2 new links to store items (`Entitlement` and `Sku` could implement methods that could return these URLs):
  - `https://discord.com/application-directory/:appID/store`
  - `https://discord.com/application-directory/:appID/store/:skuID`

mkrasnitski:
Fixes #2904

Adds support for the new premium buttons that replace the `PREMIUM_REQUIRED` interaction response type, which has been deprecated.

### DIFF
--- a/src/builder/create_components.rs
+++ b/src/builder/create_components.rs
@@ -53,6 +53,21 @@ impl CreateButton {
         })
     }
 
+    /// Creates a new premium button associated with the given SKU.
+    ///
+    /// Clicking this button _will not_ trigger an interaction event in your bot.
+    pub fn new_premium(sku_id: impl Into<SkuId>) -> Self {
+        Self(Button {
+            kind: ComponentType::Button,
+            data: ButtonKind::Premium {
+                sku_id: sku_id.into(),
+            },
+            label: None,
+            emoji: None,
+            disabled: false,
+        })
+    }
+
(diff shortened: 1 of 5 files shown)


vicky5124:
Are there any use cases for the store link without a SKU? If there are, it may be worth implementing it, otherwise, it is fine to be left out IMO.

mkrasnitski:
The store link does in fact get rendered as a rich embed too ([docs](https://discord.com/developers/docs/monetization/managing-your-store#linking-to-your-store)), however I don't really know which struct would be best suited to have a method that returns the url.

vicky5124:
Both could either have an additional method, or make the existing `url()` method require a `with_sku: bool` argument.

Alternatively, or additively, `IntegrationApplication`, `PartialCurrentApplicationInfo` and `CurrentApplicationInfo` could have the method, as they hold the application ID.

mkrasnitski:
Sorry for the delay, should be good for merge now.

arqunis (approved):

PR #2914, opened 2024-07-02
Add display_name method to user

jamesbt365:
Yeah this annoyed me when I was implementing this in a bot. So I stole the method from `Member`!


GnomedDev (commented):
LGTM, but can you add a note that this does not take into account any guild specific information as only a member has that information.

PR #2915, opened 2024-07-03
Implement From<Button> for CreateButton

thou-vow:


GnomedDev:
This was already merged into `current` in #2913 and will be rebased soon, merging this would just cause a massive headache when next rebasing.

Issue #2883, opened 2024-05-30
Error in CurrentApplicationInfo causing crash in certain case


ValgulNecron:
In current branch if we try to access application info with `ctx.http.get_current_application_info().await.unwrap();` we get the error

 `thread 'tokio-runtime-worker' panicked at src\main.rs:162:65:
called "Result::unwrap()" on an "Err" value: Json(Error("missing field "oauth2_install_params"", line: 1, column: 706))
stack backtrace: ` 

this is due by the fact that bot are not required to have "installation link" (option are none, Discord provided link and custom url).
if we check the `CurrentApplicationInfo `there is a `integration_types_config `which take an `InstallationContextConfig` and inside it there is `oauth2_install_params: InstallParams` wich should be from my understanding of [discord docs](https://discord.com/developers/docs/resources/application#application-object-application-integration-types) be an option since there is the ? next to the field name.

PR #2917, opened 2024-07-15
Fix null error in InstallationContextConfig

ValgulNecron:
Added #[serde(default)] to oauth2_install_params in InstallationContextConfig and added a Default implementation to InstallParams #2883 and https://discord.com/developers/docs/resources/application#application-object-application-structure.

### DIFF
--- a/src/model/application/mod.rs
+++ b/src/model/application/mod.rs
@@ -124,7 +124,8 @@ enum_number! {
 #[cfg(feature = "unstable_discord_api")]
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct InstallationContextConfig {
-    pub oauth2_install_params: InstallParams,
+    #[serde(default)]
+    pub oauth2_install_params: InstallParams, // TODO: Replace with an option instead of #[serde(default)]
 }
 
 /// Information about the Team group of the application.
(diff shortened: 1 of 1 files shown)


src/model/application/mod.rs
```diff
@@ -124,7 +124,8 @@ enum_number! {
 #[cfg(feature = "unstable_discord_api")]
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct InstallationContextConfig {
-    pub oauth2_install_params: InstallParams,
+    #[serde(default)]
+    pub oauth2_install_params: InstallParams, // TODO: Replace with an option instead of #[serde(default)]
```
mkrasnitski, line 128:
```suggestion
    // TODO: Replace with an option instead of #[serde(default)]
    #[serde(default)]
    pub oauth2_install_params: InstallParams,
```
Formatting nit

mkrasnitski:
Since `CurrentApplicationInfo::integration_types_config` is gated behind `unstable_discord_api`, it's actually ok to break the serenity API and change its `oauth2_install_params` field to `Option<InstallParams>`.

ValgulNecron:
Hi thanks for the info i changed it to an option and removed the change I made before

mkrasnitski (approved):

PR #2918, opened 2024-07-19
Change type of `Http::application_id` to `OnceLock<ApplicationId>`

mkrasnitski:
To avoid converting `HttpBuilder::application_id` into u64 and then back again, we can use `OnceLock` to provide atomic access to the field. We don't actually ever need to re-assign the value, and because it never makes sense for an application id to change, we panic if the value is ever different than what is already stored.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1,9 +1,7 @@
 #![allow(clippy::missing_errors_doc)]
 
 use std::borrow::Cow;
-use std::num::NonZeroU64;
-use std::sync::atomic::{AtomicU64, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 
 use percent_encoding::{utf8_percent_encode, NON_ALPHANUMERIC};
 use reqwest::header::{HeaderMap as Headers, HeaderValue};
@@ -141,7 +139,10 @@ impl HttpBuilder {
     /// Use the given configuration to build the `Http` client.
     #[must_use]
     pub fn build(self) -> Http {
-        let application_id = AtomicU64::new(self.application_id.map_or(0, ApplicationId::get));
+        let application_id = match self.application_id {
+            Some(id) => id.into(),
+            None => OnceLock::new(),
+        };
 
         let client = self.client.unwrap_or_else(|| {
             let builder = configure_client_backend(Client::builder());
@@ -197,7 +198,7 @@ pub struct Http {
     pub ratelimiter: Option<Ratelimiter>,
     pub proxy: Option<String>,
     token: SecretString,
-    application_id: AtomicU64,
+    application_id: OnceLock<ApplicationId>,
     pub default_allowed_mentions: Option<CreateAllowedMentions>,
 }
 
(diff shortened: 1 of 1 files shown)


GnomedDev:
I'm a little confused by this, are there any other benefits than panicking if the application id changes?

mkrasnitski:
The main benefit of storing `ApplicationId` is not constructing it over again after setting it `HttpBuilder` or `ClientBuilder`, or after receiving it from the gateway (plus using a sentinel value to signify "not set", i.e. `0` on current and `u64::MAX` on next). Obviously we still need atomic access but `Mutex` would be a bad choice due to possible contention, and `RwLock` might serve fine but cleanly dealing with potential poisoning was annoying and easy to get wrong. Therefore `OnceLock` provides the lock-free atomic access we want as well as interior mutability ala `Cell`.

I added the panic to encode what I'm assuming is an invariant of the application ID never changing, given that authenticating to the Discord API is done per-application (with the token corresponding 1-to-1 with the application ID). It's also kind of a sanity check for the call to `OnceLock::set` - once the inner value is set, it won't actually be replaced by the value passed in, but if the new value would have been the same it's a no-op in my mind.

GnomedDev:
Okay, so if this PR is:

Pros:
- No need to check the ApplicationId's validity (NonZero, or NonMax) every time it is fetched
- Extra sanity check

Cons:
- Extra atomic overhead because OnceLock

I don't see that it's worth it? Checking the validity of `ApplicationId` is trivial and this isn't a complexity win as the extra sanity check adds mental overhead.

mkrasnitski:
Compared to `AtomicU64` is there any extra overhead to `OnceLock`? I would also argue that the sanity check is less mental overhead than checking a sentinel, whose value is really just an implementation detail.

GnomedDev:
Yes, somewhat significant compared to the AtomicU64's single load and set instructions. https://godbolt.org/z/T78sGM8YY

mkrasnitski:
I see what you mean - however, that overhead is incurred only once on initialization. For easier comparison, I split the `get` and `set` cases into separate godbolts.

 - get: https://godbolt.org/z/K5TbhWqez
   - Here the compiled output is almost identical. There is an opportunity for a niche optimization to be introduced to the compiler to signify "not initialized". 
 - set: https://godbolt.org/z/b5qroETbG
   - Yes, the first initialization is more expensive and involves a `futex()` syscall (on linux at least). But subsequent calls take the fast path through `.LBB3_4` with very little overhead.

Additionally, I was thinking about the potential maintenance burden of exposing the representation of the sentinel ID value. Since that value is actually changed on `next`, cases like these bring the risk of mistakenly leaving the sentinel value unchanged if the internal representation were ever changed to something else.

Side note, but does `nonmax` make use of any niche optimizations, or is the compiler not smart enough to do so?

GnomedDev:
Okay, so the pro instead is less maintenance burden in case the internal representation is changed again (doubtful, and all calls to Id::new would have to be audited anyway), and the cons stay the same.

For your side note, yes, `NonMax*` is just `NonZero*` with the `0` represented as `*::MAX`.

mkrasnitski:
Oh well, I guess it's not worth it. If rust had `AtomicNonZero` or maybe even some sort of generic `Atomic<T>` for cheap T, then this would work.

mkrasnitski:
Closing for now.

PR #2920, opened 2024-07-20
Add support for Application Emojis

mkrasnitski:


src/client/context.rs
```diff
@@ -308,6 +309,61 @@ impl Context {
...
+    ///
+    /// # Errors
+    ///
+    /// See [`Guild::create_emoji`] for information about name and filesize requirements. This
+    /// method will error if said requirements are not met.
+    pub async fn create_application_emoji(&self, name: &str, image: &str) -> Result<Emoji> {
+        let map = json!({
+            "name": name,
+            "image": image,
+        });
```
GnomedDev: Can you please use a struct with `Deserialize` instead of `json!`? I've been removing this kind of pattern as it allocates and leads to the http interface being restricted to only `Value` instead of the `&impl Serialize` it actually needs.

src/client/context.rs
```diff
@@ -308,6 +309,61 @@ impl Context {
...
+
+    /// Changes the name of an application emoji.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the emoji does not exist.
+    pub async fn edit_application_emoji(&self, emoji_id: EmojiId, name: &str) -> Result<Emoji> {
+        let map = json!({
+            "name": name,
+        });
```
GnomedDev: Ditto to above.

src/http/client.rs
```diff
@@ -511,6 +511,25 @@ impl Http {
         .await
     }
 
+    /// Creates an application emoji with the given data.
+    ///
+    /// See [`Context::create_application_emoji`] for required fields.
+    ///
+    /// [`Context::create_application_emoji`]: crate::client::Context::create_application_emoji
+    pub async fn create_application_emoji(&self, map: &Value) -> Result<Emoji> {
```
GnomedDev: Change to `map: &impl Serialize`.

src/http/client.rs
```diff
@@ -1609,6 +1644,26 @@ impl Http {
         .await
     }
 
+    /// Changes application emoji information.
+    ///
+    /// See [`Context::edit_application_emoji`] for required fields.
+    ///
+    /// [`Context::edit_application_emoji`]: crate::client::Context::edit_application_emoji
+    pub async fn edit_application_emoji(&self, emoji_id: EmojiId, map: &Value) -> Result<Emoji> {
```
GnomedDev: Change to `map: &impl Serialize`.

src/client/context.rs
mkrasnitski: Done.

GnomedDev (approved):

arqunis (approved):

PR #2931, opened 2024-08-05
Fix deserialisation failure of audit logs when the user ID is absent

### LINKED ISSUE #2929: Null error in audit logs.
if we try to get the audit log like this
```rs
guild_id
        .audit_logs(&ctx.http, None, None, None, Some(100))
 ```
and a webhook updated itself before with something like this
```rs
let attachment = CreateAttachment::bytes(decoded_bytes, "avatar");
let edit_webhook = EditWebhook::new().name(trimmed_name).avatar(&attachment);
webhook.edit(&ctx.http, edit_webhook)
```
we will get the error
```
invalid type: null, expected a non-zero string or integer snowflake at line 1 column 2363
```
because the user_id field is null
```json
{"id":"1253136095328337940","action_type":51,"user_id":null,"target_id":"1253136028441510109","changes":[{"key":"avatar_hash","new_value":"8120e9729eda8cfec8fcf84b7f69833f","old_value":"68787e5ed700348a5eb2a25f274d2dcb"},{"key":"name","new_value":"Jellyfish Can\\u2019t Swim in the Night / Yoru no Kura","old_value":"THE NEW GATE / THE NEW GATE"}]}
```

arqunis:
Kind of fixes #2929.

Real fix would be to account for this at the type level with `Option<T>`, but this change cannot be committed to `current` as it is a breaking change.

### DIFF
--- a/src/model/guild/audit_log/mod.rs
+++ b/src/model/guild/audit_log/mod.rs
@@ -344,6 +344,8 @@ pub struct AuditLogEntry {
     pub action: Action,
     /// What was the reasoning by doing an action on a target? If there was one.
     pub reason: Option<String>,
+    // FIXME: Remove this attribute and change the type to `Option<UserId>` on `next`.
+    #[serde(default)]
     /// The user that did this action on a target.
     pub user_id: UserId,
     /// What changes were made.


arqunis:
Actually not as simple of a fix as I initially thought. The `#[serde(default)]` attribute doesn't take into account the case where the field exists, but a `null` was provided as its value. And the [real fix is already present on `next`](https://github.com/serenity-rs/serenity/blob/640fa8a518f04e49a7dc61281c6b13080dba2dc7/src/model/guild/audit_log/mod.rs#L386).

I'll just close this. While I could add custom deserialisation behaviour to allow the `null` case, it would have to be removed anyway afterwards.

Issue #2475, opened 2023-06-24
Support for oauth2 token exchanges?


ghost:
Would it be possible to add oauth2 token support? Having a library that supports the discord api, but having to manually make access token exchanges is tedious.

[Here's an example of what has to be sent](https://discord.com/developers/docs/topics/oauth2#authorization-code-grant-access-token-exchange-example)

[Here's an example response](https://discord.com/developers/docs/topics/oauth2#authorization-code-grant-access-token-response)

It would be much appreciated!

valentinegb:
I second this! It seems like I can't rely on Serenity for OAuth2 stuff at all since it doesn't even have the OAuth2 endpoints in [`Route`](https://docs.rs/serenity/latest/serenity/http/enum.Route.html)...

GnomedDev:
This looks like a duplicate of #762.

Issue #762, opened 2020-01-02
Better oauth2 support


SOF3:
I tried using serenity with the web flow login, and I have facing these problems:
- A new `reqwest::Client` has to be created for each session, which is not optimal for performance
- The authorization and token endpoints are not available via this library.

PR #2935, opened 2024-08-06
Add `OAuth2Token`, `OAuth2TokenRevocation`, and `OAuth2AuthorizationCurrent` to `Route`

valentinegb:
Adds routes to `Route` which are related to OAuth2 and token exchanges. This is a step towards resolving #762 and #2475.

### DIFF
--- a/src/http/routing.rs
+++ b/src/http/routing.rs
@@ -361,8 +361,20 @@ routes! ('a, {
     api!("/invites/{}", code),
     Some(RatelimitingKind::Path);
 
-    Oauth2ApplicationCurrent,
+    OAuth2Token,
+    api!("/oauth2/token"),
+    None;
+
+    OAuth2TokenRevocation,
+    api!("/oauth2/token/revoke"),
+    None;
+
+    OAuth2ApplicationCurrent,
     api!("/oauth2/applications/@me"),
+    None;
+
+    OAuth2AuthorizationCurrent,
+    api!("/oauth2/@me"),
     None;
 
     StatusIncidentsUnresolved,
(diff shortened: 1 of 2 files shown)


src/http/routing.rs
```diff
@@ -496,4 +499,69 @@ routes! ('a, {
...
+            format!("&token_type_hint={token_type_hint}")
+        } else {
+            String::new()
+        },
+    ),
+    None;
+
+    OAuth2AuthorizationCurrent,
+    api!("/oauth2/@me"),
+    None;
```
mkrasnitski: Query params for endpoints shouldn't be defined in this way. Instead, they should be passed using the `Request::params` field, see [example](https://github.com/serenity-rs/serenity/blob/658b6a7261fbde7cf59f84613f870ef7a48735c6/src/http/client.rs#L2749-L2782).

valentinegb: Oh whoops, wish I knew that before lol, that makes this PR *way* simpler

src/http/routing.rs
```diff
@@ -496,4 +496,16 @@ routes! ('a, {
...
+    api!("/oauth2/token"),
+    None;
+
+    OAuth2TokenRevocation,
+    api!("/oauth2/token/revoke"),
+    None;
+
+    OAuth2AuthorizationCurrent,
+    api!("/oauth2/@me"),
+    None;
```
mkrasnitski: Could you group these together with the other OAuth route?

valentinegb: Sure

mkrasnitski (approved):

PR #2954, opened 2024-08-23
Fix and make `edit_role_position` plural

mkrasnitski:
There was a regression introduced to the `edit_role_position` endpoint in #2806 - this endpoint expects an array of json objects but a single object was being sent to the API, not inside of an array. This means the response would always error due to invalid format.

This PR fixes said bug by pluralizing the endpoint to take a list of roles to edit. The list will be serialized as a JSON array.

### DIFF
--- a/src/model/guild/guild_id.rs
+++ b/src/model/guild/guild_id.rs
@@ -794,15 +794,28 @@ impl GuildId {
         builder.execute(http, self, sticker_id).await
     }
 
-    /// Edit the position of a [`Role`] relative to all others in the [`Guild`].
+    /// Edits the position of [`Role`]s relative to all others in the [`Guild`].
     ///
     /// **Note**: Requires the [Manage Roles] permission.
     ///
     /// # Examples
     ///
-    /// ```rust,ignore
-    /// use serenity::model::{GuildId, RoleId};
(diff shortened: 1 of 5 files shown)


src/model/guild/guild_id.rs
```diff
@@ -827,12 +839,15 @@ impl GuildId {
...
-            id: role_id,
-            position,
-        };
+        let map = roles
+            .into_iter()
+            .map(|(id, position)| EditRole {
+                id,
+                position,
+            })
+            .collect::<Vec<_>>();
```
GnomedDev: Use [`SerializeIter`](<https://github.com/serenity-rs/serenity/blob/8b201e9708aa15d2a45b35f4457908aeb7a52d16/src/model/utils.rs#L75-L93>) here

GnomedDev (changes requested):
Either change this to use `SerializeIter`, or change the arguments to take a slice instead of an iterator, as otherwise this is just taking something different to what is needed.

src/model/guild/guild_id.rs
mkrasnitski: Done

GnomedDev (approved):

Issue #2948, opened 2024-08-17
Poll results added to the API


vicky5124:
A new poll result message has been added. This includes a new message type with the ID 46 called `POLL_RESULT`, as well as a new embed type with the field name `poll_result`, and a new type to accompany it.

Note that the types of the fields in the new type have not been documented, the documentation only shows if they are optional fields or not, as well as their description. A review of the pull request notes that all the values are strings, but this has not been validated, and for ergonomics, it may be best to parse the values to their appropriate type, which is something the review also mentioned.

[Discord API docs change commit (Poll)](https://github.com/discord/discord-api-docs/commit/74c2762bb4f707119622fd3a141c97144960428c) | [PR](https://github.com/discord/discord-api-docs/pull/7050)

There's also a new deletable message type with the ID 44 called `PURCHASE_NOTIFICATION` that was missed, that could be easily added along with this change.

[Discord API docs change commit (Purchase)](https://github.com/discord/discord-api-docs/commit/174c1a18933e068d70d56c2340b9936eed8b366a)

ivinjabraham:
Working on this right now. From what I understand, the `fields` in an Embed are optional and we provide support for them as [documented](https://discord.com/developers/docs/resources/message#embed-object-embed-field-structure). This appears to be the first time we've been given the [possible fields](https://discord.com/developers/docs/resources/message#embed-fields-by-embed-type-poll-result-embed-fields) for the new `poll_result` embed type. With the current implementation of `EmbedField`, these new fields should be covered as well right? Or is there something I'm missing and we can add in the new information provided?

PR #2956, opened 2024-08-25
Add support for new `POLL_RESULT` and `PURCHASE_NOTIFICATION` message types

ivinjabraham:
Adds the new message types as requested in #2948. Awaiting instruction on how to best proceed with implementing a new Embed kind and it's fields.


jamesbt365:
> Awaiting instruction on how to best proceed with implementing a new Embed kind and it's fields.

You don't unless anything has been added to the embed fields on the discord api docs, kind is a string sent by discord and fields should just work regardless. You just need to add the message type

GnomedDev (approved):
Only adds `PURCHASE_NOTIFICATION`, but good to merge otherwise.

Issue #2884, opened 2024-05-30
Messages should be truncated from cache when `Settings::max_messages` is lower than current cached messages 


jamesbt365:
I'm not sure if we had this at some point but we certainly don't now, if max messages is reduced at runtime any amount set over the limit is kept.

ivinjabraham:
I'll take this up.

I believe that this is the only method to change `max_messages` at runtime after creation:
```rust
    /// Sets the maximum amount of messages per channel to cache.
    ///
    /// By default, no messages will be cached.
    pub fn set_max_messages(&self, max: usize) {
        self.settings.write().max_messages = max;
    }
```

I'll make it check if the argument `max` is less than `max_messages` by reading the settings since checking it after assigning the new value should take more time. And if it is greater, delete the messages from `Cache.messages` that correspond to the front of `Cache.message_queue` for each channel's queue. Does that sound good?

PR #2957, opened 2024-08-26
Truncate cache if `max_messages` is set to a lower value at runtime

ivinjabraham:
Just a small commit that removes excess messages from cache if `Settings::max_messages` is changed at runtime through `set_max_messages`. Lot of iteration, but I couldn't think of a faster way to do it.

Address #2884.

### DIFF
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -439,6 +439,18 @@ impl Cache {
     ///
     /// By default, no messages will be cached.
     pub fn set_max_messages(&self, max: usize) {
+        // Check to see if cache has to be truncated
+        if max < self.settings.read().max_messages {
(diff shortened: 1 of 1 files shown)


GnomedDev:
Can this be retargetted and reimplemented on `next`? Although this is a non-breaking change, the layout of the MessageCache was simplified on next and isn't so fragile.

ivinjabraham:
Re-implemented on `next` as requested. There were _a few_ unintended commits from `current` pushed, which caused the bot to add the new labels.

GnomedDev (approved):
LGTM, but if you can just change to using `VecDeque::drain` instead of that manual `pop_front` loop that would be lovely.

arqunis (approved):

Issue #2635, opened 2023-11-29
Attachment names string treatment is not consistent between different functions


asibahi:
Serenity version: I am using `poise 0.6.0`

Rust version (`rustc -V`): 1.74

Minimal test case if possible:

```rust
    // inside a command function
    let attachment_name = format!("{}'s_{}.png", "some", "one"); // mind the apostrophe

    let attachment = serenity::CreateAttachment::bytes(image_data, &attachment_name); // 👈 first use here

    let embed = serenity::CreateEmbed::new()
         // other things here
        .attachment(attachment_name); // 👈 second use here

    let reply = poise::CreateReply::default()
        .attachment(attachment)
        .embed(embed);
}
```

the apostrophe in `attachment_name` causes the two functions to give different output somehow. The attachment is no longer *in* the embed.

<img width="598" alt="attachment out of the embed" src="https://github.com/serenity-rs/serenity/assets/17417266/e373d64b-2197-4b46-814a-a8b26d693eb3">

Removing the apostrophe resolved that problem.

<img width="471" alt="attachment inside the embed" src="https://github.com/serenity-rs/serenity/assets/17417266/348624be-ec60-4b16-8908-c07f9988a4ed">

ivinjabraham:
I'll try looking into this

ivinjabraham:
This is caused by a URL sanitation issue. The `url` field in `CreateEmbed(Embed)` shouldn't contain invalid characters. Perhaps `serde` replaces the invalid characters, but somehow that messes up where the attachment should go.

This is a temporary fix:
```rust
    pub fn attachment(self, filename: impl Into<String>) -> Self {
        let mut filename = filename.into();
        filename.retain(|c| c != '\'');
        filename.insert_str(0, "attachment://");
        self.image(filename)
    }
```

Will try and check if there are better ways to ensure `url` is valid and also why it causes the attachment to go out of the embed.

UPDATE: Found that when there are invalid characters, the `Message` struct has attachments, and the embed does not have an image. As expected, when it does not have invalid characters, there are no attachments and the embed as an `EmbedImage`.

asibahi:
Thanks for looking into this. I know of the work around (just not add invalid characters .. duh!). Yes what you're describing is the same behavior I found.

jamesbt365:
Discord converts the names when they include characters it doesn't like, which means that when supplying the attachment name to the embed, it doesn't point to the real attachment anymore.

[...]

Issue #2776, opened 2024-02-25
Embed Footer, Thumbnail & Image can be local files, not only HTTP(S)


keiveulbugs:
In the documentation it states that the Footer, Thumbnail & Image only support HTTP(S).

https://docs.rs/serenity/latest/serenity/builder/struct.CreateEmbed.html#method.image
https://docs.rs/serenity/latest/serenity/builder/struct.CreateEmbed.html#method.thumbnail
https://docs.rs/serenity/latest/serenity/builder/struct.CreateEmbedFooter.html#method.icon_url

This is a bit misleading as you can use local files when attached to the message. I tried fixing the docs, but have not a setup ready at the moment to test if the docs are working and if it works in pure Serenity. (In poise it works, but uses the `CreateReply` function.)

Would love if someone could check this and add something to the docs about how to use local files.

```rust
use serenity::builder::{CreateEmbed, CreateAttachment, CreateEmbedFooter};
// First we create an attachment
let attachment = CreateAttachment::path("images/loss.jpg").await.expect("Could not find image");

// Then we create an embed where we add the image/footer/thumbnail
let embed = CreateEmbed::default()
     .image("attachment://loss.jpg")
     .thumbnail("attachment://loss.jpg")
     .footer(
            CreateEmbedFooter::new("Loss footer")
                .icon_url("attachment://loss.jpg"),
     );

// I think this should work in Serenity:
 let builder = CreateMessage::new().embed(embed).add_file(attachment);

// In Poise we do the following and that works for sure:
 ctx.send(CreateReply::default().embed(embed).attachment(attachment))
       .await?;
  ```

**Comment by floris-xlx** [comment]
You can create an attachment an attach it as such

```rust
let img_path: &Path = Path::new("./img.png");
let attachment_bytes: Vec<u8> = std::fs::read(img_path).expect("Failed to read card file");
let attachment: CreateAttachment = CreateAttachment::bytes(
    attachment_bytes,
    "img.png"
);
// attachment can be routed thru an embed aswell

CreateMessage::default()
    .add_file(attachment)
    .content("@everyone")
```

ivinjabraham:
This seems correct, not sure why the docs say it only supports HTTPS for those methods. `CreateEmbed::attachment` even specifies `CreateEmbed::image` works with local attachments by using the `attachment://` syntax. So do the [Discord docs](https://discord.com/developers/docs/resources/message#embed-object-embed-structure).

ivinjabraham:
[...]

PR #2959, opened 2024-08-27
Document rule for attachment names for proper functioning in embeds

ivinjabraham:
Address #2635 by documenting [Discord's rules](https://discord.com/developers/docs/reference#uploading-files) for attachment URLS in the `attachment` and `image` methods of `CreateEmbed`.

It is worth noting that these rules apply to all attachment URLs. I chose to document only in `CreateEmbed` as this is, as far as I know, the only case where unexpected behaviour occurs when the name contains invalid characters. In other cases, Discord handles it by stripping or replacing the invalid characters with underscores and no error is thrown.

### DIFF
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -101,6 +101,9 @@ impl CreateEmbed {
     }
 
     /// Set the image associated with the embed. This only supports HTTP(S).
+    ///
+    /// Refer [Discord Documentation](https://discord.com/developers/docs/reference#uploading-files)
+    /// for rules on naming local attachments.
     #[inline]
     pub fn image(mut self, url: impl Into<String>) -> Self {
         self.0.image = Some(EmbedImage {
@@ -162,6 +165,8 @@ impl CreateEmbed {
     ///
     /// Note however, you have to be sure you set an attachment (with [`ChannelId::send_files`])
     /// with the provided filename. Or else this won't work.
+    ///
+    /// Refer [`Self::image`] for rules on naming local attachments.
     ///
     /// [`ChannelId::send_files`]: crate::model::id::ChannelId::send_files
     #[inline]
@@ -300,6 +305,8 @@ impl CreateEmbedFooter {
     }
 
     /// Set the icon URL's value. This only supports HTTP(S).
+    ///
+    /// Refer [`CreateEmbed::image`] for rules on naming local attachments.
     pub fn icon_url(mut self, icon_url: impl Into<String>) -> Self {
         self.0.icon_url = Some(icon_url.into());
         self
--- a/src/builder/create_embed.rs
+++ b/src/builder/create_embed.rs
@@ -100,7 +100,7 @@ impl CreateEmbed {
         self
     }
 
-    /// Set the image associated with the embed. This only supports HTTP(S).
+    /// Set the image associated with the embed.
     ///
     /// Refer [Discord Documentation](https://discord.com/developers/docs/reference#uploading-files)
     /// for rules on naming local attachments.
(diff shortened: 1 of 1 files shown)


src/builder/create_embed.rs
```diff
@@ -101,6 +101,11 @@ impl CreateEmbed {
     }
 
     /// Set the image associated with the embed. This only supports HTTP(S).
+    ///
+    /// The URL must be ASCII alphanumeric characters ('a-z', 'A-Z', '0-9').
+    /// Underscores ('\_'), dashes ('-') and dots ('.') are allowed but they must not
+    /// appear as the first character. Otherwise, the attachment will be sent separately from the
+    /// Embed.
```
jamesbt365: This is a bit of a misfit with the line above saying it only supports HTTP(S), either adjust that line or adjust what you have written.

jamesbt365 (changes requested):

GnomedDev:
I feel like this would be better served by a PR to the discord developer docs, then simply linking there, as we cannot guarantee the rules keep up to date.

ivinjabraham:
Discord does mention the rules for [naming here](https://discord.com/developers/docs/reference#uploading-files). Though you have to dig a little to see it. Should I just link this?

As for the changes requested, I was awaiting verdict on #2776 before removing the misleading HTTP(S) comment. Will remove them now.

GnomedDev:
Yeah, please do link to it, if there is Discord documentation on something we are moving to just link to that to prevent our docs rotting so much.

ivinjabraham:
Updated as requested

src/builder/create_embed.rs
```diff
@@ -100,7 +100,10 @@ impl CreateEmbed {
         self
     }
 
-    /// Set the image associated with the embed. This only supports HTTP(S).
+    /// Set the image associated with the embed.
+    ///
+    /// Refer [Discord Documentation](https://discord.com/developers/docs/reference#uploading-files)
+    /// for allowed characters in `url`.
```
jamesbt365: The url isn't the problem, is it not? Its Discord filtering the characters allowed in filenames on their cdn.

src/builder/create_embed.rs
```diff
@@ -299,7 +305,7 @@ impl CreateEmbedFooter {
         self
     }
 
-    /// Set the icon URL's value. This only supports HTTP(S).
+    /// Set the icon URL's value.
```
jamesbt365, line 307: I'm almost certainly sure you can use local files here too, so showing the rules would be ideal, to prevent duplicated docs you can just tell users to refer to `Self::image` for the rules surrounding local attachments.

src/builder/create_embed.rs
ivinjabraham: Yeah, I mentioned `url` because I thought it would make sense to the user when calling the function. Changed it to just rules for naming local attachments as you said.

src/builder/create_embed.rs
ivinjabraham: Done :+1:

ivinjabraham:
Rebased to make sure the edits are in the right commits.

ivinjabraham:
On merging, please don't forget to close #2635 as well as #2776.

GnomedDev:
If you want to make that automatic, put "closes/fixes #xxx" in the PR description. See the docs for that [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

ivinjabraham:
> If you want to make that automatic, put "closes/fixes #xxx" in the PR description. See the docs for that [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue).

Thanks for the tip!

jamesbt365 (approved):

PR #2964, opened 2024-09-01
Add enum for JSON error codes in DiscordJsonError

### LINKED ISSUE #2752: feat(http): Http Error Codes as enum?
I searched through the issue tracker for issues linked to the error code field on the ``serenity::http::error::DiscordJsonError`` and couldn't find any, but this feels like an issue that should've come up, so I apologize if it has done so.

The error field feels it can be improved by either being replaced or supplemented by an enum that covers all currently existing error-codes, with (depending on implementation) either a variant or an Option for if the error-code is unknown.

A list from discords official docs is here:

https://github.com/discord/discord-api-docs/blob/main/docs/topics/Opcodes_and_Status_Codes.md

Would this be wanted? I could potentially implement it myself, with instructions of how the maintainers prefer such implementations.

ivinjabraham:
fixes #2752.


arqunis:
This changes the type of the `DiscordJsonError::code` field, which is a breaking change. Could you rebase on top of the `next` branch instead?

ivinjabraham:
> This changes the type of the `DiscordJsonError::code` field, which is a breaking change. Could you rebase on top of the `next` branch instead?

Yep, rebased.

GnomedDev (commented):
Looks good, but I'm not going to check every error code so I'm not merging.

arqunis (approved):

PR #2966, opened 2024-09-01
Use u64 serialization of Snowflake for non human-readable serializers

majkrzak:
Snowflake deserialization uses [`serde::de::Deserializer deserialize_any`](https://docs.rs/serde/latest/serde/de/trait.Deserializer.html#tymethod.deserialize_any) method, which may not be provided by some formats and according the documentation should not be relied on. This for example causes `UserID` fail to serialize with `bincode`, effectively preventing storing it in `native_db`.

Currently `serde` seems not to support easy way to determinate if `deserialize_any` is supported or not. Although [`is_human_readable`](https://docs.rs/serde/latest/serde/de/trait.Deserializer.html#method.is_human_readable) is provided for both `Serializer` and `Deserializer` traits.

This PR bypasses requirement of having `deserialize_any` available, by serializing and deserializing Snowflake values as `u64` for non human-readable formats. This additionally optimizes its storage size for such formats.

### DIFF
--- a/src/model/id.rs
+++ b/src/model/id.rs
@@ -343,12 +343,20 @@ mod snowflake {
     use serde::{Deserializer, Serializer};
 
     pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<NonZeroU64, D::Error> {
-        deserializer.deserialize_any(SnowflakeVisitor)
+        if deserializer.is_human_readable() {
+            deserializer.deserialize_any(SnowflakeVisitor)
+        } else {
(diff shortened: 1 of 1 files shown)


GnomedDev (commented):
Please can you add a shorter version of this explanation as comments to the code, and rebase to make it mergable. Adding tests to make sure this works would also be nice, but isn't necessary.

GnomedDev:
Closing due to a lack of activity, if you would like to continue this PR, just let me know and I'll re-open.

PR #2968, opened 2024-09-11
Allowed embeds to be optional in followup messages

TapGhoul:
This pairs with https://github.com/serenity-rs/poise/pull/301 and is in the vein of #2216

This marks the Embeds field as optional in `CreateInteractionResponseFollowup`, which if unset in an edit message context, prevents embeds from being removed. This is a behavioral change, but not an API-incompatible one.


TapGhoul:
Rebased onto next as per request of the Serenity discord.

PR #2982, opened 2024-10-03
Add support for voice status

rhgndf:
Haven't seen any docs to change voice status in this crate.

API here: https://github.com/discord/discord-api-docs/pull/6400

This patch makes changing voice status possible. Not familiar with the code base, so I'm not sure if this is the best way to do it.

### DIFF
--- a/src/builder/edit_channel.rs
+++ b/src/builder/edit_channel.rs
@@ -5,7 +5,6 @@ use super::CreateForumTag;
 use crate::http::CacheHttp;
 #[cfg(feature = "http")]
 use crate::internal::prelude::*;
-use crate::json::*;
 use crate::model::prelude::*;
 
 /// A builder to edit a [`GuildChannel`] for use via [`GuildChannel::edit`].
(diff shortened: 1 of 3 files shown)


src/builder/edit_channel.rs
```diff
@@ -324,6 +337,13 @@ impl<'a> Builder for EditChannel<'a> {
             }
         }
 
+        if let Some(ref status) = self.status {
+            cache_http
+                .http()
+                .edit_voice_status(ctx, &json!({ "status": status }), self.audit_log_reason)
+                .await?;
+        }
```
arqunis:
```suggestion
        if let Some(status) = &self.status {
            #[derive(Serialize)]
            struct EditVoiceStatusBody<'a> {
                status: &'a str,
            }

            cache_http
                .http()
                .edit_voice_status(ctx, &EditVoiceStatusBody { status: status.as_str() }, self.audit_log_reason)
                .await?;
        }
```

arqunis (changes requested):
Looks good, but could you change the use of the `json!` macro to an intermediary struct? We like to avoid temporary heap allocations as much as possible.

arqunis (approved):

PR #2986, opened 2024-10-07
Remove more json! usage

GnomedDev:


src/builder/create_components.rs
```diff
@@ -17,19 +16,20 @@ pub enum CreateActionRow<'a> {
...
-                Self::Buttons(x) => serde_json::to_value(x).map_err(S::Error::custom)?,
-                Self::SelectMenu(x) => serde_json::to_value(vec![x]).map_err(S::Error::custom)?,
-                Self::InputText(x) => serde_json::to_value(vec![x]).map_err(S::Error::custom)?,
-            }
-        })
-        .serialize(serializer)
+        use serde::ser::SerializeMap as _;
+
+        let mut map = serializer.serialize_map(Some(2))?;
+        map.serialize_entry("kind", &1_u8)?;
```
arqunis: Shouldn't this be `type` instead of `kind`?

GnomedDev: Indeed, fixed.

src/builder/create_components.rs
```diff
@@ -17,19 +16,20 @@ pub enum CreateActionRow<'a> {
...
-        .serialize(serializer)
+        use serde::ser::SerializeMap as _;
+
+        let mut map = serializer.serialize_map(Some(2))?;
+        map.serialize_entry("type", &1_u8)?;
+
+        match self {
+            CreateActionRow::Buttons(buttons) => map.serialize_entry("components", buttons)?,
+            CreateActionRow::SelectMenu(select) => map.serialize_entry("components", select)?,
+            CreateActionRow::InputText(input) => map.serialize_entry("components", input)?,
```
arqunis: These two should serialise as an array.
```suggestion
            CreateActionRow::SelectMenu(select) => map.serialize_entry("components", [select])?,
            CreateActionRow::InputText(input) => map.serialize_entry("components", [input])?,
```

GnomedDev: Fixed

arqunis (approved):

PR #2996, opened 2024-10-18
Add `MessageReferenceKind` enum

### LINKED ISSUE #2994: Update `MessageReference` Struct
the [`MessageReference`](https://github.com/serenity-rs/serenity/blob/current/src/model/channel/message.rs#L1095) struct is missing the type field (see here: https://discord.com/developers/docs/resources/message#message-reference-object)

I am happy to implement this myself, the question is do you want this added as a struct field, or make a union enum from this?

MarkusTheOrt:
Fixes #2994 

also adds additional `From<>` implementation for `(MessageReferenceKind, ChannelId, MessageId)`


GnomedDev:
I don't think we should continue adding `From`s for this type as it will just get silly if there are more message reference types added. Can you remove the one you added in this PR and add `#[deprecated]` to the existing one?

MarkusTheOrt:
also right now the `kind` is an option, however the discord API docs say that this is a required field in the future. So maybe just make it a Non-Optional type now?

GnomedDev:
Sure, chuck a serde default on there as-well to make sure old `MessageReference`s without the field set still deserialize.

src/model/channel/message.rs
```diff
@@ -1115,9 +1152,12 @@ impl From<&Message> for MessageReference {
     }
 }
 
+#[allow(useless_deprecated)]
```
GnomedDev: Remove the deprecated if it is useless. Maybe try applying it to `from`?

If nothing works here, don't worry about it, just put `// TODO(next): Remove this`.

GnomedDev (changes requested):

MarkusTheOrt:
20241018_16h51m35s_grim
I can confirm its working

MarkusTheOrt:
To use the forward message feature the bot needs to have at least the "Read Message History" permissions, ~~(I don't know about reply but I believe that too)~~, should I put a comment there to clarify things?

MarkusTheOrt:
so while this code is all fine, I think its better to go with a builder and use a `CreateMessageReference` struct instead of the `MessageReference` struct, that way one can set all the fields they want in the way it works with everything else.

GnomedDev:
We are moving away from documenting stuff like that as it may change in future. If it's documented on ddev docs, it wouldn't be the end of the world to duplicate here, but if not we shouldn't.

MarkusTheOrt:
this looks a lot more reasonable to me, let me know what you think

src/builder/create_message.rs
```diff
@@ -209,7 +210,7 @@ impl CreateMessage {
     }
 
     /// Set the reference message this message is a reply to.
-    pub fn reference_message(mut self, reference: impl Into<MessageReference>) -> Self {
+    pub fn reference_message(mut self, reference: impl Into<CreateMessageReference>) -> Self {
```
GnomedDev: This is a breaking change, and either requires targeting to the `next` branch or reverting. I would prefer reverting, as we should have support for new discord features on `current`.

src/builder/create_message.rs
```diff
@@ -209,7 +210,7 @@ impl CreateMessage {
     }
 
     /// Set the reference message this message is a reply to.
```
GnomedDev:
```suggestion
    /// Set the message this reply or forward is referring to.
```

GnomedDev (changes requested):

src/builder/create_message.rs
MarkusTheOrt: sure, I'll open a PR for `next` once this is done so we don't forget :-)

jamesbt365: Dont open multiple PRs, the 2 branches are rebased on each other to keep feature parity so whoever rebases it will remove it. Opening multiple PRs will just diverge the branches and make rebasing in the future a pain.

src/builder/create_message_reference.rs
GnomedDev, line 1: Isn't this completely unused now? This shouldn't be added for `current`, you can make a separate PR after this is merged and `next` is rebased to add this builder and swap `CreateMessage::reference_message` to use this.

MarkusTheOrt: I can remove this if you want, but I made an from impl to convert a CreateMessageReference into a MessageReference, so not 100% unused but I get your point.

GnomedDev: Yep, please remove it for now.

GnomedDev (approved):

PR #2997, opened 2024-10-20
Replace outdated secrecy library with own impl

GnomedDev:
`secrecy` has had major changes in it's transition from `0.8` -> `0.10` that would make it even less useful than it is now.

This just brings the basic implementation in-house and exposes it publicly, which is somehow less lines of code that trying to use the `0.8` API, let alone `0.10`.

### DIFF
--- a/src/secret_string.rs
+++ b/src/secret_string.rs
@@ -1,0 +1,41 @@
+use std::sync::Arc;
+
+/// A cheaply clonable, zeroed on drop, String.
+///
+/// This is a simple newtype of `Arc<str>` that uses [`zeroize::Zeroize`] on on last drop to avoid
+/// keeping it around in memory.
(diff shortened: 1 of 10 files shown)


jamesbt365 (approved):

src/secret_string.rs
```diff
@@ -0,0 +1,41 @@
+use std::sync::Arc;
+
+/// A cheaply clonable, zeroed on drop, String.
+///
+/// This is a simple newtype of `Arc<str>` that uses [`zeroize::Zeroize`] on on last drop to avoid
```
jamesbt365:
```suggestion
/// This is a simple newtype of `Arc<str>` that uses [`zeroize::Zeroize`] on last drop to avoid
```

PR #3001, opened 2024-10-23
Add partial_member_permissions on Guild and PartialGuild

jamesbt365:
buh

### DIFF
--- a/src/model/guild/mod.rs
+++ b/src/model/guild/mod.rs
@@ -1918,6 +1918,24 @@ impl Guild {
         )
     }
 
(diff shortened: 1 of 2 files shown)


jamesbt365:
Should I mimic the panic if the UserId is wrong (if its Some) like what currently happens on `partial_member_permissions_in`?

arqunis:
> Should I mimic the panic if the UserId is wrong (if its Some) like what currently happens on partial_member_permissions_in?

Yes, as that would help make the two functions consistent since their roles are similar, but only slightly different (one calculates based on roles only, whereas the other takes into consideration channel overrides).

arqunis (approved):

PR #3005, opened 2024-10-28
Make fields of BulkBanResponse public

jamesbt365:
How are we supposed to use them if they're private?


GnomedDev:
Good point! Add non_exhaustive and it will be good.

GnomedDev (approved):

PR #3009, opened 2024-10-31
Make Full Event serializable

cheesycod:


cheesycod:
Remaking due to git issues

PR #3010, opened 2024-10-31
feat: Make FullEvent serializeable

cheesycod:


jamesbt365:
...Whats the usecase here? Seems like you're trying to do something the library isn't designed for and you would probably have a better experience using twilight instead of creating a hacky solution with serenity.

GnomedDev:
I've talked in DMs about this, it's for serializing to an interpreter and removes 3k LOC with the only downside being slightly longer fresh compile times. I'm good with this.

GnomedDev:
I'll enable auto merge, but you can have the final say on approving. @jamesbt365

src/model/id.rs
```diff
@@ -339,7 +339,7 @@ id_u64! {
 /// This identifier is special, it simply models internal IDs for type safety,
 /// and therefore cannot be [`Serialize`]d or [`Deserialize`]d.
```
jamesbt365: If you're gonna do it, this comment is outdated?

GnomedDev: Yep, this comment should be `cannot be deserialized.`

cheesycod: Fixed

jamesbt365 (approved):

PR #3013, opened 2024-11-03
Add polls to interaction responses

jamesbt365:
This was not added when poll support was added to serenity. I have not tested this, but it is modeled after existing support for `CreateMessage`.


jamesbt365:
CI failure is unrelated to the changes made in this PR.

GnomedDev (approved):
Tested locally, seems to work fine.

PR #3015, opened 2024-11-04
Fix before for archived public and private threads having incorrect before

### LINKED ISSUE #3014: get_archived_public_threads and get_archived_private_threads use a u64 before instead of a timestamp
According to https://github.com/discord/discord-api-docs/blob/main/docs/topics/Threads.md#DOCS_RESOURCES_CHANNEL/list-public-archived-threads, getting archived threads, both public and private, are sorted by archived timestamp. The code instead takes an ID of some form, but the API expects an ISO timestamp. This can easily be fixed by simply switching before from an Option<u64> to an Option<Timestamp>. Tested locally, and confirmed that this fix works.

It would also be nice if a version of the messages iterator was provided for public/private archived threads.

BigBadE:
Fixes #3014

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2987,7 +2987,7 @@ impl Http {
     pub async fn get_channel_archived_public_threads(
         &self,
         channel_id: ChannelId,
-        before: Option<u64>,
+        before: Option<Timestamp>,
         limit: Option<u64>,
     ) -> Result<ThreadsData> {
         let mut params = vec![];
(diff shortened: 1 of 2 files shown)


jamesbt365:
This is a breaking change, please rebase to `next`.

BigBadE:
Already fixed on next, I just didn't see it:
https://github.com/Serenity-rs/serenity/blob/next/src/http/client.rs#L2665
Found the fix in
https://github.com/Serenity-rs/serenity/commit/ec346762ad3ccbeb02252f7dfcbdd5f552f3ee45
guess I just didn't see the original issue because I was searching for the specific method name.

PR #3020, opened 2024-11-07
Add Consume Entitlement endpoint and missing fields for SKUs and Entitlements

mkrasnitski:
I have not tested this endpoint so I'm not completely sure of the failure condition, but I think it's a reasonable assumption.


mkrasnitski:
This is ready for merge IMO. CI failures are unrelated.

arqunis (approved):

PR #3021, opened 2024-11-08
Add support for Entry Point commands

mkrasnitski:
The discord docs are a little confusing for this feature, but I think it's basically pretty simple. Applications with activities enabled can define a global command of type `PrimaryEntryPoint` - based on the value of the `handler` field, this command can be treated essentially like normal, allowing the application to respond with a followup message when called; or, the behavior of the command can be automatically handled by Discord.

### DIFF
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -64,6 +64,13 @@ pub enum CreateInteractionResponse {
     /// Corresponds to Discord's `PREMIUM_REQUIRED'.
     #[deprecated = "use premium button components via `CreateButton::new_premium` instead"]
     PremiumRequired,
+    /// Not valid for autocomplete and Ping interactions. Only available for applications with
+    /// Activities enabled.
+    ///
+    /// Responds to the interaction by launching the Activity associated with the app.
+    ///
+    /// Corresponds to Discord's `LAUNCH_ACTIVITY`.
+    LaunchActivity,
 }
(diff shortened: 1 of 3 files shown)


src/builder/create_interaction_response.rs
```diff
@@ -64,6 +64,13 @@ pub enum CreateInteractionResponse {
     /// Corresponds to Discord's `PREMIUM_REQUIRED'.
     #[deprecated = "use premium button components via `CreateButton::new_premium` instead"]
     PremiumRequired,
+    /// Not valid for autocomplete and Ping interactions. Only available for applications with
+    /// Activities enabled.
+    ///
+    /// Responds to the interaction by launching the Activity associated with the app.
```
GnomedDev, line 70: Can these lines be flipped? Seems to be backwards.

GnomedDev (commented):
Small nitpick, but otherwise good to merge.

src/builder/create_interaction_response.rs
mkrasnitski: This mirrors the docs for the other variants, like `Acknowedge`, `UpdateMessage`, `Autocomplete`, and `Modal`. If this should be changed, I think the others should too.

GnomedDev: Ah, no worries then.

GnomedDev (approved):

PR #3026, opened 2024-11-12
Split GuildThread from GuildChannel

### LINKED ISSUE #2991: Thread channels don't appear to be cached
When running a starboard bot with pretty much a default configuration, the bot is unable to find thread channels in its cache causing events to be discarded.

The bot is present in the thread's member list and can receive events for the channel.
Can provide more information on request but at present, not sure what else to mention.

GnomedDev:
This fixes #2991, via the solution voted for on the serenity discord.


GnomedDev:
This is ready for review, the main changes are:
- `GuildChannel` is now:
- - `BaseGuildChannel` (shared fields)
- - `GuildChannel` (channel-exclusive)
- - `GuildThread` (thread-exclusive)
- - `GenericGuildChannelRef` (an enum over references to both)

- `PartialChannel` is now:
- - `BaseInteractionChannel` (shared fields)
- - `InteractionChannel` (channel-exclusive)
- - `InteractionThread` (thread-exclusive)
- - `GenericInteractionChannel` (an enum over the both)

- `ChannelId` is now:
- - `GenericChannelId` (shared purpose/methods) 
- - `ChannelId` (channel-exclusive)
- - `ThreadId` (thread-exclusive)

The IDs have methods to convert between, such as:
-  `GenericChannelId::expect_channel` and `GenericChannelId::expect_thread` which casts the ID types without touching the value
- `GenericChannelId::split` which calls both and returns both for when something needs to handle both cases
- `ThreadId::widen` and `Channelid::widen` which cast to `GenericChannelId` for when you have a specific ID and need to "widen" the number of methods you can call.

`GenericGuildChannelRef` is returned from the new `Guild::channel` method, which looks the `GenericChannelId` up in both `Guild::channels` and `Guild::threads`.

jamesbt365 (approved):
This looks really good, not 100% sure if I like the name changes for PartialChannel, but I can't think of anything better right now so I think its passable.

peanutbother:
I also think the naming could be improved since we are explicitly talking about channels and threads.
My proposal in discord was the following:

> How about
> ```rust
> enum MessageOrigin {
>     Channel(ChannelId),
>     Thread(ThreadId)
> }
> ```
> seems a better naming to me since I see Threads as their own thing so I would not name a message's origin a _channel_

In that naming scheme I would name `Base*` rather `Origin*` and in some contexts append or prepend `Message*` or `Interaction*` accordingly.

GnomedDev:
@peanutbother  I don't fully understand how this "origin" naming improves things? From how I processed that I think you are saying `OriginGuildChannel` which doesn't make too much sense, and we already have prepended `Interaction` where possible (as the other types are truly generically used everywhere).

peanutbother:
I mean something like `MessageOrigin` and `InteractionOrigin` as base names instead of the chosen base names as for e.g. threads and channels the base name is still channel.

PR #2988, opened 2024-10-08
Fix clippy warnings

GnomedDev:
I swapped the CI to nightly for clippy, as otherwise I couldn't ignore newly-added lints.

PR #3027, opened 2024-11-12
Improve some docs

mkrasnitski:
Improved wording and removed some sentences that are no longer relevant. These docs are mostly the ones that were touched by #2988.

### DIFF
--- a/src/gateway/sharding/mod.rs
+++ b/src/gateway/sharding/mod.rs
@@ -1,4 +1,4 @@
-//! Module containing types for gateway sharding.
+//! Mechanisms for configuring and managing sharded gateway connections.
 //!
 //! Sharding is a method for load-balancing bots across separate threads or processes. Sharding is
 //! enforced on bots by Discord once they reach a certain number of guilds (2500). Once this
@@ -62,16 +62,16 @@ use crate::model::gateway::{GatewayIntents, ShardInfo};
 use crate::model::id::{ApplicationId, GuildId, ShardId};
 use crate::model::user::OnlineStatus;
 
-/// A Shard is a higher-level handler for a websocket connection to Discord's gateway.
+/// An abstract handler for a websocket connection to Discord's gateway.
 ///
-/// The shard allows for sending and receiving messages over the websocket,
-/// such as setting the active activity, reconnecting, syncing guilds, and more.
+/// Allows a user to send and receive messages over said websocket, including:
+///   * setting the current activity
+///   * setting the current online status
+///   * receiving gateway events
+///   * connection management via heartbeating
 ///
-/// Refer to the [module-level documentation][module docs] for information on effectively using
-/// multiple shards, if you need to.
-///
-/// Note that there are additional methods available if you are manually managing a shard yourself,
-/// although they are hidden from the documentation since there are few use cases for doing so.
+/// Shard management (including starting, restarting, heartbeating), is performed by the [`Client`]
+/// automatically on the user's behalf.
 ///
 /// # Stand-alone shards
 ///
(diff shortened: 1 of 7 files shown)


src/model/guild/member.rs
```diff
@@ -559,8 +559,10 @@ pub struct PartialThreadMember {
 
 /// A model representing a user in a Guild Thread.
 ///
-/// [Discord docs](https://discord.com/developers/docs/resources/channel#thread-member-object),
-/// [extra fields](https://discord.com/developers/docs/topics/gateway-events#thread-member-update-thread-member-update-event-extra-fields).
+/// [Discord docs], [extra fields].
+///
+/// [Discord docs]: https://discord.com/developers/docs/resources/channel#thread-member-object,
+/// [extra fields]: https://discord.com/developers/docs/topics/gateway-events#thread-member-update-thread-member-update-event-extra-fields
```
GnomedDev, line 565: Is this not the same but longer?

src/model/guild/role.rs
```diff
@@ -10,10 +10,10 @@ use crate::model::utils::is_false;
...
 /// Information about a role within a guild.
 ///
-/// A role represents a set of permissions, and can be attached to one or multiple users. A role has
-/// various miscellaneous configurations, such as being assigned a colour. Roles are unique per
-/// guild and do not cross over to other guilds in any way, and can have channel-specific permission
-/// overrides in addition to guild-level permissions.
+/// A role represents a set of permissions, and can be attached to one or multiple users. A role
+/// has various miscellaneous configurations, such as being assigned a colour. Roles are unique per
+/// guild and do not cross over to other guilds in any way, and can have channel-specific
+/// permission overrides in addition to guild-level permissions.
```
GnomedDev, line 16: Doesn't rustfmt deal with this? This also doesn't make any difference to the userfacing docs, so I don't see a point?

src/gateway/sharding/mod.rs
```diff
@@ -62,34 +62,28 @@ use crate::model::gateway::{GatewayIntents, ShardInfo};
 use crate::model::id::{ApplicationId, GuildId, ShardId};
 use crate::model::user::OnlineStatus;
 
-/// A Shard is a higher-level handler for a websocket connection to Discord's gateway.
+/// An abstract handler for a websocket connection to Discord's gateway.
 ///
-/// The shard allows for sending and receiving messages over the websocket,
-/// such as setting the active activity, reconnecting, syncing guilds, and more.
+/// Allows a user to send and receive messages over said websocket, such as: setting the current
+/// activity, reconnecting, syncing guilds, and more.
```
GnomedDev: Could this be expanded to bullet points with links to the methods?

src/model/guild/member.rs
mkrasnitski: Kind of, but since the links are so long it's better not to have them inline as the text of the doc comment is less readable in code otherwise.

GnomedDev: Fair enough.

src/model/guild/role.rs
mkrasnitski: Yeah I guess you're right - the `/// ` at the start of the line isn't counted by rustfmt for the line length, so that's why vim still wraps the line. Should I undo this change?

src/gateway/sharding/mod.rs
mkrasnitski: Done.

src/model/guild/role.rs
GnomedDev: It's fine if your editor is wrapping it anyway, because it would just cause pain to keep it as-is for you.

GnomedDev (approved):

PR #3028, opened 2024-11-12
Add `EditCommand` builder

mkrasnitski:
This separates the the builders for creating versus editing a command, since it's not possible to change the `type` of a command (in serenity this is the `kind` field). Also, the `name` field is not required when editing a command.


src/builder/create_command.rs
```diff
@@ -320,56 +320,28 @@ impl<'a> CreateCommandOption<'a> {
...
     #[serde(rename = "type")]
     kind: Option<CommandType>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    integration_types: Option<Vec<InstallationContext>>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    contexts: Option<Vec<InteractionContext>>,
-    nsfw: bool,
+
+    #[serde(flatten)]
+    fields: EditCommand<'a>,
```
GnomedDev, line 336: I like this, a lot. We should do this more.

GnomedDev (approved):

PR #3030, opened 2024-11-12
Deprecate `GuildChannel::permissions_for_user` and add `Message::author_permissions`

GnomedDev:
This deprecates `GuildChannel::permissions_for_user` as:
- It is unnecessary API surface
- Performs a cache lookup when most users will have `Guild` if they have `GuildChannel`

It also adds `Message::author_permissions` to match `Interaction::permissions` and hopefully be a replacement for a lot of the usage of `GuildChannel::permissions_for_user`. This also takes advantage of the lesser known `Message::member` field to not require member intents for calculating the permissions from a message create event.

For `Message::author_permissions`, this also adds `Permissions::dm_permissions`, which returns the permissions that are implicitly enabled in a DM.

### DIFF
--- a/examples/testing/src/main.rs
+++ b/examples/testing/src/main.rs
@@ -196,12 +196,7 @@ async fn message(ctx: &Context, msg: Message) -> Result<(), serenity::Error> {
         channel.edit(ctx, EditChannel::new().category(None)).await?;
         channel.edit(ctx, EditChannel::new().category(Some(parent_id))).await?;
     } else if msg.content == "channelperms" {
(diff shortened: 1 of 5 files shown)


jamesbt365 (approved):
LGTM. Some stuff in poise uses `Permissions::all()` for DMs instead of setting a list of permissions like has been done here, I don't think theres really a "best" way of handling this but worth a mention I guess.

GnomedDev:
That poise stuff can be moved to using `Permissions::dm_permissions` then, great.

PR #3031, opened 2024-11-12
customizable wait time on shard queuer

cheesycod:


jamesbt365 (commented):
I don't think its clear enough that in normal cases you should not change the shard start time (as this will rate limit you).

cheesycod:
> I don't think its clear enough that in normal cases you should not change the shard start time (as this will rate limit you).

True, will try adding a warning to the builder function

src/gateway/client/mod.rs
```diff
@@ -153,6 +163,13 @@ impl ClientBuilder {
         self.framework.as_deref()
     }
 
+    /// Sets the time to wait between starting shards. This is useful when using proxies such
+    /// as twilight or sandwich
```
GnomedDev:
```suggestion
    /// Sets the time to wait between starting shards.
    ///
    /// This should only be used when using a gateway proxy, such as [Sandwich] or [Twilight Gateway Proxy], as otherwise this will lead to gateway disconnects if the shard start rate limit is not respected.
    ///
    /// [Sandwich]:  https://github.com/WelcomerTeam/Sandwich-Daemon
    /// [Twilight Gateway Proxy]: https://github.com/Gelbpunkt/gateway-proxy
```

GnomedDev (changes requested):

GnomedDev (approved):

PR #3035, opened 2024-11-13
Update `CurrentApplicationInfo` and add support for editing it

### LINKED ISSUE #2977: Missing endpoint: Edit Current Application
Discord docs: https://discord.com/developers/docs/resources/application#edit-current-application

ivinjabraham:
Fixes #2977


jamesbt365 (changes requested):
Use the provided to_vec function instead of importing it from serde_json.

jamesbt365:
Also you haven't added anything but the low level http interface? Http usually shouldn't be directly used and you should probably add a builder like how our other stuff interacts with http.

ivinjabraham:
> Use the provided to_vec function instead of importing it from serde_json.

Pesky auto-import. Should be fixed now.

> Also you haven't added anything but the low level http interface? Http usually shouldn't be directly used and you should probably add a builder like how our other stuff interacts with http.

Right, so I'll make a `edit_current_application_info.rs` and impl Builder along with the required structs and methods?

ivinjabraham:
It seems the `CurrentApplicationInfo` struct is out of date. There are few fields such as `event_webhook_url` and `event_webhook_status` that are missing. I'll send in a PR for that first.

[Doc for reference](https://discord.com/developers/docs/resources/application#application-object)

jamesbt365 (changes requested):
edit_current_app**ll**ication_info.rs -> edit_current_application_info.rs

src/builder/edit_current_application_info.rs
```diff
@@ -0,0 +1,182 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    event_webhooks_url: Option<Cow<'a, str>>,
+
+    /// If webhook events are enabled for the app. 1 to disable, and 2 to enable.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    event_webhook_status: Option<EventWebhookStatus>,
+
+    /// List of Webhook event types to subscribe to.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    event_webhook_types: Option<Vec<EventWebhookType>>,
```
GnomedDev: These fields are private, so these docs are only visible to us working on serenity. Since these are just on the ddevs page anyway, can you remove the docs?

src/model/application/mod.rs
```diff
@@ -49,39 +49,52 @@ pub struct CurrentApplicationInfo {
...
+    pub install_params: Option<InstallParams>,
     #[serde(default)]
     pub integration_types_config: HashMap<InstallationContext, InstallationContextConfig>,
-    pub approximate_guild_count: Option<u32>,
-    pub approximate_user_install_count: Option<u32>,
-    pub guild: Option<PartialGuild>,
-    pub redirect_uris: Option<Vec<String>>,
-    pub interactions_endpoint_url: Option<String>,
+    #[serde(default)]
+    pub custom_install_url: String,
```
GnomedDev: You have replaced a bunch of `FixedString` here with `String`, I'm assuming it's just a leftover from when this PR targeted `current`, can you change it back to use `FixedString` (for performance reasons).

src/builder/edit_current_application_info.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    /// Default rich presence invite cover image for the app.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    cover_image: Option<ImageHash>,
+
+    /// Interactions endpoint URL for the app.
+    ///
+    /// To update an Interactions endpoint URL via the API, the URL must be valid according
+    /// to the [Receiving an Interaction]
+    /// (https://discord.com/developers/docs/interactions/receiving-and-responding#receiving-an-interaction) documentation.
```
mkrasnitski: These docs are private and therefore unnecessary. I'd recommend instead incorporating them into the docs for the methods themselves.

src/builder/edit_current_application_info.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    /// Sets the cover image hash.
+    pub fn cover_image(mut self, hash: &str) -> Self {
+        if let Ok(image_hash) = hash.parse() {
+            self.cover_image = Some(image_hash);
+        } else {
+            println!("Error parsing image hash while constructing EditCurrentApplicationInfo");
+        }
+        self
+    }
```
GnomedDev: 1. This shouldn't be taking one type, parsing it into another, then printlning on the user's behalf.
2. This shouldn't be taking ImageHash, as discord documents it as "image data", therefore it should be base64 string input.

src/builder/edit_current_application_info.rs
ivinjabraham: Ah, of course. Fixed it now.

src/model/application/mod.rs
```diff
@@ -262,13 +296,32 @@ bitflags! {
...
+        self.scopes = FixedArray::from_vec_trunc(scopes);
+        self
+    }
+
+    #[must_use]
+    pub fn permissions(mut self, permissions: Permissions) -> Self {
+        self.permissions = permissions;
+        self
+    }
+}
```
GnomedDev: Please make this an `CreateInstallParams` in the new module to avoid mixing memory-optimised (FixedArray) and compute-optimised (Vec) types. This would also allow for `new` to take both the required fields, `scopes` and `permissions`.

src/builder/edit_current_application_info.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    /// Default rich presence invite cover image for the app.
+    pub fn cover_image(mut self, hash: &str) -> Self {
+        if let Ok(image_hash) = hash.parse() {
+            self.cover_image = Some(image_hash);
+        } else {
+            println!("Error parsing image hash while constructing EditCurrentApplicationInfo");
+        }
+        self
+    }
```
GnomedDev: 1. Please don't take types just to parse them into different types.
2. Please don't print on the user's behalf.
3. This is meant to be base64 image data, not an image hash, as the docs say.

GnomedDev (changes requested):

src/model/application/mod.rs
ivinjabraham: Yep, they were leftovers. I've changed them back to `FixedString` now.

jamesbt365:
I think everything in `CurrentApplicationInfo` can use FixedString (String replacment) and FixedArray (Vec replacement)

src/model/application/mod.rs
ivinjabraham: Made the requested changes, I've also squashed the two commits since they're intertwined now.

src/builder/edit_current_application_info.rs
ivinjabraham: I've removed the unnecessary code and changed it to accept base64 data instead.

src/builder/edit_current_application_info.rs
```diff
@@ -0,0 +1,177 @@
...
+    pub fn icon(mut self, base64_image_data: String) -> Self {
+        self.icon = Some(base64_image_data);
+        self
+    }
+
+    /// Sets the default rich presence invite cover image using base64 image data.
+    pub fn cover_image(mut self, base64_image_data: String) -> Self {
+        self.cover_image = Some(base64_image_data);
+        self
+    }
```
arqunis, line 101: Rather than passing in base64 data manually as a `String`, we should allow the user to decide the place they want to source the image from and convert it to base64 on their behalf, similar to how [`EditProfile::avatar`](https://github.com/serenity-rs/serenity/blob/current/src/builder/edit_profile.rs#L49-L52) does it using `CreateAttachment`.

src/model/application/mod.rs
```diff
@@ -45,43 +45,56 @@ pub struct CurrentApplicationInfo {
...
     #[serde(default)]
+    pub guild: Option<PartialGuild>,
+    #[serde(default)]
     pub primary_sku_id: Option<SkuId>,
     #[serde(default)]
-    pub slug: Option<FixedString>,
+    pub slug: FixedString,
     #[serde(default)]
-    pub cover_image: Option<FixedString>,
+    pub cover_image: FixedString,
```
arqunis, line 70: Similar to the application icon, the API returns a hash of a constant size, so this should use [`ImageHash`](https://serenity-rs.github.io/serenity/next/serenity/model/misc/struct.ImageHash.html) to store the hash inline. Also should be wrapped in an `Option<...>`.

arqunis (changes requested):

GnomedDev:
This PR looks to be idle and therefore I am closing it to clean up the PR list, if anyone wants to pick up this and resolve the existing review comments here feel free to open a new PR.

PR #3036, opened 2024-11-14
Transport compression support

### LINKED ISSUE #693: Implement transport compression
Implement [`zlib-stream` transport compression](https://discordapp.com/developers/docs/topics/gateway#transport-compression) which compresses everything and therefore is more effective.

fgardt:
Reworks how the current payload compression is handled and adds support for transport compression.

- [x] implement `zlib-stream` decompression
- [x] implement `zstd-stream` decompression

Should close #693.
_And unlike #1508 it concats the WS query params with `&` instead of [`?`](https://github.com/serenity-rs/serenity/pull/1508/files?diff=split&w=0#diff-d1e3faa80a031e801cfb6cfac377a10139019f0133cf7fedb5cc33cc9c3b4d45R826-R831)._

#1508 had some logic to periodically resize/shrink buffers (even tho it seems like it was not hooked up to be called anywhere?) which this PR does not ~_yet_~ have.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -46,6 +46,9 @@ pub enum Error {
     /// An error from the `tungstenite` crate.
     #[cfg(feature = "gateway")]
     Tungstenite(Box<TungsteniteError>),
+    #[cfg(feature = "transport_compression_zlib")]
+    /// A decompression error from the `flate2` crate.
+    DecompressZlib(flate2::DecompressError),
 }
 
 #[cfg(feature = "gateway")]
@@ -101,6 +104,13 @@ impl From<ReqwestError> for Error {
     }
 }
 
+#[cfg(feature = "transport_compression_zlib")]
+impl From<flate2::DecompressError> for Error {
+    fn from(e: flate2::DecompressError) -> Error {
+        Error::DecompressZlib(e)
+    }
+}
+
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
@@ -113,6 +123,8 @@ impl fmt::Display for Error {
             Self::Http(inner) => fmt::Display::fmt(&inner, f),
             #[cfg(feature = "gateway")]
             Self::Tungstenite(inner) => fmt::Display::fmt(&inner, f),
+            #[cfg(feature = "transport_compression_zlib")]
+            Self::DecompressZlib(inner) => fmt::Display::fmt(&inner, f),
         }
     }
 }
@@ -130,6 +142,8 @@ impl StdError for Error {
             Self::Http(inner) => Some(inner),
             #[cfg(feature = "gateway")]
             Self::Tungstenite(inner) => Some(inner),
+            #[cfg(feature = "transport_compression_zlib")]
+            Self::DecompressZlib(inner) => Some(inner),
         }
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -46,9 +46,6 @@ pub enum Error {
     /// An error from the `tungstenite` crate.
     #[cfg(feature = "gateway")]
     Tungstenite(Box<TungsteniteError>),
-    #[cfg(feature = "transport_compression_zlib")]
-    /// A decompression error from the `flate2` crate.
-    DecompressZlib(flate2::DecompressError),
 }
 
 #[cfg(feature = "gateway")]
@@ -104,13 +101,6 @@ impl From<ReqwestError> for Error {
     }
 }
 
-#[cfg(feature = "transport_compression_zlib")]
-impl From<flate2::DecompressError> for Error {
-    fn from(e: flate2::DecompressError) -> Error {
-        Error::DecompressZlib(e)
-    }
-}
-
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
@@ -123,8 +113,6 @@ impl fmt::Display for Error {
             Self::Http(inner) => fmt::Display::fmt(&inner, f),
             #[cfg(feature = "gateway")]
             Self::Tungstenite(inner) => fmt::Display::fmt(&inner, f),
-            #[cfg(feature = "transport_compression_zlib")]
-            Self::DecompressZlib(inner) => fmt::Display::fmt(&inner, f),
         }
     }
 }
@@ -142,8 +130,6 @@ impl StdError for Error {
             Self::Http(inner) => Some(inner),
             #[cfg(feature = "gateway")]
             Self::Tungstenite(inner) => Some(inner),
-            #[cfg(feature = "transport_compression_zlib")]
-            Self::DecompressZlib(inner) => Some(inner),
         }
     }
 }
--- a/src/gateway/sharding/mod.rs
+++ b/src/gateway/sharding/mod.rs
@@ -130,7 +130,7 @@ impl Shard {
     /// use std::num::NonZeroU16;
     /// use std::sync::Arc;
     ///
-    /// use serenity::gateway::Shard;
+    /// use serenity::gateway::{Shard, TransportCompression};
     /// use serenity::model::gateway::{GatewayIntents, ShardInfo};
     /// use serenity::model::id::ShardId;
     /// use serenity::secret_string::SecretString;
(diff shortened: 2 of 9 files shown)


src/gateway/ws.rs
```diff
@@ -102,17 +188,11 @@ impl WsClient {
...
-                    debug!("Failing bytes: {bytes:?}");
-
-                    why
-                })?;
+                let Some(decompressed) = self.compression.inflate(&bytes)? else {
+                    return Ok(None);
+                };
 
-                decompressed
+                String::from_utf8_lossy(decompressed).to_string()
```
GnomedDev: This shouldn't be using the lossy variant, as that is just going to fail serde deserialisation right after.

src/gateway/sharding/mod.rs
```diff
@@ -954,3 +970,24 @@ impl PartialEq for CollectorCallback {
...
+}
+
+impl TransportCompression {
+    fn query_param(self) -> &'static str {
+        match self {
+            Self::None => "",
+            #[cfg(feature = "transport_compression_zlib")]
+            Self::Zlib => "&compress=zlib-stream",
+        }
+    }
```
GnomedDev: Define a mutable empty `ArrayString`, `aformat_into` it, then return that `ArrayString`. `aformat_into` will error if the buffer is too small, so you can return a perfectly sized buffer and not have the CapStr after the call.

src/error.rs
```diff
@@ -46,6 +46,9 @@ pub enum Error {
     /// An error from the `tungstenite` crate.
     #[cfg(feature = "gateway")]
     Tungstenite(Box<TungsteniteError>),
+    #[cfg(feature = "transport_compression_zlib")]
+    /// A decompression error from the `flate2` crate.
+    DecompressZlib(flate2::DecompressError),
```
GnomedDev: Can this be put inside the `Gateway` variant?

src/gateway/ws.rs
```diff
@@ -75,25 +77,109 @@ struct WebSocketMessage<'a> {
...
+        decompressed: Vec<u8>,
+    },
+}
 
-const TIMEOUT: Duration = Duration::from_millis(500);
 const DECOMPRESSION_MULTIPLIER: usize = 3;
+#[cfg(feature = "transport_compression_zlib")]
+const ZLIB_SUFFIX: [u8; 4] = [0x00, 0x00, 0xFF, 0xFF];
+#[cfg(feature = "transport_compression_zlib")]
+const ZLIB_BUFFER_SIZE: usize = 32 * 1024;
```
GnomedDev: This should be a part of the `Compression` type, I feel, as a function.

src/gateway/ws.rs
```diff
@@ -75,25 +77,109 @@ struct WebSocketMessage<'a> {
...
+        inflater: Decompress,
+        compressed: Vec<u8>,
+        decompressed: Vec<u8>,
+    },
+}
 
-const TIMEOUT: Duration = Duration::from_millis(500);
 const DECOMPRESSION_MULTIPLIER: usize = 3;
+#[cfg(feature = "transport_compression_zlib")]
+const ZLIB_SUFFIX: [u8; 4] = [0x00, 0x00, 0xFF, 0xFF];
```
GnomedDev: This should just be part of the `inflate` method, since it's not used anywhere else. That also gets rid of a feature gate.

src/gateway/sharding/mod.rs
```diff
@@ -954,3 +970,24 @@ impl PartialEq for CollectorCallback {
         Arc::ptr_eq(&self.0, &other.0)
     }
 }
+
+/// The transport compression method to use.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum TransportCompression {
```
GnomedDev: `#[non_exhaustive]`

GnomedDev (commented):
Just some initial comments, alongside this I don't think you should implement buffer resizing logic if not necessary as lowering the buffer size isn't going to lead to a noticeable memory usage improvement compared to the code complexity.

src/error.rs
fgardt: `flate2::DecompressError` is not `Clone` but the `Gateway` error requires it so its unfortunately not that simple

src/gateway/ws.rs
fgardt: thoughts about changing the `serde_json::from_str` call to `serde_json::from_slice` and not having to allocate a string + doing the utf8 checking shenanigans? Problem is that the original string is needed for `GatewayEvent::Dispatch` and possible deserialization error message.

src/error.rs
GnomedDev: Remove the Clone implementation for GatewayError then, if possible.

src/gateway/ws.rs
fgardt: I got rid of that constant. Was only used in the initial construction of the decompressed zlib data buffer. The other 2 constants are moved into the `inflate` method.

src/gateway/ws.rs
GnomedDev: If it can skip an allocation, defer the String until error handling and use the lossy variant there.

fgardt:
`zlib-stream` decompression does work now.
Tested by running `e01_basic_ping_bot` with all gateway intents.

Trying to get `zstd-stream` decompression working next.

src/gateway/ws.rs
fgardt: It requires a clone on the close frame and I'm not sure if there is a better solution to create a `FixedString` from a `[u8]` slice.
```diff
@@ -188,40 +188,36 @@ impl WsClient {
             Ok(None) | Err(_) => return Ok(None),
         };
 
-        let json_str = match message {
-            Message::Text(payload) => payload,
+        let json_str = match &message {
+            Message::Text(payload) => payload.as_bytes(),
             Message::Binary(bytes) => {
-                let Some(decompressed) = self.compression.inflate(&bytes)? else {
+                let Some(decompressed) = self.compression.inflate(bytes)? else {
                     return Ok(None);
                 };
 
-                String::from_utf8(decompressed.to_vec()).map_err(|why| {
-                    warn!("Err decompressing bytes: Decompression resulted in invalid UTF-8 data: {why}");
-                    debug!("Failing bytes: {bytes:?}");
-                    debug!("Decompressed bytes: {decompressed:?}");
-
-                    GatewayError::DecompressUtf8(why)
-                })?
+                decompressed
             },
             Message::Close(Some(frame)) => {
-                return Err(Error::Gateway(GatewayError::Closed(Some(frame))));
+                return Err(Error::Gateway(GatewayError::Closed(Some(frame.clone()))));
             },
             _ => return Ok(None),
         };
 
-        match serde_json::from_str(&json_str) {
+        match serde_json::from_slice(json_str) {
             Ok(mut event) => {
                 if let GatewayEvent::Dispatch {
                     original_str, ..
                 } = &mut event
                 {
-                    *original_str = FixedString::from_string_trunc(json_str);
+                    *original_str = FixedString::from_string_trunc(
+                        String::from_utf8_lossy(json_str).to_string(),
+                    );
                 }
 
                 Ok(Some(event))
             },
             Err(err) => {
-                debug!("Failing text: {json_str}");
+                debug!("Failing text: {}", String::from_utf8_lossy(json_str));
                 Err(Error::Json(err))
             },
         }
```

GnomedDev: Couldn't you get rid of the close frame clone by matching on owned message and returning `Cow` from the match?

src/gateway/ws.rs
```diff
@@ -75,25 +79,169 @@ struct WebSocketMessage<'a> {
...
+            TransportCompression::Zlib => Compression::Zlib {
+                inflater: ZlibInflater::new(true),
+                compressed: Vec::new(),
+                decompressed: vec![0; Self::DECOMPRESSED_CAPACITY].into_boxed_slice(),
+            },
+
+            #[cfg(feature = "transport_compression_zstd")]
+            TransportCompression::Zstd => {
+                let mut inflater = ZstdInflater::create();
+                inflater.init().expect("Failed to initialize Zstd decompressor");
```
GnomedDev, line 211: This kind of `create` and `init` pattern is making me iffy about `zstd_safe`, especially not being able to find any documentation on what it possibly errors on. Is there another library for it?

GnomedDev (commented):
Hmm, this looks good, just a small concern with the zstd library.

src/gateway/ws.rs
fgardt: I tried to figure that out but couldn't get a working solution with `Cow` :/

GnomedDev: Sorted it out and pushed to the PR.

src/gateway/ws.rs
fgardt: <sub>this is restating things that have been said on discord for completeness sake.</sub>

[`zstd_safe`](https://github.com/gyscos/zstd-rs) is using bindings to the zstd reference implementation in C (https://github.com/facebook/zstd). The (to my knowledge) only other crate for zstd decompression that is not relying on `zstd_safe` in some way is [`ruzstd`](https://github.com/KillingSpark/zstd-rs). Looking at the docs for `ruzstd` it does not look that nice to work with in our context at least but maybe I'm wrong about that. The performance hit compared to  the C implementation is unclear, should probably be tested.

GnomedDev (approved):
Okay, I'm good with merging this as-is, if it has been tested.

fgardt:
I retested the current state with `e01_basic_ping_bot` and all compression settings worked without any issues.

PR #3043, opened 2024-11-16
feat(ci): move rustflags to specific targets

Erk-:
This would cause a cascade of failures if you developed on anything,
but x86_64.

This is mostly to ease the development on non-x86_64 machines.


GnomedDev:
These have already been removed on next, it's just a problem on current to keep simd-json working, btw.

PR #3044, opened 2024-11-17
Replace bans delete_message_days with delete_message_seconds

cheesycod:
Delete_message_days is deprecated by Discord and is right now being internally converted by serenity. Fix this to allow it to be directly passed


src/model/guild/guild_id.rs
```diff
@@ -182,7 +182,7 @@ impl GuildId {
     /// # let http: Http = unimplemented!();
     /// # let user = UserId::new(1);
     /// // assuming a `user` has already been bound
-    /// let _ = GuildId::new(81384788765712384).ban(&http, user, 4, None).await;
+    /// let _ = GuildId::new(81384788765712384).ban(&http, user, 345600, None).await;
```
GnomedDev: This should be using a const, such as the following, as to not hard code magic numbers.
```rs
const FOUR_DAYS_IN_SECONDS: u32 = 4 * 60 * 60 * 24
```

src/model/guild/guild_id.rs
```diff
@@ -195,13 +195,27 @@ impl GuildId {
...
+        reason: Option<&str>,
+    ) -> Result<()> {
+        // Convert to usize for check overflow
+        let delete_message_seconds_usize =
+            usize::try_from(delete_message_seconds).map_err(|_| {
+                Error::Model(ModelError::TooLarge {
+                    maximum: Maximum::DeleteMessageSeconds,
+                    value: usize::MAX,
+                })
+            })?;
```
GnomedDev: I'm good with merging this as-is, but I would prefer a `u32 -> usize` helper with an unwrap which contained a
```rust
#[cfg(target_pointer_width = 16)]
const { panic!("Serenity is not supported on 16 bit platforms.") };
```

cheesycod: Changed to use a constant u32 and added the target_pointer_width to the overflow case since IDK what ``as usize`` does on 16 bit platforms

src/model/guild/guild_id.rs
```diff
@@ -195,13 +196,33 @@ impl GuildId {
...
+
+        if delete_message_seconds > SEVEN_DAYS_MAXIMUM {
+            #[cfg(target_pointer_width = "16")]
+            const {
+                panic!("Serenity is not supported on 16 bit platforms.")
+            };
+
+            return Err(Error::Model(ModelError::TooLarge {
+                maximum: Maximum::DeleteMessageSeconds,
+                value: SEVEN_DAYS_MAXIMUM as usize,
```
GnomedDev: 1. This is meant for `delete_message_days`, not the maximum that was exceeded.
2. If you are doing `as` with the const panic you might as well use `.try_into().unwrap()`.

src/model/guild/guild_id.rs
```diff
@@ -177,12 +177,13 @@ impl GuildId {
     /// ```rust,no_run
     /// use serenity::model::id::{GuildId, UserId};
     ///
+    /// # const FOUR_DAYS_IN_SECONDS: u32 = 4 * 60 * 60 * 24;
```
GnomedDev: This probably shouldn't be hidden (`#`).

src/model/error.rs
```diff
@@ -117,7 +120,7 @@ impl fmt::Display for Minimum {
 /// #[serenity::async_trait]
 /// impl EventHandler for Handler {
 ///     async fn guild_ban_removal(&self, ctx: Context, guild_id: GuildId, user: User) {
-///         match guild_id.ban(&ctx.http, user.id, 8, Some("No unbanning people!")).await {
+///         match guild_id.ban(&ctx.http, user.id, 691200, Some("No unbanning people!")).await {
```
jamesbt365: could you do the const in the docs here too?

GnomedDev: Same "use a const" comment here.

src/model/guild/guild_id.rs
cheesycod: Regarding point 1, I'm using Maximum::DeleteMessageSeconds here though (?) which should be the maximum exceeded here? Point 2 though, yeah, will fix

GnomedDev: Yes, the maximum exceeded is stored in the `maximum` field. The value that exceeded it is stored in `value`. You have passed the maximum (`SEVEN_DAYS_MAXIMUM`) into the `value` field.

jamesbt365:
I'm not sure the lib should be checking if the values are too large (over 7 days) and sending errors if they are, while it is unlikely that discord changes the maximum value in the future, it is still a possibility and we should not artificially limit it if that is the case. Just my personal take on this

cheesycod:
> I'm not sure the lib should be checking if the values are too large (over 7 days) and sending errors if they are, while it is unlikely that discord changes the maximum value in the future, it is still a possibility and we should not artificially limit it if that is the case. Just my personal take on this

Will remove the check then

src/model/error.rs
```diff
@@ -111,13 +111,16 @@ impl fmt::Display for Minimum {
 /// use serenity::prelude::*;
 /// use serenity::Error;
 ///
+/// const BAN_DURATION: u32 = 691200;
```
GnomedDev: ....

cheesycod: fixed

src/model/error.rs
```diff
@@ -111,28 +111,30 @@ impl fmt::Display for Minimum {
...
+///                 value, ..
+///             })) => {
+///                 println!("Failed deleting {value} days' worth of messages");
+///             },
+///             Err(why) => {
+///                 println!("Unexpected error: {why:?}");
+///             },
+///         }
+///     }
+/// }
```
GnomedDev: Why rework this test to rely on gateway again?

GnomedDev (changes requested):

src/model/error.rs
cheesycod: I used GitHub’s rebase

cheesycod: Should be resolved

src/model/error.rs
```diff
@@ -111,20 +111,22 @@ impl fmt::Display for Minimum {
 /// use serenity::prelude::*;
 /// use serenity::Error;
 ///
+/// const BAN_DURATION: u32 = 8 * 24 * 60 * 60;
```
jamesbt365: Can we name it what it is? BAN_DURATION is so generic and forces you to use math to figure out how long this actually is.

cheesycod: Renamed to EIGHT_DAYS_IN_SECONDS

jamesbt365 (approved):
Other than a nitpick in the [lib-development](https://canary.discord.com/channels/381880193251409931/673965002805477386/1310325902672138300) channel from another user, I think this is fine now.

cheesycod:
Any update here, I have incorporated the nitpick

jamesbt365:
Rebase and I'll merge

cheesycod:
IDK how to rebase this ATP so I'll close it for now and remake the whole branch

PR #3046, opened 2024-11-18
Clean up permissions.rs

GnomedDev:
This PR combines the `bitflags` and `generate_permission_names` macros into a single macro, which also generates the shorthand methods. The first commit in this PR is the actual change but with weird formatting to make it nicer to diff, and the second commit fixes that formatting.

This should make no user-facing diff, except from fixing missing entries in `get_permission_names`.


cheesycod:
I can confirm this works

jamesbt365 (approved):

arqunis (approved):

PR #3049, opened 2024-11-18
Get rid of redundant status argument in Http::wind

GnomedDev:
Turns out this was always being passed 204, except one case where it should have been deserialising the `Member` return type anyway. This fixes both issues.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1878,16 +1878,14 @@ impl Http {
         .await
     }
 
-    /// Edits the current user's nickname for the provided [`Guild`] via its Id.
-    ///
-    /// Pass [`None`] to reset the nickname.
-    pub async fn edit_nickname(
+    /// Edits the current member for the provided [`Guild`] via its Id.
+    pub async fn edit_current_member(
         &self,
         guild_id: GuildId,
         map: &impl serde::Serialize,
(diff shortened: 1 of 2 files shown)


src/http/client.rs
```diff
@@ -4416,16 +4414,16 @@ impl Http {
     ///
     /// This is a function that performs a light amount of work and returns an empty tuple, so it's
     /// called "self.wind" to denote that it's lightweight.
```
```suggestion
    /// Performs a request and, if successful, verifies that the response
    /// status code is equal to the `NO_CONTENT` status code. 
    /// 
    ///
    /// This is a function that performs a light amount of work and returns an empty tuple, so it's
    /// called "self.wind" to denote that it's lightweight.
```

This is meant to include the two lines above, but Github isn't allowing me to select them.

GnomedDev:
```rust
    /// Performs a request and verifies that Discord responds with [`StatusCode::NO_CONTENT`].
    ///
    /// This is a function that performs a light amount of work and returns the unit type, so it's
    /// called "self.wind" to denote that it's lightweight.
```
Is this good for you?

arqunis: Yeah, that's fine.

arqunis (approved):

PR #3052, opened 2024-11-19
Improve tokens semantics and clean up gateway feature gates

mkrasnitski:
This divorces the gateway feature from the model/http code such that there is no cross-contamination (setting the stage for a crate split). A few doctests got rewritten to not rely on the gateway, and the `Http::token` method was removed. The gateway structs now carry their own copy of the token, and the Http token field was made optional for users who only care about webhooks.

Furthermore, renamed the `secret_string` module to `secrets` and added a `Token` type that can partially validate token strings. This type is what is now passed in to the client. The user is expected to call either `str::parse::<Token>()` or `Token::from_str` in order to create a Client.


GnomedDev (commented):
This looks mainly good just some minor issues:
- Can you add a `Token::from_env("DISCORD_TOKEN")` helper for the examples?
- Can you remove the `Http::default` implementation, and have `Http::new` take the token with a `Http::wthout_token` constructor for webhook-only use, since this is the super minority of use cases.

mkrasnitski:
Done.

src/http/client.rs
```diff
@@ -71,34 +69,25 @@ where
...
-            ratelimiter: None,
-            ratelimiter_disabled: false,
-            token: parse_token(token),
-            proxy: None,
-            application_id: None,
-            default_allowed_mentions: None,
-        }
+    /// Construct a new builder.
+    pub fn new() -> Self {
+        Self::default()
```
GnomedDev: This is still defaulting to no token, which is iffy for me. Can this have the same `without_auth` treatment?

src/http/client.rs
```diff
@@ -241,15 +218,23 @@ pub struct Http {
...
+    /// Construct an authorized HTTP client.
     #[must_use]
-    pub fn new(token: &str) -> Self {
-        HttpBuilder::new(token).build()
+    pub fn new(token: Token) -> Self {
+        HttpBuilder::new().token(token).build()
+    }
+
+    /// Construct an unauthorized HTTP client, with no token. Few things will work, but webhooks
+    /// are one exception.
```
GnomedDev:
```suggestion
    /// Construct an unauthorized HTTP client, with no token.
    /// 
    /// Most Discord functionality requires a logged-in Bot token, but there are some exceptions such as webhook endpoints.
```

src/http/client.rs
mkrasnitski: Sure. Do you also think `without_auth` is a better name than `Http::without_token`?

GnomedDev: `without_token` sounds better imo

mkrasnitski: Done. Removed `HttpBuilder::token` as well.

src/error.rs
```diff
@@ -46,6 +47,10 @@ pub enum Error {
     /// An error from the `tungstenite` crate.
     #[cfg(feature = "gateway")]
     Tungstenite(Box<TungsteniteError>),
+    /// An error from the [`secrets`] module.
+    ///
+    /// [`secrets`]: crate::secrets
+    Token(TokenError),
```
GnomedDev, line 53: Does this have to be a part of this enum? I can't see a spot that returns this instead of just `Result<T, TokenError>`.

mkrasnitski: I figured this would be useful when using `?` so that `TokenError` could coerce to `Error`.

GnomedDev (approved):

mkrasnitski:
Should be marked breaking btw

PR #3053, opened 2024-11-20
Remove ArgumentConvert trait and implementations

GnomedDev:
This historically was added for poise usage, but then poise still had to implement it's own trait which it uses when possible instead of this. Since this kind of logic is usually quite framework-specific let's get rid of it from serenity.


arqunis (approved):

GnomedDev:
Thinking about it, let's hold off merging this until poise@serenity-next doesn't fall apart catastrophically from this. That would probably need https://github.com/serenity-rs/poise/pull/320 merged.

mkrasnitski:
This should be unblocked now that serenity-rs/poise#357 is merged

src/utils/argument_convert/mod.rs
```diff
@@ -1,195 +0,0 @@
...
-        if let Some(parts) = s.strip_prefix(prefix.as_str()) {
-            let mut parts = parts.splitn(2, '/');
-
-            let guild_id = parts.next()?.parse().ok()?;
-            let channel_id = parts.next()?.parse().ok()?;
-            return Some((guild_id, channel_id));
-        }
-    }
-    None
-}
```
mkrasnitski, line 179: Can you move this to `utils/mod.rs` as well?

GnomedDev: Done.

mkrasnitski (approved):

PR #3055, opened 2024-11-22
Factor out collector methods on models into traits

mkrasnitski:
Moves the feature-gated collector methods on `UserId`, `MessageId`, etc. into four traits named `CollectMessages`, `CollectReactions`, `CollectModalInteractions` and `CollectComponentInteractions`. Also, move the quick modal machinery into the collector module and define a `QuickModal` trait. This fully removes any collector feature gates from the model types.


examples/testing/src/main.rs
```diff
@@ -227,6 +228,18 @@ async fn message(ctx: &Context, msg: Message) -> Result<(), serenity::Error> {
             .await?;
         tokio::time::sleep(std::time::Duration::from_secs(10)).await;
         channel_id.delete(&ctx.http, None).await?;
+    } else if msg.content.starts_with("suppressembeds") {
+        // let mut msg = channel_id
+        //     .send_message(
+        //         &ctx.http,
+        //         CreateMessage::new().content("https://www.youtube.com/watch?v=bsDcVtJCAsA"),
+        //     )
+        //     .await?;
```
GnomedDev: Can you remove the commented out code?

GnomedDev (approved):
Looks good, just a small nitpick

examples/testing/src/main.rs
mkrasnitski: Whoops didn't mean to commit that

arqunis (approved):

PR #3056, opened 2024-11-24
Adds "unstable-exhaustive-types" as new feature.

TheCataliasTNT2k:
This feature deactivates all "#[non_exhaustive]" added into the code. This can be used by developers to ensure, that they match all enum variants or struct fields during pattern matching.

This should not break any existing code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -153,6 +153,10 @@ native_tls_backend = [
     "bytes",
 ]
 
+# Disables all "#[non_exhaustive]" macros, to ensure that all enum variants or struct fields are required when pattern matching.
+unstable_exhaustive_types = ["serenity-voice-model/unstable_exhaustive_types"]
+
+
 
 [package.metadata.docs.rs]
 features = ["full"]
--- a/src/builder/create_attachment.rs
+++ b/src/builder/create_attachment.rs
@@ -17,7 +17,7 @@ use crate::model::id::AttachmentId;
 ///
 /// [Discord docs](https://discord.com/developers/docs/resources/channel#attachment-object-attachment-structure).
 #[derive(Clone, Debug, Serialize, PartialEq)]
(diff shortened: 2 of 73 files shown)


GnomedDev:
Why is this pointed at next? You yourself say it isn't a breaking change.

TheCataliasTNT2k:
This is pointed as next because
- I use next instead of current
- it saves you from doing it again for next, which would have to be done at some point.

GnomedDev (changes requested):
Okay, so there is no technical reason it cannot be on `current`. Please rebase the PR on `current`.

It does not save anyone work to point a PR at the wrong branch, as if it is pointed at next but could be on current someone is going to backport it for current and lead to rebase annoyances.

TheCataliasTNT2k:
I could also create another PR for current if you want, so you have both.
But I need this feature for next.
Is there any way to get it for next?

GnomedDev:
Do not create two PRs, simply rebase this one on current. When a maintainer (probably me) feels like it (usually meaning not many PRs to `next`), next will be rebased on current.

TheCataliasTNT2k:
Ok, that makes sense.
However, I an not going to make a rebase, I have like 30 files conflicting.
I will just recreate the commit really quick on current, that is simpler :D

TheCataliasTNT2k:
Done

GnomedDev (commented):
Why disable this feature when doing `doc` builds?

TheCataliasTNT2k:
To not alte the documentation.
If it was active, when the documentation is built, the "non_exhaustive" would be deactivated when the documentation is built.

Joshix-1:
I like this feature. Why are you removing `#[serde(from = "u8", into = "u8")]` from multiple enums? Seems to be an unrelated change.

PR #3058, opened 2024-11-24
Add guild_id to ReactionRemoveAll

jamesbt365:
This was missing on FullEvent but was present within the event itself.

### DIFF
--- a/src/gateway/client/dispatch.rs
+++ b/src/gateway/client/dispatch.rs
@@ -325,6 +325,7 @@ fn update_cache_with_event(
             removed_reaction: event.reaction,
         },
         Event::ReactionRemoveAll(event) => FullEvent::ReactionRemoveAll {
+            guild_id: event.guild_id,
             channel_id: event.channel_id,
             removed_from_message_id: event.message_id,
         },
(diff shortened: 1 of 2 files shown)


src/client/event_handler.rs
```diff
@@ -295,8 +295,8 @@ event_handler! {
 
     /// Dispatched when all reactions of a message are detached from a message.
     ///
-    /// Provides the channel's id and the message's id.
-    ReactionRemoveAll { channel_id: ChannelId, removed_from_message_id: MessageId } => async fn reaction_remove_all(&self, ctx: Context);
+    /// Provides the channel's id, message's id, and guild's id if in a guild.
+    ReactionRemoveAll { guild_id: Option<GuildId>, channel_id: ChannelId, removed_from_message_id: MessageId } => async fn reaction_remove_all(&self, ctx: Context);
```
GnomedDev: This, sadly, is a breaking change. `FullEvent` does not generate `non_exhaustive` on the enum variants therefore code which destructures `FullEvent::ReactionRemoveAll` will break with this.

GnomedDev (changes requested):

src/client/event_handler.rs
jamesbt365: Yikes, resolved.

GnomedDev (approved):

PR #3059, opened 2024-11-25
Fix incorrect parameters for retrieving invites

archanyhm:
It seems that during https://github.com/serenity-rs/serenity/pull/2288 the route parameters for retrieving guild invite metadata haven't been mapped to the correct values outlined in the Discord API documentation causing the member counts to be reported as 0 erroneously no matter whether member_counts was set or not.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -4166,11 +4166,11 @@ impl Http {
         let code = crate::utils::parse_invite(code);
 
         let mut params = vec![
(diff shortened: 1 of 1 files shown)


archanyhm:
Would it help the testing if I provided a minimal reproducible example of the bug with and without fix?

GnomedDev (approved):
Tested this myself, looks like `with_expiration` just doesn't work but that looks to be a discord bug so let's get this merged in.

PR #3063, opened 2024-11-25
Remove broken Message::super_react

jamesbt365:
Turns out you can't send these anymore, and they weren't documented to begin with, I don't know how I ended up implementing it.

This is on current despite it removing a public method because the method does not work.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -836,7 +836,8 @@ impl Http {
         .await
     }
 
-    async fn create_reaction_(
+    /// Reacts to a message.
+    pub async fn create_reaction(
         &self,
         channel_id: ChannelId,
         message_id: MessageId,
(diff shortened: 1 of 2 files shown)


src/http/client.rs
```diff
@@ -865,19 +864,8 @@ impl Http {
         message_id: MessageId,
         reaction_type: &ReactionType,
     ) -> Result<()> {
-        self.create_reaction_(channel_id, message_id, reaction_type, false).await
+        self.create_reaction_(channel_id, message_id, reaction_type).await
```
GnomedDev: The `create_reaction_` function can be merged back into `create_reaction` now.

GnomedDev (changes requested):

GnomedDev (approved):

PR #3064, opened 2024-11-25
Make FullEvent fields non exhaustive

jamesbt365:
Right now we can't add fields to events without it being a breaking change, unless I'm missing anything this should let us do that.

### DIFF
--- a/src/gateway/client/event_handler.rs
+++ b/src/gateway/client/event_handler.rs
@@ -57,6 +57,7 @@ macro_rules! event_handler {
                 $( #[doc = $doc] )*
                 $( #[cfg(feature = $feature)] )?
                 $( #[deprecated = $deprecated] )?
+                #[cfg_attr(not(feature = "unstable"), non_exhaustive)]
                 $variant_name {
                     $( $arg_name: $arg_type ),*
                 },


cheesycod:
I personally dislike this change, I would prefer for all FullEvent changes to be breaking so I am informed of new stuff and can add those features to my bot as they're added to serenity

Then again, I suppose it doesnt matter much, given I use the ``unstable`` feature flag anyways...

DPlayer234:
Would this even fix the issue? The same macro is used to generate the `EventHandler` trait's functions, so adding a field would still break users/implementors of `EventHandler`. Not exposing all fields of `FullEvent` in `EventHandler` also doesn't sound good.

cheesycod:
Honestly, if you really wanted this, maybe the fields should just put in a struct instead like ``FullEventMessageData`` and so on, then adding new fields would just add to the struct and not break the functions.

jamesbt365:
Right, i forgot about the functions, I wonder if theres a nice way to do this without throwing everything into a struct.

jamesbt365:
> I personally dislike this change, I would prefer for all FullEvent changes to be breaking so I am informed of new stuff and can add those features to my bot as they're added to serenity
> 
> Then again, I suppose it doesnt matter much, given I use the `unstable` feature flag anyways...

And I would prefer not having to do a breaking version bump to add fields to events!

jamesbt365:
The only way I can see this working is if we drop the function method and make users match the fullevent themselves, not sure if this is a warranted change.

PR #3065, opened 2024-11-25
Encapsulate Message Cache in it's own type

GnomedDev:
Draft as based on #3061.

This keeps the invariant (channel ordering and max messages setting) encapsulated and correctly handled, while speeding up message create handling and pushing messages into the cache from edits.


GnomedDev:
I am not interested in pushing this over the finish line, especially since I do not use the message cache. If anyone wants to pick this up, feel free to open a PR based on this.

PR #3066, opened 2024-11-26
Fix CacheUpdate implementations for extra FullEvents

mkrasnitski:
The gateway was generating extra FullEvents by directly inspecting the contents of the cache, however this functionality can be incorporated as part of the `CacheUpdate` implementation for those events. The main cache fields are also private now. The temp_cache fields could be made private too, but I'm unsure if it's worth it.

### DIFF
--- a/src/gateway/client/dispatch.rs
+++ b/src/gateway/client/dispatch.rs
@@ -11,8 +11,6 @@ use crate::internal::tokio::spawn_named;
 use crate::model::channel::ChannelType;
 use crate::model::event::Event;
 use crate::model::guild::Member;
-#[cfg(feature = "cache")]
-use crate::model::id::GuildId;
 
 #[cfg(feature = "cache")]
 macro_rules! if_cache {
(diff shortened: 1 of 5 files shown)


mkrasnitski:
Blocked on #3053 for now. Don't feel like fixing code that's gonna get removed anyways.

GnomedDev:
Since this is blocked on another PR, I am changing to draft.

src/gateway/client/dispatch.rs
```diff
@@ -206,7 +195,7 @@ fn update_cache_with_event(
             }
         },
         Event::GuildDelete(mut event) => {
-            let full = if_cache!(event.update(cache));
+            let full = if_cache!(update_cache!(cache, event));
```
GnomedDev, line 198: Why change this? I'm fine with merging as-is, just curious.

GnomedDev (approved):
LGTM

src/gateway/client/dispatch.rs
mkrasnitski: Simply because `update_cache!` is used elsewhere in the file to do the same thing. Figured I'd make it consistent.

PR #3075, opened 2024-12-08
Allows the usage of "channel urls" for parameters of type "GuildChannel"

TheCataliasTNT2k:
Similar to message urls, Discord also provides urls for guild channels.
This PR adds this method to pass a url as argument for the Channel type.
Private channels are not affected by this change.


TheCataliasTNT2k:
The linter is not complaining about any of my changes.
Maybe this can be merged nevertheless?

arqunis (approved):

PR #3077, opened 2024-12-08
Specify shared crate metadata at the workspace level

mkrasnitski:
This keeps crate metadata in sync across the whole workspace, including MSRV and edition, plus adds some missing metadata fields to serenity's subcrates.

Feedback wanted: should the `authors` field also be applied to the whole workspace? Note that workspace fields are opt-in via `field.workspace = true` and can be overridden by simply providing a `field = ...` value.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,20 +1,30 @@
 [package]
+name = "serenity"
+version = "0.12.4"
 authors = ["Alex M. M. <acdenissk69@gmail.com>"]
 description = "A Rust library for the Discord API."
-documentation = "https://docs.rs/serenity"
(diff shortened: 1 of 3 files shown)


arqunis (approved):

arqunis:
> Feedback wanted: should the `authors` field also be applied to the whole workspace? Note that workspace fields are opt-in via `field.workspace = true` and can be overridden by simply providing a `field = ...` value.

I believe that should be set individually per crate, especially since we want to integrate poise into the Serenity repository, [which has a different `authors` field](https://github.com/serenity-rs/poise/blob/db10b126c8b07f7e1924cba1672f04ff24ed4ec7/Cargo.toml#L2).

PR #3088, opened 2025-01-07
replace dmd with dms

cheesycod:
Take 2 on the delete_message_seconds PR. This also removes the pre-validation code as discussed in #lib-development and by extension currently removes the broken example in src/model/error.rs


mkrasnitski:
The code itself is fine. I don't know if this has been brought up, but I think there's a small footgun here that users might forget to switch from days to seconds and they'll then be banning people for 7 seconds instead of 7 days... Would it be better to add a `ban_seconds` or equivalent so that people can switch?

jamesbt365:
> The code itself is fine. I don't know if this has been brought up, but I think there's a small footgun here that users might forget to switch from days to seconds and they'll then be banning people for 7 seconds instead of 7 days... Would it be better to add a `ban_seconds` or equivalent so that people can switch?

I don't think that its a big issue, especially because it'll be in the changelog. And a small correction here, its not banning them for 7 seconds, its deleting their messages in the past 7 seconds. So at most developers wouldn't delete as many messages if they didn't read the changelog.

I'm not really a fan of implementing a new method, deprecating the old one and then just switching in internally for the next major release under the old name? I don't know, could be fine though?

src/http/client.rs
```diff
@@ -332,19 +332,17 @@ impl Http {
     }
 
     /// Bans a [`User`] from a [`Guild`], removing their messages sent in the last X number of
-    /// days.
+    /// seconds.
     ///
-    /// Passing a `delete_message_days` of `0` is equivalent to not removing any messages. Up to
-    /// `7` days' worth of messages may be deleted.
+    /// Passing a `delete_message_seconds` of `0` is equivalent to not removing any messages. Up to
+    /// `604800` seconds' (or 7 days') worth of messages may be deleted.
```
tazz4843:
```suggestion
    /// `604800` seconds (or 7 days) worth of messages may be deleted.
```

src/model/guild/guild_id.rs
```diff
@@ -161,8 +161,9 @@ impl GuildId {
         builder.execute(http, self, user_id).await
     }
 
-    /// Ban a [`User`] from the guild, deleting a number of days' worth of messages (`dmd`) between
-    /// the range 0 and 7.
+    /// Ban a [`User`] from the guild, deleting a number of seconds' worth of messages (`dms`)
```
tazz4843:
```suggestion
    /// Ban a [`User`] from the guild, deleting `dms` seconds worth of messages from them.
```

src/model/guild/guild_id.rs
```diff
@@ -161,8 +161,9 @@ impl GuildId {
...
+    /// between the range 0 and 604800 (subject to change). To ban someone for X days', multiply
```
tazz4843:
```suggestion
    /// `dms` should be between 0 and 604800 (subject to change). To ban someone for X days, multiply
```

Also is "subject to change" here really valid? If anything it would be Discord changing it.

src/model/guild/member.rs
```diff
@@ -134,18 +134,17 @@ impl Member {
     }
 
     /// Ban a [`User`] from the guild, deleting a number of days' worth of messages (`dmd`) between
```
tazz4843:
```suggestion
    /// Ban a [`User`] from the guild, deleting `dms` seconds worth of messages from them.
```

src/model/guild/member.rs
```diff
@@ -134,18 +134,17 @@ impl Member {
...
-    /// the range 0 and 7.
+    /// the range 0 and 604800. To ban someone for X days', multiply X by 86400
```
tazz4843:
```suggestion
    /// `dms` should be between 0 and 604800. To ban someone for X days, multiply X by 86400.
```

tazz4843 (commented):
I see a few comment issues here, but besides that looks fine. I still don't like swapping it out without at least making a relatively big deal in the changelog. For some people this might trigger a mismatched types error, hopefully that helps soften impact here.

cheesycod:
> The code itself is fine. I don't know if this has been brought up, but I think there's a small footgun here that users might forget to switch from days to seconds and they'll then be banning people for 7 seconds instead of 7 days... Would it be better to add a `ban_seconds` or equivalent so that people can switch?

This should always cause a type error anyways unless you're doing a try_into/into thing. But yeah, should be in changelog

tazz4843:
Implicit casting is what I'm thinking of with hardcoded integer values specifically.

mkrasnitski:
Yeah, since the type is now widened, then integer literals will always typecheck correctly.

> @jamesbt365: I'm not really a fan of implementing a new method, deprecating the old one and then just switching in internally for the next major release under the old name? I don't know, could be fine though?

I'd even say that the two methods could potentially live side by side, with `ban` just converting to seconds and then wrapping `ban_seconds`.

cheesycod:
> Implicit casting is what I'm thinking of with hardcoded integer values specifically.

Technically, couldnt this be solved by just making a new DeleteMessageSeconds(u32) wrapper type just like how we have wrapper types for ShardId and other id types?

Just wondering out of curiosity

jamesbt365:
I'd say this is fine, but I'll leave it to someone else to make the final decision on if this is okay (as it may confuse bot developers and lead them to delete messages far less than they intended to)

arqunis (approved):

arqunis:
Manually merged in https://github.com/serenity-rs/serenity/commit/09bea5638cc4c6a2150901c6dfa700b092bdbfc0 to rebase personally.

PR #3095, opened 2025-01-19
Update two emoji functions with roles field

### LINKED ISSUE #3328: GuildId::edit_emoji missing setting allowed roles
We [don't currently support](https://docs.rs/serenity/0.12.4/serenity/model/id/struct.GuildId.html#method.edit_emoji) changing the allowed roles for the emoji, we should add support for this

relevant docs: https://discord.com/developers/docs/resources/emoji#modify-guild-emoji

nwerosama:
As was discussed/brought up [in Discord](https://discord.com/channels/381880193251409931/381912587505500160/1317791118813036564) a month ago for missing support for roles field in `create_emoji` and `edit_emoji`.

I feel like I am missing something prior to creating this PR but feel free to correct me on it though.

Yes, it is also tested on the bot before forking and creating the PR.

### DIFF
--- a/src/model/guild/guild_id.rs
+++ b/src/model/guild/guild_id.rs
@@ -318,10 +318,13 @@ impl GuildId {
         builder.execute(http, self).await
     }
 
-    /// Creates an emoji in the guild with a name and base64-encoded image.
+    /// Creates an emoji in the guild with a name and base64-encoded image and set of roles.
     ///
     /// The name of the emoji must be at least 2 characters long and can only contain alphanumeric
     /// characters and underscores.
+    ///
+    /// The emoji will only be visible by anyone that has the role(s) defined in the `roles`
+    /// parameter.
     ///
     /// Requires the [Create Guild Expressions] permission.
     ///
@@ -343,17 +346,21 @@ impl GuildId {
         http: &Http,
         name: &str,
         image: &str,
+        roles: Option<Vec<RoleId>>,
         reason: Option<&str>,
     ) -> Result<Emoji> {
         #[derive(serde::Serialize)]
         struct CreateEmoji<'a> {
             name: &'a str,
             image: &'a str,
+            #[serde(skip_serializing_if = "Option::is_none")]
+            roles: Option<Vec<RoleId>>,
         }
 
         let body = CreateEmoji {
             name,
             image,
+            roles,
         };
 
         http.create_emoji(self, &body, reason).await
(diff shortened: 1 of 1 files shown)


src/builder/create_command.rs
```diff
@@ -344,7 +344,6 @@ impl<'a> CreateCommand<'a> {
         Self {
             kind: None,
             handler: None,
-
```
GnomedDev: Can you undo this unnecessary change?

src/model/guild/guild_id.rs
```diff
@@ -317,7 +317,7 @@ impl GuildId {
         builder.execute(http, self).await
     }
 
-    /// Creates an emoji in the guild with a name and base64-encoded image.
+    /// Creates an emoji in the guild with a name and base64-encoded image and set of roles.
```
GnomedDev, line 321: Can this documentation be expanded a little to explain what the roles passed do? I had to check ddev docs to realize it restricts the emoji to only be usable by those roles.

src/model/guild/guild_id.rs
```diff
@@ -342,17 +342,20 @@ impl GuildId {
         http: &Http,
         name: &str,
         image: &str,
+        roles: Vec<RoleId>,
```
GnomedDev: `Option<Vec<RoleId>>`?

src/model/guild/guild_id.rs
```diff
@@ -582,15 +585,18 @@ impl GuildId {
         http: &Http,
         emoji_id: EmojiId,
         name: &str,
+        roles: Option<Vec<RoleId>>,
         reason: Option<&str>,
     ) -> Result<Emoji> {
         #[derive(serde::Serialize)]
         struct EditEmoji<'a> {
             name: &'a str,
+            roles: Option<Vec<RoleId>>,
```
GnomedDev, line 603: Can you add a `skip_serialising_if` for Option::none?

GnomedDev (changes requested):

src/model/guild/guild_id.rs
nwerosama: How does one find/do this?

src/model/guild/guild_id.rs
nwerosama: Should I add the section about it above the permission line comment? Something like "The emoji will only be visible to those that has the role to use it."? It's the furthest I can think of right now while understanding the Discord docs.

src/model/guild/guild_id.rs
mkrasnitski: Annotate the field with `#[serde(skip_serializing_if = "Option::is_none")]`

GnomedDev (approved):
LGTM

PR #3096, opened 2025-01-20
Add kind to all interaction enum variants to fix interaction serde

cheesycod:
Currently, serializing an Interaction strips its ``type`` entirely making said interaction not only not deserializable but also need manual patching to confirm with discords own JSON structure (e.g. when serializing it for storage/use in other systems, the resulting JSON loses the type entirely)

Fix this issue the same way Components are implemented: by adding a ``kind`` to each interaction variant.


jamesbt365:
So this just increases model sizes... for what? The struct *is* the kind of interaction and there isn't really a reason to serialize these types? Not even sure why they even implement serialize tbh

cheesycod:
> So this just increases model sizes... for what? The struct _is_ the kind of interaction and there isn't really a reason to serialize these types? Not even sure why they even implement serialize tbh

Right now, i have to manually inject type into Interaction before serializing them into ``serde_json::Value`` (and finally ``mlua::LuaValue`` (at the end of the day) and dumping it there. I actually rely heavily on the fact that you can serialize interactions throughout my bot

E.g:

```
                    // Inject in type
                    if let serde_json::Value::Object(ref mut map) = value {
                        let typ: u8 = interaction.kind().0;
                        map.insert("type".to_string(), serde_json::Value::Number(typ.into()));
                    }

                    serde_json::json!({
                        "InteractionCreate": {
                            "interaction": value
                        }
                    })
```

GnomedDev:
If this is going to be done, I would like it to be a private ZST using const generics to encode the number, in order to keep it internal and not touch type sizes. As this PR has been left idle for months, I'll close for now, although if you would like to get this to a mergable state I can reopen.

PR #3108, opened 2025-02-08
Merge `ShardQueuer` functionality into `ShardManager`

mkrasnitski:
Untangles the spaghetti of shard management code by removing the middleman `ShardQueuer` struct, which used a disorienting mess of mpsc channels to facilitate message passing between the `ShardManager` and each of its `ShardRunner`s.

The new `ShardManager::run` function takes care of creating/spawning new runners as needed. This is fine to call without spawning a separate task (which was being done for `ShardQueuer`) because `Client::start_connection` previously already waited for the `ShardQueuer` loop to (essentially) finish before returning. In other words, the Client would start the `ShardManager` and then block, but the waiting was just being facilitated through channels, which wasn't obvious.

The only messages passed across channels now are `ShardRunnerMessage` (sent from the manager to a given runner), and a new `ShardManagerMessage` (sent from a runner back to the manager). For example, the manager can send a message to a runner telling it to restart, then the runner will shut itself down and ask the manager to reboot the shard by sending it a message back.

The shard queue now keeps track of if it should dispatch in concurrent batches or not. Since we only use concurrent startup when first initializing all shards, and then switch it off once and don't re-enable it, this should be fine.

I've done some cursory testing of this using the repo examples, but I'd like some real apps to test this out before it's merged so that I know nothing is broken. Especially I'd like to make sure that concurrent shard starting works correctly.


src/gateway/sharding/shard_manager.rs
```diff
@@ -170,41 +116,76 @@ impl ShardManager {
...
-    /// responsible for the given ID.
+    /// The main interface for starting the management of shards. Initializes the shards by
+    /// queueing them for starting, and then listens for [`ShardManagerMessage`]s in a loop.
     ///
-    /// If a shard has been queued but has not yet been initiated, then this will return `false`.
-    pub async fn has(&self, shard_id: ShardId) -> bool {
-        self.runners.lock().await.contains_key(&shard_id)
+    /// This function takes care of:
+    ///   1. Starting [`ShardRunner`]s and spawning tasks.
+    ///   2. Restarting shards (by shuttind them down and spawning a new task) when requested.
```
GnomedDev: Typo
```suggestion
    ///   2. Restarting shards (by shutting them down and spawning a new task) when requested.
```

src/gateway/sharding/shard_manager.rs
```diff
@@ -390,14 +379,70 @@ pub struct ShardManagerOptions {
...
     pub intents: GatewayIntents,
     pub presence: Option<PresenceData>,
-    pub max_concurrency: NonZeroU16,
-    /// Number of seconds to wait between starting each shard/set of shards start
-    pub wait_time_between_shard_start: Duration,
-    pub compression: TransportCompression,
+}
+
+pub enum ShardManagerMessage {
+    PollShardQueue,
```
GnomedDev: Can this variant be documented, or even document all the variants? You know they need explanation as you did it in the PR description.

src/gateway/sharding/shard_manager.rs
```diff
@@ -390,14 +379,70 @@ pub struct ShardManagerOptions {
...
+    }
+
+    /// Calculates the corresponding bucket for the given `ShardId` and **prepends** to it.
+    pub fn push_front(&mut self, shard_id: ShardId) {
+        let bucket = self.calculate_bucket(shard_id);
+        self.buckets[bucket].push_front(shard_id);
+    }
+
+    /// Pops a `ShardId` from every bucket containing at least one and returns them all as a `Vec`.
+    /// If the queue is in concurrent mode, will wait until all buckets are filled before returning
```
GnomedDev: "will wait" makes me think this method is doing some kind of async locking/waiting, when it instead just returns `None`. Can this wording be improved?

src/gateway/sharding/shard_manager.rs
```diff
@@ -214,169 +195,177 @@ impl ShardManager {
...
-
-    /// Returns the [`ShardId`]s of the shards that have been instantiated and currently have a
-    /// valid [`ShardRunner`].
-    ///
-    /// [`ShardRunner`]: super::ShardRunner
-    #[cfg_attr(feature = "tracing_instrument", instrument(skip(self)))]
-    pub async fn shards_instantiated(&self) -> Vec<ShardId> {
-        self.runners.lock().await.keys().copied().collect()
+        self.shutdown(shard_id, 4000);
+        self.boot(shard_id, false).await;
```
GnomedDev: This is no longer waiting for the shard to disconnect before restarting, as `shutdown` is now just firing off a message, which worries me quite a bit as a possible race condition. The `boot` call being async is also misleading as this code path is interchangably using unbounded and bounded senders for reasons?

Can you implement the waiting for shutdown, by using something like tokio::Notify (or a simpler solution: sending a oneshot zst channel over)?

src/gateway/sharding/shard_manager.rs
```diff
@@ -214,169 +195,177 @@ impl ShardManager {
...
+    /// This shuts down all active [`ShardRunner`]s.
     ///
     /// [`ShardRunner`]: super::ShardRunner
     fn drop(&mut self) {
-        drop(self.shard_queuer.unbounded_send(ShardQueuerMessage::Shutdown));
+        info!("Shutting down all shards");
+
+        for shard_id in self.shards_instantiated() {
+            self.shutdown(shard_id, 1000);
+        }
```
GnomedDev: Nit: This is making a Vec of ShardIds just to iterate through them. This could probably be improved by outlining the inner logic (well, just the warn tbh) of `ShardManager::shutdown` and using `drain` or `take`.

src/gateway/sharding/shard_manager.rs
```diff
@@ -390,14 +379,70 @@ pub struct ShardManagerOptions {
...
+    Err(GatewayError),
+}
+
+/// A queue of [`ShardId`]s that is split up into multiple buckets according to the value of
+/// [`max_concurrency`](crate::model::gateway::SessionStartLimit::max_concurrency).
+#[must_use]
+pub struct ShardQueue {
+    buckets: FixedArray<VecDeque<ShardId>, u16>,
+    concurrent: bool,
+}
```
GnomedDev: I feel like ShardQueue should be split out into it's own file, because it's super easy to decouple and already seems a little out of place at the end of this file.

src/gateway/sharding/shard_runner.rs
```diff
@@ -487,14 +467,29 @@ impl ShardRunner {
...
+    fn make_context(&self) -> Context {
+        Context::new(
+            Arc::clone(&self.data),
+            self.messenger(),
+            self.shard.shard_info().id,
+            Arc::clone(&self.http),
+            #[cfg(feature = "cache")]
+            Arc::clone(&self.cache),
+        )
+    }
```
GnomedDev, line 482: Why move this?

src/gateway/sharding/shard_runner.rs
```diff
@@ -487,14 +467,29 @@ impl ShardRunner {
...
-    async fn update_manager(&self) {
-        self.manager
-            .update_shard_latency_and_stage(
-                self.shard.shard_info().id,
-                self.shard.latency(),
-                self.shard.stage(),
-            )
-            .await;
+    async fn update_runner_info(&self) {
+        let mut runner_info = self.runner_info.lock().await;
```
GnomedDev: Can you convert this to `try_lock` and just ignore if it can't grab the lock? This makes me worried that user code could hold the runner_info lock too long and just lead to a shard dying.

Thinking about it, the runners lock could also be changed to a sync lock as it's never needed to be held over an await sensibly.

src/gateway/sharding/shard_runner.rs
```diff
@@ -475,10 +455,10 @@ impl ShardRunner {
     async fn request_restart(&mut self) {
         debug!("[ShardRunner {:?}] Requesting restart", self.shard.shard_info());
 
-        self.update_manager().await;
+        self.update_runner_info().await;
```
GnomedDev: I don't see a reason why the runner info should be updated before sending a message that removes this info from the runners map, can you remove this?

GnomedDev (changes requested):
Looks very good, a lot better, just some concerns I spotted.

src/gateway/sharding/shard_manager.rs
mkrasnitski: Done. Renamed the `Err` variant to `Quit` as well.

src/gateway/sharding/shard_manager.rs
mkrasnitski: I inlined the logic but I wonder if we should try waiting for shutdown to finish or not.

src/gateway/sharding/shard_manager.rs
mkrasnitski: Done

src/gateway/sharding/shard_runner.rs
mkrasnitski: It was kind of out of place in the middle of the method list. It's a small method that is unrelated to direct shard state management.

src/gateway/sharding/shard_runner.rs
mkrasnitski: Changed it - can you think of a circumstance where one of the locks might end up getting held across an await point? Just so I have an example in mind

src/gateway/sharding/shard_runner.rs
mkrasnitski: Done

src/gateway/sharding/shard_manager.rs
mkrasnitski: I'm actually having a hard time figuring out how the manager was waiting for the shutdown to finish before. Do you know what it was doing?

mkrasnitski: Oh, seems like the runner was calling `self.manager.shutdown_finished()` which would make the manager send a message to itself through a dedicated shutdown channel saying the shutdown was done. Meanwhile inside `ShardManager::shutdown`, the recv end of the channel was waiting for the message to come in before rebooting the shard.

src/gateway/sharding/shard_manager.rs
GnomedDev: We should follow File IO stuff, have a `close` or `shutdown_all` method which is async and has error handling but have a Drop impl which just spawns shutdown_all and hope it completes.

src/gateway/sharding/shard_runner.rs
GnomedDev: A user could loop over the runners reporting statuses and forget to drop the lock before sending the messages.

src/gateway/sharding/shard_queue.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// filled and a maximally-sized batch can be popped, otherwise will return `None`.
+    ///
+    /// If the queue is _not_ in concurrent mode, this will return `Some` as long as at least one
+    /// bucket is filled.
+    pub fn pop_batch(&mut self) -> Option<Vec<ShardId>> {
+        if self.concurrent && !self.buckets.iter().all(|b| !b.is_empty()) {
+            None
+        } else {
+            Some(self.buckets.iter_mut().filter_map(VecDeque::pop_front).collect())
+        }
```
GnomedDev, line 54: Performance nit: This currently iterates twice when concurrent is false, when it could just do the one iteration and check if the Vec is empty.

mkrasnitski: How would I do this without actually popping the buckets? If I'm iterating only once, I can't know whether or not to pop from a given bucket because a later bucket might not be filled yet.

GnomedDev: I'm silly, I misread how the logic flows when concurrent is false.

src/gateway/sharding/shard_manager.rs
mkrasnitski: Ok, I was able to simplify the process of booting and restarting shards by a fair bit. Added a `ShardManagerMesssage::Boot` variant that the runner sends after it finishes shutting down. Also, I realized the `PollShardQueue` variant is not strictly needed, and removing it doesn't change control flow nor performance. Maybe you could cause fewer linear scans of the ShardQueue by avoiding checking for a batch every time a new shard is queued, but the queue is small anyways so it doesn't matter.

src/gateway/sharding/shard_manager.rs
mkrasnitski: Well, sending a shutdown message is sync already with `unbounded_send`, and the error handling is performed on the runner side, which will result in logging if errors occur. Is this sufficient you think?

src/gateway/sharding/shard_runner.rs
mkrasnitski: Do you think this is likely enough that the Mutexes should remain async?

GnomedDev: Nope, that's why I said to make them sync lol

src/gateway/sharding/shard_manager.rs
GnomedDev: This is sufficent, a future PR could add a shutdown_all which waits for a response.

GnomedDev (approved):
Okay, I am comfortable with this being merged now and getting some testing on next especially with follow up changes on the way. 

People like me with big bots reliant on next may want to be ready to roll back after this PR as it is likely to cause issues that code review and small scale testing cannot find.

arqunis (approved):

PR #3112, opened 2025-02-14
Rebase `next` on `current`

mkrasnitski:
diff: https://gist.github.com/mkrasnitski/9931c7d9d99da23c9e4fa38d2c84f7dc


jamesbt365:
When MessageSnapshot was rebased last time, it wasn't changed to use `FixedString` and `FixedArray` instead of `String` and `Vec`, these can be committed directly if you don't want to do them here though.

mkrasnitski:
I think it's fine as a followup PR

PR #3113, opened 2025-02-14
Use `foldhash` in `ExtractMap<K, V>` within model/cache instead of `std`'s ` hasher

xacrimon:
This PR replaces `std` hashers with the `foldhash` variant optimized for hash-based datastructures. It doesn't make a whole lot of sense to use the `std` hasher here since the keys aren't user controlled and follow a known pattern that prevents any sort of HashDOS type of attack. This is significantly faster than `std` for our keys and due to reduced size of the hasher factory type, reduces the size of any type containing `ExtractMap`'s such as `Guild`, which goes from 708 to 648 bytes.


mkrasnitski:
Writing out the explicit random state parameter each time is a bit noisy. Could you add a typedef to `src/internal/prelude.rs` that assigns it once and then the whole lib can use it?

xacrimon:
@mkrasnitski gnome suggested I move the hasher wrapper in cache into model and use that instead, going to try that.

GnomedDev:
Closing due to being an inactive draft. Let me know if you want this re-opened.

PR #3114, opened 2025-02-14
Avoid temporarily deserializing gateway messages to an untyped enum+map tree

xacrimon:
Avoid temporarily deserializing gateway messages to a `serde_json::Map<String, Value>` before typed deserialization to an `Event`. Previously, this meant the creation of a serde_json::Value tree, causing creation and immediately after the destruction of upwards of hundreds to thousands of owned strings and btreemaps for every handled gateway event.

### DIFF
--- a/src/gateway/sharding/mod.rs
+++ b/src/gateway/sharding/mod.rs
@@ -447,7 +447,7 @@ impl Shard {
         match event {
             Ok(GatewayEvent::Dispatch {
                 seq,
-                event
+                event,
             }) => self.handle_gateway_dispatch(seq, &event).map(|e| Some(ShardAction::Dispatch(e))),
             Ok(GatewayEvent::Heartbeat) => {
                 info!("[{:?}] Received shard heartbeat", self.shard_info);
--- a/src/gateway/sharding/mod.rs
+++ b/src/gateway/sharding/mod.rs
@@ -47,7 +47,6 @@ use std::time::{Duration as StdDuration, Instant};
 #[cfg(any(feature = "transport_compression_zlib", feature = "transport_compression_zstd"))]
 use aformat::aformat_into;
 use aformat::{aformat, ArrayString, CapStr};
-use serde::Deserialize;
 use tokio_tungstenite::tungstenite::error::Error as TungsteniteError;
 use tokio_tungstenite::tungstenite::protocol::frame::CloseFrame;
 #[cfg(feature = "tracing_instrument")]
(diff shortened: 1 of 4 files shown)


src/gateway/sharding/mod.rs
```diff
@@ -749,8 +740,8 @@ async fn connect(base_url: &str, compression: TransportCompression) -> Result<Ws
     WsClient::connect(url, compression).await
 }
 
-fn deserialize_and_log_event(map: JsonMap, original_str: &str) -> Result<Event> {
-    Event::deserialize(Value::Object(map)).map_err(|err| {
+fn deserialize_and_log_event(event: &[u8]) -> Result<Event> {
+    serde_json::from_slice(event).map_err(|err| {
         let err = serde::de::Error::custom(err);
```
mkrasnitski:
```suggestion
```
For some reason we're round-tripping `serde_json::Error` through `Error::custom`, which serde_json actually accounts for internally by parsing strings that look like the output of the `Display` impl. We can remove this line without changing the log output in any way.

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fa6d353e7b6cace83719a17e07d80318)

mkrasnitski (changes requested):
Overall this is definitely an improvement. It does suck that we're cloning the json data once, but for now we have to defer deserialization of `Event` in order to process the value of `seq` even if we fail to parse the event.

This deserves a rework but for now it's fine.

src/model/event.rs
```diff
@@ -948,30 +947,47 @@ pub enum GatewayEvent {
...
+            op: Opcode,
+            #[serde(rename = "s")]
+            seq: Option<u64>,
+            #[serde(rename = "d")]
+            data: &'a RawValue,
+            #[serde(rename = "t")]
+            ty: Option<&'a str>,
+        }
+
+        let raw: GatewayEventRaw<'_> = Deserialize::deserialize(deserializer)?;
```
mkrasnitski:
```suggestion
        let raw = GatewayEventRaw::deserialize(deserializer)?;
```
More readable/simpler.

xacrimon: fixed

src/gateway/sharding/mod.rs
xacrimon: Thank you! Fixed now

xacrimon:
@mkrasnitski ready for another look now

mkrasnitski:
Can you rebase onto latest `next`?

xacrimon:
rebased

mkrasnitski (approved):

arqunis (approved):

PR #3119, opened 2025-02-16
Avoid deferring gateway event deserialization

mkrasnitski:
Followup to #3114.

By using an intermediate untagged enum with an `Unknown` variant, we can process sequence numbers for unknown events without cloning the json payload.


GnomedDev:
@mkrasnitski  You can probably solve the typesize error using `typesize(with = "")` and providing a function with the signature `fn(Box<str>) -> usize` returning `str::len`.

GnomedDev (approved):

arqunis (approved):

PR #3121, opened 2025-02-20
Fix ShardQueue batch size bug

mkrasnitski:
If the last batch to be started was not of maximum size, it wouldn't get popped off the queue. Instead, we should just pop off the queue periodically to ensure that we fully flush it. On client startup, we queue all shards before popping the first batch, so suboptimal batch sizes are not a concern. Now, the queue also doesn't have to worry about "concurrent mode" and saturating the size of its batches.

### DIFF
--- a/src/gateway/sharding/shard_manager.rs
+++ b/src/gateway/sharding/shard_manager.rs
@@ -141,16 +141,18 @@ impl ShardManager {
         shard_init: u16,
         shard_total: NonZeroU16,
     ) -> Result<(), GatewayError> {
-        self.initialize(shard_index, shard_init, shard_total).await;
+        self.initialize(shard_index, shard_init, shard_total);
         loop {
             if let Ok(Some(msg)) =
                 timeout(self.wait_time_between_shard_start, self.manager_rx.next()).await
             {
                 match msg {
-                    ShardManagerMessage::Boot(shard_id) => self.boot(shard_id, false).await,
+                    ShardManagerMessage::Boot(shard_id) => self.queue_for_start(shard_id),
                     ShardManagerMessage::Quit(err) => return Err(err),
                 }
             }
+            let batch = self.queue.pop_batch();
+            self.checked_start(batch).await;
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/gateway/sharding/shard_manager.rs
```diff
@@ -141,16 +141,18 @@ impl ShardManager {
...
                 timeout(self.wait_time_between_shard_start, self.manager_rx.next()).await
             {
                 match msg {
-                    ShardManagerMessage::Boot(shard_id) => self.boot(shard_id, false).await,
+                    ShardManagerMessage::Boot(shard_id) => self.queue_for_start(shard_id),
                     ShardManagerMessage::Quit(err) => return Err(err),
                 }
             }
+            let batch = self.queue.pop_batch();
+            self.checked_start(batch).await;
```
GnomedDev, line 155: Currently this is fine, but seems fragile. I worry that this will end up being called too often in future, if more ShardManagerMessages are added.

src/gateway/sharding/shard_queue.rs
```diff
@@ -9,7 +9,6 @@ use crate::internal::prelude::*;
 #[must_use]
 pub struct ShardQueue {
     buckets: FixedArray<VecDeque<ShardId>, u16>,
-    concurrent: bool,
```
GnomedDev, line 12: I like the removal of `concurrent` though, it seemed weird previously.

GnomedDev (changes requested):

src/gateway/sharding/shard_manager.rs
mkrasnitski: Could you elaborate? I could put a check for if the queue is empty, which would require a linear scan of the buckets but would save creating an empty `Vec` each time.

Btw, how should the check be implemented? Should `pop_batch` return `None` if the queue is empty, or should I add a `is_empty` method? Empty batches are already accounted for in `checked_start` so it's a matter of API design.

GnomedDev: Currently, this code will start a new batch of shards if it receives a message or the time between shard starts has been hit.

Taking a look at this again, it is problematic today, as if someone sends `ShardManagerMessage::Boot` too quickly it will start new batches of shards faster than the wait time, and therefore hit rate limit and be disconnected.

mkrasnitski: Well, `checked_start` already takes care of sleeping between shard batches. If restarts come in quickly, then they'll be caught and spaced out accordingly.

GnomedDev: Ah, that's what I was missing. This is fine then.

GnomedDev (approved):

arqunis (approved):

Issue #2963, opened 2024-08-31
Stop expecting top level components to always be ActionRow


GnomedDev:
This promise is being removed from the discord documentation without a breaking release, so we must start filtering for `ActionRow`s on current and change the public API on next.

jamesbt365:
This is done on current by #3127 and the componentsV2 pr ultimately changes the public api (#3123)

PR #3123, opened 2025-02-21
ComponentsV2 support

jamesbt365:
Opened early, but getting a PR open sooner rather than later doesn't hurt.

TODO:
- [x] Base typing
- [x] Test thoroughly to make sure everything deserializes
- [x] Assure that I'm typing this correctly and not making things vague where it shouldn't be
- [x] Make `Component` & others more resilient to discord changes
  - Right now any change discord makes to components could knock out our Message deserialization.
- [x] Mark under `unstable` feature flag
  - This is still getting breaking changes and we should wait until rollout before marking this as stable.
- [x] Use FixedString everywhere relevant.
- [x] Add relevant builders
- [x] Document, including updating limits.

Additionally, we *really* need to PR to current and modify the deserialization to not fail if the top level component is not an ActionRow or it will not deserialize. (See #2963) (Done in #3127)

Once merged, as usual a PR to poise@serenity-next to add support over there too!


jamesbt365:
Clippy failure does not seem like my fault? But this should be complete now.

jamesbt365:
I forgot the new field for webhooks, will add later.

jamesbt365:
> I forgot the new field for webhooks, will add later.

Seems like this applies prior to ComponentsV2 too, though the restriction here is that they can only use **non-interactive** components on non application owned webhooks when sending this query param.

This can be implemented separately as this works on link buttons from my understanding and the information on https:discord/discord-api-docs#7368.

So... this PR should be done completely and now just needs a review (and fixing whatever funky stuff is going on with our CI?)

jamesbt365:
rebased for review reasons, but due to a timecrunch i haven't added V2 support on webhooks yet, will do in a couple hours when i get back.

jamesbt365:
> rebased for review reasons, but due to a timecrunch i haven't added V2 support on webhooks yet, will do in a couple hours when i get back.

Disregard this comment, i didn't have time to properly check but what was needed was added cleanly in the rebase.

jamesbt365:
Rebased, history will be a little weird because i removed the macro and stuff entirely from the history, doesn't really matter though, makes even sense while going through.

Will refer to unofficial docs and get this in as best of a state I can before monday, when the draft PR for docs will drop.

jamesbt365:
A little note to myself (i'm about to test if i messed this up) i'm not sure if putting an enum over the media items is required/better, nullablity is not confirmed completely yet and we won't know until the docs on monday, i'll go over everything one last time then and hopefully nail it.

jamesbt365:
I have gone ahead and updated it to match official documentation in discord/discord-api-docs#7487, I am technically doing undocumented stuff right now as for whatever reason they didn't add the extra fields for the unfurled media item, not sure if thats intentional or not yet.

jamesbt365:
This feature has now been officially released! I (hopefully) have nailed it now and it should be ready for any reviews before merge!

I did remove the loading state in the most recent commit, we don't reflect that for embeds (which i think have one too, under the hood) as the docs do not show this.

I will leave it under the unstable feature gate until a little bit after merge (even though its on the `next` branch) as a precaution as I don't know how stable the de-serialization and stuff is. I've been running it on my bot for a few days and it has been stable though!

GnomedDev:
Can we remove the unstable gate? It is expected for next to be unstable, the gate is just for undocumented features.

jamesbt365:
> Can we remove the unstable gate? It is expected for next to be unstable, the gate is just for undocumented features.

Done as its documented now, hopefully I removed all the flags lol

src/builder/create_components.rs
```diff
@@ -4,6 +4,15 @@ use serde::Serialize;
...
 
+#[derive(Clone, Debug)]
+struct StaticU8<const VAL: u8>;
+
+impl<const VAL: u8> Serialize for StaticU8<VAL> {
+    fn serialize<S: serde::Serializer>(&self, ser: S) -> Result<S::Ok, S::Error> {
+        ser.serialize_u8(VAL)
+    }
+}
+
```
mkrasnitski, line 15: Shouldn't uses of this be replaced with `ComponentType`?

GnomedDev: This avoids having to store the ComponentType in the builder and handle matching on it when serializing. It's a small optimisation.

mkrasnitski: The matching actually isn't a problem because `enum_number!` makes `ComponentType` a newtype around u8, and the variants are actually associated constants. Also, the existing component types already store `ComponentType` inside, so whichever approach we choose we should be consistent.

src/model/application/component.rs
```diff
@@ -19,10 +20,258 @@ enum_number! {
...
+    /// The spacing of the separator.
+    pub spacing: Option<SeparatorSpacingSize>,
+}
+
+enum_number! {
+    /// The size of a separator component.
+    #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Deserialize, Serialize)]
+    #[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
+    #[non_exhaustive]
+    pub enum SeparatorSpacingSize {
```
DPlayer234, line 230: This seems like it's a duplicate of the `Spacing` struct in `create_components.rs`. I don't see an immediate reason why these need to be distinct, but I may be missing something.

src/builder/create_components.rs
```diff
@@ -47,6 +56,462 @@ impl serde::Serialize for CreateActionRow<'_> {
...
+    Container(CreateContainer<'a>),
+}
+
+/// A builder to create a section component, supports up to a max of **3** components with an
+/// accessory.
+#[derive(Clone, Debug, Serialize)]
+#[must_use]
+pub struct CreateSection<'a> {
+    #[serde(rename = "type")]
+    kind: StaticU8<9>,
```
DPlayer234, line 116: This may have been intentionally omitted, but every component type is documented to have an optional integer `id` field (documented as being limited to 32-bits). I'll just note this once here, but this affects every `Create*` for a component, as well as the "received" versions in `model/application/component.rs`.

DPlayer234 (commented):
I personally only see some minor issues, if these aren't intentional. There are a few other things I've considered (i.e. convenience functions to skip needing explicit construction of `CreateUnfurledMediaItem` for `CreateThumbnail` and such), but that may be better motivated by necessity and would be appropriate in a separate PR.

src/builder/create_components.rs
jamesbt365: I intentionally omitted this because the existing components did not already have them and they can be PRed to current in the future, as its a non breaking change.

src/model/application/component.rs
jamesbt365: There isn't really a reason, I'll combine them if its wanted but maybe Discord will do something in the future with this and then it would be a breaking change to do, idk what the rest of the lib looks like in this regard but I'll do as a maintainer asks.

arqunis (approved):

jamesbt365:
Merging due to approving review, also tested it myself for about a month and other users have also been using it for a while and have not reported any issues. I'll get a PR ready for poise soon:tm: (or somebody else can do that)

PR #3127, opened 2025-02-28
filter component deserialization for messages

jamesbt365:
Will prevent components v2 from causing messages to not deserialize, it is important that a new serenity release is released before the 2nd week of March as that is when these components will be enabled.

Not the most familiar with doing stuff like this so this may not be optimal, other implementations are welcome.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -176,7 +176,7 @@ where
         where
             A: serde::de::SeqAccess<'de>,
         {
-            let mut components = Vec::new();
+            let mut components = Vec::with_capacity(seq.size_hint().unwrap_or_default());
 
             while let Some(raw) = seq.next_element::<&serde_json::value::RawValue>()? {
                 // We deserialize only the `kind` field to determine the component type.
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -185,9 +185,7 @@ where
                 let min_component =
                     MinComponent::deserialize(raw).map_err(serde::de::Error::custom)?;
 
-                // This is an action row, the only top level supported component in serenity at this
-                // time, we only have deseliazed the kind until now to avoid parsing unsupported
-                // components.
+                // Action rows are the only top level component supported in serenity at this time.
                 if min_component.kind == 1 {
                     match ActionRow::deserialize(raw) {
                         Ok(valid_row) => components.push(valid_row),
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -179,11 +179,15 @@ where
             let mut components = Vec::new();
 
             while let Some(raw) = seq.next_element::<&serde_json::value::RawValue>()? {
+                // We deserialize only the `kind` field to determine the component type.
+                // We later use this to check if its a supported component before deserializing the
+                // entire payload.
                 let min_component =
                     MinComponent::deserialize(raw).map_err(serde::de::Error::custom)?;
 
                 // This is an action row, the only top level supported component in serenity at this
-                // time.
+                // time, we only have deseliazed the kind until now to avoid parsing unsupported
+                // components.
                 if min_component.kind == 1 {
                     match ActionRow::deserialize(raw) {
                         Ok(valid_row) => components.push(valid_row),
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -179,13 +179,13 @@ where
             let mut components = Vec::new();
 
             while let Some(raw) = seq.next_element::<&serde_json::value::RawValue>()? {
-                let min_component: MinComponent =
-                    serde_json::from_str(raw.get()).map_err(serde::de::Error::custom)?;
+                let min_component =
+                    MinComponent::deserialize(raw).map_err(serde::de::Error::custom)?;
 
                 // This is an action row, the only top level supported component in serenity at this
                 // time.
                 if min_component.kind == 1 {
-                    match serde_json::from_str::<ActionRow>(raw.get()) {
+                    match ActionRow::deserialize(raw) {
                         Ok(valid_row) => components.push(valid_row),
                         Err(_) => {
                             tracing::debug!("Failed to deserialize ActionRow, malformed data.");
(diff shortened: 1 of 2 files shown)


src/model/channel/message.rs
```diff
@@ -151,6 +151,41 @@ pub struct Message {
...
+            formatter.write_str("a sequence of ActionRow elements")
+        }
+
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: serde::de::SeqAccess<'de>,
+        {
+            let mut components = Vec::new();
+            while let Some(action_row) = seq.next_element::<serde_json::Value>()? {
+                match serde_json::from_value::<ActionRow>(action_row) {
```
GnomedDev: Using serde_json::Value here seems like a bad idea, is there a way to use `flatten` and `RawValue` to only deserialize the type field and whitelist type values?

jamesbt365: Should be better now?

src/model/channel/message.rs
```diff
@@ -151,6 +151,63 @@ pub struct Message {
...
+
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: serde::de::SeqAccess<'de>,
+        {
+            let mut components = Vec::new();
+
+            while let Some(raw) = seq.next_element::<&serde_json::value::RawValue>()? {
+                let min_component: MinComponent =
+                    serde_json::from_str(raw.get()).map_err(serde::de::Error::custom)?;
```
```suggestion
                let min_component =
                    MinComponent::deserialize(raw).map_err(serde::de::Error::custom)?;
```

src/model/channel/message.rs
```diff
@@ -151,6 +151,63 @@ pub struct Message {
...
+
+                // This is an action row, the only top level supported component in serenity at this
+                // time.
+                if min_component.kind == 1 {
+                    match serde_json::from_str::<ActionRow>(raw.get()) {
```
```suggestion
                    match ActionRow::deserialize(raw) {
```

src/model/channel/message.rs
```diff
@@ -151,6 +151,63 @@ pub struct Message {
...
+            A: serde::de::SeqAccess<'de>,
+        {
+            let mut components = Vec::new();
+
+            while let Some(raw) = seq.next_element::<&serde_json::value::RawValue>()? {
+                let min_component: MinComponent =
+                    serde_json::from_str(raw.get()).map_err(serde::de::Error::custom)?;
+
+                // This is an action row, the only top level supported component in serenity at this
+                // time.
```
mkrasnitski: Could you elaborate that we grab just the `ComponentKind` first to avoid parsing the entire component if it's unsupported?

mkrasnitski (changes requested):

src/model/channel/message.rs
jamesbt365: Sure, please check out d868cdf3c7f598177a8204bbe5a6b68a3404ac33 and let me know what you think.

mkrasnitski: The comment on the line containing `MinComponent::deserialize` is good. The following comment doesn't need to reiterate that we deserialize only the type. I'd suggest writing just:

```
// Action rows are the only top level component supported in serenity at this time.
```

jamesbt365: Resolved.

src/model/channel/message.rs
```diff
@@ -151,6 +151,65 @@ pub struct Message {
...
+                // entire payload.
+                let min_component =
+                    MinComponent::deserialize(raw).map_err(serde::de::Error::custom)?;
+
+                // Action rows are the only top level component supported in serenity at this time.
+                if min_component.kind == 1 {
+                    match ActionRow::deserialize(raw) {
+                        Ok(valid_row) => components.push(valid_row),
+                        Err(_) => {
+                            tracing::debug!("Failed to deserialize ActionRow, malformed data.");
```
GnomedDev: Should this be ignoring errors?

src/model/channel/message.rs
```diff
@@ -151,6 +151,65 @@ pub struct Message {
...
+
+        fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            formatter.write_str("a sequence of ActionRow elements")
+        }
+
+        fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+        where
+            A: serde::de::SeqAccess<'de>,
+        {
+            let mut components = Vec::new();
```
GnomedDev: Can you use `Vec::with_capacity` with `SeqAccess::size_hint`?

GnomedDev (commented):
Looks good, just a couple questions

src/model/channel/message.rs
jamesbt365: I think its fine? its logged and if we return out and fail the deserialization then the entire message won't deserialize which is what we are trying to avoid here.

Unless I'm mistaken and theres a better way.

src/model/channel/message.rs
jamesbt365: See 7481a6958fa9d2fcc6137f9320428099ee3f2c65.

src/model/channel/message.rs
mkrasnitski: Do you expect top-level action rows to occasionally fail to deserialize once Components V2 are introduced? If not, then the data should be parsed the same as before. Something like:

```rust
components.push(ActionRow::deserialize(raw).map_err(serde::de::Error::custom)?);
```

jamesbt365: > Do you expect top-level action rows to occasionally fail to deserialize once Components V2 are introduced?

They shouldn't, they haven't changed anything about ActionRows and the allowed component types will remain the same, but I'm not sure if this assumption will change.

The way it is currently should be safer but apply the suggestion is fine by me if you want it to.

mkrasnitski: Given that the format hasn't changed and Components V2 is shipping soon, they likely won't break it any more in the near future. I'd say the suggestion is fine to apply.

jamesbt365: Applied in 775a7f30e3a630a591a955dc9c92fec89d2029ed

mkrasnitski (approved):
LGTM

PR #3120, opened 2025-02-19
Remove `ShardMessenger` and clean up `ShardRunner`

mkrasnitski:
Follow up to #3108.

Inlines the fields and methods of `ShardMessenger` into `Context` and changes collectors to take `&Context`. Also, removes the `ShardRunnerMessage::{Close, Message}` variants, adds an `UpdateVoiceState` variant (songbird will want to use this) and combines the 3 presence-related variants into one. Plus some miscellaneous cleanup of the ShardRunner code.

PR #3310, opened 2025-03-07
Expose `ShardManager::manager_tx` field for more flexible control flow

mkrasnitski:
This allows the channel to be cloned by something other than the `Client`, (e.g. poise or any other framework that gets a client from `Framework::init`), so that messages can be sent directly to the manager.

In particular, `ShardManagerMessage::Quit` now holds a `Result` instead of just an error. This enables things like shutdown commands to send `Ok(())` to the manager to cleanly close the connection.


mkrasnitski:
Supeseded by #3120.

PR #3311, opened 2025-03-10
Avoid fully parsing Dispatch events twice

DPlayer234:
Implements `Deserializer` for the internal `GatewayEventRaw` to avoid having to parse the entire gateway event payload a second time.

This is almost certainly a minor difference in perf still. Someone should evaluate whether this is worth keeping around.

### DIFF
--- a/src/model/event.rs
+++ b/src/model/event.rs
@@ -4,9 +4,10 @@
 //! Discord documentation for the event.
 
 use serde::Serialize;
-use serde::de::Error as DeError;
+use serde::de::{DeserializeSeed, Error as DeError, IntoDeserializer, MapAccess};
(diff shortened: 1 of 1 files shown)


mkrasnitski:
I don't think this is worth the extra complexity. Deserializing to `&RawValue` is incredibly cheap because it's just a wrapper type around `str`, so no actual parsing is performed. Given that `GatewayEventRaw::data` is a raw value, it means that the gateway event payload is really "parsed" just once, when it's needed for dispatch.

DPlayer234:
It still ends up needing to parse the JSON so it knows where the object ends. That said, I do broadly agree with you. Just thought it'd be worth bringing up, since it was quick to prototype.

PR #3312, opened 2025-03-10
Fix top-level component deserialization for messages

DPlayer234:
`RawValue` can't be used as the outer deserialization doesn't use it. Instead, use `JsonMap` again.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ zerofrom = "=0.1.5"
 
 # Required dependencies
 bitflags = "2.4.2"
-serde_json = { version = "1.0.108", features = ["raw_value"] }
+serde_json = "1.0.108"
 async-trait = "0.1.74"
 tracing = { version = "0.1.40", features = ["log"] }
 serde = { version = "1.0.192", features = ["derive"] }
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -22,7 +22,7 @@ use crate::gateway::ShardMessenger;
 #[cfg(feature = "model")]
 use crate::http::{CacheHttp, Http};
 use crate::model::prelude::*;
-use crate::model::utils::{discord_colours, StrOrInt};
+use crate::model::utils::{deserialize_val, discord_colours, StrOrInt};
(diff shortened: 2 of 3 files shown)


jamesbt365:
...??? Why?

This works just fine though? Tested it, components deserialize exactly how I expect them to.

jamesbt365:
image

DPlayer234:
When I tried out the latest current commit, any component interactions and message creates with components will fail to deserialize and end up as Unknown events:

```
[DEBUG serenity::client::dispatch] An unknown event was received: UnknownEvent { kind: "INTERACTION_CREATE", value: Object { ...
```

On next, the same commit is broken only for interactions, but for the same reason.

Message updates "work", but that seems to be because `MessageUpdateEvent` doesn't use the new logic for the components.

jamesbt365:
> Message updates "work", but that seems to be because MessageUpdateEvent doesn't use the new logic for the components.

I guess that was missed, ideally something needs to be done about that.

jamesbt365:
> When I tried out the latest current commit, any component interactions and message creates with components will fail to deserialize and end up as Unknown events:

Haven't tested with component interactions because i didn't think it would affect those but messages... work fine for me?

mkrasnitski:
I tested the new logic with `e17_message_components` and can confirm the deserialization fails. Also, besides `MessageUpdateEvent`, there's also `ModalInteractionData::components` which would need to use the new logic as well.

DPlayer234:
Just a few additional notes from me:
- Against next, message creates work (but component interactions still don't).
- With #3313, next should work either way (so if both are merged, this PR shouldn't be rebased onto next).
- The alternative is to use `RawValue` instead of `JsonMap` everywhere on current, effectively backporting those changes.

mkrasnitski:
I'd say this should get merged. Backporting #3313 onto current will make the subsequent rebase a huge pain. Instead, we can just merge this and #3313 separately, and rebasing after that shouldn't be a problem.

mkrasnitski (approved):
Unfortunate that we're losing some perf with this, but avoiding it is more trouble than it's worth, at least on `current`.

arqunis (approved):

PR #3316, opened 2025-03-10
[UNTESTED]: deserialize components properly on MessageUpdateEvent

jamesbt365:
I haven't tested this as a i don't really have time now, but this should work to deserialize components properly on `MESSAGE_UPDATE`?

### DIFF
--- a/src/model/utils.rs
+++ b/src/model/utils.rs
@@ -460,7 +460,7 @@ pub fn optional_deserialize_components<'de, D>(
 where
     D: Deserializer<'de>,
 {
-    Ok(Some(deserialize_components(deserializer)?))
+    deserialize_components(deserializer).map(Some)
 }
 
 // Custom deserialize function to deserialize components safely without knocking the whole message
--- a/src/model/utils.rs
+++ b/src/model/utils.rs
@@ -450,3 +450,65 @@ where
         })
         .collect()
 }
+
+pub fn optional_deserialize_components<'de, D>(
+    deserializer: D,
+) -> Result<Option<Vec<ActionRow>>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let components = deserialize_components(deserializer)?;
+    if components.is_empty() {
+        Ok(None)
+    } else {
+        Ok(Some(components))
+    }
+}
+
+// Custom deserialize function to deserialize components safely without knocking the whole message
+// out when new components are found but not supported.
+pub fn deserialize_components<'de, D>(deserializer: D) -> Result<Vec<ActionRow>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    struct ComponentsVisitor;
+
+    impl<'de> Visitor<'de> for ComponentsVisitor {
+        type Value = Vec<ActionRow>;
+
+        fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            formatter.write_str("a sequence of ActionRow elements")
+        }
+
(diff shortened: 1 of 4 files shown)


src/model/utils.rs
```diff
@@ -450,3 +450,65 @@ where
...
+                // entire payload.
+                let raw_kind =
+                    map.get("type").ok_or_else(|| DeError::missing_field("type"))?.clone();
+                let kind: i64 = deserialize_val(raw_kind)?;
+
+                // Action rows are the only top level component supported in serenity at this time.
+                if kind == 1 {
+                    let value = Value::from(map);
+                    components
+                        .push(ActionRow::deserialize(value).map_err(serde::de::Error::custom)?);
```
mkrasnitski:
```suggestion
                    components
                        .push(ActionRow::deserialize(value).map_err(DeError::custom)?);
```
Small nit.

src/model/utils.rs
```diff
@@ -450,3 +450,64 @@ where
...
+where
+    D: Deserializer<'de>,
+{
+    let components = deserialize_components(deserializer)?;
+    if components.is_empty() {
+        Ok(None)
+    } else {
+        Ok(Some(components))
+    }
+}
```
mkrasnitski: I tested a bit and seems like the gateway sends empty arrays as part of message update events. Right now they're preserved, but this would map them to `None`. I'd say this is probably unneeded, and the `MessageUpdateEvent::components` field can just be decorated with `#[serde(default, deserialize_with = "deserialize_components")]`.

jamesbt365: Oh right, this would not allow for the devs to know that the components were removed, would need to change the behaviour a little bit to match what is expected.

And would just using the deserialize_components one even work? I'm pretty sure it would complain because its trying to deserialize to Vec<T> and not Option<Vec<T>>

mkrasnitski: If you add `#[serde(default)]` then it won't complain if the `components` field is missing. Option is treated specially by `serde` where `Option<T>` normally just means `T | null`, and adding `default` also accounts for the field being missing.

jamesbt365: image

jamesbt365: I think the way forward would just be to change the function so that if components are null return null, and when empty do the empty collection and keep the extra function?

Might actually add another to handle this all correctly and keep duplication to a minimum? Sound correct?

mkrasnitski: Yeah, I had it wrong. I guess a dedicated function will serve.

jamesbt365: Still need to distinguish between None and empty though... currently don't do that and I  can't really map it out in my head on how to do that without duplicating extra code

jamesbt365: Because Discord now returns the whole message payload on message edits, this is unnecessary and can always return Some, fef3b0aa39221f13221f20188a027bc59ca24395 does that now.

src/model/utils.rs
```diff
@@ -450,3 +450,62 @@ where
...
+// A function used for deserializing components within a MessageUpdateEvent.
+// Due to discord now sending the whole message payload, we don't need to distinguish between None
+// and empty, as such we always return Some.
+pub fn optional_deserialize_components<'de, D>(
+    deserializer: D,
+) -> Result<Option<Vec<ActionRow>>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    Ok(Some(deserialize_components(deserializer)?))
```
mkrasnitski:
```suggestion
    deserialize_components(deserializer).map(Some)
```

mkrasnitski (approved):

arqunis (approved):

PR #3318, opened 2025-03-11
Omit `InnerId` from `Debug` representation of `*Id`

dis-archive:
They're mostly just noise, the library user doesn't care about the inner representation.

The debug representation goes from `GuildId(InnerId(1))` to `GuildId(1)`.


GnomedDev (approved):
Looks good

jamesbt365 (approved):

PR #3322, opened 2025-03-15
Change `CreateAttachment` to stream data from disk when possible

mkrasnitski:
Implements #2690, as a second attempt at #2774. Defers reading from paths/files until the data is actually needed, which improves the performance of multipart file uploads by streaming directly from disk instead of copying into memory first. Cases where the file is included as a base64 string in the json payload are unaffected since the data needs to be encoded and written into the payload.

### DIFF
--- a/src/builder/create_attachment.rs
+++ b/src/builder/create_attachment.rs
@@ -183,10 +183,15 @@ impl ImageData {
     }
 }
 
+#[derive(Clone, Debug, Serialize)]
+struct ExistingAttachment {
+    id: AttachmentId,
+}
+
 #[derive(Clone, Debug)]
-enum EditAttachmentsInner<'a> {
+enum NewOrExisting<'a> {
     New(CreateAttachment<'a>),
-    Existing(AttachmentId),
+    Existing(ExistingAttachment),
 }
 
 /// You can add new attachments and edit existing ones using this builder.
@@ -247,7 +252,7 @@ enum EditAttachmentsInner<'a> {
 #[derive(Default, Debug, Clone)]
 #[must_use]
 pub struct EditAttachments<'a> {
-    inner: Vec<EditAttachmentsInner<'a>>,
+    new_and_existing_attachments: Vec<NewOrExisting<'a>>,
 }
 
 impl<'a> EditAttachments<'a> {
@@ -271,7 +276,15 @@ impl<'a> EditAttachments<'a> {
     /// Discord will throw an error!**
     pub fn keep_all(msg: &Message) -> Self {
         Self {
-            inner: msg.attachments.iter().map(|a| EditAttachmentsInner::Existing(a.id)).collect(),
+            new_and_existing_attachments: msg
+                .attachments
+                .iter()
+                .map(|a| {
+                    NewOrExisting::Existing(ExistingAttachment {
+                        id: a.id,
+                    })
+                })
+                .collect(),
         }
     }
 
@@ -280,7 +293,9 @@ impl<'a> EditAttachments<'a> {
     ///
     /// Opposite of [`Self::remove`].
     pub fn keep(mut self, id: AttachmentId) -> Self {
-        self.inner.push(EditAttachmentsInner::Existing(id));
+        self.new_and_existing_attachments.push(NewOrExisting::Existing(ExistingAttachment {
+            id,
+        }));
         self
     }
 
@@ -289,9 +304,9 @@ impl<'a> EditAttachments<'a> {
     ///
     /// Opposite of [`Self::keep`].
     pub fn remove(mut self, id: AttachmentId) -> Self {
-        self.inner.retain(|a| match a {
-            EditAttachmentsInner::Existing(existing_id) => *existing_id != id,
-            EditAttachmentsInner::New(_) => true,
+        self.new_and_existing_attachments.retain(|a| match a {
+            NewOrExisting::Existing(a) => a.id != id,
+            NewOrExisting::New(_) => true,
         });
         self
     }
@@ -299,7 +314,7 @@ impl<'a> EditAttachments<'a> {
     /// Adds a new attachment to the attachment list.
     #[expect(clippy::should_implement_trait)] // Clippy thinks add == std::ops::Add::add
     pub fn add(mut self, attachment: CreateAttachment<'a>) -> Self {
-        self.inner.push(EditAttachmentsInner::New(attachment));
+        self.new_and_existing_attachments.push(NewOrExisting::New(attachment));
         self
     }
 
@@ -308,10 +323,10 @@ impl<'a> EditAttachments<'a> {
     /// this method can only be called once.
     #[cfg(feature = "http")]
     pub(crate) fn new_attachments(&mut self) -> Vec<CreateAttachment<'a>> {
-        self.inner
+        self.new_and_existing_attachments
             .iter()
             .filter_map(|attachment| {
-                if let EditAttachmentsInner::New(attachment) = &attachment {
+                if let NewOrExisting::New(attachment) = &attachment {
                     Some(attachment.clone())
                 } else {
                     None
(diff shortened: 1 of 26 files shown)


src/builder/create_attachment.rs
```diff
@@ -107,15 +137,10 @@ impl<'a> CreateAttachment<'a> {
...
-struct ExistingAttachment {
-    id: AttachmentId,
-}
-
 #[derive(Clone, Debug)]
-enum NewOrExisting<'a> {
+enum EditAttachmentsInner<'a> {
     New(CreateAttachment<'a>),
-    Existing(ExistingAttachment),
+    Existing(AttachmentId),
```
GnomedDev: Do these changes need to be done as part of this PR? If not, can this be split out to make this PR easier to review and test.

GnomedDev (changes requested):
I like this, but I have some concerns. Alongside the review comment, I dislike how a bunch of methods have been made `async` and falliable, instead of just taking the data.

Maybe we need to stop using CreateAttachment for these methods, taking a newtype around `String` if we want to enforce the base64 encoding with the type system? If so, please do this as a separate commit.

mkrasnitski:
We could do that, unfortunately the data needs to be encoded as a Data URL, not just raw base64, and therefore needs to have a valid prefix like `data:image/png;base64,`, so there's a tradeoff. We could certainly validate the string that gets passed in, but that would keep those methods fallible, though it would make them sync again.

GnomedDev:
The newtype can enforce the prefix as well, I'm fine with that, that keeps the builders infallible.

src/builder/create_attachment.rs
```diff
@@ -115,14 +148,19 @@ impl<'a> CreateAttachment<'a> {
 }
 
 #[derive(Clone, Debug, Serialize)]
-struct ExistingAttachment {
-    id: AttachmentId,
+#[serde(transparent)]
+pub struct ImageData(String);
```
GnomedDev: Can you add some documentation to this?

src/builder/create_attachment.rs
mkrasnitski: Sure, I can separate it out into a separate PR.

src/builder/create_attachment.rs
```diff
@@ -23,50 +29,45 @@ use crate::model::id::AttachmentId;
...
-                .file_name()
-                .ok_or_else(|| std::io::Error::other("attachment path must not be a directory"))?;
-
-            Ok(CreateAttachment::bytes(data, filename.to_string_lossy().into_owned()))
-        }
-
-        inner(path.as_ref()).await
+    /// Returns [`Error::Io`] if the path is not a valid file path.
+    pub fn path(path: impl AsRef<Path>) -> Result<Self> {
+        let path = path.as_ref().to_owned();
```
GnomedDev: This is not good, just take `PathBuf` if you need it or store the `&Path` in `Self`.

src/builder/create_attachment.rs
```diff
@@ -23,50 +29,45 @@ use crate::model::id::AttachmentId;
...
+        let filename = path
+            .file_name()
+            .ok_or_else(|| std::io::Error::other("attachment path must not be a directory"))?
+            .to_string_lossy()
+            .into_owned();
```
GnomedDev, line 54: Is it possible to reuse the `path` buffer, instead of reallocating here?

src/builder/create_attachment.rs
```diff
@@ -114,6 +144,45 @@ impl<'a> CreateAttachment<'a> {
...
+        let url = Url::parse(s).map_err(|_| UrlError::InvalidDataURI)?;
+
+        let err = Error::Url(UrlError::InvalidDataURI);
+        if url.scheme() != "data" {
+            return Err(err);
+        }
+
+        let Some((mimetype, encoding)) = url.path().split_once(';') else {
+            return Err(err);
+        };
```
GnomedDev: Since this is already doing manual parsing for the data URI, can we drop the usage of `Url`? It's quite an expensive type and requires reallocating the buffer which in past has been a performance problem right?

src/builder/create_attachment.rs
```diff
@@ -235,10 +304,9 @@ impl<'a> EditAttachments<'a> {
...
     /// Opposite of [`Self::keep`].
     pub fn remove(mut self, id: AttachmentId) -> Self {
-        #[expect(clippy::match_like_matches_macro)] // `matches!` is less clear here
         self.new_and_existing_attachments.retain(|a| match a {
-            NewOrExisting::Existing(a) if a.id == id => false,
-            _ => true,
+            NewOrExisting::Existing(a) => a.id != id,
+            NewOrExisting::New(_) => true,
         });
         self
```
GnomedDev, line 301: Why change this?

src/builder/create_attachment.rs
```diff
@@ -251,22 +319,19 @@ impl<'a> EditAttachments<'a> {
...
-                let cloned_attachment = CreateAttachment::bytes(
-                    std::mem::take(&mut attachment.data),
-                    attachment.filename.clone(),
-                );
-
-                files.push(cloned_attachment);
-            }
-        }
-        files
+    pub(crate) fn new_attachments(&self) -> Vec<CreateAttachment<'a>> {
```
GnomedDev, line 314: Can this take `self`, Vec::retain for the new attachments, then return that `Vec`? I don't see where it is being used without ownership.

GnomedDev (changes requested):
Looking much better, especially like the ImageData changes. Just a few more concerns though.

src/builder/create_attachment.rs
mkrasnitski: For some reason I had run into borrowck problems before, but no longer.

src/builder/create_attachment.rs
mkrasnitski: Unfortunately, `reqwest` expects a `Cow<'static, str>` for multipart filenames - also there would be self-reference if I store the path in the `CreateAttachment::data` field while storing a reference into it inside `CreateAttachment::filename`.

src/builder/create_attachment.rs
mkrasnitski: Taking `self` wouldn't be possible, as the builder gets passed to be serialized as part of the payload, and that actually requires that metadata for both new and existing attachments gets serialized as part of the body, whereas the file data for the new attachments are passed as part of the multipart upload.

src/builder/create_attachment.rs
mkrasnitski: I figured it was a little easier to read and gets rid of the `expect`.

src/builder/create_attachment.rs
mkrasnitski: This does mean that passing string literals will require the caller explicitly do a `.as_ref()`.

GnomedDev: I'm good with taking `impl AsRef<Path>` instead of `&Path`, the main concern was taking a reference to Path when a PathBuf is needed.

src/builder/create_attachment.rs
```diff
@@ -114,6 +139,39 @@ impl<'a> CreateAttachment<'a> {
...
+    ///
+    /// let s = "data:image/png;base64,R0lGODlhAQABAIAAAP///wAAACwAAAAAAQABAAACAkQBADs=";
+    /// assert!(ImageData::from_base64(s).is_ok());
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`Error::Url`] if the string is not a valid data URI. See the [Discord
+    /// docs](https://discord.com/developers/docs/reference#image-data).
+    pub fn from_base64(s: &str) -> Result<Self> {
```
GnomedDev: Can this be changed to either take `String` or take `impl Into<Cow<'a, str>>`?

GnomedDev (approved):
Looking good, once someone does some testing on this I'm good with merging.

src/builder/create_attachment.rs
mkrasnitski: Unfortunately that would cause lifetime problems since I'd be storing a reference to a `Path` that lives inside the current stack frame.

GnomedDev: Oh that's super annoying, but I don't see a nice workaround, good with merging like this.

src/builder/create_attachment.rs
mkrasnitski: How come? I would expect any type that implements `Into<Cow<'a, str>>` also implements `AsRef<str>`.

GnomedDev: Am I missing something? This is taking a borrowed string in, performing a check which is expected to always pass, then cloning the borrowed string into an owned string? I am saying take the owned string originally so you don't have to clone it.

mkrasnitski: Ah sorry, right.

mkrasnitski:
Tested this with message attachments and modifying a webhook's avatar, works fine in both cases.

PR #3329, opened 2025-03-25
Remove `FullEvent` macro

mkrasnitski:
The macro wasn't defining anything else besides the variants of `FullEvent`, so we can just use native syntax for that.

### DIFF
--- a/src/gateway/client/event_handler.rs
+++ b/src/gateway/client/event_handler.rs
@@ -14,17 +14,15 @@ use crate::model::prelude::*;
 pub trait EventHandler: Send + Sync {
     /// Checks if the `event` should be dispatched or ignored. Returns `true` by default.
     ///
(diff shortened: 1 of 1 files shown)


DPlayer234:
This is now missing `#[cfg_attr(not(feature = "unstable"), non_exhaustive)]` on each variant. I believe that was in part the reason why the macro wasn't immediately removed.

mkrasnitski:
I see. The attribute was added in #3324 to guard against breakage if a new field is added to a variant, now that event handling matches directly against `FullEvent`.

jamesbt365:
> I see. The attribute was added in #3324 to guard against breakage if a new field is added to a variant, now that event handling matches directly against `FullEvent`.

Yeah its just easier to define it this way in the macro, additionally, they'll be another line too if #3056 gets merged in, probably best to keep the macro instead of just repeating this every time.

PR #3331, opened 2025-03-26
Avoid extra Guild removals in Ready handler

GnomedDev:
This code was removing guilds from cache which the current shard is responsible for, but wasn't received in the unavailable payload.

This code path could only be hit under extreme circumstances, such as unsuccessful resumes, and since we cannot rely on cache consistency after an unsuccessful resume anyway this does more bad than good.

To add to the motivation for doing this, out of the libraries checked (twilight, discord.py, hikari) none do this either.

### DIFF
--- a/src/cache/event.rs
+++ b/src/cache/event.rs
@@ -1,5 +1,3 @@
-use std::collections::HashSet;
-
 use super::{Cache, CacheUpdate};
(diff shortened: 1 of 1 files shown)


jamesbt365 (approved):
LGTM

PR #3333, opened 2025-03-26
Avoid boxed event reallocations

GnomedDev:
Currently, Event is boxed during deserialisation only to be dereferenced in `Shard::handle_gateway_dispatch` and boxed again afterwards. This changes the box dance to dereference as late as possible, during cache updating, in hope that only fields matched on are memcopied.


jamesbt365 (approved):
I think this looks good, but i'm not the most familiar with allocation shenanigans!

PR #3345, opened 2025-04-14
Fix shard runner info

GnomedDev:
Previously, example 7 didn't work as `runner_info` would clone the hashmap before it had any entries. This solution goes back to how `ShardManager::runners` is in current, except using `DashMap` and `try_get_mut`.

### DIFF
--- a/src/gateway/client/context.rs
+++ b/src/gateway/client/context.rs
@@ -1,5 +1,6 @@
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 
+use dashmap::DashMap;
 use futures::channel::mpsc::UnboundedSender as Sender;
 
 #[cfg(feature = "cache")]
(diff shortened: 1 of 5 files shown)


src/gateway/client/context.rs
```diff
@@ -46,7 +47,8 @@ pub struct Context {
     pub http: Arc<Http>,
     #[cfg(feature = "cache")]
     pub cache: Arc<Cache>,
-    pub runner_info: Arc<Mutex<ShardRunnerInfo>>,
+    /// Metadata about the initialised shards, and their control channels.
+    pub runners: Arc<DashMap<ShardId, (ShardRunnerInfo, Sender<ShardRunnerMessage>)>>,
```
mkrasnitski, line 51: How come this is changed? The context is per-shard, so including info about other shards in the context feels strange. And exposing the sender here is also unnecessary because the context already has wrapper methods around `self.shard`.

GnomedDev: Because that would require wrapping the ShardRunnerInfo in Arc and another lock, and this is the most useful as users (including me) often want access to all the runners.

mkrasnitski: Do you think turning the tuple into a struct would make sense? I understand how access to all the other shard info is useful for calculating metrics, but access to all the other senders feels unlikely to be useful.

tazz4843: I do personally use the senders to set status on every shard for per-shard details. If there's another way to set status per-shard I'd be happy with that but I can't see that anywhere.

mkrasnitski: I see, that makes sense.

mkrasnitski (approved):

arqunis (approved):

PR #3353, opened 2025-04-21
Implement ShardManager::get_shutdown_trigger

GnomedDev:
This removes the uncallable public functions on ShardManager after the rework, and adds a `get_shutdown_trigger` function which allows for shutting the shard manager down after it has been mutably borrowed for the lifetime of the bot.

### DIFF
--- a/src/gateway/sharding/shard_manager.rs
+++ b/src/gateway/sharding/shard_manager.rs
@@ -118,6 +118,15 @@ impl ShardManager {
         }
     }
 
+    /// Retrieves a function which can be used to shut down the ShardManager later.
+    ///
+    /// This function will return `true` if the ShardManager has successfully been
+    /// notified to shut down, or false if it has already shut down and been dropped.
+    pub fn get_shutdown_trigger(&self) -> impl FnOnce() -> bool + Send + use<> {
+        let manager_tx = self.manager_tx.clone();
+        move || manager_tx.unbounded_send(ShardManagerMessage::Quit(Ok(()))).is_ok()
+    }
+
     /// The main interface for starting the management of shards. Initializes the shards by
     /// queueing them for starting, and then listens for [`ShardManagerMessage`]s in a loop.
     ///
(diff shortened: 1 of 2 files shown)


mkrasnitski:
Doesn't `Context::shutdown_all` fill the role that this does? I think calling that method should always be possible, right?

GnomedDev:
Nope, what about shutting down the bot cleanly on control C?

mkrasnitski:
> Nope, what about shutting down the bot cleanly on control C?

Do you mean shutting down the bot from within a signal handler?

GnomedDev:
Yep, like https://github.com/Discord-TTS/Bot/blob/7ebedabfaded165740725573a7057a4e318f0bf3/src/main.rs#L177-L184

src/gateway/sharding/shard_manager.rs
```diff
@@ -118,6 +118,15 @@ impl ShardManager {
...
 
+    /// Retrieves a function which can be used to shut down the ShardManager later.
+    ///
+    /// This function will return `true` if the ShardManager has successfully been
+    /// notified to shut down, or false if it has already shut down and been dropped.
+    pub fn get_shutdown_trigger(&self) -> impl FnOnce() -> bool + Send + use<> {
+        let manager_tx = self.manager_tx.clone();
+        move || manager_tx.unbounded_send(ShardManagerMessage::Quit(Ok(()))).is_ok()
+    }
+
```
mkrasnitski, line 129: Do you think exposing `self.manager_tx` is maybe a more flexible solution? This function kinda sticks out, and I don't really like the name.

GnomedDev: It would be more flexible, but it's exposing too much API to public consumers imo, leading to defensive or defective programming inside the currently quite simple `ShardManager`. This is the bare minimum interface that users (at least I) need, and future refactors/reworks cannot be merged if they cannot replicate this behavior.

fgardt (approved):

arqunis (approved):

Issue #3356, opened 2025-04-27
Tracking issue for reducing future sizes


GnomedDev:
Serenity's future sizes are quite large, leading to `clippy::large_futures` warnings and suboptimal performance.

This is a list of known problems leading to these large futures.

- The `Http` interface is forced to keep builders in futures even after serializing them, due to the `&impl serde::Serialize` signature.
- Some functions take ownership of a type, then only use a couple fields even though it's carrying over an await point.
- - This probably requires enabling `struct_lit_single_line` in rustfmt settings.
- *More to be found*

mkrasnitski:
What is the value of `future-size-threshold` that you set to produce warnings for you? Sticking with the default value produces none for me.

GnomedDev:
The default, but my message event handler is `tokio::join`ing 3 futures together all doing serenity stuff, which peaked at 19kb before #3357 and around 15k after.

PR #3357, opened 2025-04-27
Solve low hanging fruit causing large future sizes

GnomedDev:
See #3356.

This PR also removes `CreateMessage::reactions` as this brought my attention to it and it is bloating future sizes for all users while not following discord's API.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -4381,7 +4381,7 @@ impl Http {
     pub async fn request(&self, req: Request<'_>) -> Result<ReqwestResponse> {
         let method = req.method.reqwest_method();
         let response = if let Some(ratelimiter) = &self.ratelimiter {
-            ratelimiter.perform(req).await?
+            ratelimiter.perform(&req).await?
         } else {
             let request = req
                 .build(
--- a/src/http/multipart.rs
+++ b/src/http/multipart.rs
@@ -49,7 +49,8 @@ impl Multipart<'_> {
             },
             MultipartUpload::Attachments(attachment_files) => {
                 for (idx, file) in attachment_files.into_iter().enumerate() {
-                    multipart = multipart.part(format!("files[{idx}]"), file.into_part().await?);
+                    let part = file.into_part().await?;
+                    multipart = multipart.part(format!("files[{idx}]"), part);
                 }
             },
         }
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -75,6 +75,13 @@ impl<'a> Request<'a> {
         token: Option<&str>,
         proxy: Option<&str>,
     ) -> Result<ReqwestRequestBuilder> {
+        // Build the multipart first, as to keep future size small.
+        let multipart = if let Some(multipart) = self.multipart {
+            Some(multipart.build_form().await?)
+        } else {
+            None
+        };
+
         let mut path = self.route.path().into_owned();
 
         if let Some(proxy) = proxy {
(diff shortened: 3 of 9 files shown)


src/http/client.rs
```diff
@@ -4381,7 +4381,7 @@ impl Http {
     pub async fn request(&self, req: Request<'_>) -> Result<ReqwestResponse> {
         let method = req.method.reqwest_method();
         let response = if let Some(ratelimiter) = &self.ratelimiter {
-            ratelimiter.perform(req).await?
+            Box::pin(ratelimiter.perform(&req)).await?
```
mkrasnitski: I think boxing large futures does nothing to actually address the size problem. The data is just now on the heap and incurs an allocation.

src/http/multipart.rs
```diff
@@ -8,13 +8,17 @@ use crate::internal::prelude::*;
 
 impl CreateAttachment<'_> {
     async fn into_part(self) -> Result<Part> {
-        let mut part = match self.data {
+        #[rustfmt::skip]
+        let Self { filename, data, description } = self;
+        drop(description);
```
mkrasnitski: This should wait until after we've changed our rustfmt settings to include `struct_lit_single_line`.

src/http/request.rs
```diff
@@ -75,6 +75,13 @@ impl<'a> Request<'a> {
         token: Option<&str>,
         proxy: Option<&str>,
     ) -> Result<ReqwestRequestBuilder> {
+        // Build the multipart first, as to keep future size small.
+        let multipart = if let Some(multipart) = self.multipart {
+            Some(multipart.build_form().await?)
+        } else {
+            None
+        };
+
```
mkrasnitski, line 84: I'm confused here. The function call doesn't capture any locals, so why would it matter when it gets awaited? Or does it capture all of `self`?

src/http/client.rs
GnomedDev: This shrinks size of the future for the branch where a rate limiter is disabled, which is required for large bots using nirn/another proxy.

src/http/request.rs
GnomedDev: The function call doesn't use any of the locals, but the state machine for `Request::build` still needs to store the locals. By moving this back, we don't have to store the `path`/`builder`/`headers` in the state machine as there are no awaits after this.

src/http/client.rs
DPlayer234: I was curious what this would actually save. With the changes in the PR, this should really only reduce the size of the `Http::request` future by about 200 bytes -- while allocating almost 2kb for the ratelimiter case. That doesn't seem worth it.

Partial output with `Zprint-type-sizes`:
```
type: `{async fn body of http::client::Http::request()}`: 2264 bytes, alignment: 8 bytes
    variant `Suspend0`: 711 bytes // boxed ratelimiter await
    variant `Suspend1`: 2263 bytes // second largest await
type: `{async fn body of http::ratelimiting::Ratelimiter::perform()}`: 1744 bytes, alignment: 8 bytes
```

That said, why not put the ratelimiter behind the feature gate?

xacrimon: My impression with futures & async is that there are no sweeping giant single gains to be made. That the problems just come from a zillion smaller bits adding on top of eachother and you have to look at that and see what systematic changes can be made in many places, little by little.

xacrimon: That said, from my POV, this does not seem like the right solution. Rate limiting is very common, and we're going to pay a bunch of cost in the common case for a lot of people as a result. What do you instead think about doing some strategic cold-path future outlining/boxing inside of the rate limiter code? That definently has hot and cold paths.

arqunis (approved):

PR #3359, opened 2025-04-28
Support manual resuming

### LINKED ISSUE #3093: Manually resume
It would be nice to be able to save the current last seq id and session id to disk, then have serenity try to send a `Resume` instead of `Identify` on startup to make sure that no messages were missed when restarting bots.

This was asked in #2932 but I'm not sure why it was closed when `Resume` exists.

GnomedDev:
Closes #3093

Draft as based on #3353 and unfinished.


GnomedDev:
I don't currently have the bandwidth to work on this, but the general problem that needs fixing is ensuring that stale caches are not reloaded.

PR #3360, opened 2025-05-01
Specialize BuildHasher and switch hashers in ExtractMap and HashMap

xacrimon:
This PR specializes our `BuildHasher` and `Hasher` to provide the integer hash types, as most of our map usage in serenity uses integer keys, this allows for a far simpler and smaller codepath. Switching the hasher everywhere reduces the size of some types quite significantly as every single map in serenity is now 40 bytes instead of 48 bytes.

On my MacBook, this reduces the binary size significantly. For the e01 example, it is 550 KiB in release saved and 1.9 MiB in dev.


GnomedDev:
Why is this draft? I'm happy for this to merged in after a rebase.

Issue #3341, opened 2025-04-08
Get Guild Ban endpoint not supported


cheesycod:
Seems like serenity does not support Get Guild Ban endpoint yet

PR #3362, opened 2025-05-07
Add support for GET method to Route::GuildBan (#3341)

AshleyBilbrey:
This pull request allows Serenity to get a single `model::guild::Ban` utilizing the GET `/guilds/{guild.id}/bans/{user.id}` route.
Discord Documentation: https://discord.com/developers/docs/resources/guild#get-guild-ban

This should resolve issue #3341.

Currently the only option to get a single ban for a specific user is to use `get_bans()` and target the specific user with a `http::UserPagination` and then verify through the returned `vec<Ban>`.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2794,6 +2794,40 @@ impl Http {
         .await
     }
 
+    /// Gets a [`Ban`] for a specific user in a guild. Returns [`None`] if no ban was found
+    /// matching both the [`GuildId`] and [`UserId`].
+    ///
+    /// **Note**: Requires that you have the [Ban Members] permission
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::Http`] if the current user lacks permission.
+    ///
+    /// [Ban Members]: Permissions::BAN_MEMBERS
+    pub async fn get_ban(&self, guild_id: GuildId, user_id: UserId) -> Result<Option<Ban>> {
+        let result = self
+            .fire(Request {
+                body: None,
+                multipart: None,
+                headers: None,
+                method: LightMethod::Get,
+                route: Route::GuildBan {
+                    guild_id,
+                    user_id,
+                },
+                params: None,
+            })
+            .await;
+
+        match result {
+            Ok(ban) => Ok(Some(ban)),
+            Err(Error::Http(ref err)) if err.status_code() == Some(StatusCode::NOT_FOUND) => {
+                Ok(None)
+            },
+            Err(e) => Err(e),
+        }
+    }
+
     /// Gets all audit logs in a specific guild.
     pub async fn get_audit_logs(
         &self,
--- a/src/model/guild/guild_id.rs
+++ b/src/model/guild/guild_id.rs
@@ -300,6 +300,21 @@ impl GuildId {
         http.as_ref().get_bans(self, target, limit).await
     }
 
+    /// Gets a user's ban from the guild.
+    /// See [`Http::get_ban`] for details.
+    ///
+    /// **Note**: Requires the [Ban Members] permission.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::Http`] if the current user lacks permission.
+    ///
+    /// [Ban Members]: Permissions::BAN_MEMBERS
+    #[inline]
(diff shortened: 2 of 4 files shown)


src/http/client.rs
```diff
@@ -2794,6 +2794,60 @@ impl Http {
...
+                headers: None,
+                method: LightMethod::Get,
+                route: Route::GuildBan { guild_id, user_id },
+                params: None,
+            })
+            .await;
+
+        match result {
+            Ok(ban) => Ok(Some(ban)),
+            Err(Error::Http(ref err)) if err.status_code().unwrap() == StatusCode::NOT_FOUND => {
```
jamesbt365: Do NOT use unwrap here. Panicking is not an acceptable outcome.

src/http/client.rs
```diff
@@ -2794,6 +2794,60 @@ impl Http {
...
+    ///
+    /// match http.get_ban(guild, user).await {
+    ///     Ok(Some(ban)) => println!("Ban: {}", ban.reason.unwrap_or("No reason provided".to_string())),
+    ///     Ok(None) => println!("No ban found"),
+    ///     Err(serenity::Error::Http(ref err)) if err.status_code().unwrap() == serenity::http::StatusCode::FORBIDDEN => {
+    ///         println!("Missing Permissions::BAN_MEMBERS")
+    ///     }
+    ///     Err(err) => println!("Unknown error occurred: {:?}", err),
+    /// }
+    /// ```
```
jamesbt365: Not really a big deal but do we really need an example for this? 90% of your example is just setting up the conditions to use this and I already got rid of a lot of self explanatory examples on the `next` branch.

This isn't really a complex feature or anything that requires a specific setup to use, it'll be fine to leave without an example imo.

jamesbt365 (changes requested):

src/http/client.rs
AshleyBilbrey: In my opinion, examples are important for library adoption and developer experience, so I would like this to stay. Examples are how people best learn a new library or programming language, and a common complaint about Rust and its ecosystem is a steep learning curve. I'm not familiar with your work on the `next` branch but my input would be for those examples to stay as well.

If not for that, handling the missing permission through `Error::Http` may be useful for beginners.

That being said, I can remove the example if you still would prefer. I just need the feature for my own project. 😅

jamesbt365: > examples are important for library adoption and developer experience

Yes, but this doesn't really cover the usual usecase for a bot anyway (use of serenity::Client) and all these examples usually result in just doing unimplemented!() on http, which doesn't really explain how to get it either.

We really don't need all these examples? They don't exactly cover how to actually do all this stuff (that isn't shown) so they aren't really that useful, especially when this is literally just a method you call with 2 snowflakes.

It is a self explanatory method, and even then, users ideally shouldn't have to touch the http interface themselves and they should be using the snowflake types (in this case guild id) as they provide actual sugar and a more user friendly interface.

Interacting with http itself is a pretty "low level" way of interacting with the API in serenity, usually not for "beginners" that would just be starting out touching this lib.

jamesbt365 (approved):
Looks fine for me, the option handling on the ban seems intuitive but others may think its not right for this lib, i'll let them decide before merging.

src/model/guild/guild_id.rs
```diff
@@ -300,6 +300,21 @@ impl GuildId {
         http.as_ref().get_bans(self, target, limit).await
     }
 
+    /// Gets a user's ban from the guild.
+    /// [`Http::get_ban`] for details.
```
```suggestion
    /// See [`Http::get_ban`] for details.
```

arqunis (approved):
If you could add the missing word in `GuildId::get_ban`'s docs and run `cargo fmt`, this should be fine to merge. The failing CI jobs (besides Format) are unrelated to this PR. 

I also agree that returning an `Option` is likely the better choice over the user checking the status code themselves (although it does make the function inconsistent with other `get_*` endpoint functions that only return `Result`s).

AshleyBilbrey:
I have added the missing word and ran `cargo +nightly fmt`.

jamesbt365:
> I have added the missing word and ran `cargo +nightly fmt`.

In the future, please stop force pushing its hard to see what you actually changed and its all squashed on merge anyway.

PR #3368, opened 2025-05-23
Add banner model to Member and PartialMember

### LINKED ISSUE #3354: Add `avatar_decoration_data` to users and members
`avatar_decoration_data` is part of user and member objects and, unlike banners, sent to bots.

The field is documented for [users](https://discord.com/developers/docs/resources/user#user-object) and [members](https://discord.com/developers/docs/resources/guild#guild-member-object), and by extension is also sent in various related gateway events.

TapGhoul:
Adds missing fields in member data.
Annoyingly, discord does not allow us to update this data in guilds at the moment (that I can find) so it's just this for now.


src/model/user.rs
```diff
@@ -371,7 +371,7 @@ impl User {
     /// [`crate::http::Http::get_user`].
```
GnomedDev: This doc comment is wrong now, can you update it/remove the note?

GnomedDev (commented):
Looks good, just a small doc update update needed.

Is there a reason this is targeting `next` instead of `current`?

TapGhoul:
> Looks good, just a small doc update update needed.
> 
> Is there a reason this is targeting `next` instead of `current`?

From previous discussion, this was in `next` because the struct member is `pub` and therefore it's a breaking interface change.

GnomedDev:
I don't see where this is breaking? It's not changing existing fields and where new fields are added they are already `non_exhaustive`?

src/model/user.rs
TapGhoul: Donezo

TapGhoul:
> I don't see where this is breaking? It's not changing existing fields and where new fields are added they are already `non_exhaustive`?

Oh wait sorry I got wires crossed, I was thinking of the other old PR - not this one. You're right. I can rebase.

arqunis (approved):

PR #3369, opened 2025-06-04
Don't wait for websocket to close on Shard disconnect

GnomedDev:
This could be the cause and fix for the shard issues that have been happening on failed resumes. I am testing it out on TTS Bot before I remove the `needs-testing` label.

### DIFF
--- a/src/gateway/sharding/shard_runner.rs
+++ b/src/gateway/sharding/shard_runner.rs
@@ -3,7 +3,6 @@ use std::sync::Arc;
 use dashmap::DashMap;
 use dashmap::try_result::TryResult;
 use futures::channel::mpsc::{self, UnboundedReceiver as Receiver, UnboundedSender as Sender};
-use tokio_tungstenite::tungstenite;
 use tokio_tungstenite::tungstenite::error::Error as TungsteniteError;
 use tokio_tungstenite::tungstenite::protocol::frame::CloseFrame;
 #[cfg(feature = "tracing_instrument")]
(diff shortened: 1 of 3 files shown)


GnomedDev:
I have confirmed with TTS Bot that no shards have broken since applying this patch, and logs show the shards hitting Ready after resume failures instead of simply never showing up in logs again.

src/gateway/sharding/shard_runner.rs
```diff
@@ -248,22 +253,6 @@ impl ShardRunner {
...
-                Some(Err(_)) => {
-                    warn!(
-                        "[ShardRunner {:?}] Received an error awaiting close frame",
-                        self.shard.shard_info(),
-                    );
-                    return;
-                },
-                _ => {},
-            }
-        }
```
mkrasnitski, line 266: Do you think there might end up being a race condition during a restart where a reidentify is attempted before the close frame is received, and the reidentify fails as a result? That wouldn't be catastrophic as the shard would get re-queued and maybe a 5 second penalty would be incurred. However, maybe this could be controlled with a boolean parameter that indicates whether to wait for a close frame or not. Then, in the case where a resume fails, we can decide not to wait for the close frame since the behavior of Discord's gateway would be in question.

GnomedDev: I don't see why the reidentify could possibly fail due to a race condition with the "poisoned" websocket. The reidentify would be using an entirely new `Shard` with an entirely separate websocket connection.

mkrasnitski: I'm referring to a reidentify being sent for a Shard that hasn't fully shutdown and is still considered connected. It's not about the websocket being poisoned, but rather the connection status of a specific Shard id.

GnomedDev: Multiple connections can be opened with the same Shard ID at the same time. 
> Do you think there might end up being a race condition during a restart where a reidentify is attempted before the close frame is received, and the reidentify fails as a result?

I don't see why a reidentify could fail due to an old connection being in an indeterminate state?

mkrasnitski: Oh, I actually had no idea that you could be connected to the same shard multiple times simultaneously. I must've been thinking back to my work on max-concurrency.

arqunis (approved):

PR #3377, opened 2025-06-30
Fix Ord impl for Role

cheesycod:
Discord does position DESC, id ASC


jamesbt365 (approved):

jamesbt365:
clippy failure is unrelated, will merge if everything else passes.

PR #3378, opened 2025-06-30
Clean up Result types for WsClient

mkrasnitski:
Follow-up to #3369 - using `StdResult` isn't necessary because we can already coerce `tungstenite::Error` into serenity's `Error`. Worth noting that this incurs a `Box` call for the error type, but this isn't a hot path by any means.

### DIFF
--- a/src/gateway/sharding/mod.rs
+++ b/src/gateway/sharding/mod.rs
@@ -698,8 +698,7 @@ async fn connect(base_url: &str, compression: TransportCompression) -> Result<Ws
         Error::Gateway(GatewayError::BuildingUrl)
     })?;
 
-    let client = WsClient::connect(url, compression).await?;
-    Ok(client)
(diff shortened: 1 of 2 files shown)


GnomedDev (changes requested):
This was a purposeful optimisation to avoid returning the large enum of serenity's `Error` and instead just return the smaller tungstenite error. Since these are internal methods, it doesn't matter to be hiding the error type, so I cannot see the benefit here.

If you have a benefit to reverting this, let me know, otherwise I would like to see this PR closed.

mkrasnitski:
Turns out `serenity::Error` is pretty small, 72 bytes. And, `tokio_tungstenite::tungstenite::Error` is 136 bytes large. So, unless we box it, we're not getting any benefit. Plus, mixing result types like this is odd imo.

GnomedDev (approved):
Fair enough.

PR #3382, opened 2025-07-05
Gradient roles, guild tags, avatar decorations and nameplates

### LINKED ISSUE #3375: GuildMemberUpdateEvent is outdated
the flags and avatar decoration fields is missing.
https://docs.rs/serenity/0.12.4/serenity/model/event/struct.GuildMemberUpdateEvent.html 

https://discord.com/developers/docs/events/gateway-events#guild-member-update-guild-member-update-event-fields

jamesbt365:
Fixes #3375.

Adds support for the mentioned discord features, 

Copied the guild features from the discord docs, its unclear what other ones exist (i could go digging for them, but its pretty useless because it'll never be non exhaustive and even their list isn't complete)

Derived `Copy` on RoleColours which is fine and it works on avatar decorations too but if discord adds a field its probably best to not have it copy? Let me know what you think and if i should remove this derive.

For some reason Discord didn't add nameplate image formats to the docs for whatever reason, but links said docs, so I peeked at devtools to see what the discord client is doing and found the links. I've left this under unstable though as i don't know if discord will be changing anything later. But given its just links we could probably just remove the feature flag?

Heres a peak at the object:
```json
    "collectibles": {
        "nameplate": {
            "asset": "nameplates/nameplates/cherry_blossoms/",
            "skuId": "1349849614102036581",
            "expiresAt": null,
            "label": "COLLECTIBLES_NAMEPLATES_CHERRY_BLOSSOMS_A11Y",
            "palette": "berry"
        }
    }
```

And the links for example
https://cdn.discordapp.com/assets/collectibles/nameplates/nameplates/cherry_blossoms/asset.webm
https://cdn.discordapp.com/assets/collectibles/nameplates/nameplates/cherry_blossoms/static.png

The test fix I have NO idea what to do, why are we parsing a role from json here?

### DIFF
--- a/src/builder/edit_role.rs
+++ b/src/builder/edit_role.rs
@@ -54,6 +54,9 @@ pub struct EditRole<'a> {
     #[serde(rename = "color")]
     colour: Option<Colour>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "colors")]
+    colours: Option<CreateRoleColours>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     hoist: Option<bool>,
     #[serde(skip_serializing_if = "Option::is_none")]
     icon: Option<Option<String>>,
@@ -86,6 +89,7 @@ impl<'a> EditRole<'a> {
             colour: Some(role.colour),
             unicode_emoji: role.unicode_emoji.as_ref().map(|v| Some(v.clone())),
             audit_log_reason: None,
+            colours: Some(role.colours.into()),
             // TODO: Do we want to download role.icon?
             icon: None,
         }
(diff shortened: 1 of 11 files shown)


jamesbt365:
Also I have no idea how to fix the large future, have fun with that one!

src/builder/edit_role.rs
```diff
@@ -150,6 +154,55 @@ impl<'a> EditRole<'a> {
...
 
+/// The colours of a Discord role, secondary_colour and tertiary_colour may only be set if
+/// the [Guild] has the `ENHANCED_ROLE_COLORS` feature.
+///
+/// Note: 2024-07-05 - tertiary_colour is currently enforced to be set with a specific pair of
+/// primary and secondary colours, for current validation see
+/// [Discord docs](https://discord.com/developers/docs/topics/permissions#role-object-role-colors-object).
+#[derive(Clone, Debug, Default, Serialize)]
+#[must_use]
+#[expect(clippy::struct_field_names)]
```
DPlayer234: `#[expect]` isn't supported in the MSRV for `current`.

src/model/user.rs
```diff
@@ -355,6 +364,99 @@ bitflags! {
...
+/// [Discord docs](https://discord.com/developers/docs/resources/user#user-object-user-primary-guild)
+#[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[non_exhaustive]
+pub struct PrimaryGuild {
+    /// the id of the user's primary guild.
+    pub identity_guild_id: Option<GuildId>,
+    // whether the user is displaying the primary guild's server tag. This can be null if the
+    // system clears the identity, e.g. because the server no longer supports tags.
+    pub identity_enabled: Option<bool>,
```
DPlayer234: Is this meant to be a doc comment?

src/model/colour.rs
```diff
@@ -19,10 +19,12 @@
 /// # use serenity::model::id::RoleId;
 /// # use serenity::model::id::GuildId;
 /// # use serenity::model::permissions;
+/// # use serenity::model::guild::RoleColours;
 /// #
 /// # fn main() {
 /// # let role = from_value::<Role>(json!({
 /// #     "color": Colour::BLURPLE,
+///       "colors": RoleColours::default(),
```
DPlayer234: This line should also be prefixed with `#`.

DPlayer234 (commented):
The large future seems to be due to increasing the size of some structs here, but since it only complains about the shard runner, it's probably fine to `#[allow]` it.

PR #3385, opened 2025-07-09
New minimal dependency for reqwest

Jasper-Bekkers:
Reqwest has been released for a while and this crate seems to be fully compatible with both versions. In order to shrink our list of duplicate dependencies downstream I'm PRing this to keep folks from being able to use 0.11 but also jump on 0.12 if that fits their dependencies better.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ uwl = { version = "0.6.0", optional = true }
 levenshtein = { version = "1.0.5", optional = true }
 chrono = { version = "0.4.31", default-features = false, features = ["clock", "serde"], optional = true }
 flate2 = { version = "1.0.28", optional = true }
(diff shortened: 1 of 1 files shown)


GnomedDev (changes requested):
This is failing CI due to what looks to be incompatibilities. Since we are releasing `next` soon™ this is probably not worth hacking around to make compatible with both. I'll leave this open if you can fix it nicely, but I recommend closing it.

DerPlayer2001:
I think @Jasper-Bekkers just missed bumping the http_crate dev dependency that is only used during testing.

edit:
I need the interface feature of 0.12 request so it would be great if this could be added 👍

arqunis:
Closing in favour of https://github.com/serenity-rs/serenity/pull/3404

PR #3386, opened 2025-07-12
Accept `ImageData` when passing an image for an emoji in `GuildId::create_emoji`

arqunis:
Previously, users were able to create the base64-encoded blob of data via `CreateAttachment::to_base64`, which returned a string. On `next`, this method had been replaced by `encode`, which returns an `ImageData` instead.

It is still the base64-encoded blob of data as before, but users were no longer able to access it directly. This wasn't too much of a problem, as many places that accepted a plain string switched to expecting `ImageData`. So for users, nothing practically has changed. _Except_ `GuildId::create_emoji`, which still wanted a string, yet users couldn't reuse any `ImageData` they might have already had.

This rectifies the issue by not only switching `image: &str` to `image: ImageData` in `GuildId::create_emoji`, but also adding an accessor method for the blob of data in `ImageData`.

### DIFF
--- a/src/builder/create_attachment.rs
+++ b/src/builder/create_attachment.rs
@@ -146,6 +146,12 @@ impl<'a> CreateAttachment<'a> {
 pub struct ImageData<'a>(Cow<'a, str>);
 
 impl<'a> ImageData<'a> {
+    /// Accesses the stored base64-encoded image data.
+    #[must_use]
+    pub fn as_base64(&self) -> &str {
+        &self.0
+    }
+
     /// Constructs image data from a base64-encoded blob of data. The string must be a valid data
     /// URI, for example:
     ///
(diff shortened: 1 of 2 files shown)


src/builder/create_attachment.rs
```diff
@@ -146,6 +146,12 @@ impl<'a> CreateAttachment<'a> {
 pub struct ImageData<'a>(Cow<'a, str>);
 
 impl<'a> ImageData<'a> {
+    /// Accesses the stored base64-encoded image data.
+    #[must_use]
+    pub fn as_base64(&self) -> &Cow<'a, str> {
+        &self.0
```
Blui42:
```suggestion
    pub fn as_base64(&self) -> &str {
        self.0.as_ref()
```

PR #3387, opened 2025-07-17
Split core functionality into a `serenity-core` subcrate

mkrasnitski:
This PR splits serenity into two crates, `serenity` and `serenity-core`. 

 * The core library contains the `builder`, `cache`, `http`, `model`, and `utils` modules. 
 * The `serenity` crate depends on the core and re-exports everything in it, plus it also contains the `gateway`, `collector`, `framework`, and `interactions_endpoint` modules. It also re-exposes all of the core's feature flags, though I'm not sure if that's the right approach.
 
Further splitting may be possible in the future but would require significant refactoring to remove circular dependencies between modules. For error propagation, a bunch of variants on `serenity::Error` got moved into the core, and are propagated using `Error::Core` which wraps a `serenity_core::Error`. Though, I wonder if it's better to keep the variants and convert 1-1 from a `serenity_core::Error` so as to keep things more flat.


mkrasnitski:
Ran some compile-time benchmarks on my machine using `hyperfine` and came up with some numbers. Before this PR:
```
Benchmark 1: cargo check
  Time (mean ± σ):     14.868 s ±  0.049 s    [User: 60.332 s, System: 9.855 s]
  Range (min … max):   14.760 s … 14.931 s    10 runs

Benchmark 2: cargo build
  Time (mean ± σ):     19.035 s ±  0.059 s    [User: 83.263 s, System: 11.827 s]
  Range (min … max):   18.945 s … 19.098 s    10 runs

Benchmark 3: cargo build --release
  Time (mean ± σ):     24.560 s ±  0.291 s    [User: 195.193 s, System: 12.501 s]
  Range (min … max):   24.159 s … 25.139 s    10 runs
```

After this PR:
```
Benchmark 1: cargo check
  Time (mean ± σ):     14.478 s ±  0.054 s    [User: 59.951 s, System: 9.903 s]
  Range (min … max):   14.377 s … 14.572 s    10 runs

Benchmark 2: cargo build
  Time (mean ± σ):     17.117 s ±  0.036 s    [User: 82.604 s, System: 12.095 s]
  Range (min … max):   17.079 s … 17.185 s    10 runs

Benchmark 3: cargo build --release
  Time (mean ± σ):     22.155 s ±  0.143 s    [User: 199.097 s, System: 12.608 s]
  Range (min … max):   21.930 s … 22.431 s    10 runs
```

If we compare mean execution times, that gives a 2.6% improvement for `cargo check`,  10.1% improvement for `cargo build`, and 9.8% improvement for `cargo build --release`.

mkrasnitski:
This PR is ready for review now

serenity-core/Cargo.toml
```diff
@@ -0,0 +1,73 @@
...
+bytes = { version = "1.5.0", optional = true }
+chrono = { version = "0.4.31", default-features = false, features = ["clock", "serde"], optional = true }
+dashmap = { version = "6.1.0", optional = true }
+foldhash = { version = "0.1.4", optional = true }
+futures = { version = "0.3.29", default-features = false, features = ["std"], optional = true }
+mime_guess = { version = "2.0.4", optional = true }
+percent-encoding = { version = "2.3.0", optional = true }
+reqwest = { version = "0.12.2", default-features = false, features = ["multipart", "stream", "json"], optional = true }
+typesize = { version = "0.1.13", optional = true, features = ["url", "time", "serde_json", "secrecy", "parking_lot", "nonmax"] }
+mini-moka = { version = "0.10.2", optional = true }
```
GnomedDev: Can we use workspace dependencies here?

GnomedDev (changes requested):
Looks good, and a good start to splitting apart serenity.

I would be interested in seeing what the difference in `cargo build --timings` looks like, specifically with a poise or songbird based bot to see if they can depend only on core and kick off their build earlier.

I haven't taken a deep look over this, but spotted this comment immediately.

serenity-core/Cargo.toml
mkrasnitski: Yeah, not sure why I didn't think to do so.

mkrasnitski:
Closing and opening a new PR which will supersede this one.

PR #3392, opened 2025-07-22
add `len` and `is_empty` methods to v2 components

KiloOscarSix:
I'm working with a v2 component message that's built dynamically based on user inputs. It would be helpful to have a way to check the total number of components within a message to avoid hitting the 40-component limit.

I'm not sure what the standard approach is for this kind of access, so I'm open to feedback or alternative suggestions. Primarily looking for a lightweight way to introspect the current state of a component message.

### DIFF
--- a/src/builder/create_components.rs
+++ b/src/builder/create_components.rs
@@ -134,6 +134,16 @@ impl<'a> CreateSection<'a> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


KiloOscarSix:
Actually, upon reflection, this won't work for my use case or the intention perceived.

PR #3394, opened 2025-07-23
Add `get_user_voice_state` endpoint

KiloOscarSix:
Noticed this endpoint was missing - https://discord.com/developers/docs/resources/voice#get-user-voice-state

I don't know if there is any standard on where methods are placed, so do let me know if you'd rather it be placed somewhere else.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -4111,13 +4111,17 @@ impl Http {
         guild_id: GuildId,
         user_id: UserId,
     ) -> Result<VoiceState> {
-        self.fire::<VoiceState>(Request::new(
-            Route::GuildVoiceStates {
+        self.fire(Request {
+            body: None,
+            multipart: None,
+            headers: None,
+            method: LightMethod::Get,
+            route: Route::GuildVoiceStates {
                 guild_id,
                 user_id,
             },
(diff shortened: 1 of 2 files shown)


src/http/client.rs
```diff
@@ -4104,6 +4104,23 @@ impl Http {
...
+        guild_id: GuildId,
+        user_id: UserId,
+    ) -> Result<VoiceState> {
+        self.fire::<VoiceState>(Request::new(
+            Route::GuildVoiceStates {
+                guild_id,
+                user_id,
+            },
+            LightMethod::Get,
+        ))
```
```suggestion
        self.fire(Request {
            body: None,
            multipart: None,
            headers: None,
            method: LightMethod::Get,
            route: Route::GuildVoiceStates {
                guild_id,
                user_id,
            },
            params: None
        })
```

arqunis (changes requested):

arqunis (approved):
Clippy failure is unrelated

jamesbt365:
Why did this target next?

KiloOscarSix:
> Why did this target next?

Truthly, was the branch I had open as I was doing some stuff on the V2 Components

PR #3404, opened 2025-08-10
bump http and reqwest versions

DerPlayer2001:
Fix for #3385 also bump test dep to fix pipeline


DerPlayer2001:
The clippy error(implicit-clone) is not related to my changes. I could add the .clone() if desired.

arqunis (approved):

PR #3410, opened 2025-08-27
Handle query parameters for webhook requests

pasmophobia:
This pull request updates how query parameters are handled when sending webhook requests in the `Http` client. The main change is that query parameters from the URL are now parsed and passed along with the request, rather than always being set to `None`.

Improvements to webhook request handling:

* Query parameters from the URL are now parsed using `url.query_pairs()`, collected into a vector, and passed to the request if present, otherwise set to `None` (`src/http/client.rs`).
* The `params` field in the `Request` struct is now set to the parsed query parameters instead of always being `None` (`src/http/client.rs`).

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -4642,6 +4642,11 @@ impl Http {
         let url = Url::parse(url).map_err(HttpError::Url)?;
         let (webhook_id, token) =
             crate::utils::parse_webhook(&url).ok_or(HttpError::InvalidWebhook)?;
+        let mut params = url.query_pairs().collect::<Vec<(&str, String)>>();
+        let params = match params.len() {
+            0 => None,
+            _ => Some(params),
+        };
         self.fire(Request {
             body: None,
             multipart: None,
@@ -4651,7 +4656,7 @@ impl Http {
                 webhook_id,
                 token,
             },
-            params: None,
+            params,
         })
         .await
     }


mkrasnitski:
In serenity, the `get_webhook_from_url` function is equivalent to `get_webhook_with_token` in terms of functionality, and it makes requests to the same API endpoint. According to the [Discord docs](https://discord.com/developers/docs/resources/webhook#get-webhook-with-token), the endpoints for getting a webhook object don't use any query params. What's the reason you want to do this?

pasmophobia:
The reason for handling query params here is that webhook URLs can include them (e.g. ?thread_id=).
Right now get_webhook_from_url drops that information, but the query string is semantically part of the webhook URL.
I think preserving it makes the behavior more intuitive.

mkrasnitski:
The problem here is that the `thread_id` is not stored in the actual `Webhook` object returned from Discord when you ask for one, because Discord itself doesn't let you create a webhook directly in a thread. The user is instead expected to provide the query parameter each time a relevant API call is made, such as [Execute Webhook](https://discord.com/developers/docs/resources/webhook#execute-webhook) or [Get Webhook Message](https://discord.com/developers/docs/resources/webhook#get-webhook-message).

Joshix-1:
Handling the thread_id query param would make sense for the parse_webhook function in utils

pasmophobia:
Sorry, I misunderstood the code. I’ll open a new PR later.

PR #3412, opened 2025-08-29
Rename `ImageData` to `DataUri` and allow setting its mimetype

mkrasnitski:
Back in September 2024, Discord added [Soundboard](https://discord.com/developers/docs/resources/soundboard#soundboard-resource) APIs that allow a bot to create a new sound by uploading raw mp3 or ogg data. This data is specified as part of the JSON rather than a multipart upload (an odd decision IMO), however that means that the semantics of `ImageData` are wrong as it makes too many assumptions about its contents. In particular, we assume that the base64 data will always be an image, and the `CreateAttachment::encode` method always assigns the `image/png` mimetype.

This PR removes these assumptions by renaming the struct to simply `DataUri`, and now requiring the user to specify the mimetype when calling `CreateAttachment::encode`.

### DIFF
--- a/src/builder/create_attachment.rs
+++ b/src/builder/create_attachment.rs
@@ -115,21 +115,21 @@ impl<'a> CreateAttachment<'a> {
     /// # Errors
     ///
     /// See [`CreateAttachment::get_data`] for details.
-    pub async fn encode(&self) -> Result<ImageData<'_>> {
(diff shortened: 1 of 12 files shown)


mkrasnitski:
I somehow forgot we already had support for the soundboard API.

arqunis (approved):

PR #3414, opened 2025-09-01
Refactor CI workflows

mkrasnitski:
This improves the ergonomics of the CI workflows and cuts down on CI runtime by adding a new `check` job for feature flags which only need to be tested using `cargo check` rather than `cargo test`. I removed the dedicated `nightly` job by including it in the `test` job.

Also added `temp_cache` to the list of features to check; for some reason it wasn't included in CI before.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,8 @@ jobs:
           - builder without model
           - unstable + typesize
           - unstable (no default features)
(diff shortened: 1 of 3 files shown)


mkrasnitski:
Seems the MSRV job isn't being skipped - will fix tomorrow

arqunis (approved):

PR #3418, opened 2025-09-09
Improve usage of FixedString

Joshix-1:


src/model/channel/reaction.rs
```diff
@@ -368,7 +368,7 @@ impl From<char> for ReactionType {
     /// # fn main() {}
     /// ```
     fn from(ch: char) -> ReactionType {
-        ReactionType::Unicode(ch.to_string().trunc_into())
+        ReactionType::Unicode(ch.into())
```
Joshix-1, line 371: this should remove a heap allocation

GnomedDev:
@Joshix-1 I have pushed a new and fixed version of small_fixed_array, after that this should be mergable.

Cargo.toml
```diff
@@ -41,7 +41,7 @@ base64 = { version = "0.22.0" }
 zeroize = { version = "1.7" } # Not used in serenity, but bumps the minimal version from secrecy
 arrayvec = { version = "0.7.4", features = ["serde"] }
 serde_cow = { version = "0.1.0" }
-small-fixed-array = { version = "0.4", features = ["serde"] }
+small-fixed-array = { version = "0.4, >=0.4.10", features = ["serde"] }
```
GnomedDev:
```suggestion
small-fixed-array = { version = "0.4.10", features = ["serde"] }
```
This is equal and simpler.

GnomedDev (changes requested):
Just one quick fix.

GnomedDev (approved):

PR #3426, opened 2025-10-02
Add support for gradient roles

mkrasnitski:
Splitting out from #3382 to rebase and fix it up. One unresolved question is if `RoleColours` should be `Copy` or not.


arqunis (approved):
> One unresolved question is if `RoleColours` should be Copy or not.

I think it is fine for it to be `Copy`. It doesn't appear to be a structure that they are going to be updating frequently, so we shouldn't have to worry about them introducing a heavy type into it, and the overall size turns out to be `20` bytes, which is dirt cheap to copy.

PR #3430, opened 2025-10-07
Add support for Label and File Upload components

mkrasnitski:
Labels are a new top-level component only supported in modals that can hold a Text Input, Select Menu, or File Upload (another new component) inside it. Oddly enough, the label and description are removed from the label payload when sent back as part of a modal response. See the [docs](https://discord.com/developers/docs/components/reference#label).

Text Inputs inside Action Rows are deprecated, therefore I removed the `InputText` variant for `ActionRowComponent`. Also, at the moment modals can include Action Rows, Text Displays, and Labels as top-level components, so I changed modal responses to store `Component` instead of `ActionRow` and added a `text` method to quick modals. I thought about adding select menu support to quick modals but the methods felt a little too verbose.

### DIFF
--- a/src/builder/create_components.rs
+++ b/src/builder/create_components.rs
@@ -523,6 +523,16 @@ impl<'a> CreateLabel<'a> {
         }
     }
 
+    /// Create a file upload with a specific label.
+    pub fn file_upload(label: impl Into<Cow<'a, str>>, file_upload: CreateFileUpload<'a>) -> Self {
+        Self {
+            kind: ComponentType::Label,
+            label: label.into(),
+            description: None,
+            component: CreateLabelComponent::FileUpload(file_upload),
+        }
+    }
+
     /// Sets the description of this component, which will display underneath the label text.
     pub fn description(mut self, description: impl Into<Cow<'a, str>>) -> Self {
         self.description = Some(description.into());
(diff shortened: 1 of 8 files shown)


GnomedDev:
Why remove `StaticU8`? I'm fine with it, as it's a minor performance opt, but I just want the reason written down somewhere.

mkrasnitski:
The `CreateButton` and `CreateInputText` structs have a `kind` field of type `ComponentType` already, so this is an inconsistency that I wanted to resolve. Also, `ComponentType` is a fake enum since it uses the `enum_number!` macro, meaning it's actually a newtype around `u8`, same as `StaticU8`. I'll make a follow-up PR removing the `kind` fields entirely and just inserting it at serialization time, since it doesn't serve any other purpose.

DPlayer234:
Wasn't that the point of `StaticU8`? It was a ZST that would serialize an integer but not take up any memory at runtime.

mkrasnitski:
I've gone ahead and also added support for the new File Upload component, as it can only be found inside a Label.

> Wasn't that the point of `StaticU8`? It was a ZST that would serialize an integer but not take up any memory at runtime.

I didn't realize it was a ZST, however I do still think that for now, `ComponentType` is clearer, given that the variants can be named instead of being bare integers.

calum4:
I've been using this branch (thank you) and unfortunately discovered an issue.

In my testing, all [Select Menu](https://discord.com/developers/docs/components/reference#string-select) interaction responses originating from a modal include the `values` field, however this is not currently present in the serenity [`SelectMenu`](https://github.com/mkrasnitski/serenity/blob/current/src/model/application/component.rs#L205) model.

On a fork of this branch, I have applied a fix which you can find [here](https://github.com/calum4/serenity/commit/890192636b3dca7d0f45cf1f83108ed43a952d78).

I have also created an MRE which I used to test said fix which you can find [here](https://github.com/calum4/serenity-mre/tree/mkrasnitski/serenity/label-component/modal-values). By default the serenity crate is built from your latest commit https://github.com/serenity-rs/serenity/pull/3430/commits/bc37312a56f83bb7cb220d857d97f2d58e212713. In [`Cargo.toml`](https://github.com/calum4/serenity-mre/blob/mkrasnitski/serenity/label-component/modal-values/Cargo.toml) I have included a commented out a dependency to use my fix.

Thanks!

arqunis (approved):

arqunis:
@calum4 I [cherry-picked](https://github.com/serenity-rs/serenity/commit/7cf845b281933e651007430957ee30392aac2289) your commit after the merge.

PR #3431, opened 2025-10-11
Make `ClientBuilder` directly accept boxed handler types

DPlayer234:
Rather than having the `ClientBuilder` accept `impl Into<Arc<H>> where H: HandlerTrait`, directly accept the corresponding `Arc<dyn HandlerTrait>`.
This allows accepting non-concrete handler types (f.e. if constructed by a function that returns `Arc<dyn HandlerTrait>` already) and fixes type inference failures when passing an `Arc<H>` rather than `H`.

This affects the setters for `EventHandler`, `RawEventHandler`, `Framework`, and `VoiceGatewayManager`.

`Framework` is in a bit of a weird spot because it is only put in an `Arc` after `Framework::init` has been called, so a `&mut` can be taken to it for that call, so this PR has `ClientBuilder` accept `Box<dyn Framework>` for it.

Most of the PR diff is related to fixing examples.

### DIFF
--- a/src/gateway/client/mod.rs
+++ b/src/gateway/client/mod.rs
@@ -231,26 +231,25 @@ impl ClientBuilder {
         self.intents
     }
 
-    /// Adds an event handler with multiple methods for each possible event.
+    /// Sets the event handler where all received gateway events will be dispatched.
     pub fn event_handler(mut self, event_handler: Arc<dyn EventHandler>) -> Self {
         self.event_handler = Some(event_handler);
         self
     }
 
-    /// Gets the added event handlers. See [`Self::event_handler`] for more info.
+    /// Gets the added event handler. See [`Self::event_handler`] for more info.
     #[must_use]
     pub fn get_event_handler(&self) -> Option<Arc<dyn EventHandler>> {
         self.event_handler.clone()
     }
 
-    /// Adds an event handler with a single method where all received gateway events will be
-    /// dispatched.
+    /// Sets the raw event handler where all received gateway events will be dispatched.
     pub fn raw_event_handler(mut self, raw_event_handler: Arc<dyn RawEventHandler>) -> Self {
         self.raw_event_handler = Some(raw_event_handler);
         self
     }
 
-    /// Gets the added raw event handlers. See [`Self::raw_event_handler`] for more info.
+    /// Gets the added raw event handler. See [`Self::raw_event_handler`] for more info.
     #[must_use]
     pub fn get_raw_event_handler(&self) -> Option<Arc<dyn RawEventHandler>> {
         self.raw_event_handler.clone()
--- a/src/gateway/client/mod.rs
+++ b/src/gateway/client/mod.rs
@@ -187,7 +187,7 @@ impl ClientBuilder {
     /// Sets the voice gateway handler to be used. It will receive voice events sent over the
     /// gateway and then consider - based on its settings - whether to dispatch a command.
     #[cfg(feature = "voice")]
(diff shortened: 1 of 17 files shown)


src/gateway/client/mod.rs
```diff
@@ -236,34 +232,28 @@ impl ClientBuilder {
     }
 
     /// Adds an event handler with multiple methods for each possible event.
```
Joshix-1: isn't that comment wrong?

src/gateway/client/mod.rs
```diff
@@ -236,34 +232,28 @@ impl ClientBuilder {
...
-    where
-        H: RawEventHandler + 'static,
-    {
-        self.raw_event_handler = Some(raw_event_handler.into());
+    pub fn raw_event_handler(mut self, raw_event_handler: Arc<dyn RawEventHandler>) -> Self {
+        self.raw_event_handler = Some(raw_event_handler);
         self
     }
 
     /// Gets the added raw event handlers. See [`Self::raw_event_handler`] for more info.
```
Joshix-1: `handlers` -> `handler`

src/gateway/client/mod.rs
```diff
@@ -236,34 +232,28 @@ impl ClientBuilder {
...
-    where
-        H: EventHandler + 'static,
-    {
-        self.event_handler = Some(event_handler.into());
+    pub fn event_handler(mut self, event_handler: Arc<dyn EventHandler>) -> Self {
+        self.event_handler = Some(event_handler);
         self
     }
 
     /// Gets the added event handlers. See [`Self::event_handler`] for more info.
```
Joshix-1: `handlers` -> `handler`

src/gateway/client/mod.rs
DPlayer234: Must have been forgotten when `EventHandler` was changed a while back.

src/gateway/client/mod.rs
DPlayer234: Must have been forgotten when it changed to only allow registering one of each.

src/gateway/client/mod.rs
DPlayer234: Must have been forgotten when it changed to only allow registering one of each.
That all said, I can include the comment change here.

src/gateway/client/mod.rs
```diff
@@ -188,11 +187,8 @@ impl ClientBuilder {
     /// Sets the voice gateway handler to be used. It will receive voice events sent over the
     /// gateway and then consider - based on its settings - whether to dispatch a command.
     #[cfg(feature = "voice")]
-    pub fn voice_manager<V>(mut self, voice_manager: impl Into<Arc<V>>) -> Self
-    where
-        V: VoiceGatewayManager + 'static,
-    {
-        self.voice_manager = Some(voice_manager.into());
+    pub fn voice_manager<V>(mut self, voice_manager: Arc<dyn VoiceGatewayManager>) -> Self {
```
Joshix-1: > type parameter `V` goes unused in function definition

Joshix-1:
the readme mentions the event_handler method. You should update that and check if there are other markdown files mentioning the methods you changed

DPlayer234:
The example in the README was horribly outdated, I've replaced it with a trimmed down version of the up-to-date ping bot example.
Didn't find any other places where these functions are used in non-Rust files.

KiloOscarSix (approved):

arqunis (approved):

PR #3432, opened 2025-10-11
Fix clippy warnings on current

GnomedDev:
Just some simple warnings added to nightly clippy recently.


src/client/event_handler.rs
```diff
@@ -21,6 +21,7 @@ macro_rules! event_handler {
                 $( #[deprecated = $deprecated] )?
                 async fn $method_name(&self, $($context: Context,)? $( $arg_name: $arg_type ),*) {
                     // Suppress unused argument warnings
+                    #[allow(clippy::double_parens)]
                     drop(( $($context,)? $($arg_name),* ))
```
mkrasnitski:
```suggestion
                    drop(( $($context,)? $($arg_name,)* ))
```
This way the macro will produce 1-tuples instead of parenthesized single values.

PR #3434, opened 2025-10-28
Fix clippy

mkrasnitski:


KiloOscarSix:
This not a duplicate of #3432?

mkrasnitski:
Yeah, you're right I didn't realize.

PR #3452, opened 2025-12-12
Add guild role member counts endpoint

nwerosama:
Recently introduced and documented: https://github.com/discord/discord-api-docs/pull/8025

I did test the endpoint with an example bot. However, I'm not sure if `HashMap` is a good idea to use but I'm open to other options instead.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -1,6 +1,7 @@
 #![allow(clippy::missing_errors_doc)]
 
 use std::borrow::Cow;
+use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


arqunis (approved):
> However, I'm not sure if `HashMap` is a good idea to use but I'm open to other options instead.

The [`CommandData::resolved`](https://serenity-rs.github.io/serenity/current/serenity/model/application/struct.CommandDataResolved.html) field is a collection of maps, for which we use `HashMap`. Should be appropriate here since the endpoint also returns a map.

PR #3461, opened 2025-12-23
Add `content` to `TextDisplay`

### LINKED ISSUE #3460: Add `content` field to `TextDisplay` component
Although the Discord documentation omits the `content` field from the documented response structure, it is present in the actual HTTP response. This field contains meaningful data directly associated with the `TextDisplay` component, and excluding it results in a loss of relevant information.

Including `content` would improve completeness and better reflect both actual and expected API behaviour, even if it is currently undocumented.

Adding it as `Option<String>` would make the most sense, allowing Serenity to handle its removal from the HTTP response gracefully without requiring a breaking change to the struct.

KiloOscarSix:
Fixes #3460

Payload from message component:
```rust
RawValue({
    "type":17,
    "spoiler":false,
    "id":1,
    "components":[
        {
            "type":10,
            "id":2,
            "content":"### Blackjack\nYour bet: 100 <:coin:1383692085529415680>\n\n**Your Hand**\n<:5:1383692755917733888> <:9:1383692701987246080> - 14\n\n**Dealer Hand**\n<:7:138369276838727
2704> <:blank:1390357737011155024> - 7"
        },
        {
            ...
        }
    ]
})
```

### DIFF
--- a/src/model/application/component.rs
+++ b/src/model/application/component.rs
@@ -233,6 +233,7 @@ pub struct TextDisplay {
     /// Discord’s official documentation does not mention this field; however, it is currently
     /// returned by the API and represents meaningful data, so it is included here. This
     /// behavior is undocumented and may change or be removed by Discord at any time.
+    #[cfg(feature = "unstable_discord_api")]
     pub content: Option<String>,
 }
 
--- a/src/model/application/component.rs
+++ b/src/model/application/component.rs
@@ -229,6 +229,11 @@ pub struct TextDisplay {
     /// Always [`ComponentType::TextDisplay`]
     #[serde(rename = "type")]
     pub kind: ComponentType,
+
+    /// Discord’s official documentation does not mention this field; however, it is currently
+    /// returned by the API and represents meaningful data, so it is included here. This
+    /// behavior is undocumented and may change or be removed by Discord at any time.
+    pub content: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


KiloOscarSix:
I'm checking in to see if this could be reviewed when you have a moment. I’m currently relying on my fork, which has led to issues with Songbird integration.

src/model/application/component.rs
```diff
@@ -229,6 +229,11 @@ pub struct TextDisplay {
     /// Always [`ComponentType::TextDisplay`]
     #[serde(rename = "type")]
     pub kind: ComponentType,
+
+    /// Discord’s official documentation does not mention this field; however, it is currently
+    /// returned by the API and represents meaningful data, so it is included here. This
+    /// behavior is undocumented and may change or be removed by Discord at any time.
+    pub content: Option<String>,
```
arqunis, line 237: [We have a feature](https://github.com/serenity-rs/serenity/blob/1809beb0fc24f3942c500058ad4fa47e6a97d3f9/Cargo.toml#L116-L118) for these kinds of things that Discord does not document, but still offers them as inputs/outputs from their API.

```suggestion
    #[cfg(feature = "unstable_discord_api")]
    pub content: Option<String>,
```

arqunis (changes requested):

KiloOscarSix:
Thanks, didn't consider the feature flags 👍

arqunis (approved):

PR #3462, opened 2025-12-24
Fix transport compression

lucascompython:
I didn't run `cargo fmt --all` because it formatted tens of files, maybe that should be done in another PR.
From my own testing, both zlib and zstd work now.
I also enabled the `std` feature of the `zstd-safe` crate so that we don't have to manually manage the buffer, hope that's ok.

Before these changes the decompressed buffer had 64KB of fixed size, so larger messages would be truncated and that caused the parsing errors.
The code before for zlib also assumed decompression happened in one call, but `flate2::Decompress::decompress_vec` doesn't manage the vec so it only writes as much as it can currently fit in it.

### DIFF
--- a/src/gateway/ws.rs
+++ b/src/gateway/ws.rs
@@ -111,17 +111,15 @@ enum Compression {
 
 impl Compression {
     #[cfg(any(feature = "transport_compression_zlib", feature = "transport_compression_zstd"))]
-    const DECOMPRESSED_CAPACITY: usize = 256 * 1024;
+    const DECOMPRESSED_CAPACITY: usize = 174_504;
 
     fn inflate(&mut self, slice: &[u8]) -> Result<Option<&[u8]>> {
         match self {
             Compression::Payload {
                 decompressed,
             } => {
-                const DECOMPRESSION_MULTIPLIER: usize = 3;
-
                 decompressed.clear();
-                decompressed.reserve(slice.len() * DECOMPRESSION_MULTIPLIER);
+                decompressed.reserve(Self::DECOMPRESSED_CAPACITY);
 
                 ZlibDecoder::new(slice).read_to_end(decompressed).map_err(|why| {
                     warn!("Err decompressing bytes: {why:?}");
--- a/src/gateway/ws.rs
+++ b/src/gateway/ws.rs
@@ -1,10 +1,10 @@
 use std::env::consts;
-use std::io::Read;
+use std::io::{Read, Write};
 use std::time::{Duration, SystemTime};
 
-#[cfg(feature = "transport_compression_zlib")]
-use flate2::Decompress as ZlibInflater;
 use flate2::read::ZlibDecoder;
+#[cfg(feature = "transport_compression_zlib")]
+use flate2::write::ZlibDecoder as ZlibWriter;
 use futures::{SinkExt, StreamExt};
 use tokio::net::TcpStream;
 use tokio::time::timeout;
@@ -16,7 +16,7 @@ use tracing::instrument;
 use tracing::{debug, trace, warn};
 use url::Url;
 #[cfg(feature = "transport_compression_zstd")]
-use zstd_safe::{DStream as ZstdInflater, InBuffer, OutBuffer};
+use zstd::stream::write::Decoder as ZstdWriter;
 
 use super::{ActivityData, ChunkGuildFilter, GatewayError, PresenceData, TransportCompression};
 use crate::constants::{self, Opcode};
@@ -99,22 +99,17 @@ enum Compression {
 
     #[cfg(feature = "transport_compression_zlib")]
     Zlib {
-        inflater: ZlibInflater,
+        decoder: ZlibWriter<Vec<u8>>,
         compressed: Vec<u8>,
-        decompressed: Box<[u8]>,
     },
 
     #[cfg(feature = "transport_compression_zstd")]
     Zstd {
-        inflater: ZstdInflater<'static>,
-        decompressed: Box<[u8]>,
+        decoder: ZstdWriter<'static, Vec<u8>>,
     },
 }
 
 impl Compression {
-    #[cfg(any(feature = "transport_compression_zlib", feature = "transport_compression_zstd"))]
-    const DECOMPRESSED_CAPACITY: usize = 64 * 1024;
-
     fn inflate(&mut self, slice: &[u8]) -> Result<Option<&[u8]>> {
         match self {
             Compression::Payload {
@@ -127,8 +122,6 @@ impl Compression {
 
                 ZlibDecoder::new(slice).read_to_end(decompressed).map_err(|why| {
                     warn!("Err decompressing bytes: {why:?}");
-                    debug!("Failing bytes: {slice:?}");
-
                     why
                 })?;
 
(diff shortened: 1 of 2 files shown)


KiloOscarSix:
fmt uses the nightly version, you'll need to install the nightly toolchain and then run `cargo +nightly fmt`

lucascompython:
> fmt uses the nightly version, you'll need to install the nightly toolchain and then run `cargo +nightly fmt`

Didn't know about that, oops.

mkrasnitski:
I haven't touched the compression code myself, but looking at the docs for `flate2` and `zstd`, both libraries provide higher-level decompression utilities which avoid having to perform your own bookkeeping as well as having to call the decompressor multiple times. Specifically, `flate2::ZlibDecoder` (which comes in 3 different flavors depending on if you want to wrap a `BufRead`, `Read`, or `Write` object), and `zstd::Decoder` which also comes in `Read` and `Write` flavors. Both of these support their respective standard library traits for consuming an entire stream, and decompress data as they do so. How come we aren't using these APIs to avoid any potential bookkeeping bugs?

lucascompython:
@mkrasnitski I've now implemented the changes with higher-level APIs. I don't really know why it wasn't being used in the first place.

src/gateway/error.rs
```diff
@@ -1,5 +1,7 @@
 use std::error::Error as StdError;
 use std::fmt;
+#[cfg(feature = "transport_compression_zstd")]
+use std::io;
```
mkrasnitski: This import isn't needed, just use the fully qualified `std::io::Error` inside `Error::DecompressZstd`.

src/gateway/ws.rs
```diff
@@ -134,67 +129,33 @@ impl Compression {
...
             } => {
                 const ZLIB_SUFFIX: [u8; 4] = [0x00, 0x00, 0xFF, 0xFF];
 
-                compressed.extend_from_slice(slice);
-                let length = compressed.len();
-
-                if length < 4 || compressed[length - 4..] != ZLIB_SUFFIX {
+                if slice.len() < 4 || slice[slice.len() - 4..] != ZLIB_SUFFIX {
                     return Ok(None);
                 }
```
mkrasnitski, line 144: I think this isn't needed as technically a zlib compressed payload might not end with a flush code (`00 00 FF FF`), but instead with a finish sequence. And I wouldn't expect a payload less than 4 bytes long either, as that will never be valid zlib data.

Removing this branch would also allow changing the function return type to just `Result<&[u8]>`.

src/gateway/error.rs
fgardt: This feature gated import seems to be only used in the `DecompressZstd(io::Error)` Error enum variant.
Could get rid of this import here then and just fully qualify `std::io::Error` in the Enum variant.

src/gateway/ws.rs
```diff
@@ -134,67 +129,33 @@ impl Compression {
...
-                inflater
-                    .decompress(compressed, decompressed, flate2::FlushDecompress::Sync)
-                    .map_err(GatewayError::DecompressZlib)?;
-                compressed.clear();
-                let produced = (inflater.total_out() - pre_out) as usize;
+                decoder.get_mut().clear();
+                decoder.write_all(slice).map_err(|e| {
+                    warn!("Err decompressing bytes: {e:?}");
+                    e
+                })?;
```
fgardt: Why does the Zlib decoder print warnings here while the Zstd one does not?
Thoughts on matching both Zlib and Zstd with the way the Payload compression prints such a warning + debug message?

fgardt (commented):
Thanks for looking into this, I got a couple minor nits.

> I don't really know why it wasn't being used in the first place.

I also honestly don't remember my reasoning for it back then, its been over a year at this point.. 😅
Checking the original transport compression PR and related discord messages also didn't reveal any bigger reasoning so this change is completely fine along with fixing the underlying message size issue.

src/gateway/ws.rs
lucascompython: I think I left the `debug!` on zlib by mistake.

> Thoughts on matching both Zlib and Zstd with the way the Payload compression prints such a warning + debug message?

I think it makes sense to make it a warning just like it is on payload compression.

fgardt (approved):

mkrasnitski (dismissed):
One question I had about the implementation @fgardt is how the flush flag should be handled. The [discord docs](https://discord.com/developers/docs/events/gateway#zlibstream) mention that zlib-encoded payloads should be decompressed as soon as `Z_SYNC_FLUSH` is first seen, whereas compressed data streams might include multiple of these if the compressor was flushed multiple times. I tested appending multiple short compressed payloads together separated by flushes and decompressing it using `write_all` followed by `flush` causes the entire payload to be decompressed rather than just the first chunk.

I figured I'd just note this in case Discord starts including multiple JSON payloads in compressed packets, delimited by flushes. But for now, I don't think this is really an issue as long as we have proper error reporting if/when it happens.

src/gateway/ws.rs
fgardt: Well this condition and suffix check is based on the provided [example by discord themselves](https://discord.com/developers/docs/events/gateway#zlibstream-transport-compression-example).

lucascompython: I'm sorry, I didn't really think all that much about this, the `compressed` buffer that was previously implemented, handles this possible issue.

Although, from my limited testing, the condition is never true.

I can revert if you want to.

fgardt: Oh right you're always immediately decompressing a given payload and don't buffer it in case discord splits it up into multiple packets.

Yeah that should be reverted, so `Compression::Zlib` should have a compressed buffer again.

I never tested / checked if this case can happen but discord themselves are explicitly describing it and even provide a concrete code example that behaves like this so we should definitely match that behaviour.

lucascompython:
Please review the changes.

Also, would it be possible to enable the `zlib-rs` feature of the `flate2` crate? It has some performance benefits at the cost of some unsafe code.

fgardt (approved):

arqunis:
> Please review the changes.
> 
> Also, would it be possible to enable the `zlib-rs` feature of the `flate2` crate? It has some performance benefits at the cost of some unsafe code.

>> `zlib-rs` - this implementation utilizes the zlib-rs crate, a Rust rewrite of zlib.

If the `unsafe` portions of the Rust rewrite are sound, I suppose why not?

src/gateway/ws.rs
```diff
@@ -99,36 +99,30 @@ enum Compression {
...
-        inflater: ZstdInflater<'static>,
-        decompressed: Box<[u8]>,
+        decoder: ZstdWriter<'static, Vec<u8>>,
     },
 }
 
 impl Compression {
     #[cfg(any(feature = "transport_compression_zlib", feature = "transport_compression_zstd"))]
-    const DECOMPRESSED_CAPACITY: usize = 64 * 1024;
+    const DECOMPRESSED_CAPACITY: usize = 174_504;
```
lucascompython, line 113: From my local and small testing, just from starting a bot and doing pretty much nothing else, the decoder buffer would get reallocated until the capacity hit 174KiB

Image

fgardt (approved):

GnomedDev (approved):
Tested with 32 shards of events for a couple minutes.

PR #3467, opened 2026-01-04
Add into_* methods to Components

calum4:


mkrasnitski:
Could you describe your usecase for these functions? You could definitely just match directly on the enum variant if you need to destructure a `Component`. Either way, similar to the destructuring methods on `Channel`, these should return owned types rather than references, and drop the `as_` prefix. This way, any further operations do not have to go through an immutable reference, which might require extra clones.

calum4:
> Could you describe your usecase for these functions? You could definitely just match directly on the enum variant if you need to destructure a Component.

Yeah absolutely, it's just unnecessarily verbose in my opinion. For an example of my use case:
```rust
fn foo(interaction: &ModalInteraction) {
    let mut components_iter = interaction.data.components.iter();

    // AS_*

    let _channel_id = match components_iter
        .next()
        .and_then(|c| c.as_label())
        .and_then(|c| c.component.as_select_menu())
        .and_then(|c| c.values.first())
        .map(|s| GenericChannelId::from_str(s))
    {
        Some(Ok(channel_id)) => channel_id,
        _ => todo!(),
    };

    // LET ELSE

    let Some(Component::Label(label)) = components_iter.next() else {
        todo!()
    };

    let LabelComponent::SelectMenu(select_menu) = &label.component else {
        todo!()
    };

    let Some(Ok(_channel_id)) = select_menu.values.first().map(|s| GenericChannelId::from_str(s)) else {
        todo!()
    };

    // MATCH

    let _channel_id = match components_iter.next() {
        Some(Component::Label(label)) => {
            match &label.component {
                LabelComponent::SelectMenu(select_menu) => {
                    match select_menu.values.first().map(|s| GenericChannelId::from_str(s)) {
                        Some(Ok(channel_id)) => channel_id,
                        _ => todo!(),
                    }
                }
                _ => todo!(),
            }
        }
        _ => todo!(),
    };
}
```

> Either way, similar to the destructuring methods on Channel, these should return owned types rather than references, and drop the as_ prefix.

I agree with this criticism, expect another commit to implement this. In my use case I am working with a reference to said `ModalInteraction` and the thought hadn't crossed my mind.

calum4:
> these should return owned types rather than references, and drop the as_ prefix.

Note, rather than simply omitting the `as_` prefix, I followed the convention outlined [here](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) and used `into_`.

If this contravenes an overriding convention for this crate, I'll drop the `into_` prefix altogether.

Thanks!

mkrasnitski:
One way of match you've left out is using let-chains, which does make the code less verbose:

```rust
fn foo(interaction: &ModalInteraction) {
    if let Some(Component::Label(label)) = interaction.data.components.iter().next()
        && let LabelComponent::SelectMenu(select_menu) = &label.component
        && let Some(value) = select_menu.values.first()
        && let Ok(channel_id) = GenericChannelId::from_str(value)
    {
        todo!()
    } else {
        todo!()
    }
}
```

In fact, I'd say I prefer this to chaining `.and_then` together. Ideally the additional (single) level of indentation this introduces would go away if the let-else and let-chain features could be combined, but in this case you could simply return `channel_id` from the block like you already do in your example.

calum4:
I had no idea let-chains existed! That's certainly significantly better than my let-else and match examples.

I'll leave this PR open regardless, but thanks again!

mkrasnitski:
> I had no idea let-chains existed!

They were only stabilized in Rust 1.88 in June, so have only been usable on stable for a little while.

PR #3469, opened 2026-01-07
Update cache for reaction events

paring-chan:
Added `CacheUpdate` implementation for:
- `ReactionAddEvent`
- `ReactionRemoveEvent`
- `ReactionRemoveEmojiEvent`
- `ReactionRemoveAllEvent`

PR #3470, opened 2026-01-07
Update cache for reaction events (next)

### LINKED ISSUE #3468: Message cache doesn't update reactions
(on `next` branch)
Reaction events don't make the cache updated, so the cache always returns content when created

paring-chan:
next branch port of #3469

## Changes

Converted message reactions to Vec<MessageRection> to modify

Added `CacheUpdate` implementation for:
- `ReactionAddEvent`
- `ReactionRemoveEvent`
- `ReactionRemoveEmojiEvent`
- `ReactionRemoveAllEvent`

Fixes #3468

### DIFF
--- a/src/cache/event.rs
+++ b/src/cache/event.rs
@@ -4,6 +4,14 @@ use std::num::NonZeroU16;
 use extract_map::entry::Entry;
 
 use super::{BaseGuildChannel, Cache, CacheUpdate, GenericChannelId, GuildThread};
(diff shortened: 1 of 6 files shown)


mkrasnitski:
At first glance looks ok, but you should update the `FullEvent` variants to contain the old data that gets returned from the cache update.

arqunis (approved):

PR #3479, opened 2026-01-28
Fix push_named_link_safe

### LINKED ISSUE #3440: push_named_link_safe is sometimes unsafe
If you try to push a name like `BROKEN[EXAMPLE` with `push_named_link_safe`, you will get something the following: `[BROKEN[EXAMPLE](https://www.example.com)` which will wrap the link around `EXAMPLE` like this: [BROKEN[EXAMPLE](https://www.example.com)

meditationmind:
Previously, only the closing square bracket was replaced, which would not actually prevent broken/altered link markdown.

While only the opening square bracket needs to be replaced to prevent breakage, I opted for replacing both opening and closing brackets. Consider, for example, the attempted legitimate usage: `[1] My Link [PDF]`. While an outcome of ` 1  My Link  PDF ` may be unexpected, there is minimal impact on meaning. ` 1] My Link  PDF]`, on the other hand, is arguably far more jarring.

Since escaping parentheses in the URL works as expected, I have opted for escaping the closing parenthesis, rather than replacing it with a space.

Fixes #3440.

### DIFF
--- a/src/utils/message_builder.rs
+++ b/src/utils/message_builder.rs
@@ -852,8 +852,8 @@ pub trait EmbedMessageBuilding {
     fn push_named_link<'a>(self, name: impl Into<Content<'a>>, url: impl Into<Content<'a>>)
     -> Self;
 
-    /// Pushes a named link intended for use in an embed, but with a normalized name to avoid
-    /// escaping issues.
+    /// Pushes a named link intended for use in an embed, normalizing content and escaping or
+    /// removing characters that would break the link markdown.
     ///
     /// Refer to [`Self::push_named_link`] for more information.
     ///
@@ -864,10 +864,10 @@ pub trait EmbedMessageBuilding {
     ///
     /// let content = MessageBuilder::new()
     ///     .push("A weird website name: ")
-    ///     .push_named_link_safe("Try to ] break links (](", "https://rust-lang.org")
+    ///     .push_named_link_safe("[Try] to [ break links ([(", "https://rust-lang.org")
     ///     .build();
     ///
-    /// assert_eq!(content, "A weird website name: [Try to   break links ( (](https://rust-lang.org)");
+    /// assert_eq!(content, "A weird website name: [ Try  to   break links ( (](https://rust-lang.org)");
     /// ```
     #[must_use]
     fn push_named_link_safe<'a>(
@@ -895,9 +895,9 @@ impl EmbedMessageBuilding for MessageBuilder {
         url: impl Into<Content<'a>>,
     ) -> Self {
         self.0.push('[');
-        self.push_safe_(name.into(), |c| normalize(c).replace(']', " "));
+        self.push_safe_(name.into(), |c| normalize(c).replace(['[', ']'], " "));
         self.0.push_str("](");
-        self.push_safe_(url.into(), |c| normalize(c).replace(')', " "));
+        self.push_safe_(url.into(), |c| normalize(c).replace(')', "\\)"));
         self.0.push(')');
 
         self
--- a/src/utils/message_builder.rs
+++ b/src/utils/message_builder.rs
@@ -867,7 +867,10 @@ pub trait EmbedMessageBuilding {
     ///     .push_named_link_safe("[Try] to [ break links ([(", "https://rust-lang.org")
     ///     .build();
     ///
-    /// assert_eq!(content, "A weird website name: [ Try  to   break links ( (](https://rust-lang.org)");
+    /// assert_eq!(
+    ///     content,
+    ///     "A weird website name: [ Try  to   break links ( (](https://rust-lang.org)"
+    /// );
     /// ```
     #[must_use]
     fn push_named_link_safe<'a>(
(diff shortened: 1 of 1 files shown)


Joshix-1:
Why not use url escapes and replace `)` with `%29`?

Joshix-1:
I think you need to url escape existing backslashes in urls.

I tried to create a named link to a random url with backslash I found and it seems to be the only way to make it work properly.

`https://wiki.selfhtml.org/wiki/\_(Sonderzeichen)`

meditationmind:
Thanks for the feedback!

I admit, this was only intended to fix #3440 so I didn't think about it too deeply. I did notice that the implementation is rather simplistic, though. I suppose it's a question of the intent of the function. Currently, it normalizes the content and prevents markdown breakage, and it does this successfully, I believe.

The issue with the backslash that you mentioned is also present when using the `push_named_link` variant, and also occurs when creating link markdown manually in chat, so escaping backslashes would change the nature/purpose of the method. Is it meant to simply prevent abuse/breakage, or to do extra processing to ensure that a URL is properly escaped/encoded?

For example, if I want to create a named link using your URL in Discord, I need to use `[Link](https://wiki.selfhtml.org/wiki/%5C_(Sonderzeichen))`. Otherwise, Discord will break the link. Knowing this, I would probably attempt to input the same thing using this method, given its description.

I'm fine with adding the percent-encoding, if that's what we want, but do we arbitrarily draw the line at backslashes, or do we percent-encode other characters too? There is a [percent-encoding](https://docs.rs/percent-encoding/latest/percent_encoding/) crate, but I doubt we want to bring in another dependency just for a coupe of methods that most likely see very little usage.

Joshix-1:
Are you using LLMs to write? If yes, please stop.


> Currently, it normalizes the content and prevents markdown breakage, and it does this successfully, I believe.

What if the URL ends with a backslash?

> is also present when using the `push_named_link` variant

That doesn't end with `_safe`

> Is it meant to simply prevent abuse/breakage, or to do extra processing to ensure that a URL is properly escaped/encoded?

What is the difference?

> I need to use `[Link](https://wiki.selfhtml.org/wiki/%5C_(Sonderzeichen))`

no `[Link](https://wiki.selfhtml.org/wiki/%5C_(Sonderzeichen\))` or  `[Link](https://wiki.selfhtml.org/wiki/%5C_(Sonderzeichen%29)`

> but do we arbitrarily draw the line at backslashes, or do we percent-encode other characters too? 

Only characters that break the markdown formatting or the URL should be escaped `\` and `)` come to mind.

> There is a [percent-encoding](https://docs.rs/percent-encoding/latest/percent_encoding/) crate

Yes, and it's already in the dependency tree of serenity. But I'm not sure a crate is required to replace 2? chars with their percent encoding.

meditationmind:
> Are you using LLMs to write? If yes, please stop.

I'm not sure how English usage has anything to do with a PR, but no, I do not use LLMs to write. I write and edit text professionally, and I've been using proper English since long before ChatGPT and the other popular LLMs even existed.

> What if the URL ends with a backslash?

Yes, good catch. That would indeed break the markdown.

> That doesn't end with _safe

Yes, true. However, all the other `_safe` functions only normalize content, so I suppose in that sense this one is already an outlier.

> What is the difference?

The difference is that you can have unbroken markdown, and yet Discord still breaks the link, as demonstrated in your backslash link example. However, as you demonstrated above, the backslash at the end does indeed break the markdown, so it needs to be escaped.

> no `[Link](https://wiki.selfhtml.org/wiki/%5C_(Sonderzeichen\))`

It works fine for me in Discord without escaping the closing parenthesis here. This is because a pair of opening and closing parentheses will not break the markdown. If it were a closing parenthesis only, it would break the markdown, which is why I added the escape for the closing parenthesis.

> Yes, and it's already in the dependency tree of serenity. But I'm not sure a crate is required to replace 2? chars with their percent encoding.

Good to know! And agreed. No need to use a crate if we only escape two characters.

> Only characters that break the markdown formatting or the URL should be escaped \ and ) come to mind.

Cool, I'll add percent-encoding for these two! Thanks again.

meditationmind:
Discovered a couple new things following testing:

- Percent-encoding before three forward slashes have been used breaks markdown.
- Percent-encoding only closing parentheses led to breakage in certain cases.

So the function now removes any parentheses or backslashes before three forward slashes have been used and percent-encodes all others.

This does result in one edge case. Using your backslash link as an example:
> `[Link](https://wiki.selfhtml.org\wiki/%5C_(Sonderzeichen))`

Typing this in chat will produce unbroken markdown with a working link, because Discord turns the backslash preceding "wiki" into a forward slash. This function removes that backslash, which results in working markdown, but a broken link.

However, it's unlikely that someone will be switching back and forth between forward and back slashes, so this seems a bit too niche to try to account for. Simply replacing backslashes with forward slashes until reaching three forward slashes results in unexpected behavior, because a backslash before a hyphen or period (`wiki.selfhtml\.org`) does not break the link in the client, while a forward slash does.

arqunis (approved):

PR #3482, opened 2026-02-06
Remove mentions of shuttle from README

SirzBenjie:
Shuttle has shut down: https://docs.shuttle.dev/docs/shuttle-shutdown

The docs should be updated to reflect this.


arqunis:
Merged as https://github.com/serenity-rs/serenity/commit/79e7872e376d4ab34ddd04d18e41ade92139316b

PR #3483, opened 2026-02-13
RadioGroup, CheckboxGroup and Checkbox Components

calum4:
[Changelog](https://docs.discord.com/developers/change-log#radio-groups%2C-checkbox-groups%2C-and-checkboxes-in-modals)

### DIFF
--- a/src/builder/create_components.rs
+++ b/src/builder/create_components.rs
@@ -548,6 +548,39 @@ impl<'a> CreateLabel<'a> {
         }
     }
 
+    /// Create a radio group with a specific label.
+    pub fn radio_group(label: impl Into<Cow<'a, str>>, radio_group: CreateRadioGroup<'a>) -> Self {
+        Self {
+            kind: ComponentType::Label,
+            label: label.into(),
+            description: None,
+            component: CreateLabelComponent::RadioGroup(radio_group),
+        }
+    }
+
+    /// Create a checkbox group with a specific label.
+    pub fn checkbox_group(
+        label: impl Into<Cow<'a, str>>,
+        checkbox_group: CreateCheckboxGroup<'a>,
+    ) -> Self {
+        Self {
+            kind: ComponentType::Label,
+            label: label.into(),
+            description: None,
+            component: CreateLabelComponent::CheckboxGroup(checkbox_group),
+        }
+    }
+
+    /// Create a checkbox with a specific label.
+    pub fn checkbox(label: impl Into<Cow<'a, str>>, checkbox: CreateCheckbox<'a>) -> Self {
+        Self {
+            kind: ComponentType::Label,
+            label: label.into(),
+            description: None,
+            component: CreateLabelComponent::Checkbox(checkbox),
+        }
+    }
+
     /// Sets the description of this component, which will display underneath the label text.
     pub fn description(mut self, description: impl Into<Cow<'a, str>>) -> Self {
         self.description = Some(description.into());
@@ -563,6 +596,9 @@ enum CreateLabelComponent<'a> {
     SelectMenu(CreateSelectMenu<'a>),
     InputText(CreateInputText<'a>),
     FileUpload(CreateFileUpload<'a>),
+    RadioGroup(CreateRadioGroup<'a>),
+    CheckboxGroup(CreateCheckboxGroup<'a>),
+    Checkbox(CreateCheckbox<'a>),
 }
 
 /// A builder for creating a file upload in a modal.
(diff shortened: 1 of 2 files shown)


src/builder/create_components.rs
```diff
@@ -611,6 +636,199 @@ impl<'a> CreateFileUpload<'a> {
...
+    kind: ComponentType,
+    custom_id: Cow<'a, str>,
+    options: Cow<'a, [CreateRadioGroupOption<'a>]>,
+    required: bool,
+}
+
+impl<'a> CreateRadioGroup<'a> {
+    /// Creates a builder with the given custom id and options.
+    ///
+    /// Note - As of 2025-02-13 a minimum of 2 and max of 10 radio group options is required,
```
meditationmind: Year should be `2026`.

src/builder/create_components.rs
```diff
@@ -611,6 +636,199 @@ impl<'a> CreateFileUpload<'a> {
...
+    options: Cow<'a, [CreateCheckboxGroupOption<'a>]>,
+    min_values: u8,
+    max_values: u8,
+    required: bool,
+}
+
+impl<'a> CreateCheckboxGroup<'a> {
+    /// Creates a builder with the given custom id and options.
+    ///
+    /// Note - As of 2025-02-13 a minimum of 0 and max of 10 checkbox group options is required,
```
meditationmind: Year should be `2026` and minimum should be `1`.

src/builder/create_components.rs
```diff
@@ -611,6 +636,199 @@ impl<'a> CreateFileUpload<'a> {
...
+    }
+
+    /// The minimum number of checkboxes that must be checked. Must be a number from 0 through 10,
+    /// and defaults to 1.
+    pub fn min_values(mut self, min_values: u8) -> Self {
+        self.min_values = min_values;
+        self
+    }
+
+    /// The maximum number of checkboxes that can be checked. Defaults to 1, but can be at most 10.
```
meditationmind: Did you intend to change the default behavior? According to the docs, this defaults to the number of options but you've set it to `1` by default. Would it be better to make this `Option` with `#[serde(skip_serializing_if = "Option::is_none")]` in the struct and set it to `None` in `CreateCheckboxGroup::new`?

src/builder/create_components.rs
```diff
@@ -611,6 +636,199 @@ impl<'a> CreateFileUpload<'a> {
...
+            min_values: 1,
+            max_values: 1,
+            required: true,
+        }
+    }
+
+    /// The minimum number of checkboxes that must be checked. Must be a number from 0 through 10,
+    /// and defaults to 1.
+    pub fn min_values(mut self, min_values: u8) -> Self {
+        self.min_values = min_values;
```
meditationmind: Docs say `required` must be `false` if this is set to `0`. Should we add a check here that does this?
```rust
if min_values == 0 { self.required = false }
```
Then we might also want a check in `CreateCheckboxGroup::required`, either to skip setting `required` to `true` if `self.min_values == 0`, or to set `self.min_values` back to the default of `1` if it has been set to `0`.

Or just note the requirement here and make it the dev's responsibility to choose a valid combination?

src/builder/create_components.rs
```diff
@@ -611,6 +647,228 @@ impl<'a> CreateFileUpload<'a> {
...
+    default: bool,
+}
+
+impl<'a> CreateCheckbox<'a> {
+    /// Creates a builder with the given custom id.
+    pub fn new(custom_id: impl Into<Cow<'a, str>>) -> Self {
+        Self {
+            kind: ComponentType::Checkbox,
+            custom_id: custom_id.into(),
+            default: false,
```
meditationmind: Any reason you defaulted this to `false`, instead of matching the API reference like you did with the CheckboxGroup options?

meditationmind (commented):
Haven't tested this yet, but a few comments below.

src/builder/create_components.rs
calum4: I will get used to it being 2026 in time for Jan 2027.... 

Thank you

src/builder/create_components.rs
calum4: I completely missed this.

In the case of `CreateCheckboxGroup::min_values`, I personally think setting required to false would be the preferential option since it just seems most logical to me.

```rust
    /// The minimum number of checkboxes that must be checked. Must be a number from 0 through 10,
    /// and defaults to 1.
    ///
    /// Note - In the event `0` is passed, `self.required` will be set to `false`.
    pub fn min_values(mut self, min_values: u8) -> Self {
        if min_values == 0 {
            self.required = false;
        }

        self.min_values = min_values;
        self
    }
```

As for `CreateCheckboxGroup::required`, I think `min_values` should be set to `1` in the event `true` is passed and `min_values` is currently `0`. Respectively, I think `min_values` should be set to `0` if `false` is passed.

```rust
    /// Whether the checkbox group is required.
    ///
    /// Note - if `false` is passed, `self.min_values` will be set to `0`. If `true` is passed and
    /// `self.min_values` is `0`, it will be set to `1`.
    pub fn required(mut self, required: bool) -> Self {
        if required && self.min_values == 0 {
            self.min_values = 1;
        } else if !required {
            self.min_values = 0;
        }

        self.required = required;
        self
    }
```

src/builder/create_components.rs
calum4: Was not my intention, evidently I need to read the docs better. (Apologies)

I agree with your suggestion.

src/builder/create_components.rs
calum4: Nope! Just another shoddy job unfortunately, will follow the API reference.

src/builder/create_components.rs
calum4: In light of your later comment pointing out I had deviated from the API reference, I realised I had here as well with both `min_values` and `required` being optional fields.

The code snippets from my previous comment don't reflect it, but I have changed them both to be optional fields as well in my upcoming commit.

calum4:
Returning to draft whilst I test the changes I just made.

calum4:
Should be all good now, thanks for the review!

src/builder/create_components.rs
```diff
@@ -611,6 +647,247 @@ impl<'a> CreateFileUpload<'a> {
...
+    /// The maximum number of checkboxes that can be checked. Defaults to the number of options
+    /// provided, and can be at most 10.
+    pub fn max_values(mut self, max_values: u8) -> Self {
+        self.max_values = Some(max_values);
+        self
+    }
+
+    /// Whether the checkbox group is required.
+    ///
+    /// Note - if `false` is passed, `self.min_values` will be set to `0`. If `true` is passed and
```
meditationmind: Remove ```if `false` is passed, `self.min_values` will be set to `0`.``` (Second sentence should stay.)

The function is not doing this, and more importantly, the current behavior is correct. It's a bit unintuitive, but the combination of `!required` and `min_values > 1` is valid. In such a case, the user can choose zero options, or they can choose *at least* `min_values`. (Forgot about this when I commented earlier!)

src/builder/create_components.rs
```diff
@@ -611,6 +647,247 @@ impl<'a> CreateFileUpload<'a> {
...
+    }
+
+    /// Whether the radio group is required.
+    pub fn required(mut self, required: bool) -> Self {
+        self.required = Some(required);
+        self
+    }
+}
+
+/// A builder for creating an option of a radio group component in a message
```
meditationmind: Missing a period here and after `Discord docs`. I think you copied the pattern from an existing builder, but that seems to deviate from the usual pattern.

src/builder/create_components.rs
```diff
@@ -611,6 +647,247 @@ impl<'a> CreateFileUpload<'a> {
...
+        if required && self.min_values == Some(0) {
+            self.min_values = Some(1);
+        }
+
+        self.required = Some(required);
+        self
+    }
+}
+
+/// A builder for creating an option of a checkbox group component in a message
```
meditationmind: Again, period here and after `Discord docs`.

meditationmind (commented):
Did some quick testing and everything seems to function properly!

Just a few small issues with comments below. Once fixed, everything should be good to go!

src/builder/create_components.rs
calum4: Yeah, this was the original functionality when I coded it but realised when testing it was wrong and evidently forgot to remove the comment when fixing the code.

src/builder/create_components.rs
calum4: Guilty as charged, fixed in all 3 of my cases

calum4:
> Did some quick testing and everything seems to function properly!
> 
> Just a few small issues with comments below. Once fixed, everything should be good to go!

Should be all fixed now, appreciate your diligence and apologies this was initially so sloppy!

meditationmind:
Awesome! LGTM!

arqunis (approved):

PR #3486, opened 2026-02-13
Update Discord docs links

meditationmind:
Discord migrated their API docs to Mintlify, which broke all the links to their docs in the Serenity docs.

This PR fixes the broken links, and I have manually verified that each one works. However, note that a significant number of the subheadings no longer have anchors, likely due to migration issues that will hopefully be resolved. For links that would be broken due to a missing subheading anchor, I have:
- linked to the closest relevant anchor (usually the section heading),
- and added a `// TODO: <expected subheading link>` comment on the line below.

This should make updating the links easier if/when the subheading anchors are restored.


meditationmind:
Fixed Clippy and CI failures.

arqunis (approved):

PR #3491, opened 2026-02-18
Restore Discord docs links

meditationmind:
Now that all the anchors have been added in the new Discord docs, this restores the links to their original intended targets.


meditationmind:
I fixed the Clippy failure by replacing the deprecated method, but since `current` is frozen, it may be preferable to fix it by adding `#[allow(deprecated)]` instead. If so, let me know and I'll amend the change.

arqunis (approved):

PR #3493, opened 2026-02-20
Refactor and simplify `ShardRunner`

mkrasnitski:
Moves responsibility for dispatching events to the `ShardManager` and adds a `ShardManagerMessage::DispatchEvent`, rendering most of the fields on `ShardRunner` no longer necessary and removing the need for `ShardRunnerOptions`. In order to facilitate this, I added a `ShardStageUpdateEvent` variant to `Event` - even though connection stage updates are purely internal, they are still a consequence of gateway activity. Note that sending a `Box<Event>` across a channel is cheap and doesn't move the `Event` on the heap.

Also, renames `ShardManagerMessage::Boot` to `RestartShard`, and refactors `ShardManager::runners` to store additional metadata about each runner that simplifies how the `Context` is initialized. By wrapping `ShardRunnerInfo` in `Arc<RwLock<T>>`, we can give a copy to both the runner and to the `Context` without cloning the entire `runner` map.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ tracing = { version = "0.1.44", features = ["log"] }
 serde = { version = "1.0.228", features = ["derive", "rc"] }
 url = { version = "2.5.7", features = ["serde"] }
 tokio = { version = "1.48.0", features = ["macros", "rt", "sync", "time", "io-util"] }
-futures = { version = "0.3.31", default-features = false, features = ["std"] }
+futures = { version = "0.3.32", default-features = false, features = ["std"] }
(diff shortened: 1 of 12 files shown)


meditationmind:
I'm not familiar enough with the internals to fully review the changes, but I did quickly test this with my bot and didn't run into any issues. The `ShardStageUpdateEvent` fired on startup, and I was able to access `ShardRunnerInfo`. I had to fix `Context::ping()` breakage in Poise, so I'll PR that once this is merged.

You can fix the failing CI test by bumping `futures` to version `0.3.32` in Cargo.toml.

arqunis (approved):

PR #3498, opened 2026-02-25
Update CV2 to match API

meditationmind:
This PR introduces several small fixes to V2 components:

- Remove the `unstable` flag from the `content` field of `TextDisplay`, since it is now [documented](https://docs.discord.com/developers/components/reference#text-display-text-display-structure).
- Align the `Separator` builder with the API by making `divider` optional.
- Remove the redundant `Spacing` enum and use the more appropriate `SeparatorSpacingSize` instead. This mismatch made using a `Separator` component to configure a `CreateSeparator` builder awkward since you had to convert between the types for no reason.
- Update `Button` builder to align with [requirements for Premium buttons](https://docs.discord.com/developers/components/reference#button-button-structure).
- Various documentation fixes.

### DIFF
--- a/examples/e01_basic_ping_bot/src/main.rs
+++ b/examples/e01_basic_ping_bot/src/main.rs
@@ -1,3 +1,4 @@
+#![expect(clippy::collapsible_match)]
 use std::sync::Arc;
 
 use serenity::async_trait;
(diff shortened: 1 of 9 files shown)


meditationmind:
Rebased; should be ready for review.

As noted in the final commit, I fixed Clippy by using `expect(clippy::collapsible_match)` to maintain consistent event handler structure in the examples. The errors are caused by [#16560: Enhance collapsible_match to cover if-elses](https://github.com/rust-lang/rust-clippy/pull/16560). If collapsing the matches is preferable, let me know and I can do that.

arqunis (approved):

PR #3505, opened 2026-03-02
Fix `is_new` returning `true` for existing guilds

arqunis:
Regression introduced in https://github.com/serenity-rs/serenity/pull/3457.

The `is_new` parameter indicates whether the guild notified by `GUILD_CREATE` is a brand new guild that the bot has just joined, or an existing one that it was in already. #3457 separated out the virtual event `CacheReady` for clarity, but this broke the parameter since to tell whether a guild is new or old it checks the list of unavailable guilds supplied from the `READY` event. When a `GUILD_CREATE` is received, the guild is removed from this list. The parameter was being calculated after the guild had already been removed from the list.

This merges the virtual events back as they were before in order to precisely control when the cache gets updated.

### DIFF
--- a/src/gateway/client/dispatch.rs
+++ b/src/gateway/client/dispatch.rs
@@ -556,33 +556,27 @@ mod tests {
     use extract_map::ExtractMap;
     use small_fixed_array::{FixedArray, FixedString};
 
-    use super::is_guild_new;
-    use crate::all::dispatch::update_cache_with_event;
-    use crate::all::{
-        ApplicationFlags,
-        ApplicationId,
-        Cache,
-        CurrentUser,
+    use super::{is_guild_new, update_cache_with_event};
+    use crate::cache::Cache;
+    use crate::gateway::client::FullEvent;
+    use crate::model::Timestamp;
+    use crate::model::application::{ApplicationFlags, PartialCurrentApplicationInfo};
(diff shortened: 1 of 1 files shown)


GnomedDev:
Can we add a regression test for this behaviour? This is the second time it has broken in this release cycle iirc

arqunis:
Added a test via synthetic `READY` and `GUILD_CREATE` events.

src/gateway/client/dispatch.rs
```diff
@@ -36,6 +38,24 @@ macro_rules! update_cache {
...
+    {
+        &ctx.cache
+    }
+
+    #[cfg(not(feature = "cache"))]
+    {
+        &()
+    }
+}
+
```
mkrasnitski, line 58: Why was this added? Isn't passing a `&Context` to `update_cache_with_event` fine?

arqunis: No, because then the function would become invalid if the `cache` feature is disabled. I changed the first parameter to accept a `Cache` directly, since `update_cache_with_event` did not use any other field from `Context`, which made using it in the test much easier; I did not want to bother having to create a `Context` in the test.

mkrasnitski: I see, then we can keep this for now.

mkrasnitski (approved):
LGTM

PR #3506, opened 2026-03-02
Revert ShardRunner refactor/simplification

GnomedDev:
This causes all events to be dispatched from a single thread, which caused a single core to hang at 100% when deploying to 1.1m guilds. I am not quite sure the exact performance hotspot, but the general idea of dispatching all events from the ShardManager is a bad idea.

This is running on TTS Bot now.


mkrasnitski:
I have a PR that supersedes this almost ready, which moves dispatch back into per-shard tasks but keeps the cleanup to `ShardRunner`.

GnomedDev:
Fair enough, just opening this in case others need it as a known good configuration.

PR #3511, opened 2026-03-08
Fix deserialization of User::collectibles

GnomedDev:
I do not understand why `#[serde(default)]` doesn't already do this.


KiloOscarSix:
Serde is strict, and the JSON spec defines null as a concrete type that's different to the absence of a key :sweat_smile:

arqunis (approved):

Joshix-1:
What about PresenceUser::collectibles?

PR #3512, opened 2026-03-08
Add edit roles positions

corentin-godefroy:
Add edit roles positions to reorder multiple roles at once according to https://docs.discord.com/developers/resources/guild#modify-guild-role-positions

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2227,7 +2227,7 @@ impl Http {
             headers: None,
             method: LightMethod::Patch,
             route: Route::GuildRoles {
-                guild_id
+                guild_id,
             },
             params: None,
         })
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -2221,7 +2221,7 @@ impl Http {
     ) -> Result<Vec<Role>> {
         let body = to_vec(value)?;
 
-        self.fire(Request{
+        self.fire(Request {
             body: Some(body),
             multipart: None,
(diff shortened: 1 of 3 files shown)


nwerosama:
Wouldn't this add a duplicated method? As there's already an existing method for that endpoint.

https://github.com/serenity-rs/serenity/blob/87b2bd1a773dbdbba9982a00b05896a12a2f6929/src/http/client.rs#L2175-L2210

corentin-godefroy:
@nwerosama 
> Wouldn't this add a duplicated method? As there's already an existing method for that endpoint.
> 
> https://github.com/serenity-rs/serenity/blob/87b2bd1a773dbdbba9982a00b05896a12a2f6929/src/http/client.rs#L2175-L2210

That method is not compliant to the discord specification, and multiple single role position modification can conduct to inconsistent positions. 
So this old method should be deprecated since I implement this version that match better with discord's specifications. I don't want to break things.

meditationmind:
This is already implemented on the `next` branch (see [#2954](https://github.com/serenity-rs/serenity/pull/2954)). The `current` branch is frozen since the next release will be `0.13.0`, and the only changes being accepted to `current` are bug fixes. I'm not a maintainer so I can't say whether this PR would be considered a fix, but if you want to use this endpoint (and other new features), it's probably best to start migrating to `next` (or `serenity-next` if you use Poise).

PR #3514, opened 2026-03-11
Add push/extend methods for CreateAllowedMentions users/roles

calum4:


KiloOscarSix:
Would the `Extend` trait be more appropriate? also provides `.extend_one()` to cover "push" as well

calum4:
> Would the `Extend` trait be more appropriate? also provides `.extend_one()` to cover "push" as well

The thought hadn't crossed my mind honestly.

Personally I prefer the current API given `CreateAllowedMentions` contains multiple collection types. I think it makes the usage of the `Extend` trait slightly ambiguous. Although I'll admit I'm not particularly passionate about that viewpoint.

[`Extend::extend_one()`](https://doc.rust-lang.org/stable/std/iter/trait.Extend.html#method.extend_one) is currently unstable though, so regardless I do think `push_*` ought to remain.

KiloOscarSix:
> > Would the `Extend` trait be more appropriate? also provides `.extend_one()` to cover "push" as well
> 
> The thought hadn't crossed my mind honestly.
> 
> Personally I prefer the current API given `CreateAllowedMentions` contains multiple collection types. I think it makes the usage of the `Extend` trait slightly ambiguous. Although I'll admit I'm not particularly passionate about that viewpoint.
> 
> [`Extend::extend_one()`](https://doc.rust-lang.org/stable/std/iter/trait.Extend.html#method.extend_one) is currently unstable though, so regardless I do think `push_*` ought to remain.

Likewise, I'm not particularly passionate either way; I mostly bought it up so it's on record.
Forgot `extend_one` was only on nightly, so totally agree on `push` remaining

arqunis (approved):

PR #3516, opened 2026-03-17
Remove default rustls provider

DPlayer234:
Changes the `reqwest` feature enabled by `rustls_backend` to `rustls-no-provider`, therefore requiring that a provider is otherwise selected by the user. This allows opting out of the default `rustls` provider, which is now `aws-lc-rs`.

The README is updated to include the simplest possible guidance: Depending on `rustls` and registering `aws-lc-rs` explicitly, or to use the `native_tls_backend`.

Most of the diff is due to adjusting the examples to include registering the crypto provider.

# Notes

Unless the user directly enables the `reqwest/rustls` feature, a provider has to be registered because `reqwest` does not use `rustls` defaults if `reqwest` is initialized first. Noting this quirk feels too specific and instead guiding the user towards always registering a provider is probably best.

Some examples use an outdated version of `sqlx` that still depends on `rustls` 0.21, so it technically doesn't inherit the provider registered for the bot in the example. Someone that knows `sqlx` better may want to update the examples using it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,7 +138,7 @@ tracing_instrument = ["tracing/attributes"]
 # Backends to pick from:
 # - Rustls Backends
 rustls_backend = [
-    "reqwest/rustls",
+    "reqwest/rustls-no-provider",
     "tokio-tungstenite/rustls-tls-webpki-roots",
 ]
 
--- a/README.md
+++ b/README.md
@@ -71,6 +71,9 @@ impl EventHandler for Handler {
 
 #[tokio::main]
 async fn main() {
+    // Register the crypto provider used for TLS by this application.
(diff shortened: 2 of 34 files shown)


mkrasnitski:
This is just my personal opinion, but I feel like needing to provide a magic incantation at the start of the program kind of conflicts with serenity's batteries-included approach. Is there any way to make custom crypto providers an opt-in feature of the `Client`?

DPlayer234:
This was made after a brief discussion here: https://discord.com/channels/381880193251409931/673965002805477386/1483407313095102609

I can make an alternate PR that simply provides an alternative `rustls_backend_no_provider` feature (essentially mirroring the underlying reqwest features), so there is still a provider by default but it lets the user opt out if there are conflicts.

mkrasnitski:
Closing in favor of #3517.

PR #3517, opened 2026-03-18
Allow opting out of a default Rustls crypto provider

DPlayer234:
Alternative to #3516.

Instead of completely removing the implicit default crypto provider when using the Rustls backend, instead provide an alternative feature set to opt out, essentially mirroring the features `reqwest` provides.

This adds the feature `rustls_backend_no_provider` which is identical to `rustls_backend` except that it includes no provider. Additionally adds `default_rustls_no_provider`, mirroring `default` with the `rustls_backend` replaced by `rustls_backend_no_provider` for convenience.

cc @mkrasnitski 

Note: The goal here is less to not to allow opting into custom crypto providers (that's already possible), it's to avoid compiling in the default when a different provider is included.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,7 +140,7 @@ tracing_instrument = ["tracing/attributes"]
 # - Rustls Backends
 rustls_backend = [
     "reqwest/rustls",
-    "rustls_backend_no_provider",
+    "tokio-tungstenite/rustls-tls-webpki-roots",
 ]
 rustls_backend_no_provider = [
     "reqwest/rustls-no-provider",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,7 @@ package = "http"
 [features]
 # Defaults with different backends
 default = ["default_no_backend", "rustls_backend"]
+default_rustls_no_provider = ["default_no_backend", "rustls_backend_no_provider"]
 default_native_tls = ["default_no_backend", "native_tls_backend"]
 
 # Serenity requires a backend, this picks all default features without a backend.
@@ -139,6 +140,10 @@ tracing_instrument = ["tracing/attributes"]
 # - Rustls Backends
 rustls_backend = [
     "reqwest/rustls",
+    "rustls_backend_no_provider",
+]
+rustls_backend_no_provider = [
+    "reqwest/rustls-no-provider",
     "tokio-tungstenite/rustls-tls-webpki-roots",
 ]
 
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -139,6 +140,10 @@ tracing_instrument = ["tracing/attributes"]
 # - Rustls Backends
 rustls_backend = [
     "reqwest/rustls",
+    "rustls_backend_no_provider",
```
mkrasnitski:
```suggestion
    "tokio-tungstenite/rustls-tls-webpki-roots",
```

In case the `rustls` and `rustls-no-provider` features of `reqwest` become mutually exclusive in the future, I think it's better to duplicate `tokio-tungstenite/rustls-tls-webpki-roots` here instead.

README.md
```diff
@@ -167,6 +168,8 @@ the HTTP functions.
 [lavalink-rs][project:lavalink-rs] or [Songbird][project:songbird] are recommended voice plugins.
 - **default_native_tls**: Default features but using `native_tls_backend`
 instead of `rustls_backend`.
+- **default_no_backend**: Defaults features, but does not enable any backend.
+- **default_rustls_no_provider**: Defaults features but using `rustls_backend_no_provider` instead of `rustls_backend`, requiring manual registration of a Rustls crypto provider.
```
```suggestion
- **default_rustls_no_provider**: Default features but using `rustls_backend_no_provider` instead of `rustls_backend`, requiring manual registration of a Rustls crypto provider.
- **default_no_backend**: Default features, but does not enable any backend.
```

Switching the line order reads a little better here. Also fixed a typo.

build.rs
```diff
@@ -1,8 +1,12 @@
-#[cfg(all(feature = "http", not(any(feature = "rustls_backend", feature = "native_tls_backend"))))]
+#[cfg(all(
+    feature = "http",
+    not(any(feature = "rustls_backend_no_provider", feature = "native_tls_backend"))
+))]
```
mkrasnitski, line 8:
```suggestion
#[cfg(all(
    feature = "http",
    not(any(
        feature = "rustls_backend",
        feature = "rustls_backend_no_provider",
        feature = "native_tls_backend"
    ))
))]
```

Given previous review, `rustls_backend` no longer implies `rustls_backend_no_provider`.

mkrasnitski (changes requested):
I agree this is the better approach than always requiring manual registration of a crypto provider. Just some small changes and this can get merged.

mkrasnitski (approved):

PR #3519, opened 2026-03-19
Add shared themes to Message

nwerosama:
Adds a new field to Message object for `shared_client_themes`.
Discord docs: https://docs.discord.com/developers/resources/message#shared-client-theme-object

I'm just going to open this as draft until I could try to get the `colours` field to work without serenity throwing the deserialization error, `Failed to deserialize event data: invalid type: string "CA48C8", expected u32 at line 1 column 1887`.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -139,6 +139,10 @@ pub struct Message {
     ///
     /// Only present in [`MessageCreateEvent`].
     pub poll: Option<Box<Poll>>,
+    /// The custom client-side theme attached to a message.
+    ///
+    /// Only present in [`MessageCreateEvent`]
+    pub shared_client_theme: Option<Box<SharedClientTheme>>,
 }
 
 #[cfg(feature = "model")]
@@ -1197,6 +1201,41 @@ pub struct MessagePin {
 pub struct MessagePinsPage {
     pub items: Vec<MessagePin>,
     pub has_more: bool,
+}
+
+enum_number! {
+    /// The mode of the shared client theme.
+    ///
+    /// [Discord docs](https://docs.discord.com/developers/resources/message#base-theme-types)
+    #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Deserialize, Serialize)]
+    #[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
+    #[non_exhaustive]
+    pub enum BaseTheme {
+        Unset = 0,
+        Dark = 1,
+        Light = 2,
+        Darker = 3,
+        Midnight = 4,
+        _ => Unknown(u8),
+    }
+}
+
+/// A shared client theme that is attached to [`Message`].
+///
+/// [Discord docs](https://docs.discord.com/developers/resources/message#shared-client-theme-object)
+#[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[non_exhaustive]
+pub struct SharedClientTheme {
+    /// The hexadecimal-encoded colours of the theme.
+    #[serde(rename = "colors", deserialize_with = "discord_colours")]
+    pub colours: FixedArray<Colour>,
+    /// The direction of the theme's colours.
+    pub gradient_angle: u16,
+    /// The intensity of the theme's colours.
+    pub base_mix: u16,
+    /// The mode of the theme.
+    pub base_theme: Option<BaseTheme>,
 }
 
 // all tests here require cache, move if non-cache test is added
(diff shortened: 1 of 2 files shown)


nwerosama:
So I can apparently use `deserialize_with = "discord_colours"`, but given how Discord populates the `colours` field, it's a valid hex code without the hashtag and the `discord_colours` would dump straight to "Invalid colour data" instead.

Comparing this with `burst_colours` for reference though.

src/model/channel/message.rs
```diff
@@ -1199,6 +1203,35 @@ pub struct MessagePinsPage {
...
+/// Base theme for the client
+#[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub enum BaseClientTheme {
+    #[default]
+    Dark = 1,
+    Light = 2,
+    Darker = 3,
+    Midnight = 4,
+}
```
meditationmind: Need to wrap this in `enum_number! {}` to deserialize the value. Also, docs don't specify a default value, so probably best to drop that. Docs don't mention the possibility of new variants, but might not hurt to add `non_exhaustive` anyway.

```suggestion
enum_number! {
    /// Base theme for the shared client theme.
    ///
    /// [Discord docs](https://docs.discord.com/developers/resources/message#base-theme-types).
    #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Deserialize, Serialize)]
    #[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
    #[non_exhaustive]
    pub enum BaseTheme {
        Dark = 1,
        Light = 2,
        Darker = 3,
        Midnight = 4,
        _ => Unknown(u8),
    }
}
```

src/model/channel/message.rs
```diff
@@ -1199,6 +1203,35 @@ pub struct MessagePinsPage {
...
+pub struct SharedClientTheme {
+    /// The hexadecimal-encoded colours of the theme
+    #[serde(rename = "colors", deserialize_with = "discord_colours")]
+    pub colours: FixedArray<Colour>,
+    /// The direction of the theme's colours
+    pub gradient_angle: u64,
+    /// The intensity of the theme's colours
+    pub base_mix: u64,
+    /// The mode of the theme
+    pub base_theme: Option<BaseClientTheme>,
```
meditationmind: Probably good to derive `Default` here and since int value max is 360/100, `u16` should be sufficient unless there's another reason to use `u64`.

```suggestion
#[derive(Clone, Debug, Default, Deserialize, Serialize)]
#[non_exhaustive]
pub struct SharedClientTheme {
    /// The hexadecimal-encoded colours of the theme
    #[serde(rename = "colors", deserialize_with = "discord_colours")]
    pub colours: FixedArray<Colour>,
    /// The direction of the theme's colours
    pub gradient_angle: u16,
    /// The intensity of the theme's colours
    pub base_mix: u16,
    /// The mode of the theme
    pub base_theme: Option<BaseTheme>,
```

meditationmind (commented):
To fix the `discord_colours` issue, change L246 of utils.rs from
```rust
let s = s.0.strip_prefix('#').ok_or_else(|| DeError::custom("Invalid colour data"))?;
```
to
```rust
let s = s.0.strip_prefix('#').unwrap_or(s.0.as_ref());
```

Alternatively, could create a separate function for this, but seems like overkill?

src/model/channel/message.rs
nwerosama: I had `u16` before, I'm not sure why they are `u64` now but will revert them back. I chose `BaseClientTheme` so it matches similarly to the docs and Shared, it sounds quite off if I rename it to `BaseTheme`.

nwerosama:
> To fix the `discord_colours` issue, change L246 of utils.rs from
> 
> ```rust
> let s = s.0.strip_prefix('#').ok_or_else(|| DeError::custom("Invalid colour data"))?;
> ```
> 
> to
> 
> ```rust
> let s = s.0.strip_prefix('#').unwrap_or(s.0.as_ref());
> ```
> 
> Alternatively, could create a separate function for this, but seems like overkill?

Yeah I think updating that would be better instead of creating another function for same purpose.

src/model/channel/message.rs
nwerosama: I cannot seem to wrap it inside `enum_number` for whatever the reason, seems to complain about the arguments. (`expected 0 arguments, found 1` for the `Dark` and others and `derive`?)

meditationmind: You need to add the `_ => Unknown(u8)` variant at the end because the macro is expecting it.

nwerosama: Yeah, I did that as well. Doesn't seem to make a difference, did compare with other enum_number usage and looks identical, yet still errors. I did copy the derive to ensure but still nothing.

nwerosama: Restarted rust-analyzer and it went away, welp.

meditationmind:
LGTM

arqunis (approved):

PR #3520, opened 2026-03-21
Document CacheRef deadlock risk across await points

### LINKED ISSUE #3425: `CacheRef` does not document how it should not be carried over an await point
If a channel is created using a `GuildRef` from a task other than the gateway task (e.g. an event handler task), a deadlock can ensue. This is because `GuildRef` wraps `Ref`. If the `GuildRef` is still being held when the Channel Create event is received, then serenity will attempt to update the `Guild` in the cache with new channels. This will call `get_mut` on the `guilds` `DashMap`, which [may deadlock if a `Ref` is already held](https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html#method.get_mut), thus causing a deadlock if the two tasks happen to be running on the same system thread at that particular time (which depends on tokio internals). 

The root cause here is attempting to hold an equivalent of a sync lock across `await` boundaries. I don't really see a good fix for the root cause, but it may be good to keep an eye on https://github.com/xacrimon/dashmap/issues/150.

Sim-hu:
Closes #3425.

`CacheRef` holds a `DashMap` shard lock for its lifetime, so holding it across an `.await` can deadlock if a gateway event tries to update the same shard. The type is `!Send` which prevents this in most cases, but contexts that allow `!Send` futures (like the `#[tokio::main]` top-level future) aren't caught by the compiler.

This adds a doc comment to `CacheRef` explaining the deadlock risk and recommending users extract the data they need before awaiting. Also includes a short code example showing the suggested pattern.

### DIFF
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -64,6 +64,34 @@ enum CacheRefInner<'a, K, V, T> {
     ReadGuard(parking_lot::RwLockReadGuard<'a, V>),
 }
 
+/// A reference to data in the [`Cache`].
+///
+/// This type wraps a reference to cached data and dereferences to `V`. It is deliberately
+/// `!Send` to prevent holding it across `.await` points in most situations, since doing so can
+/// cause deadlocks: the underlying [`DashMap`] shard lock is held for the lifetime of this
+/// reference, so any cache update (e.g. from a gateway event) that needs the same shard will
+/// block indefinitely.
+///
+/// # Deadlock warning
+///
+/// Although `CacheRef` is `!Send`, which makes the compiler reject it across `.await` in
+/// spawned tasks, there are contexts where `!Send` futures are permitted (notably the
+/// `#[tokio::main]` top-level future). In those contexts the compiler will **not** prevent you
+/// from holding a `CacheRef` over an `.await`, but a deadlock can still occur.
+///
+/// As a general rule, never hold a `CacheRef` across an `.await` point. Instead, clone or copy
+/// the data you need and drop the reference before awaiting:
+///
+/// ```rust,no_run
+/// # use serenity::cache::Cache;
+/// # use serenity::model::prelude::*;
+/// # fn example(cache: &Cache, guild_id: GuildId) {
+/// // Good: extract what you need, then drop the reference.
+/// let channel_count = cache.guild(guild_id).map(|g| g.channels.len());
+/// # }
+/// ```
+///
+/// [`DashMap`]: dashmap::DashMap
 pub struct CacheRef<'a, K, V, T = ()> {
     inner: CacheRefInner<'a, K, V, T>,
     phantom: std::marker::PhantomData<*const NotSend>,
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -68,11 +68,8 @@ enum CacheRefInner<'a, K, V, T> {
 ///
 /// This type wraps a reference to cached data and dereferences to `V`. It is deliberately
 /// `!Send` to prevent holding it across `.await` points in most situations, since doing so can
-/// cause deadlocks: the underlying [`DashMap`] shard lock is held for the lifetime of this
-/// reference, so any cache update (e.g. from a gateway event) that needs the same shard will
-/// block indefinitely.
-///
-/// # Deadlock warning
+/// cause deadlocks: the underlying read lock is held for the lifetime of this reference, so any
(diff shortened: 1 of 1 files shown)


src/cache/mod.rs
```diff
@@ -63,6 +63,34 @@ enum CacheRefInner<'a, K, V, T> {
     ReadGuard(parking_lot::RwLockReadGuard<'a, V>),
 }
 
+/// A reference to data in the [`Cache`].
+///
+/// This type wraps a reference to cached data and dereferences to `V`. It is deliberately
+/// `!Send` to prevent holding it across `.await` points in most situations, since doing so can
+/// cause deadlocks: the underlying [`DashMap`] shard lock is held for the lifetime of this
+/// reference, so any cache update (e.g. from a gateway event) that needs the same shard will
+/// block indefinitely.
```
GnomedDev: I feel like this duplicates the section below, and exposes internal details that don't need to be (DashMap shards). If the mechanism in which deadlocks occur is going to be documented, I would talk of it in abstract terms of RwLocks.

GnomedDev (changes requested):

Sim-hu:
Good point — removed the DashMap internals and the duplicate heading. Now just talks about the read lock in abstract terms.

GnomedDev (approved):

GnomedDev (changes requested):
Actually, why is this targeting the `next` branch? This should be okay to apply to `current`.

Sim-hu:
Changed base to current.

GnomedDev:
You need to rebase now to drop all the commits from next.

Sim-hu:
Rebased onto current, dropped the next commits.

src/cache/mod.rs
```diff
@@ -64,6 +64,29 @@ enum CacheRefInner<'a, K, V, T> {
     ReadGuard(parking_lot::RwLockReadGuard<'a, V>),
 }
 
+/// A reference to data in the [`Cache`].
+///
+/// This type wraps a reference to cached data and dereferences to `V`. It is deliberately
+/// `!Send` to prevent holding it across `.await` points in most situations, since doing so can
+/// cause deadlocks: the underlying read lock is held for the lifetime of this reference, so any
+/// cache update (e.g. from a gateway event) that needs the same lock will block indefinitely.
```
mkrasnitski:
```suggestion
/// This type wraps a reference to cached data and dereferences to `V`. It is deliberately marked
/// `!Send` to prevent holding it across an `.await` point. Doing so can cause a deadlock: the
/// underlying read lock is held for the lifetime of the `CacheRef`, so any task that tries taking
/// the same lock while it is still held (i.e. across an `.await` point) will block indefinitely.
```

src/cache/mod.rs
```diff
@@ -64,6 +64,29 @@ enum CacheRefInner<'a, K, V, T> {
...
+///
+/// Although `CacheRef` is `!Send`, which makes the compiler reject it across `.await` in
+/// spawned tasks, there are contexts where `!Send` futures are permitted (notably the
+/// `#[tokio::main]` top-level future). In those contexts the compiler will **not** prevent you
+/// from holding a `CacheRef` over an `.await`, but a deadlock can still occur.
```
mkrasnitski:
```suggestion
/// In general, the compiler will reject holding a `CacheRef` across an `.await` point in spawned
/// tasks because it is `!Send`. Notably, however, the `#[tokio::main]` top-level future permits
/// `.await`-ing `!Send` futures in its body. In this case the compiler will **not** prevent you
/// from holding a `CacheRef` over an `.await` point, and a deadlock may still occur.
```

src/cache/mod.rs
```diff
@@ -64,6 +64,29 @@ enum CacheRefInner<'a, K, V, T> {
...
+///
+/// As a general rule, never hold a `CacheRef` across an `.await` point. Instead, clone or copy
+/// the data you need and drop the reference before awaiting:
```
mkrasnitski:
```suggestion
/// As a rule of thumb, never hold a `CacheRef` across an `.await` point. Instead, clone or copy the
/// data you need and drop the reference before awaiting:
```

mkrasnitski (changes requested):
Adjusted formatting (line length) and wording.

Sim-hu:
Applied all three suggestions — wording and line length look much better now, thanks.

mkrasnitski (approved):

PR #3521, opened 2026-03-21
Update docs for `User::global_name`

nwerosama:
It was removed from the Discord docs as bots do not have that field set.
https://github.com/discord/discord-api-docs/commit/49bf68b66884c858f2027f9adbea7522b6693b0f


GnomedDev (approved):
LGTM

PR #3523, opened 2026-03-24
Impl From/TryFrom for message components

meditationmind:
This PR implements the From/TryFrom traits for all message components lacking the implementation.

Currently, working with existing CV2 messages (editing, etc.) can be quite painful since the conversion from each `Component` to its builder must be done manually. This makes the process much easier.

In accordance with the [std documentation](https://doc.rust-lang.org/std/convert/trait.From.html), I only implemented `From` for lossless conversions. Furthermore, while `SectionComponent` and `SectionAccessory` would currently be lossless conversions, I chose to impl `TryFrom` instead since the [Discord docs](https://docs.discord.com/developers/components/reference#section-section-structure) note that these types should not be hardcoded to contain only the currently available components.

Impls have not been added for modal components as I couldn't think of a valid use case and the conversions would be extremely lossy.

The PR also amends an outdated doc comment for `CreateInputText`.

### DIFF
--- a/src/builder/create_components.rs
+++ b/src/builder/create_components.rs
@@ -48,6 +48,28 @@ impl serde::Serialize for CreateActionRow<'_> {
     }
 }
 
+impl TryFrom<ActionRow> for CreateActionRow<'_> {
+    type Error = ComponentConversionError;
+
+    /// Attempts to convert to this type from an [`ActionRow`].
+    ///
+    /// # Errors
+    /// Returns [`ComponentConversionError::LossyComponent`] if the [`ActionRow`] contains a
+    /// non-string [`SelectMenu`], as lossless conversion is not possible.
+    fn try_from(action_row: ActionRow) -> StdResult<Self, Self::Error> {
+        let mut buttons: Vec<CreateButton> = Vec::with_capacity(5);
+        for component in action_row.components {
+            match component {
+                ActionRowComponent::Button(button) => buttons.push(button.into()),
+                ActionRowComponent::SelectMenu(select_menu) => {
+                    return Ok(Self::SelectMenu(select_menu.try_into()?));
+                },
+            }
+        }
+        Ok(Self::Buttons(buttons.into()))
+    }
+}
+
 /// A builder for creating components in a structured way.
 ///
 /// This enum supports both V1 and V2 components, with the exception of `ActionRow`, which is a V1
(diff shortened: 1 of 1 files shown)


src/builder/create_components.rs
```diff
@@ -40,6 +41,23 @@ impl serde::Serialize for CreateActionRow<'_> {
     }
 }
 
+impl TryFrom<ActionRow> for CreateActionRow<'_> {
+    type Error = ComponentConversionError;
+
+    fn try_from(action_row: ActionRow) -> StdResult<Self, Self::Error> {
+        let mut buttons: Vec<CreateButton> = Vec::new();
```
KiloOscarSix: Micro Optimisation: `Vec::with_capacity`

KiloOscarSix (changes requested):
`TryFrom<SelectMenu>` feels rather busy with similar code on each match arm. Wonder if a generic helper function can help

src/builder/create_components.rs
```diff
@@ -157,6 +220,25 @@ pub enum CreateSectionComponent<'a> {
...
+            // Discord indicates that others may be added in the future. If the below arm is
+            // matched, a new component has been added and this code should be updated.
+            //
+            // [Discord docs](https://docs.discord.com/developers/components/reference#section-section-structure)
+            #[expect(unused)]
+            _ => Err(ComponentConversionError::UnknownType),
+        }
+    }
+}
+
```
mkrasnitski: This should really just be an infallible conversion, and likewise for `SectionAccessory`. The wildcard branch in the match will in fact ***never*** be matched, precisely because the definition of `SectionComponent` is the following:

```rust
#[non_exhaustive]
pub enum SectionComponent {
    TextDisplay(TextDisplay),
}
```

If we really want to relax how strict we are about accommodating new variants being added in the future, then we should add an `Unknown` variant. However, I think that the `Deserialize` impl for `SectionComponent` and `SectionAccessory` provides enough information to know when something has gone wrong. In particular, deserialization errors will contain the `ComponentType` of the unknown new variant, and will be emitted via `warn!` when the error is passed to `Shard::handle_event` [here](https://github.com/serenity-rs/serenity/blob/next/src/gateway/sharding/mod.rs#L446-L449).

src/builder/create_components.rs
```diff
@@ -196,6 +287,25 @@ pub enum CreateSectionAccessory<'a> {
...
+            // the below arm is matched, a new accessory has been added and this code should be
+            // updated.
+            //
+            // [Discord docs](https://docs.discord.com/developers/components/reference#section-section-structure)
+            #[expect(unused)]
+            _ => Err(ComponentConversionError::UnknownType),
+        }
+    }
+}
+
```
mkrasnitski: See above.

src/builder/create_components.rs
```diff
@@ -95,6 +113,34 @@ pub enum CreateComponent<'a> {
...
+                Ok(CreateComponent::MediaGallery(media_gallery.into()))
+            },
+            Component::File(file_component) => {
+                Ok(CreateComponent::File(file_component.try_into()?))
+            },
+            Component::Separator(separator) => Ok(CreateComponent::Separator(separator.into())),
+            Component::Container(container) => {
+                Ok(CreateComponent::Container(container.try_into()?))
+            },
+            Component::Label(_) => Err(ComponentConversionError::ModalComponent),
```
mkrasnitski: Why does this return an error if `CreateComponent::Label` is a thing?

src/builder/create_components.rs
```diff
@@ -1232,6 +1474,82 @@ impl<'a> CreateSelectMenu<'a> {
...
+                default_channels: {
+                    let mut channels = Vec::with_capacity(select_menu.values.len() as usize);
+                    for value in &select_menu.values {
+                        let channel_id = GenericChannelId::from_str(value.as_str())?;
+                        channels.push(channel_id);
+                    }
+                    if channels.is_empty() { None } else { Some(channels.into()) }
+                },
+            },
+            _ => return Err(ComponentConversionError::UnknownType),
```
mkrasnitski: Unfortunate that this wildcard needs to be here. We should probably make `SelectMenu` an enum at this point - I'll look into implementing that. Then, the match could be exhaustive and the `values` field would be type-safe, and this conversion could be made infallible.

mkrasnitski (changes requested):

src/builder/create_components.rs
meditationmind: Nice catch, thanks.

src/builder/create_components.rs
meditationmind: That makes sense, thanks. Updated all the section components to make them infallible.

src/builder/create_components.rs
meditationmind: As I mentioned in the description, I couldn't think of a use case for converting modal components to their builders, and doing so would be extremely lossy.

For example, with `InputText`, you get the submitted value, but you don't know if it was originally included as default. You also lose the style. Similar situation for `Checkbox`. For `CheckboxGroup` and `RadioGroup`, you get the submitted values, but you lose the original options, so the builder output may look nothing like the original. You also lose info like `max_values`, `min_values`, `required`, etc.

I don't mind implementing the modal components too, but the builders will end up being very minimal shells, with very little of the original data.

src/builder/create_components.rs
meditationmind: I simplified the impl here, but when doing a quick test, I realized we're looking at the same issue here as with the modal components. The values from the component being converted are not the same as the defaults that are being populated here. Not sure how we should approach this. Could either ignore all the values, or proceed with populating the defaults with the submitted values. Will need to note the behavior, either way. Thoughts?

mkrasnitski: As it turns out, if you look at the [String Select Interaction Response](https://docs.discord.com/developers/components/reference#string-select), then you'll see that `placeholder`, `min_values`, `max_values`, `required`, and `disabled` are not returned in the response either, meaning they'll always be `None` in the `SelectMenu` struct. The same goes for all the other select menu types. So, idk if it makes sense to have the impl in the first place.

meditationmind: Yeah, agreed. I initially overlooked the select menu issues because they weren't part of my use case. I grouped select menus with modal components to throw a `LossyComponent` error and allow the dev to decide how to handle them. Also documented the errors on the relevant impls.

meditationmind:
Re-implemented TryFrom for select menus, following changes in [#3531](https://github.com/serenity-rs/serenity/pull/3531).

arqunis (approved):

PR #3531, opened 2026-04-03
Remove component fields which aren't included in interaction responses

mkrasnitski:
These fields will never be `Some`, and so do not serve a purpose. Also, flattened `SelectMenu` into its separate variants for clarity, and fixed implementations of `Serialize` by making some enums untagged.

### DIFF
--- a/src/model/application/component.rs
+++ b/src/model/application/component.rs
@@ -46,6 +46,7 @@ enum_number! {
 ///   other message limitations are applied.
 #[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
 #[derive(Clone, Debug, Serialize)]
+#[serde(untagged)]
 #[non_exhaustive]
 pub enum Component {
     ActionRow(ActionRow),
@@ -115,6 +116,7 @@ pub struct Section {
 /// [Discord docs](https://docs.discord.com/developers/components/reference#section-section-child-components)
 #[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
 #[derive(Clone, Debug, Serialize)]
+#[serde(untagged)]
 #[non_exhaustive]
 pub enum SectionComponent {
     TextDisplay(TextDisplay),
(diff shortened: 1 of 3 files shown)


src/model/application/component.rs
```diff
@@ -232,9 +236,6 @@ pub struct TextDisplay {
     /// Always [`ComponentType::TextDisplay`]
     #[serde(rename = "type")]
     pub kind: ComponentType,
-    /// The text content of this component.
-    #[serde(default)]
-    pub content: FixedString<u16>,
```
KiloOscarSix, line 237: Wasn't there a discussion where `TextDisplay::content` is returned on a message interaction?

meditationmind: Yes, this should not be removed, as it would render the content of the `TextDisplay` inaccessible.

mkrasnitski: Sorry, I was going off the discord documentation and forgot that message components return the content field whereas modals do not. This is undocumented, so someone should make a PR to the discord docs.

mkrasnitski: @meditationmind When you removed the unstable flag from `TextDisplay::content`, were you going off the "Text Display Structure" table in the [docs](https://docs.discord.com/developers/components/reference#text-display)? The correct table would have been "Text Display Interaction Response Structure".

meditationmind: Hmm, yeah, I was, but taking another look now, I don't think doing so was a mistake. Rather, I think the original assumption that it was undocumented was mistaken.

We're treating `TextDisplay` like an interactive component, but it's not. `TextDisplay` is a content-type component, so there is technically no interaction. However, since a modal can contain a `TextDisplay` only, there needs to be an interaction response. (Notice the message-only content-type components do not have interaction response structures documented.) `TextDisplay` is just a bit odd since it can be used in both messages and modals, but I don't think our model should be based on the interaction response structure for a non-interactive component.

mkrasnitski: Yeah, I'm also realizing that the fields I removed from select menus do actually appear when they are sent in messages. I'll have to rethink my approach here. Seems like I was interpreting the docs wrong, as the "Interaction Response" structure essentially applies only in modal submits.

mkrasnitski:
Rethought my approach and after some testing, what I have now should reflect better what data gets sent. Ideally there would be separate types for components in messages versus modal interactions, however that should be done separately, if at all.

KiloOscarSix (approved):

arqunis (approved):

PR #3532, opened 2026-04-08
bumped tokio-tungstenite to 0.29

MoistPyro:
updating this dependancy to newest version.


MoistPyro:
MSRV failure might be false a flag, as we already have 2.1.1 in deps.
Clippy failure is code I have not touched.
Format failure I can easily fix.

nwerosama:
I brought up the MSRV issue in CI the other day outside of GitHub but nothing happened since then. While I'm here, I'm not sure what is the reason for enabling the `url` feature in this crate. (Haven't really tested the PR myself to confirm though)

MoistPyro:
URL is an optional dependency in tungstenite 0.29, so the connect_async_with_config function is now behind a feature flag

meditationmind:
Except for bug fixes, the `current` branch is frozen and will not be seeing another release. tokio-tungstenite has already been bumped in the `next` branch.

PR #3538, opened 2026-04-18
Reduce model sizes

GnomedDev:
This reduces the type sizes of member counts to u32, and some other minor improvements, and also reduces the flags to their minimum sizes which causes flag additions to become breaking changes if they overflow but new features needing a jump to next is nothing new for serenity.

| Model | Before | After |
| -- | -- | -- |
| Guild | 848 | 752 |
| GuildChannel | 172 | 164 |  
| GuildThread | 156 | 152 |

This should save 500mb of ram for TTS Bot.


KiloOscarSix:
From memory, Serenity doesn't have any other type aliases like this - `type MemberCount = NonMaxU32;` instead using the plain type. 
For future contributions, does Serenity want to encourage such aliasing or prefer using the plain type?

*(Not that I can think of any downside in using type aliases)*

GnomedDev:
I think doing these aliases in strategic positions is useful, as it stops assumptions like "there will never be more than u32::MAX members in a guild" getting desynced across the library. I don't think they should be publically exposed though.

arqunis (approved):

PR #3541, opened 2026-04-19
feat(model): add recurrence_rule to scheduled events

Dramex:
Exposes Discord's guild scheduled event recurrence rule on `ScheduledEvent`.

- `ScheduledEvent::recurrence_rule: Option<RecurrenceRule>`.
- `RecurrenceRule` mirroring the API shape (`start` / `end`, `frequency`, `interval`, `by_weekday`, `by_n_weekday`, `by_month`, `by_month_day`, `by_year_day`, `count`).
- `RecurrenceRuleFrequency` (Yearly / Monthly / Weekly / Daily), `RecurrenceRuleWeekday` (Monday-Sunday, with Discord's `0 = Monday` numbering), and `RecurrenceRuleMonth` enums.
- `RecurrenceRuleNWeekday` helper struct for "the Nth weekday of a month" entries.

Docs: https://discord.com/developers/docs/resources/guild-scheduled-event#guild-scheduled-event-recurrence-rule-object

`cargo check --all-features` passes.

### DIFF
--- a/src/model/guild/scheduled_event.rs
+++ b/src/model/guild/scheduled_event.rs
@@ -1,4 +1,4 @@
-use nonmax::NonMaxU64;
+use nonmax::{NonMaxU32, NonMaxU64};
 
 use crate::model::prelude::*;
 
@@ -141,21 +141,21 @@ pub struct RecurrenceRule {
     pub interval: u32,
     /// Weekdays the event recurs on.
     #[serde(default)]
-    pub by_weekday: Option<Vec<RecurrenceRuleWeekday>>,
+    pub by_weekday: Option<FixedArray<RecurrenceRuleWeekday, u8>>,
     /// Specific weekdays within a month the event recurs on.
     #[serde(default)]
-    pub by_n_weekday: Option<Vec<RecurrenceRuleNWeekday>>,
+    pub by_n_weekday: Option<FixedArray<RecurrenceRuleNWeekday, u8>>,
     /// Months within a year the event recurs on (1-12).
     #[serde(default)]
-    pub by_month: Option<Vec<RecurrenceRuleMonth>>,
+    pub by_month: Option<FixedArray<RecurrenceRuleMonth, u8>>,
     /// Days within a month the event recurs on (1-31).
     #[serde(default)]
-    pub by_month_day: Option<Vec<u8>>,
+    pub by_month_day: Option<FixedArray<u8, u8>>,
     /// Days within a year the event recurs on (1-364).
     #[serde(default)]
-    pub by_year_day: Option<Vec<u16>>,
+    pub by_year_day: Option<FixedArray<u16, u16>>,
     /// Total amount of times the event is allowed to recur before stopping.
-    pub count: Option<u32>,
+    pub count: Option<NonMaxU32>,
 }
 
 enum_number! {
--- a/src/model/guild/scheduled_event.rs
+++ b/src/model/guild/scheduled_event.rs
@@ -53,6 +53,8 @@ pub struct ScheduledEvent {
     pub user_count: Option<NonMaxU64>,
     /// The hash of the event's cover image, if present.
     pub image: Option<ImageHash>,
+    /// The recurrence rule of the event, if the event recurs.
+    pub recurrence_rule: Option<RecurrenceRule>,
 }
 
 impl extract_map::ExtractKey<ScheduledEventId> for ScheduledEvent {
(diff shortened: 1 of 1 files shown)


KiloOscarSix:
I don't want you to feel accused, but I do have a feeling your PRs are AI-generated. 

Not that it automatically means it's wrong, however, if this is generated by AI (Claude), it is common courtesy to add a footnote about how AI was used in generating this PR.

Dramex:
> I don't want you to feel accused, but I do have a feeling your PRs are AI-generated.
> 
> Not that it automatically means it's wrong, however, if this is generated by AI (Claude), it is common courtesy to add a footnote about how AI was used in generating this PR.


What part specifically looks off? The struct is just the API object one to one, Monday=0 is Discord's numbering not mine, and the enums use the same enum_number! pattern as the rest of the repo.

src/model/guild/scheduled_event.rs
```diff
@@ -120,3 +122,110 @@ enum_number! {
...
+pub struct RecurrenceRule {
+    /// Starting time of the recurrence interval.
+    pub start: Timestamp,
+    /// Ending time of the recurrence interval, if any.
+    pub end: Option<Timestamp>,
+    /// How often the event occurs.
+    pub frequency: RecurrenceRuleFrequency,
+    /// Spacing between recurrences, in units of [`Self::frequency`]. Must be `1` except for
+    /// weekly events, where `2` is also valid to indicate bi-weekly recurrence.
+    pub interval: u32,
```
GnomedDev, line 141: This should not be a u32 for a type with two valid values.

src/model/guild/scheduled_event.rs
```diff
@@ -120,3 +122,110 @@ enum_number! {
...
+    pub by_n_weekday: Option<Vec<RecurrenceRuleNWeekday>>,
+    /// Months within a year the event recurs on (1-12).
+    #[serde(default)]
+    pub by_month: Option<Vec<RecurrenceRuleMonth>>,
+    /// Days within a month the event recurs on (1-31).
+    #[serde(default)]
+    pub by_month_day: Option<Vec<u8>>,
+    /// Days within a year the event recurs on (1-364).
+    #[serde(default)]
+    pub by_year_day: Option<Vec<u16>>,
```
GnomedDev: If this is targeting next, which I disagree with, this should use FixedArray with a sensible LenT generic.

It might actually be more sensible to do some deserialisation here to convert the vector of enums to a bitflags value.

src/model/guild/scheduled_event.rs
```diff
@@ -120,3 +122,110 @@ enum_number! {
...
+    /// Total amount of times the event is allowed to recur before stopping.
+    pub count: Option<u32>,
```
GnomedDev: Does this need to be a u32? If so, make it a NonMaxU32 to save the Option::None tag.

src/model/guild/scheduled_event.rs
```diff
@@ -120,3 +122,110 @@ enum_number! {
...
+
+/// Specific weekday within a month for a recurrence rule (e.g. "the 2nd Tuesday").
+///
+/// [Discord docs](https://docs.discord.com/developers/resources/guild-scheduled-event#guild-scheduled-event-recurrence-rule-object-guild-scheduled-event-recurrence-rule-nweekday-structure).
+#[cfg_attr(feature = "typesize", derive(typesize::derive::TypeSize))]
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+#[non_exhaustive]
+pub struct RecurrenceRuleNWeekday {
+    /// The week to recur on (1-5).
+    pub n: u8,
```
GnomedDev, line 228: NonMaxU8

GnomedDev (changes requested):
Some review comments from me while laid in bed. Do not consider this approval to continue using generative AI, as slop cannot be merged for licensing reasons.

Dramex:
> Some review comments from me while laid in bed. Do not consider this approval to continue using generative AI, as slop cannot be merged for licensing reasons.

Just trying to help close a gap in the model. If the PR isn't welcome, feel free to close, the diff is small enough that nothing is blocked by leaving it on the floor.

src/model/guild/scheduled_event.rs
Dramex: > This should not be a u32 for a type with two valid values.

Discord's schema types this as `integer` and currently only accepts 1 or 2, but that's a semantic constraint they've set and can relax. If they allow interval=3 for a new frequency later, a `bool`-typed field silently breaks at deserialize time. u32 matches the wire format and leaves us forward-compatible. Happy to revisit if the spec ever hardens.

src/model/guild/scheduled_event.rs
Dramex: > If this is targeting next, which I disagree with, this should use FixedArray with a sensible LenT generic.

Switched all five by_* fields to FixedArray, u8 LenT for the small ones (by_weekday / by_n_weekday / by_month / by_month_day) and u16 for by_year_day. Left the bitflags conversion out of scope, I'd rather land the straight API shape first and let a follow-up layer pack it if needed.

src/model/guild/scheduled_event.rs
Dramex: > Does this need to be a u32? If so, make it a NonMaxU32 to save the Option::None tag.

Fair, switched to `Option<NonMaxU32>`. Matches the `user_count: Option<NonMaxU64>` niche already on `ScheduledEvent` itself.

src/model/guild/scheduled_event.rs
Dramex: > NonMaxU8

Left as `pub n: u8`. It's a required field, not wrapped in any `Option<NonMaxU8>`, so the niche saves zero bytes here. The only effect would be rejecting `n == 255`, which Discord won't send for a 1-to-5 value. Happy to change if there's a use site I'm missing.

GnomedDev:
Obvious AI pfp, canned and evasive responses to review.

PR #3543, opened 2026-04-20
Added AI usage guidelines in CONTRIBUTING.md

KiloOscarSix:
As Gnome stated here: https://discord.com/channels/381880193251409931/673965002805477386/1495545376348831886

Had to Google what "prose content" meant, first time coming across that terminology. So feel free to correct me if I understood incorrectly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,6 +120,13 @@ Set `blame.ignoreRevsFile` to ignore [noise commits][noise-commits] in `git blam
 git config blame.ignoreRevsFile .git-blame-ignore-revs
 ```
 
+# AI Usage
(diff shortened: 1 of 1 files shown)


KiloOscarSix:
Here is a link to Linux's AI policy if the need arises to further define what is and is not allowed.

https://docs.kernel.org/process/coding-assistants.html

GnomedDev (approved):
LGTM, I'll defer to another maintainer to give another approval though as this is a reworded version of my text.

meditationmind:
Looks good. Would it be good to require disclosure of AI use too? Or is limiting usage to line-based autocomplete enough that we don't care about disclosure?

GnomedDev:
I think disclosure would be an unenforceable burden that wouldn't benefit us due to the existing limitations.

arqunis (approved):

PR #3544, opened 2026-04-23
feat: allow const methods on CreateAllowedMentions struct

Dueen:
Converts CreateAllowedMentions::new and the flag-toggle methods (all_users, all_roles, everyone, replied_user) into const fn, so a fully-configured allowed-mentions value can be constructed at compile time.

### DIFF
--- a/src/builder/create_allowed_mentions.rs
+++ b/src/builder/create_allowed_mentions.rs
@@ -93,7 +93,6 @@ pub struct CreateAllowedMentions<'a> {
 }
 
 impl<'a> CreateAllowedMentions<'a> {
-    /// Equivalent to [`Self::default`]. Usable in `const` contexts.
     pub const fn new() -> Self {
         Self {
             parse: Parse::new(),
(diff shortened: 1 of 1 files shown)


src/builder/create_allowed_mentions.rs
```diff
@@ -72,43 +80,35 @@ impl ParseAction {
...
-            _ => {},
-        }
-
-        self
+    /// Equivalent to [`Self::default`]. Usable in `const` contexts.
+    pub const fn new() -> Self {
+        Self { parse: Parse::new(), users: Vec::new(), roles: Vec::new(), replied_user: None }
     }
 
     /// Toggles mentions for all users. Overrides [`Self::users`] if it was previously set.
```
Joshix-1: this comment seems wrong

src/builder/create_allowed_mentions.rs
```diff
@@ -72,43 +80,35 @@ impl ParseAction {
...
-    pub fn all_users(self, allow: bool) -> Self {
-        self.handle_parse_unique(ParseValue::Users, ParseAction::from_allow(allow))
+    pub const fn all_users(mut self, allow: bool) -> Self {
+        self.parse.users = allow;
+        self
     }
 
     /// Toggles mentions for all roles. Overrides [`Self::roles`] if it was previously set.
```
Joshix-1: this comment seems wrong

Issue #3530, opened 2026-04-02
Unknown event: VOICE_CHANNEL_START_TIME_UPDATE event handler


meltinglava:
Just had a look at some logs while debugging and this came up. 

```log
serenity::client::dispatch: An unknown event was received: UnknownEvent { kind: "VOICE_CHANNEL_START_TIME_UPDATE", value: Object {"guild_id": String("81730624..."), "id": String("817306..."), "voice_start_time": Null} }
serenity::client::dispatch: An unknown event was received: UnknownEvent { kind: "VOICE_CHANNEL_START_TIME_UPDATE", value: Object {"guild_id": String("81730624..."), "id": String("817306..."), "voice_start_time": Number(1775127082)} }
```

nwerosama:
That event has been around for months and hasn't been officially documented by Discord themselves.

PR #3545, opened 2026-04-23
Implement voice channel status and start time

meditationmind:
Implements all changes from [Document voice channel status and start time (#6400)](https://github.com/discord/discord-api-docs/commit/258deb498cd63e7dbe43e5a6a4f3c5c44dcc0ce4).

Fixes [#3530](https://github.com/serenity-rs/serenity/issues/3530).

### DIFF
--- a/src/model/channel/guild_channel.rs
+++ b/src/model/channel/guild_channel.rs
@@ -138,7 +138,7 @@ pub struct GuildChannel {
     /// Unix timestamp (in seconds) of when a voice session started.
     ///
     /// **Note**: This is only available in voice channels and will only be [`Some`] when the cache
-    /// is enabled.  To manually retrieve the value and trigger population in the cache, see
+    /// is enabled. To manually retrieve the value and trigger population in the cache, see
     /// [`Context::channel_info`].
     ///
     /// [`Context::channel_info`]: crate::gateway::client::Context::channel_info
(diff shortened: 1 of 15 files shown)


meditationmind:
I initially had `ChannelInfoChannel`/`ChannelInfoField` in `gateway/mod.rs` since the docs have all the `ChannelInfo` structures in the gateway section, but that caused issues with feature flags. `model/event/mod.rs` makes sense for the same reasons, but since we only have events there, it seemed out of place. I decided to go with `model/channel/mod.rs` since the types do hold channel data, albeit ephemeral and only retrievable via the gateway. Happy to receive feedback about alternate solutions.

src/model/channel/guild_channel.rs
```diff
@@ -129,8 +129,20 @@ pub struct GuildChannel {
...
+    /// **Note**: This is only available in voice channels and will only be [`Some`] when the cache
+    /// is enabled. To manually retrieve the value and trigger population in the cache, see
+    /// [`Context::channel_info`].
+    ///
+    /// [`Context::channel_info`]: crate::gateway::client::Context::channel_info
     pub status: Option<FixedString<u16>>,
+    /// Unix timestamp (in seconds) of when a voice session started.
+    ///
+    /// **Note**: This is only available in voice channels and will only be [`Some`] when the cache
+    /// is enabled.  To manually retrieve the value and trigger population in the cache, see
```
kshitijanurag:
```suggestion
    /// is enabled. To manually retrieve the value and trigger population in the cache, see
```

meditationmind: Nice catch, thanks.

arqunis (approved):

PR #3546, opened 2026-04-25
Add full support for EditCurrentMember endpoint

GnomedDev:
Previously this was only supported via Http and only the nickname was exposed, this fixes that.

Tested setting and unsetting avatars, but no other functionality.


KiloOscarSix:
@GnomedDev Did you intend to include your flake (nix) files?

GnomedDev:
God dammit. Removed

src/builder/edit_current_member.rs
```diff
@@ -0,0 +1,83 @@
...
+    audit_log_reason: Option<&'a str>,
+}
+
+impl<'a> EditCurrentMember<'a> {
+    /// Equivalent to [`Self::default`].
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Changes the bots's nickname. Pass an empty string to reset the nickname.
```
meditationmind: Passing `None` will also reset the nickname. That seems more intuitive to me than an empty string. Or perhaps mention both? `None` also works to reset the others, so might be worth adding a note to those too.

src/builder/edit_current_member.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    /// Changes the bot's guild-specific base64 encoded avatar image.
+    ///
+    /// The `avatar` must be base64-encoded png/jpeg image data.
+    pub fn avatar(mut self, avatar: Option<DataUri<'a>>) -> Self {
+        self.avatar = Some(avatar);
+        self
+    }
+
+    /// Changes the bot's (about me) in the guild.
```
meditationmind: Is this supposed to be `bio (about me)`?

src/builder/edit_current_member.rs
```diff
@@ -0,0 +1,83 @@
...
+    pub fn bio(mut self, bio: Option<impl Into<Cow<'a, str>>>) -> Self {
+        self.bio = Some(bio.map(Into::into));
```
meditationmind: [See conversation here.](https://discord.com/channels/381880193251409931/673965002805477386/1497688482724974895)

GnomedDev: Fixed in 463283b.

src/builder/edit_current_member.rs
GnomedDev: Fixed in 463283b.

src/builder/edit_current_member.rs
GnomedDev: Fixed in 931c8e09ca.

meditationmind:
LGTM

arqunis (approved):

PR #3549, opened 2026-05-04
Fix Clippy and CI on current

meditationmind:
Fixed Clippy and MSRV CI errors on `current`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,6 +170,8 @@ jobs:
           cargo update base64ct --precise 1.7.3
           cargo update deranged --precise 0.5.3
           cargo update time --precise 0.3.40
+          cargo update rustc-hash --precise 2.1.1
+          cargo update hyper-rustls@0.27 --precise 0.27.7
 
       - name: Check
         run: cargo check --features full
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ tokio = { version = "1.34.0", features = ["fs", "macros", "rt", "sync", "time", 
 futures = { version = "0.3.32", default-features = false, features = ["std"] }
 time = { version = "0.3.36", features = ["formatting", "parsing", "serde-well-known"] }
 base64 = { version = "0.22.0" }
+# serde feature only allows for serialisation
 secrecy = { version = "0.8.0", features = ["serde"] }
 arrayvec = { version = "0.7.4", features = ["serde"] }
 serde_cow = { version = "0.1.0" }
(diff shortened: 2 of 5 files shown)


meditationmind:
Hope this was the right way to approach this.

The `rustc-hash` change is to preserve MSRV (2.1.2 bumps MSRV to 1.77). `reqwest` started pulling in a copy of `hyper-rustls` that breaks MSRV (0.27.8 bumps MSRV to 1.85), so I added the compatible version to the `rustls_backend` feature. That broke the minimal versions CI, so I also added the minimum version of `hyper-util` to the feature to fix that.

mkrasnitski:
Our approach to supporting an MSRV as old as 1.74 is more hands off - we maintain a list of locked versions in [`.github/workflows/ci.yml`](https://github.com/serenity-rs/serenity/blob/f84addf/.github/workflows/ci.yml#L167-L172) which allows the MSRV job to succeed, and don't claim to support building with the MSRV out of the box without those version pins. I tested on my own fork, and seems like adding the following pins is sufficient to build:

```sh
cargo update hyper-rustls@0.27 --precise 0.27.7
cargo update rustc-hash --precise 2.1.1
```

The changes to Cargo.toml itself aren't necessary.

meditationmind:
Ah, yes, that's a much better solution! I knew Poise used the MSRV-specific `Cargo.lock`, but I hadn't noticed the pins in the workflow file for Serenity. Thanks!

Unfortunately, something seems to have changed in nightly over the past few days, and Clippy is now complaining about the visibility of `error::Result` and `error::Error`.

```rust
error[E0365]: `Result` is only public within the crate, and cannot be re-exported outside
   --> src/lib.rs:185:9
    |
185 |         model::prelude::*,
    |         ^^^^^^^^^^^^^^^^^ re-export of crate public `Result`
    |
    = note: consider declaring type or module `Result` with `pub`
```

Even though they're public, they have a `pub use` in `internal::prelude`, which is `pub(crate)`, and that seems to be causing issues since it's included in `model::prelude`. I'm assuming the `error` module has been intentionally kept private to allow for controlling where and how `Result` and `Error` are exported. I tried adjusting visibility in several places, but all of the re-exporting via various preludes is a bit of a maze and I keep running into new errors.

I've set this back to draft for now. Any advice would be appreciated.

mkrasnitski:
This is in fact a rustc bug regressed in the newest nightly, and I used `cargo bisect-rustc` to bisect to rust-lang/rust#156014. Trying to write a minimal repro atm, but not having any luck.

meditationmind:
Wow, thanks for the bisect tip. I spent a while manually combing through PRs trying to find the source (without much success).

mkrasnitski:
The newest nightly fixes the import error, so this should be good to go now.

meditationmind:
Thanks, marked as ready for review.

mkrasnitski (approved):

PR #3551, opened 2026-05-08
Move default allowed mentions into the `CacheHttp` trait and out of the `http` module

mkrasnitski:
To further remove the `http` module's dependency on the rest of serenity, the `default_allowed_mentions` field is moved into the `Client` and passed around inside each `Context`, then forwarded to requests through the `CacheHttp` trait. Unfortunately, it's now possible for a user to accidentally prevent the field from being forwarded by mistakenly passing `ctx.http` to a method expecting `impl CacheHttp`. One solution could be to gate the implementation of `CacheHttp` for Http behind `#[cfg(not(feature = "builder"))]`, but I'm not sure if this is what we want.

### DIFF
--- a/examples/e01_basic_ping_bot/src/main.rs
+++ b/examples/e01_basic_ping_bot/src/main.rs
@@ -22,7 +22,7 @@ impl EventHandler for Handler {
                     // Sending a message can fail, due to a network error, an authentication error,
                     // or lack of permissions to post in the channel, so log to
                     // stdout when some error happens, with a description of it.
-                    if let Err(why) = new_message.channel_id.say(&ctx.http, "Pong!").await {
+                    if let Err(why) = new_message.channel_id.say(ctx, "Pong!").await {
                         println!("Error sending message: {why:?}");
                     }
                 }
--- a/examples/e02_transparent_guild_sharding/src/main.rs
+++ b/examples/e02_transparent_guild_sharding/src/main.rs
@@ -30,7 +30,7 @@ impl EventHandler for Handler {
                 new_message, ..
             } => {
                 if new_message.content == "!ping" {
-                    if let Err(why) = new_message.channel_id.say(&ctx.http, "Pong!").await {
+                    if let Err(why) = new_message.channel_id.say(ctx, "Pong!").await {
                         println!("Error sending message: {why:?}");
                     }
                 }
--- a/examples/e03_struct_utilities/src/main.rs
+++ b/examples/e03_struct_utilities/src/main.rs
@@ -23,7 +23,7 @@ impl EventHandler for Handler {
                     // In this case, you can direct message a User directly by simply calling a
                     // method on its instance, with the content of the message.
                     let builder = CreateMessage::new().content("Hello!");
-                    let dm = new_message.author.id.dm(&ctx.http, builder).await;
+                    let dm = new_message.author.id.dm(ctx, builder).await;
 
                     if let Err(why) = dm {
                         println!("Error when direct messaging user: {why:?}");
--- a/examples/e04_message_builder/src/main.rs
+++ b/examples/e04_message_builder/src/main.rs
@@ -35,7 +35,7 @@ impl EventHandler for Handler {
                         .push(" channel")
                         .build();
 
-                    if let Err(why) = new_message.channel_id.say(&ctx.http, &response).await {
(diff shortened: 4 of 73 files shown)


meditationmind:
What would be the downside of gating the implementation? Because preventing an unintended side effect with potentially significant repercussions is a pretty big upside. I think many end-users of the library will just plug in the first thing that works, without paying attention to any fine text about potential footguns.

Interestingly, this PR seems to be reversing a trend I noticed when transitioning to `serenity-next`, which was that many builders that were happy with `ctx` before now wanted `ctx.http()`. In other words, `impl CacheHttp` had been replaced by `&Http`, and this is swapping them back. Very minor point, but less to deal with when upgrading is always a plus.

mkrasnitski:
The downside is that the `builder` feature would no longer be additive, in other words enabling the `builder` feature would remove the implementation of `CacheHttp` for `Http`. Overall the impact of this won't be too noticeable given that `builder` is enabled by default.

meditationmind:
Right, that makes sense. I wouldn't be against skipping the gating, but we'd need to document the issue clearly and be ready to help people troubleshoot.

Did some quick general testing and everything seems to function fine. Couple observations:

What is the purpose of `ClientBuilder::get_default_allowed_mentions`? You get the same value from `Client` as soon as you await the builder, right? Nothing wrong with having it, just couldn't think of a use case for it.

Also noticed some odd behavior with `default_allowed_mentions()`. Took a while to figure out what was going on, so noting here for later reference, since it seems Poise will need some fixing once this is merged.
```rust
ctx.default_allowed_mentions()
ctx.serenity_context().default_allowed_mentions()
ctx.framework().options.allowed_mentions
```
With `ctx` being `Poise::Context`, each of these three references a different value. The first is the `CacheHttp` impl for `Poise::Context` (missing the fn so defaults to `None`), the second is from Serenity `Context`, and the third is from Poise `FrameworkOptions`.

Related, being able to set different defaults in the `Client` and `Framework` seems a bit unintuitive. Is that intended?

mkrasnitski:
> What is the purpose of `ClientBuilder::get_default_allowed_mentions`? You get the same value from `Client` as soon as you await the builder, right? Nothing wrong with having it, just couldn't think of a use case for it.

I only added it because every other setter on `ClientBuilder` has a corresponding getter. To be honest, I don't think they're necessary either.

> Related, being able to set different defaults in the `Client` and `Framework` seems a bit unintuitive. Is that intended?

Could you clarify what you mean or give an example?

meditationmind:
> Could you clarify what you mean or give an example?

This might actually be more of a Poise issue, but perhaps worth mentioning anyway.

```rust
let options = poise::FrameworkOptions {
    allowed_mentions: my_allowed_mentions1,
    // Other options
};
let client = serenity::ClientBuilder::new(token, intents)
    .framework(Box::new(poise::Framework::new(options)))
    .default_allowed_mentions(my_allowed_mentions2)
    .await;
```
Now we have:
`ctx.framework().options.allowed_mentions` -> `my_allowed_mentions1`
`ctx.serenity_context().default_allowed_mentions()` -> `my_allowed_mentions2`

This wasn't introduced by your PR since the default is just being moved to a different place, but I hadn't noticed it before. From a quick glance, it seems the result is that `poise::CreateReply` will potentially have different default allowed mentions than the message builders in Serenity.

Not sure if it's worth doing anything about it, but perhaps worth noting since it could lead to some confusion. For example, with no defaults explicitly set:

Poise `CreateReply` / `Context::reply` default:
```rust
CreateAllowedMentions::default().all_users(true).replied_user(true)
```
Serenity overall default is `None` / `CreateAllowedMentions::default()`, `Message::reply` is:
```rust
CreateAllowedMentions::new().everyone(true).all_users(true).all_roles(true).replied_user(false)
```

nwerosama:
Imo, it would probably be best to drop Poise's one in favour of Serenity's, ~~like how the event handler was dropped from Poise prior.~~ (actually it was something else, i misremembered after posting this)

meditationmind:
Well, since Poise is a command framework, it makes sense for the default reply to commands to include `replied_user(true)`. I guess that also applies to human replies in the client. You could argue `all_users(true)` makes sense too, given that's how a message works by default in the client. The Poise default more closely mimics user behavior, but the Serenity default is safer.

Issue #3402, opened 2025-08-01
JSON deserialize error on guild invites


HyperCodec:
Running `ctx.http.get_guild_invites(guild_id).await`, I get the following:
```
thread 'tokio-runtime-worker' panicked at src/bot/mod.rs:82:22:
Failed to get guild invites: Json(Error("missing field `created_at`", line: 1, column: 770))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

According to the [api docs](https://discord.com/developers/docs/resources/invite#invite-object), it does not have a `created_at` field.

HyperCodec:
(also please make the general error type passed in to `on_error` better, i had to use `expect` in my event handler in order to actually see the error message)

jamesbt365:
> According to the [api docs](https://discord.com/developers/docs/resources/invite#invite-object), it does not have a `created_at` field.


According to the struct field docs, which links to the right type on the discord api, yes it does. https://discord.com/developers/docs/resources/invite#invite-metadata-object

Could you emulate this request externally and pull out the response? It *should* have the created_at field considering the discord docs state that it does.

jamesbt365:
> (also please make the general error type passed in to `on_error` better, i had to use `expect` in my event handler in order to actually see the error message)

Aren't you, the developer, in control of the error generic?

HyperCodec:
> > (also please make the general error type passed in to `on_error` better, i had to use `expect` in my event handler in order to actually see the error message)
> 
> Aren't you, the developer, in control of the error generic?

The error type passed into `on_error` is `FrameworkError<'_, U, E>` which makes a message like `error in ready event event handler` instead of the actual error message. It actually stores my error type inside it, the message just needs adjusted to append the actual error output after (like `error in ready event event handler: Json(Error("missing field 'created_at'", line: 1, column: 770))`, for example)

HyperCodec:
I split the off-topic error stuff into a new issue.

HyperCodec:
Did the request manually, got the following structure (string contents omitted):

```
[
  {
    "type": 0,
    "code": "...",
    "inviter": {
      "id": "...",
      "username": "...",
      "avatar": "...",
      "discriminator": "0",
      "public_flags": 128,
      "flags": 128,
      "banner": null,
      "accent_color": null,
      "global_name": "...",
      "avatar_decoration_data": null,
      "collectibles": null,
      "display_name_styles": null,
      "banner_color": null,
      "clan": null,
[...]

PR #3552, opened 2026-05-10
Update invite models

meditationmind:
Updates invite-related models and endpoints to match the current API.

Fixes [#3402](https://github.com/serenity-rs/serenity/issues/3402).

The previous `RichInvite` (now `InviteMetadata`) essentially functioned as a different kind of invite, rather than the extension of `Invite` that it's meant to be, and this resulted in a deserialization error since the same endpoint could return two different types depending on permissions.

This PR fixes the issue by consolidating all invites as `Invite` and extending that with `InviteMetadata`. Many other discrepancies with the API have been patched up in the process. Also fixed a deserialization error related to including an embedded app target, which was likely simply never tested or encountered. `role_ids` has been added to the channel invite builder, to allow for assigning roles when an invite is accepted. I tested all of the above, as well as invites with scheduled event IDs, getting invite info while banned, group DM invites, and several other combinations.

I did not add the new [target users endpoints](https://docs.discord.com/developers/change-log#new-invite-endpoints).

~~This PR also pins nightly to 2026-05-05 in the Clippy and Docs workflows to address [rust #156264](https://github.com/rust-lang/rust/issues/156264). I'll revert these pins once [#156284](https://github.com/rust-lang/rust/pull/156284) is merged.~~ nightly pins reverted.


meditationmind:
Reverted the nightly pins and moved the Clippy fixes to a separate PR.

arqunis (approved):

Issue #3060, opened 2024-11-25
Allow fetching Users who have super-reacted to a Message


DPlayer234:
`Message::reaction_users` currently only allows fetching users that added a normal reaction to a message.

That is because the [API method that returns those users](https://discord.com/developers/docs/resources/message#get-reactions) has gotten an additional parameter since. By default, it returns users that added normal reactions.

A new method to query users that added a super reaction should be added, to parallel `Message::react` and `Message::super_react` for adding one.

PR #3555, opened 2026-05-17
Add `ReactionTypes` to `get_reaction_users`

meditationmind:
Fixes [#3060](https://github.com/serenity-rs/serenity/issues/3060).

Adds `ReactionTypes` to `Http::get_reaction_users` to allow fetching users who have super reacted to a message. Also fixes `Reaction::users` so that it uses `self.reaction_type` and removed the now-redundant `Reaction::users_`.

I changed the original `reaction_type` to `emoji` so that `emoji` corresponds to `Reaction::emoji` and `reaction_type` corresponds to `Reaction::reaction_type`.

### DIFF
--- a/src/model/channel/reaction.rs
+++ b/src/model/channel/reaction.rs
@@ -90,7 +90,7 @@ impl Serialize for Reaction {
 
 #[cfg(feature = "model")]
 impl Reaction {
-    /// Retrieves the associated the reaction was made in.
+    /// Retrieves the [`Channel`] the reaction was made in.
     ///
     /// If the cache is enabled, this will search for the already-cached channel. If not - or the
     /// channel was not found - this will perform a request over the REST API for the channel.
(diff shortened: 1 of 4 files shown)


src/model/channel/reaction.rs
```diff
@@ -179,11 +179,11 @@ impl Reaction {
         }
     }
 
-    /// Retrieves the list of [`User`]s who have reacted to a [`Message`] with a certain [`Emoji`].
+    /// Retrieves the list of [`User`]s who made this reaction.
     ///
     /// The default `limit` is `50` - specify otherwise to receive a different maximum number of
```
nwerosama: The default limit seems to be 25 according to the docs, maybe it should be reflected?

nwerosama: And in other places in the changed files where it says `50` is now `25`.

https://github.com/serenity-rs/serenity/blob/97f5da5d99df844c0cb8ff20c87b52339cf2ed64/src/model/channel/channel_id.rs#L834-L836
https://github.com/serenity-rs/serenity/blob/97f5da5d99df844c0cb8ff20c87b52339cf2ed64/src/model/channel/message.rs#L389-L391

meditationmind:
Thanks @nwerosama!

I did notice the API default, but for some reason I assumed we had intentionally set it to 50 so I left that code alone. Of course, we generally honor API defaults, so most likely the API changed. The code and comments should now reflect the API limits.

I also moved the handling of the value into the `Http` method so that a future change in the API can be reflected in one place, and I changed all `limit` parameters to `NonMaxU8` (see [#2701](https://github.com/serenity-rs/serenity/pull/2701) for reasoning) for consistency.

I did not move the warning to the `Http` method because the maximum clamp is documented and logging a warning in this case seems nonstandard for the library.

arqunis (approved):

Issue #3330, opened 2025-03-26
Guild Create with an unavaliable guild will lead to deserialisation error


GnomedDev:
According to [Discord Docs](https://discord.com/developers/docs/events/gateway-events#guild-create) a guild create event may be unavailable, but we do not handle this at all. I have not tested this, but from reading the code it seems like it would just lead to a deserialisation error.

meditationmind:
> The inner payload can be:
> - An available Guild: a [guild](https://docs.discord.com/developers/resources/guild#guild-object) object with extra fields, as noted below.
> - An unavailable Guild: an [unavailable guild](https://docs.discord.com/developers/resources/guild#unavailable-guild-object) object.

We currently have:
```rust
pub struct GuildCreateEvent {
    pub guild: Guild,
}
```

Seems like the correct way to handle this would be to make `guild` optional and add an optional `unavailable_guild`:
```rust
pub struct GuildCreateEvent {
    pub guild: Option<Guild>,
    pub unavailable_guild: Option<UnavailableGuild>,
}
```

Or would a different solution be preferable?

Joshix-1:
An enum that's either Guild or UnavailableGuild would be more correct and easier to use

PR #3556, opened 2026-05-18
Handle unavailable guild object in Guild Create

meditationmind:
Fixes [#3330](https://github.com/serenity-rs/serenity/issues/3330).

Represents the inner payload of the `GuildCreate` event with a new `GuildCreateGuild` enum, which has two variants: `Present(Guild)` and `Unavailable(UnavailableGuild)`. I chose "Present" instead of "Available" since a `Guild` object sent by the event can also be marked unavailable (see [Guild Create](https://docs.discord.com/developers/events/gateway-events#guild-create)).

`is_new` behavior for the `Unavailable` variant conforms to the intention specified in [#3505](https://github.com/serenity-rs/serenity/pull/3505).

Testing `Present` variant deserialization with my own bot was simple; I added a test for the `Unavailable` variant since it requires special circumstances.

### DIFF
--- a/src/model/event/full_event.rs
+++ b/src/model/event/full_event.rs
@@ -1034,7 +1034,7 @@ mod tests {
         assert!(!is_guild_new(&cache, guild_id));
 
         let event = Box::new(Event::GuildCreate(GuildCreateEvent {
-            guild: GuildCreateGuild::Present(Guild {
+            guild: GuildCreateGuild::Available(Guild {
                 __generated_flags: GuildGeneratedFlags::default(),
                 id: guild_id,
                 name: FixedString::new(),
@@ -1104,7 +1104,7 @@ mod tests {
         let guild_id2 = GuildId::new(2);
 
         let event = Box::new(Event::GuildCreate(GuildCreateEvent {
-            guild: GuildCreateGuild::Present(Guild {
+            guild: GuildCreateGuild::Available(Guild {
                 __generated_flags: GuildGeneratedFlags::default(),
                 id: guild_id2,
                 name: FixedString::new(),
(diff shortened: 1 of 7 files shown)


KiloOscarSix (commented):
Personally, I would prefer to use `Available` instead of `Present`, as it better follows the docs wording, even with the existence of the guild property

src/model/event/full_event.rs
```diff
@@ -486,7 +489,10 @@ impl FullEvent {
             },
             Event::GuildCreate(event) => {
                 #[cfg(feature = "cache")]
-                let is_new = Some(is_guild_new(cache, event.guild.id));
+                let is_new = Some(is_guild_new(cache, match &event.guild {
+                    GuildCreateGuild::Present(guild) => guild.id,
+                    GuildCreateGuild::Unavailable(unavailable_guild) => unavailable_guild.id,
+                }));
```
Joshix-1: It would be cool if GuildCreateGuild had a getter for the id

meditationmind:
> Personally, I would prefer to use `Available` instead of `Present`, as it better follows the docs wording, even with the existence of the guild property

Hmm, that is a good point. I initially used `Available` but then changed it since it seemed odd to have an unavailable available guild. Makes sense to align with the docs, though.

Will sleep on it and see if we have any other opinions when I check back.

src/model/event/full_event.rs
meditationmind: Good idea. I'll add that.

meditationmind:
Well, I slept on it and I'm still not a fan of unavailable available guilds, but I do think it's best to mirror the Discord docs to prevent confusion, so I changed `Present` to `Available`. I also added a getter for the `unavailable` marker since it might be convenient in some cases to check whether the guild is offline without needing to know the variant.

Issue #3558, opened 2026-05-19
`HttpError::UnsuccessfulRequest` missing useful info and repeats info


ZetaTwo:
**serenity version:** 0.12.5 (latest 0.12.x)

**features:** `client, gateway, model, rustls_backend, cache`

I sometimes get this error:

```
processing message failed: fetching channel 1267480910656180244: [Serenity] Could not decode json when receiving error response from discord:, expected value at line 1 column 1: [Serenity] Could not decode json when receiving error response from discord:, expected value at line 1 column 1
```

So:
1. The error discards a lot of information making it difficult to either act on this or report bugs
2. The formatting is bugged so the error is repeated

KiloOscarSix:
Please review the "AI Usage" section in "CONTRIBUTING.md".
https://github.com/serenity-rs/serenity/blob/current/CONTRIBUTING.md#ai-usage

AI-generated prose content is not permitted.

ZetaTwo:
That's fair. I didn't interpret that as applying to issues as well but I respect that. In that case, feel free to ignore everything after the error message, or the whole report.

The very short of it is that a bunch of information is discarded in the error so it's difficult to handle it gracefully without a bunch of downcasting, etc.

Joshix-1:
Can you edit the issue to remove the ai output? If we wanted to ask ai we could do it ourselves.

ZetaTwo:
Done

meditationmind:
[#3559](https://github.com/serenity-rs/serenity/pull/3559) should make JSON decode errors more informative.

The duplicated errors issue will require more discussion and feedback before being addressed, if at all. Related discussion:
- Error Handling Project Group: [Guidelines for implementing `Display::fmt` and `Error::source`](https://blog.rust-lang.org/inside-rust/2021/07/01/What-the-error-handling-project-group-is-working-towards/#guidelines-for-implementing-display-fmt-and-error-source)
- [#210 Add section about not including sources in Display impl](https://github.com/rust-lang/api-guidelines/pull/210)

It's also worth noting both of these are 5+ years old and still there does not seem to be a clear consensus.

PR #3559, opened 2026-05-21
Improve JSON decode error reporting

meditationmind:
Partially fixes [#3558](https://github.com/serenity-rs/serenity/issues/3558).

Adds all currently missing JSON error codes except for [5300xx](https://docs.discord.com/developers/discord-social-sdk/development-guides/using-provisional-accounts#error-handling), which are related to social SDK provisional accounts.

Improves JSON decode error reporting by including the HTTP response code in the error message. Standard formatting (`{e}`) includes a brief HTTP response code, e.g.:
> `[Serenity] Could not decode json when receiving error response from discord (http 502): error decoding response body`

Alternate formatting (`{e:#}`) includes a verbose [Discord-defined response code with meaning](https://docs.discord.com/developers/topics/opcodes-and-status-codes#http), as well as the JSON decode error's source error, e.g.:
> `[Serenity] Could not decode json when receiving error response from discord (502 gateway unavailable: there was not a gateway available to process your request): error decoding response body: EOF while parsing a value at line 1 column 0`

The error duplication issue (returning source in both `Display::fmt` and `Error::source`) is a larger issue that will need to be discussed and handled in a separate PR.


src/http/error.rs
```diff
@@ -366,6 +400,40 @@ impl fmt::Display for HttpError {
...
+                if e.error.code == JsonErrorCode::Unknown(1) {
+                    f.write_str(" (")?;
+                    // Return the verbose code and meaning in the alternate (`#`) format
+                    if f.alternate() {
+                        f.write_str(&HttpResponseCode::from(e.status_code).as_str().map_or_else(
+                            || Cow::Owned(e.status_code.to_string().to_ascii_lowercase()),
+                            Cow::Borrowed,
+                        ))?;
+                    } else {
+                        f.write_str("http ")?;
```
Joshix-1:
```suggestion
                        f.write_str("HTTP ")?;
```

src/http/error.rs
```diff
@@ -494,4 +595,39 @@ mod test {
...
+        let http_error = HttpError::from(error_response);
+
+        let expected = "[Serenity] Could not decode json when receiving error response from \
+            discord (502 gateway unavailable: there was not a gateway available to process your \
+            request): error decoding response body: EOF while parsing a value at line 1 column 0";
+
+        assert_eq!(format!("{http_error:#}"), expected.to_string());
+
+        let expected = "[Serenity] Could not decode json when receiving error response from \
+            discord (http 502): error decoding response body";
```
Joshix-1:
```suggestion
            discord (HTTP 502): error decoding response body";
```

src/http/error.rs
```diff
@@ -494,4 +595,39 @@ mod test {
...
+        let http_error = HttpError::from(error_response);
+
+        let expected = "[Serenity] Could not decode json when receiving error response from \
+            discord (408 request timeout): error decoding response body: EOF while parsing a value \
+            at line 1 column 0";
+
+        assert_eq!(format!("{http_error:#}"), expected.to_string());
+
+        let expected = "[Serenity] Could not decode json when receiving error response from \
+            discord (http 408): error decoding response body";
```
Joshix-1:
```suggestion
            discord (HTTP 408): error decoding response body";
```

src/http/error.rs
```diff
@@ -494,4 +595,39 @@ mod test {
...
+
+        assert_eq!(http_error.to_string(), expected.to_string());
```
Joshix-1:
```suggestion
        assert_eq!(http_error.to_string(), expected);
```

src/http/error.rs
```diff
@@ -494,4 +595,39 @@ mod test {
...
+        let response = Builder::new().status(408).body("").unwrap();
+        let reqwest_response: reqwest::Response = response.into();
+        let error_response = ErrorResponse::from_response(reqwest_response, Method::POST).await;
+        let http_error = HttpError::from(error_response);
+
+        let expected = "[Serenity] Could not decode json when receiving error response from \
+            discord (408 request timeout): error decoding response body: EOF while parsing a value \
+            at line 1 column 0";
+
+        assert_eq!(format!("{http_error:#}"), expected.to_string());
```
Joshix-1:
```suggestion
        assert_eq!(format!("{http_error:#}"), expected);
```

src/http/error.rs
```diff
@@ -494,4 +595,39 @@ mod test {
...
+        let expected = "[Serenity] Could not decode json when receiving error response from \
+            discord (502 gateway unavailable: there was not a gateway available to process your \
+            request): error decoding response body: EOF while parsing a value at line 1 column 0";
+
+        assert_eq!(format!("{http_error:#}"), expected.to_string());
+
+        let expected = "[Serenity] Could not decode json when receiving error response from \
+            discord (http 502): error decoding response body";
+
+        assert_eq!(http_error.to_string(), expected.to_string());
```
Joshix-1:
```suggestion
        assert_eq!(http_error.to_string(), expected);
```

src/http/error.rs
```diff
@@ -494,4 +595,39 @@ mod test {
...
+        let response = Builder::new().status(502).body("").unwrap();
+        let reqwest_response: reqwest::Response = response.into();
+        let error_response = ErrorResponse::from_response(reqwest_response, Method::POST).await;
+        let http_error = HttpError::from(error_response);
+
+        let expected = "[Serenity] Could not decode json when receiving error response from \
+            discord (502 gateway unavailable: there was not a gateway available to process your \
+            request): error decoding response body: EOF while parsing a value at line 1 column 0";
+
+        assert_eq!(format!("{http_error:#}"), expected.to_string());
```
Joshix-1:
```suggestion
        assert_eq!(format!("{http_error:#}"), expected);
```

src/internal/macros.rs
```diff
@@ -182,6 +182,51 @@ macro_rules! bitflags {
...
+
+        /// Returns the Discord-defined meaning of this HTTP response code.
+        #[must_use]
+        pub const fn as_str(&self) -> Option<&'static str> {
+            Self::with_meaning(self.0)
+        }
+
+        /// Returns an HTTP response code returned by the API with its Discord-defined meaning.
+        #[must_use]
+        pub const fn with_meaning(code: u16) -> Option<&'static str> {
```
Joshix-1: this method name feels imho wrong. `meaning_of` would make more sense, but I'm not sure

src/internal/macros.rs
```diff
@@ -182,6 +182,51 @@ macro_rules! bitflags {
...
+    ) => {
+        impl HttpResponseCode {
+        $(
+            $(#[$docs])*
+            pub const $konst: Self = Self($code);
+        )+
+
+        /// Returns the Discord-defined meaning of this HTTP response code.
+        #[must_use]
+        pub const fn as_str(&self) -> Option<&'static str> {
```
Joshix-1: I feel like `as_str` should always return `&str`, but not really sure how I would've done it

src/http/error.rs
```diff
@@ -366,6 +400,40 @@ impl fmt::Display for HttpError {
             Self::UnsuccessfulRequest(e) => {
                 f.write_str(&e.error.message)?;
 
+                // Return the HTTP response code with a JSON decode error
+                if e.error.code == JsonErrorCode::Unknown(1) {
+                    f.write_str(" (")?;
+                    // Return the verbose code and meaning in the alternate (`#`) format
+                    if f.alternate() {
+                        f.write_str(&HttpResponseCode::from(e.status_code).as_str().map_or_else(
+                            || Cow::Owned(e.status_code.to_string().to_ascii_lowercase()),
```
Joshix-1: Why lowercase? If lowercase maybe using ` make_ascii_uppercase` would be better to avoid an allocation.

src/http/error.rs
```diff
@@ -292,8 +310,24 @@ impl ErrorResponse {
...
+                    DiscordJsonSingleError {
+                        code: FixedString::new(),
+                        message: e
+                            .source()
+                            .map_or(FixedString::new(), |source| source.to_string().trunc_into()),
+                        path: Arc::default(),
+                    },
+                ]
+                .trunc_into(),
+                message: FixedString::from_str_trunc(
```
Joshix-1:
```suggestion
                message: FixedString::from_static_trunc(
```

src/http/error.rs
meditationmind: The [API guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#c-good-err) recommend lowercase and since "json" is lowercase in the original error message, I just stuck with all lowercase. The lowercase "HTTP" bothers me, though, so I'm happy to change it. I see "UTF-8" and "IP" in the guideline examples too, so seems fine. If we're fine with some mixed case here (`408 Request Timeout` instead of `408 request timeout`) then I'll just drop the `_ascii_lowercase`.

src/internal/macros.rs
meditationmind: Yeah, good point. How about `to_meaning`?

src/internal/macros.rs
meditationmind: Works for me.

src/http/error.rs
Joshix-1: didn't know there was a guideline. JSON should probably be uppercase then too. But I don't know.

src/internal/macros.rs
Joshix-1: Works for me.

src/http/error.rs
meditationmind: In the [compiler development guide](https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-output-style-guide), it's clearer about only the first letter being lowercase. I think I'll switch to "JSON".

But if we're following this guideline, then perhaps it should be `[Serenity] could not...` or `[serenity] could not...`

meditationmind:
Thanks again for the review @Joshix-1. I implemented the suggested fixes.

For the formatting, I went with lowercase first letter and normal capitalization mid-sentence, in accordance with the guidelines ([API](https://rust-lang.github.io/api-guidelines/interoperability.html#c-good-err), [compiler](https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-output-style-guide)). I also decided to drop the `[Serenity]` because it's the only place in the library where we have that and it's not really necessary, IMO.

Resulting standard format error:
> ```could not decode JSON when receiving error response from Discord (HTTP 502): error decoding response body```

And alternate format error:
> ```could not decode JSON when receiving error response from Discord (502 gateway unavailable: there was not a gateway available to process your request): error decoding response body: EOF while parsing a value at line 1 column 0```

Joshix-1 (approved):

KiloOscarSix (approved):

arqunis (approved):

PR #3566, opened 2026-06-04
Update `Message` model and docs

meditationmind:
Updates the `Message` model to match the current API by adding the `call` ([discord-api-docs #6864](https://github.com/discord/discord-api-docs/pull/6864)) and `channel_type` ([discord-api-docs #8311](https://github.com/discord/discord-api-docs/pull/8311)) fields, as well as the [`MessageCall`](https://docs.discord.com/developers/resources/message#message-call-object) object as the type used by `call`.

Also updates the docs to amend or add details about when certain fields will be populated.

Verified that both new fields are properly populated and deserialized.

### DIFF
--- a/src/model/channel/message.rs
+++ b/src/model/channel/message.rs
@@ -37,6 +37,10 @@ pub struct Message {
     /// The user that sent the message.
     pub author: User,
     /// The content of the message.
+    ///
+    /// Will be empty if the bot does not have the [`MESSAGE_CONTENT`] privileged intent.
+    ///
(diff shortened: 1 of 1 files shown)


meditationmind:
Remembered that the `PollResult` message type was omitted from [#2956](https://github.com/serenity-rs/serenity/pull/2956) so I have added that as well, along with documentation for the message types I could verify.

This should complete the fix for [#2948](https://github.com/serenity-rs/serenity/issues/2948).

arqunis (approved):

Issue #3484, opened 2026-02-13
Add jiff as a backend for Timestamp


inklesspen:
The chrono maintainer has [soft-deprecated chrono in favor of jiff](https://github.com/chronotope/chrono/issues/1768).

It seems like it would be pretty straightforward to add [`jiff::Timestamp`](https://docs.rs/jiff/latest/jiff/struct.Timestamp.html) as an underlying type for `serenity::…::Timestamp`.

The only thing that might require discussion is what to do about the feature flags. I can see two possible approaches:
1. add a `jiff` feature flag, and use `#[cfg(not(any(feature = "chrono", feature = "jiff"))]` for the `time` backend.
2. add both `jiff` and `time` feature flags; since `chrono` is one of the default features anyway, this shouldn't be a huge hassle for users.

Although `jiff` isn't yet included in typesize, so maybe that would be an issue? I don't really understand what that is used for yet. But it should be pretty simple to add it if needed. I'll check with @GnomedDev on that.

I'd be happy to send a PR for this, if it'd be welcomed.

inklesspen:
I implemented approach 1 in this commit: https://github.com/inklesspen/serenity/commit/86af24c2e89e445d637247c8f022218bda2dae08

I haven't opened a PR for it yet, though, because I think it's blocked on https://github.com/GnomedDev/typesize/pull/10.

GnomedDev:
I dislike the situation currently where we have different backends for time, since it means we have non-additive features and implementation complexity. I feel like, since the Rust ecosystem seems to be moving to Jiff anyway, we should just take a stand and only support it.

jhpratt:
FYI `time` will have a `Timestamp` type soon.

PR #3567, opened 2026-06-12
Fix From impl coherence with time 0.3.48

alexandretrotel:
I work on Radion, a prediction markets trading app. We use serenity for our Discord bot.

When I updated our dependencies, `time 0.3.48` (released 2026-06-12) triggered E0119 coherence errors. This blocked us from shipping the update — we had to pin `time = "=0.3.36"` and patch serenity with a fork just to compile.

The root cause: `time 0.3.48` added generic parameters to all types in the `unit` module and new comparison impls against the generic `Unit` type ([changelog](https://github.com/time-rs/time/blob/main/CHANGELOG.md#0348-2026-06-12)). These new impls conflict with serenity's broad `From<T: Into<String>>` blanket impls on `AutocompleteChoice` and `Content`.

This PR narrows those impls to concrete `String` and `\&str` overloads, removing the overlap. `char` support for `Content` is kept so `MessageBuilder::push('a')` still works.

I saw the discussion in #3484 — I know the long-term plan is to move to Jiff. This is just a fix to unblock downstream users until that lands.

Tested with `cargo check` and `cargo test --lib utils::message_builder::test::push` (no-default-features, builder/model/utils/native_tls_backend).

### DIFF
--- a/src/builder/create_interaction_response.rs
+++ b/src/builder/create_interaction_response.rs
@@ -346,11 +346,16 @@ impl AutocompleteChoice {
     }
 }
 
-impl<S: Into<String>> From<S> for AutocompleteChoice {
-    fn from(value: S) -> Self {
-        let value = value.into();
+impl From<String> for AutocompleteChoice {
+    fn from(value: String) -> Self {
         let name = value.clone();
         Self::new(name, value)
+    }
+}
+
+impl From<&str> for AutocompleteChoice {
+    fn from(value: &str) -> Self {
+        Self::from(value.to_owned())
     }
 }
 
--- a/src/utils/message_builder.rs
+++ b/src/utils/message_builder.rs
@@ -1127,12 +1127,24 @@ impl std::fmt::Display for Content {
     }
 }
 
-impl<T: Into<String>> From<T> for Content {
-    fn from(t: T) -> Content {
+impl From<String> for Content {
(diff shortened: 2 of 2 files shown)


mkrasnitski:
This seems more like unintentional breakage on the part of the `time` crate, as this is affecting more than just `serenity`. See https://github.com/time-rs/time/issues/783

KiloOscarSix:
Time/Chrono are also likely to be discontinued in the next major release - https://github.com/serenity-rs/serenity/issues/3484#issuecomment-3935173057

alexandretrotel:
Okay thanks for the details. I guess we can close this. We'll keep maintaining our fork until this is resolved.

KiloOscarSix:
Just updated my projects and also got hit with this :rofl: 
An alternative solution is to pin time to `0.3.37` in the interim

For bot devs reading this, run `cargo update -p time --precise 0.3.47` to downgrade to 0.3.47.
However, until an update is pushed to Serenity, any blank `cargo update` will update `time` and break builds

mkrasnitski:
Looks like time v0.3.48 has been yanked, and the maintainer has said they will fix the issue before releasing v0.3.49.

alexandretrotel:
I couldn't downgrade or pin to a specific version because the way my dependencies graph resolved haha

Issue #2772, opened 2024-02-17
Missing feature dependency: http requires builder 


a-maier:
Enabling the `http` feature while disabling `builder` makes compilation fail.

Serenity version: 0.12.0

Rust version (`rustc -V`): rustc 1.76.0 (07dca489a 2024-02-04)

Minimal test case:
```
cargo add serenity --no-default-features --features rustls_backend,http
cargo check
```
I get [this error message](https://github.com/serenity-rs/serenity/files/14318081/log.txt).

Workaround:
```
cargo add serenity --no-default-features --features rustls_backend,http,builder
```

GnomedDev:
Confirmed this is still an issue.

PR #3569, opened 2026-06-14
Allow http to compile without model

soleynn:
the next-branch counterpart to #3568, both for #2772

#2496 already fixed this on current but next drifted back into it.
enabling http without model doesn't compile because:
1. http didn't depend on builder
2. some http gated builder execute methods call model gated helpers

what i did:
- http depends on builder now
- moved the plain Channel accessors (guild/thread/private/category/id/
  guild_id) out from behind the model feature, they're just enum matches.
  only Channel::delete actually needs model. same for the Mentionable
  impl on Channel
- EditRole::execute calls Http::edit_role_positions directly instead of
  the model GuildId helper

heads up: that model helper also did an audit log reason length check
that i dropped. the edit_role/create_role calls right above it already
send the reason without checking, so this just makes execute consistent.
happy to put it back if you'd rather

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,7 @@ jobs:
       matrix:
         name:
           - builder without model
+          - http without model
           - unstable + typesize
           - unstable (no default features)
           - temp cache
@@ -94,6 +95,8 @@ jobs:
         include:
           - name: builder without model
             features: builder
+          - name: http without model
+            features: http rustls_backend
           - name: unstable + typesize
             features: default unstable typesize
           - name: unstable (no default features)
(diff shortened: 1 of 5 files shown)


mkrasnitski:
We're currently already working on separating out the `http` module into its own crate as part of #3387, but some other blockers need to be PR'd first. I would also say that doing this on current will probably only cause rebasing issues and would be better done as part of a patch release to the v0.12 series after v0.13.0 is already released.

soleynn:
> We're currently already working on separating out the `http` module into its own crate as part of #3387, but some other blockers need to be PR'd first. I would also say that doing this on current will probably only cause rebasing issues and would be better done as part of a patch release to the v0.12 series after v0.13.0 is already released.

o that makes sense. ill close this one

on current, no worries abt timing its ur call. just flagging in case it matters. its an actual compile break in 12.5 and the fix itself is just one line cargo dep w/ no code but im happy to leave it for a v12 patch after 13 if thats cleaner on ur end. do u want me to leave it open or close it as well?

PR #3570, opened 2026-06-14
Fix clippy and deny warnings in CI

mkrasnitski:
Only some CI jobs were running with warnings denied before, this denies them for all jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,8 @@ jobs:
         uses: Swatinem/rust-cache@v2
 
       - name: Run tests
+        env:
+          RUSTFLAGS: -D warnings
         run: cargo test --no-default-features --features "${{ matrix.features || 'full' }}"
 
   check:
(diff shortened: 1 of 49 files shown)


meditationmind:
Looks good. Maybe a good opportunity to switch `actions/checkout@v4` to `actions/checkout@v6` to get rid of all the deprecation warnings.

mkrasnitski:
Which deprecation warnings are you referring to? Do you mean deprecated items in serenity?

meditationmind:
I mean the actions. For example, we get 21 warnings under `Annotations` [in the CI test](https://github.com/serenity-rs/serenity/actions/runs/27503749355), all for checkout v4:

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24.

arqunis (approved):

PR #3573, opened 2026-06-22
Remove most dependencies of `http` on `model`

mkrasnitski:
This PR removes some dependencies the `http` module had on various model types to prep for splitting it into its own crate. The main changes are:

* Move `http::typing` into `model`
* Move the `CacheHttp` trait into `model`
* Split the `CreateAttachment` builder into an `AttachmentData` struct which `http` now depends on instead of the builder itself
  * Rename `CreateAttachment::get_data` to `as_bytes`
  * Also change `CreateAttachment::url` to build its own reqwest client internally instead of relying on an existing `Http`, to mirror the implementation of `Attachment::download` 
* Change http methods taking `ReactionType` to take `&str` instead
* Add a `CreateTestEntitlement` builder to pass into `Http::create_test_entitlement`
* Move `constants::USER_AGENT` into `http` and rename it `SERENITY_USER_AGENT`

Supersedes #3551

### DIFF
--- a/examples/testing/src/main.rs
+++ b/examples/testing/src/main.rs
@@ -60,7 +60,7 @@ async fn message(ctx: &Context, msg: &Message) -> Result<(), serenity::Error> {
             .send_message(
                 &ctx.http,
                 CreateMessage::new()
-                    .add_file(CreateAttachment::url(&ctx.http, IMAGE_URL, "testing.png").await?),
+                    .add_file(CreateAttachment::url(IMAGE_URL, "testing.png").await?),
             )
             .await?;
         // Pre-PR, this falsely triggered a MODEL_TYPE_CONVERT Discord error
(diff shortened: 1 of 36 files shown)


meditationmind (approved):
LGTM. I didn't test every affected method, but I tested `Typing` and a handful each of reaction- and attachment-related methods. I can't create entitlements, but I tested the builder and got the expected invalid SKU error. Everything seems to work fine.

You may want to take a quick look at [Document attachment editing and is_spoiler param #8429](https://github.com/discord/discord-api-docs/pull/8429/changes). Attachment `description` and `is_spoiler` flag can now be edited and `CreateAttachment` will be getting some new fields. I don't see anything that requires a change in your PR, but figured I'd point it out in case you notice something.

mkrasnitski:
Based on feedback on Discord I renamed `as_bytes` -> `to_bytes`.

If anyone has strong feelings about bikeshedding `AttachmentData` and `AttachmentDataKind`, feel free to voice them here, but the name shouldn't be a blocker.

KiloOscarSix:
Name summary from Discord:
- Attachment{Data}Content
- FileUploadContent
- Attachment{Data}Source
- AttachmentFile / AttachmentFileContent

arqunis (approved):

PR #3576, opened 2026-06-27
Add attachment editing and `is_spoiler` param

meditationmind:
Implements changes from [Document attachment editing and is_spoiler param #8429](https://github.com/discord/discord-api-docs/pull/8429):
- Adds `is_spoiler` field and setter to `CreateAttachment`
- Adds `EditAttachments::keep_and_update` (and `keep_and_update_existing_attachment` shorthand methods) for keeping a single existing attachment and editing its `description`/`IS_SPOILER` flag
- Adds `EditAttachments::update` for editing `description`/`IS_SPOILER` flag of an existing attachment kept via `EditAttachments::keep_all` or `EditAttachments::remove` (also works with `EditAttachments::keep`)
- Updates `Attachment::duration_secs` description

Other than needing to remove `pub` from `CreateAttachment::is_spoiler`, no conflicts with [#3573](https://github.com/serenity-rs/serenity/pull/3573).

Edit: CI failures seem to be HTTP timeouts. All tests pass on my fork.


src/builder/create_attachment.rs
```diff
@@ -137,6 +141,13 @@ impl<'a> CreateAttachment<'a> {
         self.description = Some(description.into());
         self
     }
+
+    /// Whether the attachment should be marked as a spoiler and blurred until clicked.
+    #[expect(clippy::wrong_self_convention, reason = "Name reflects API")]
+    pub fn is_spoiler(mut self, is_spoiler: bool) -> Self {
```
mkrasnitski: This could probably be renamed to just `spoiler`.

mkrasnitski (changes requested):
To guard against future breakage, I think `EditAttachments::update` should take a builder, namely `AttachmentEditRequest` which would be better renamed to `EditAttachment`.

meditationmind:
> To guard against future breakage, I think `EditAttachments::update` should take a builder, namely `AttachmentEditRequest` which would be better renamed to `EditAttachment`.

Yeah, this makes sense.

I did consider `EditAttachment`, but I thought it might be too similar to `EditAttachments`. Discord added an [Attachment Request](https://docs.discord.com/developers/resources/message#attachment-object-attachment-request-structure) object, so I went with `AttachmentEditRequest`, but I admit I much prefer `EditAttachment`.

Do you think it would be a good idea to add `AttachmentRequest`/`AttachmentRequestData` and use that internally for `CreateAttachment` and `EditAttachment`? That would consolidate future changes that might otherwise affect both into one place.

meditationmind:
Changed `is_spoiler` to `spoiler` and added the `EditAttachment` builder.

Should probably change `create_attachment.rs` to `attachments.rs` since both builders use `EditAttachments`, which prevents a clean split into `create_` and `edit_`. Perhaps between this and #3573, that can be done in whichever is merged later.

mkrasnitski:
Renaming the module is fine. Since `EditAttachments` depends on `CreateAttachment`, moving it into a separate module makes less sense. Feel free to do it in this PR.

meditationmind:
Renamed the module. Thanks!

mkrasnitski (approved):
LGTM. Tested setting spoilers and it seems work fine. There's an odd quirk where marking an attachment with a spoiler while it's being viewed in the client will sometimes still keep it visible instead of immediately blurring it.

arqunis (approved):

PR #3582, opened 2026-07-19
Add support for private channel obfuscation

meditationmind:
~~Opening as draft pending merge of discord-api-docs [#8454](https://github.com/discord/discord-api-docs/pull/8454).~~

Private channel obfuscation will begin on November 16, 2026. This PR:

- Adds `CHANNEL_OBFUSCATED` channel flag
- Adds `GatewayCapabilities` bitfield for enabling gateway capabilities (see #8354 discussion/changes)
- Adds `CHANNEL_OBFUSCATION` capability flag for pre-release opt-in testing (marked as deprecated with info about release date)

Capability and flag both successfully tested.

Edit: Commit [59fa2a7](https://github.com/serenity-rs/serenity/pull/3582/commits/59fa2a7415f7979c7073a2d67feb0f07c08ca94b) also adds a distinct `ObfuscatedChannel` type to `Guild` to reduce memory usage, as discussed [in the Discord server](https://discord.com/channels/381880193251409931/673965002805477386/1529187238762512435). Cache update implementations for all relevant events have also been updated and tested. (Further testing is welcome!)


src/cache/event.rs
```diff
@@ -10,10 +10,13 @@ impl CacheUpdate for ChannelCreateEvent {
     type Output = GuildChannel;
```
GnomedDev: Return an enum of MaybeObfuscatedChannel?

src/cache/event.rs
```diff
@@ -34,10 +41,18 @@ impl CacheUpdate for ChannelUpdateEvent {
...
+            if self.channel.flags.contains(ChannelFlags::CHANNEL_OBFUSCATED) {
+                g.obfuscated_channels
+                    .insert((&self.channel).into())
+                    .map(Into::into)
+                    .or_else(|| g.channels.remove(&self.channel.id))
+            } else {
+                g.channels
+                    .insert(self.channel.clone())
+                    .or_else(|| g.obfuscated_channels.remove(&self.channel.id).map(Into::into))
+            }
```
GnomedDev: Nitpick: This logic seems correct after a bit of thinking about it, but it's hard to review and I am generally against a lot of the option methods like `or_else` as they make the code harder to read in a lot of scenarios. If you want, I would prefer this be rewritten to use some variable names and `match`.

src/cache/event.rs
```diff
@@ -174,59 +189,74 @@ impl CacheUpdate for GuildMemberUpdateEvent {
...
+            member.set_mute(self.mute());
+
+            old_member
+        });
+
+        if self.user.id == cache.current_user().id
+            && let Some(old_member) = &old_member
+        {
+            let mut to_obfuscate: Vec<ChannelId> = Vec::new();
+            if let Some(new_member) = guild.members.get(&self.user.id)
```
GnomedDev: This `guild.members.get(&self.user.id)` is confusing, as we are updating the entry in the map then getting it back out to perform this logic. I think it would be much clearer for the order of this function to be:
1. Create new Member, using `guild.members.get` to get the optional existing member.
2. Perform this obfuscation logic
3. Insert the new Member into the `guild.members` map.

src/cache/event.rs
```diff
@@ -174,59 +189,74 @@ impl CacheUpdate for GuildMemberUpdateEvent {
...
+                    if !guild.user_permissions_in(channel, new_member).view_channel() {
+                        to_obfuscate.push(channel.id);
+                    }
+                }
+            }
+            for id in to_obfuscate {
+                if let Some(channel) = guild.channels.remove(&id) {
+                    guild.obfuscated_channels.insert(channel.into());
+                }
+            }
```
GnomedDev: This inner obfuscation logic is duplicated between this and `GuildRoleUpdateEvent`, I think it would be best to move it to its own function in this file called something like `obfuscate_channels_if_not_visible` and this would provide some needed documentation (in the form of a function call) for why this logic is being done in the middle of the MemberUpdate/RoleUpdate events.

src/cache/event.rs
```diff
@@ -576,27 +625,30 @@ impl CacheUpdate for VoiceStateUpdateEvent {
...
-        } else {
-            None
+        if let Some(member) = &self.voice_state.member {
+            guild.members.insert(member.clone());
         }
+
+        let old_state = guild.voice_states.remove(&self.voice_state.user_id);
+        if self.voice_state.channel_id.is_some() {
+            guild.voice_states.insert(self.voice_state.clone());
+        }
```
GnomedDev, line 625: Missed the `guild.voice_states` removal when refactoring.

src/cache/event.rs
```diff
@@ -576,27 +625,30 @@ impl CacheUpdate for VoiceStateUpdateEvent {
...
+
+        if self.voice_state.user_id == cache.current_user().id
```
GnomedDev, line 629: Add comment something like: `Obfuscate the voice channel if the bot never had view channel access, eg, it was moved into it by someone with Permissions::MOVE_MEMBERS.`

src/cache/event.rs
```diff
@@ -263,13 +293,32 @@ impl CacheUpdate for GuildRoleUpdateEvent {
...
+        let old_role = guild
+            .roles
+            .get_mut(&self.role.id)
+            .map(|mut role| std::mem::replace(&mut *role, self.role.clone()));
+
+        if let Some(old_role) = &old_role
+            && let Some(member) = guild.members.get(&cache.current_user().id)
+            && member.roles.contains(&self.role.id)
+            && (old_role.permissions.view_channel() && !self.role.permissions.view_channel()
+                || old_role.permissions.administrator() && !self.role.permissions.administrator())
```
GnomedDev, line 302: Nitpick: Maybe comment that this admin check is necessary because this is just for avoiding unnecessarily rescanning the channels if a non-consequential role update happens.

src/gateway/client/mod.rs
```diff
@@ -231,6 +233,15 @@ impl ClientBuilder {
         self.intents
     }
 
+    /// Sets the [gateway capabilities]. Capabilities are a bitflag; you can combine them by
+    /// performing the `|`-operator.
+    ///
+    /// [gateway capabilities]: https://docs.discord.com/developers/events/gateway-events#identify-gateway-capabilities
```
GnomedDev: Nitpick:

```suggestion
    /// Sets the [gateway capabilities].
    /// 
    /// Capabilities are a bitflag; you can combine them performing the `|`-operator.
    ///
    /// [gateway capabilities]: https://docs.discord.com/developers/events/gateway-events#identify-gateway-capabilities
```

src/model/guild/mod.rs
```diff
@@ -217,11 +221,20 @@ pub struct Guild {
...
-    /// from or connect to them).
+    /// The bot must have permission to view a channel to receive its full metadata. For channels
+    /// with [obfuscated metadata], see the `obfuscated_channels` field.
+    ///
+    /// [obfuscated metadata]: https://docs.discord.com/developers/resources/channel#channel-object-obfuscated-channels
     pub channels: ExtractMap<ChannelId, GuildChannel>,
+    /// All channels with [obfuscated metadata] contained within a guild.
+    ///
+    /// Channel metadata is obfuscated when the bot does not have permission to view that channel.
+    /// Only the `id`, `type`, `position`, and `parent_id` fields are guaranteed to be available.
```
GnomedDev: This line is now redundant with ObfuscatedChannel being its own type.

src/model/guild/mod.rs
```diff
@@ -983,6 +995,142 @@ impl Guild {
...
+            unavailable: bool,
+            member_count: MemberCount,
+            voice_states: ExtractMap<UserId, VoiceState>,
+            members: ExtractMap<UserId, Member>,
+            channels: Vec<GuildChannel>,
+            threads: ExtractMap<ThreadId, GuildThread>,
+            presences: ExtractMap<UserId, Presence>,
+            stage_instances: FixedArray<StageInstance>,
+            #[serde(rename = "guild_scheduled_events")]
+            scheduled_events: ExtractMap<ScheduledEventId, ScheduledEvent>,
```
GnomedDev: I would just use the full name here and translate it when constructing Guild.

GnomedDev (changes requested):
Just thought I'd take a look over this since I requested the ObfuscatedChannel complication in the first place, looks generally good with some small comments.

src/cache/event.rs
meditationmind: Had to split it into two small functions because of issues with borrow mutability in `GuildRoleUpdateEvent`, but I think it still achieves the intended effect.

src/cache/event.rs
meditationmind: The removal was done in both if-else arms before, so I just moved it out of the statement.

meditationmind:
Thank you for taking the time to review! I implemented all the suggested changes that I didn't comment on.

Adding `MaybeObfuscated`, `GuildChannelRef` (as discussed on Discord), and changing the name of the `channels` field to `viewable_channels` expanded the scope of the PR considerably. Luckily, it also illuminated some side-effects of the channels split that got overlooked before. Some notes below.

Naming:
- I used `MaybeObfuscated` instead of `MaybeObfuscatedChannel` because it's also used for categories. While categories are technically channels, leaving "channel" out of the name may help prevent confusion.
- I used "viewable" as the counterpart to "obfuscated." I also considered "full" and "visible", but "full" could be misinterpreted as "all channels" and I don't like "visible" because obfuscated channels are still visible. "viewable" indicates that the bot has permission to view the channels.

Fixes:
- `GuildMemberUpdateEvent` was incorrectly overwriting `bool`s on the `Member` object because of `#[serde(default)]`.
- `Guild::default_channel`/`Guild::default_channel_guaranteed`: (Does anyone even use these?) These methods were not filtering out voice/stage channels and they did not sort by position, so they were just returning a random viewable channel. The `guaranteed` method was also using cached members, which meant that it could potentially return a channel that only an admin could see (just after startup if bot has admin). Both methods should work properly now, and `guaranteed` should be a lot more efficient.
- `Member::default_channel`: In addition to the above issues, this was pulling the `Member` object from the cache instead of using `self` for some reason. This meant it could unnecessarily fail if the member wasn't cached.

Feedback request:
- Returning `MaybeObfuscated` outside of `FullEvent`: Currently, `Cache::guild_categories`, `Channel::category` and `Member::default_channel` return the `MaybeObfuscated` type. I added methods and impls to make the type more convenient, but would it be better to just convert `ObfuscatedChannel` back to `GuildChannel` and return the original type?

meditationmind:
Rebased and marked ready for review since the docs PR was merged.

PR #3586, opened 2026-07-27
added member role update methods to `GuildId`

anxyum:
To modify the roles of a member you either need to get his Member representation and then update the role or to use directly the http method, so I added helper methods to `GuildId` to be able to use the directly `GuildId` and `UserId` instead of fetching the member object.

### DIFF
--- a/src/model/guild/guild_id.rs
+++ b/src/model/guild/guild_id.rs
@@ -1681,7 +1681,7 @@ impl GuildId {
         builder.execute(http, guild_id).await
     }
 
-    /// Adds a [`Role`] to the guild user.
+    /// Adds a [`Role`] to a member of the guild.
     ///
     /// **Note**: Requires the [Manage Roles] permission.
     ///
@@ -1691,17 +1691,17 @@ impl GuildId {
     /// Id does not exist.
     ///
     /// [Manage Roles]: Permissions::MANAGE_ROLES
-    #[inline]
-    pub async fn add_user_role(
+    pub async fn add_member_role(
         self,
-        http: impl AsRef<Http>,
-        user_id: impl Into<UserId>,
-        role_id: impl Into<RoleId>,
+        http: &Http,
+        user_id: UserId,
+        role_id: RoleId,
+        reason: Option<&str>,
     ) -> Result<()> {
-        http.as_ref().add_member_role(self, user_id.into(), role_id.into(), None).await
+        http.add_member_role(self, user_id, role_id, reason).await
     }
 
-    /// Removes a [`Role`] from the guild user.
+    /// Removes a [`Role`] from a member of the guild.
     ///
     /// **Note**: Requires the [Manage Roles] permission.
     ///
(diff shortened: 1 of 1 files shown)


nwerosama:
Edit: Right, I misunderstood the PR's context, my apologies in advance.

~~This duplicates an existing methods that already resides in `Member` model.~~
Also, the `current` branch is frozen to code changes as further development happens in `next`.

Joshix-1:
> This duplicates an existing methods that already resides in Member model.

Like mentioned in the pr description it is still useful as getting a Member object can be annoying and is a waste of time and resources. And as #3585 makes using http directly more annoying this is imho a welcome change.

anxyum:
> This duplicates an existing methods that already resides in `Member` model. Also, the `current` branch is frozen to code changes as further development happens in `next`.

Well, I did copy the two methods from `Member` intentionally. As @Joshix-1 pointed out, the whole point is to be able to use those methods without having to fetch the `Member`.

I also changed the PR target to the `next` branch. I'm not sure if I did it the right way, but it seems to work.

src/model/guild/guild_id.rs
```diff
@@ -1680,6 +1680,46 @@ impl GuildId {
...
+    ///
+    /// [Manage Roles]: Permissions::MANAGE_ROLES
+    #[inline]
+    pub async fn add_user_role(
+        self,
+        http: impl AsRef<Http>,
+        user_id: impl Into<UserId>,
+        role_id: impl Into<RoleId>,
+    ) -> Result<()> {
+        http.as_ref().add_member_role(self, user_id.into(), role_id.into(), None).await
```
meditationmind:
```suggestion
    /// Adds a [`Role`] to a member of the guild.
    ///
    /// **Note**: Requires the [Manage Roles] permission.
    ///
    /// # Errors
    ///
    /// Returns [`Error::Http`] if the current user lacks permission, or if a role with the given
    /// Id does not exist.
    ///
    /// [Manage Roles]: Permissions::MANAGE_ROLES
    pub async fn add_member_role(
        self,
        http: &Http,
        user_id: UserId,
        role_id: RoleId,
        reason: Option<&str>,
    ) -> Result<()> {
        http.add_member_role(self, user_id, role_id, reason).await
```

src/model/guild/guild_id.rs
```diff
@@ -1680,6 +1680,46 @@ impl GuildId {
...
+    ///
+    /// [Manage Roles]: Permissions::MANAGE_ROLES
+    #[inline]
+    pub async fn remove_user_role(
+        self,
+        http: impl AsRef<Http>,
+        user_id: impl Into<UserId>,
+        role_id: impl Into<RoleId>,
+    ) -> Result<()> {
+        http.as_ref().remove_member_role(self, user_id.into(), role_id.into(), None).await
```
meditationmind:
```suggestion
    /// Removes a [`Role`] from a member of the guild.
    ///
    /// **Note**: Requires the [Manage Roles] permission.
    ///
    /// # Errors
    ///
    /// Returns [`Error::Http`] if a role with the given Id does not exist, or if the current user
    /// lacks permission.
    ///
    /// [Manage Roles]: Permissions::MANAGE_ROLES
    pub async fn remove_member_role(
        self,
        http: &Http,
        user_id: UserId,
        role_id: RoleId,
        reason: Option<&str>,
    ) -> Result<()> {
        http.remove_member_role(self, user_id, role_id, reason).await
```

meditationmind (changes requested):
You targeted `next` but you're still using `current` conventions. See below.

src/model/guild/guild_id.rs
anxyum: Thanks for the feedback, I applied your suggestions and updated the code to match the `next` branch conventions.

meditationmind (approved):

arqunis (approved):

PR #3449, opened 2025-12-07
Add proxy support for websocket connect

Morb0:
Continue of #2709 
Fix #2479

PR #3590, opened 2026-08-01
Read `ALL_PROXY` environment variable for WebSocket connection

### LINKED ISSUE #2479: Could not establish a websocket connection through a proxy

ZiYueCommentary:
This pull request adds proxy support for WebSocket connection. Continue of #3449 

It reads `ALL_PROXY` environment variable and uses it as proxy of WebSocket. When missing this variable, the WebSocket connection works as current.

This is not a perfect solution, but it avoids breaking changes and works properly.

Fix #2479

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ typesize = { version = "0.1.2", optional = true, features = ["url", "time", "ser
 # Serenity workspace crates
 command_attr = { version = "0.5.4", path = "./command_attr", optional = true }
 serenity-voice-model = { version = "0.2.1", path = "./voice-model", optional = true }
+async-http-proxy = { version = "1.2.5", features = ["runtime-tokio", "basic-auth"] }
 
 [dev-dependencies.http_crate]
 version = ">=0.2.11"
--- a/src/gateway/ws.rs
+++ b/src/gateway/ws.rs
@@ -1,3 +1,4 @@
+use std::io::ErrorKind;
 use std::env::consts;
 #[cfg(feature = "client")]
 use std::io::Read;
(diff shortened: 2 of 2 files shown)


cheesycod:
I don’t follow serenity much anymore but id say that the correct way to do this is by having a http proxy (you’re probably using a http proxy too if you’re going for this) send the correct websocket url in /gateway and /gateway/bot APIs and then using wait_time_between_shards (or whatever it’s called now) if desired.

Using environment variables for this is not really a great use of environment variables

PR #3596, opened 2026-08-05
Add file type filtering support for command options and file uploads

KiloOscarSix:
https://docs.discord.com/developers/change-log#filter-file-types-in-file-uploads-and-attachment-options

Open Questions:
1. I went with an array of strings to best mirror the docs, but an Enum of `Image`, `Video`, `Audio`, `Extension(String)` could be argued for. 
2. I'm not a fan of the:
```
Cow::Owned(
                file_types.into_owned().into_iter().map(|f| f.into_owned().into()).collect(),
            ),
```
but kinda ties into point one with how we want the api to look.

### DIFF
--- a/src/builder/create_command.rs
+++ b/src/builder/create_command.rs
@@ -320,8 +320,10 @@ impl<'a> CreateCommandOption<'a> {
         self
     }
 
-    /// If the option is an [`Attachment`], file types to filter for; can be `image`, `video`,
-    /// `audio`, or any dot-prefixed extension such as `.pdf`; max 10.
+    /// If the option is an [`Attachment`], sets the file types to filter for.
+    ///
+    /// Can be `image`, `video`, `audio`, or any dot-prefixed extension such as `.pdf`. Maximum of
+    /// 10 types.
     ///
     /// **Note**: This only matches against the file extension. See [File Type Filtering] for
     /// details.
--- a/src/builder/create_command.rs
+++ b/src/builder/create_command.rs
@@ -40,6 +40,8 @@ pub struct CreateCommandOption<'a> {
     min_length: Option<u16>,
     #[serde(default)]
     max_length: Option<u16>,
+    #[serde(skip_serializing_if = "<[_]>::is_empty")]
+    file_types: Cow<'a, [Cow<'a, str>]>,
     #[serde(default)]
     autocomplete: bool,
 }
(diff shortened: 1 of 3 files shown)


src/builder/create_command.rs
```diff
@@ -317,6 +320,19 @@ impl<'a> CreateCommandOption<'a> {
         self
     }
 
+    /// If the option is an [`Attachment`], file types to filter for; can be `image`, `video`,
+    /// `audio`, or any dot-prefixed extension such as `.pdf`; max 10.
```
meditationmind:
```suggestion
    /// If the option is an [`Attachment`], sets the file types to filter for.
    ///
    /// Can be `image`, `video`, `audio`, or any dot-prefixed extension such as `.pdf`. Maximum of
    /// 10 types.
```

src/builder/create_components.rs
```diff
@@ -850,20 +853,36 @@ impl<'a> CreateFileUpload<'a> {
         self
     }
 
+    /// File types to filter for; can be `image`, `video`, `audio`, or any dot-prefixed extension
+    /// such as `.pdf`; max 10. Defaults to empty, which permits any file type.
```
meditationmind:
```suggestion
    /// Sets the file types to filter for.
    ///
    /// Can be `image`, `video`, `audio`, or any dot-prefixed extension such as `.pdf`. Maximum of
    /// 10 types.
```

src/model/application/command.rs
```diff
@@ -298,6 +298,16 @@ pub struct CommandOption {
     /// Maximum permitted length for String options
     #[serde(default)]
     pub max_length: Option<u16>,
+    /// If the option is an [`Attachment`], file types to filter for; can be `image`, `video`,
+    /// `audio`, or any dot-prefixed extension such as `.pdf`; max 10.
```
meditationmind:
```suggestion
    /// If the option is an [`Attachment`], the file types to filter for.
    ///
    /// Can be `image`, `video`, `audio`, or any dot-prefixed extension such as `.pdf`. Maximum
    /// of 10 types.
```

meditationmind (commented):
Overall looks fine to me. Could merge this as is, but I've noted a few docs nitpicks below, to make things more consistent with our own docs.

Also, successfully tested both slash commands and modals, and I can PR the slash changes to Poise once this is merged.

meditationmind (approved):

arqunis (approved):

PR #3599, opened 2026-08-24
Update tokio-tungstenite to 0.23.1

gak:
Updates the WebSocket TLS stack from rustls 0.22 / rustls-webpki 0.102 to rustls 0.23 / rustls-webpki 0.103. The explicit url feature preserves support for passing url::Url to connect_async_with_config after tokio-tungstenite made that feature non-default.\n\nThis is the smallest tokio-tungstenite release jump that moves to rustls 0.23.\n\nVerified with Serenity default-feature cargo check and by compiling a downstream Poise 0.6.2 application.


DPlayer234:
`tungstenite` errors are exposed in serenity's public API so this cannot be done without a breaking change release. `next` already uses newer versions of these dependencies.

PR #3606, opened 2026-09-08
Ensure URL parameters are encoded safely, using reqwest's query feature.

### LINKED ISSUE #3604: Bug: Request::build does not properly serialize params
https://github.com/serenity-rs/serenity/blob/98ec74223b0ff77fc4e8085d25569ea59e09a36f/src/http/request.rs#L100-L105

This simple logic probably works fine as long as all params are under the control of Serenity itself. However, `Http::search_guild_members` accepts an arbitrary query string, which could contain characters that must be [URL-encoded](https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_in_a_URI) to be sent safely. I'm currently working on adding support for `Http::search_guild_messages` (#3444) which has a similar query string.

I prepared a test case to demonstrate the issue: https://github.com/serenity-rs/serenity/commit/6a476a79a9daaba0955dd8462b82a69293ae762a

The test currently fails with this error:
```
assertion `left == right` failed
  left: [("query", "foo"), ("bar", ""), ("limit", "50")]
 right: [("query", "foo&bar"), ("limit", "50")]
```

I can think of a couple possible ways to resolve this:

1. Parse the path into a [`Url`](https://docs.rs/reqwest/latest/reqwest/struct.Url.html) before processing params, then call [`Url::query_pairs_mut`](https://docs.rs/url/latest/url/struct.Url.html#method.query_pairs_mut) and [`Serializer::extend_pairs`](https://docs.rs/form_urlencoded/1.2.2/form_urlencoded/struct.Serializer.html#method.extend_pairs) with the params.
2. Add a dependency on [serde_qs](https://crates.io/crates/serde_qs) or [serde_urlencoded](https://crates.io/crates/serde_urlencoded), which can be trusted to encode a sequence of params into a string.

My preference would actually be to add the dependency on `serde_qs`, because it's [capable of handling sequence values](https://docs.rs/serde_qs/1.1.3/serde_qs/enum.ArrayFormat.html) in the manner that [Discord requires](https://docs.discord.com/developers/reference#array-query-strings), and I'm working on implementing #3444 ([guild message search](https://docs.discord.com/developers/resources/message#search-guild-messages)) which has **thirteen** query arguments that are sequences, and I'd like to not have to add special support for those (via either hand-written loops or a custom-purpose macro).

However, in order to get the most use out of this new dependency, we would probably have to change the `Request::params` type to something like `&impl serde::Serialize`. (This is because neither `serde_qs` nor `serde_urlencoded` will serialize into a sequence of key-value pairs like the params field currently takes; only into a `String` or something implementing `std::io::Write`.)

Alternatively, we could allow setting the query string _as a string_ in the Request struct… Then the client method in the Http struct would be responsible for properly serializing complicated params.

So it's certainly something to think about carefully.

inklesspen:
This does introduce an indirect dependency on serde_urlencode.

Fixes #3604.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,7 +55,7 @@ foldhash = { version = "0.2.0", optional = true }
 chrono = { version = "0.4.42", default-features = false, features = ["clock", "serde"], optional = true }
 flate2 = { version = "1.1.9", features = ["zlib-rs"], optional = true }
 zstd = { version = "0.13", default-features = false, optional = true }
-reqwest = { version = ">=0.13", default-features = false, features = ["multipart", "stream", "json"], optional = true }
+reqwest = { version = ">=0.13", default-features = false, features = ["multipart", "stream", "json", "query"], optional = true }
 tokio-tungstenite = { version = "0.28.0", features = ["url"], optional = true }
 percent-encoding = { version = "2.3.2", optional = true }
 mini-moka = { version = "0.10.3", optional = true }
--- a/src/http/request.rs
+++ b/src/http/request.rs
@@ -164,3 +164,45 @@ impl<'a> Request<'a> {
         self.params
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::borrow::Cow;
+    use std::vec::Vec;
+
+    use super::*;
+    use crate::http::client::HttpBuilder;
+    use crate::model::id::GuildId;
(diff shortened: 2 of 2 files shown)


inklesspen:
I've also done an alternative approach in #3605.

PR #3608, opened 2026-09-16
Implement Guild Message Search (serde_json hacky approach)

### LINKED ISSUE #3444: Missing "Get User Message In A Guild" API Call
I wanted to get all user messages in a guild, so if they sent enough messages they'll pass the verification system I'm currently working on.
So, I found out that there is an api call that discord client makes, but its not documented anywhere. When I went and asked about it on discord developer server, they said that its available to anyone, just not documented

If I'll have some time, I can implement it and make a pull request, but for now it will be just an issue

URL of this call: /api/v10/guilds/<guild id>/messages/search
What data you need to send:
- author_id: Option<u64>
- mentions: Option<u64> 
- sort_by: i64 (unix timestamp)
- sort_order: String (idk what this means, but in default request it says "desc")
- ... i'm too lazy to analyze everything right now

inklesspen:
[Discord docs](https://docs.discord.com/developers/resources/message#search-guild-messages)

Fixes #3444.

Very important to note that this PR assumes some sort of fix for #3604 has been implemented. Depending on the nature of the fix, some refactoring may be required here.

I've added the route as a separate commit; that way, we could merge just that commit for now, while we iterate on the design of the Message Search builder.

The message search endpoint accepts 24 possible params, some of which can accept [arrays of values](https://docs.discord.com/developers/reference#array-query-strings). I don't like the idea of manually constructing a list of key-value params; that seems error-prone. The best alternative I can think of would be to use Serde.

But to use Serde, either we'd need to add a dependency on [serde_qs](https://crates.io/crates/serde_qs), or I'd need to implement a custom serializer, or we can abuse serde_json. In this PR, I've chosen to do the third thing, but I would much prefer it if we could add the dependency on serde_qs. (I'd also be willing to implement a custom serializer, but I'm not sure you'd be willing to accept that many lines of code for it.)

Happy to explore alternate approaches to this, if you have suggestions.


src/builder/search_messages.rs
```diff
@@ -0,0 +1,561 @@
...
+
+        // There must be a better way?
+        let borrowed = Vec::from_iter(params.iter().map(|(k, v)| (k.as_str(), v.as_str())));
+
+        let http = cache_http.http();
+        let messages = http.search_guild_messages(guild_id, Some(borrowed.as_slice())).await?;
+
+        // I would like to populate the cache with these messages, as is done in get_messages.rs.
+        // However this doesn't seem practical with the way fill_message_cache() is currently
+        // implemented. Thoughts?
```
arqunis: That would be nice, but given we cache not by messages themselves, but by a channel->messages mapping, it's not as straightforward as in `get_messages.rs`. `get_messages.rs` has the benefit of only fetching messages from one channel, so it easily knows how to populate the cache. Here, however, we would need to construct a `HashSet<GenericChannelId>`, iterate per unique channel ID to filter messages to that channel, and then call `fill_message_cache`. That's ugly, but it would avoid round-trips if the user were to fetch those messages again and the cache already has them. I think there are two choices here:

1. Don't -- the cache will stay blind to messages fetched via `search_guild_messages`
2. Add a third parameter, `should_cache: ShouldCache` -- `ShouldCache` would just be a `Yes`/`No` enum, so it's obvious from the call-site that `MessageQuery::default().[...].execute(..., ShouldCache::Yes)` would spend a bit more time caching messages making `execute` slower (or whether the user desires to cache those messages at all).

src/builder/search_messages.rs
```diff
@@ -0,0 +1,561 @@
...
+        // So, for example, if replied_to_user_ids has the three values [1, 2, 3],
+        // it should be encoded as replied_to_user_id=1&replied_to_user_id=2&replied_to_user_id=3.
+        // https://docs.discord.com/developers/reference#boolean-query-strings
+        // Boolean fields can be encoded as "True", "true", or "1" and "False", "false", or "0".
+
+        // We can either manually go through the params and build up a vec of key-value pairs, or
+        // rely on serde. Doing it manually seems error-prone. For serde, we could write a custom
+        // serializer for Discord's accepted format, or use serde_qs, or use serde_json to serialize
+        // to a Value and then convert the Value to querystring pairs. I've chosen to do the third
+        // thing here, but I really don't like how that's working with types.
```
arqunis, line 315: `serde_qs` would be an extra dependency, but given the complexity of the Guild Message Search endpoint's params, I think it deserves to be pulled in to avoid hacks or having to maintain a custom serialiser ourselves.

inklesspen: I've actually spent the last few hours working on a non-serde approach (#3609). I don't particularly like it but I thought it would be worth giving it a try.

The biggest obstacle to using serde_qs is that it serializes to a querystring, rather than a slice of pairs of strs. So we'd have to change the type of the params field in Request to allow this, since you can't use `impl Trait` in a struct field. (Or else we'd have to also add erased_serde, which seems like overkill?)

I don't think that's necessarily a blocker but I wanted to give you a chance to think about it.

src/builder/search_messages.rs
inklesspen: I think I like the should_cache idea; I'll give that a go.

inklesspen: Done.

src/builder/search_messages.rs
arqunis: Not a blocker, yes, but also not something we have to do right now either. We could adopt `serde_qs` later when we change `Request::params` to accept a `&str` (or perhaps `enum Params<'a> { Passthrough(&'a str), Pairs(&'a [(&'a str, &'a str)]) }`, so that it generate the query strings for us in all endpoints with queries, not just Guild Message Search. Since you already went through the effort of writing #3609, I think we can just merge #3605 and then this/#3608 with #3609's with manual slice. This way we avoid adding a new dependency for now and be consistent with how other endpoints with queries dynamically build their params. Migration to `serde_qs` can then simplify the code later.

inklesspen: Ok, so the plan is to get #3605 merged, and then get #3609 merged? Should I convert #3609 back into a non-draft PR (and add the optional caching commit to it)?

arqunis: Yeah. #3605 first, and then #3609 (with optional caching).

inklesspen: Done.

inklesspen:
Closed in favor of #3609.

PR #3609, opened 2026-09-16
Implement Guild Message Search (non-Serde approach)

inklesspen:
[Discord docs](https://docs.discord.com/developers/resources/message#search-guild-messages)

Fixes #3444.

I've added the route as a separate commit; that way, we could merge just that commit for now, while we iterate on the design of the Message Search builder.

The message search endpoint accepts 24 possible params, some of which can accept [arrays of values](https://docs.discord.com/developers/reference#array-query-strings). In #3608 I tried using Serde to serialize it to query key-value pairs, but it was pretty hacky. An improvement might be to use [serde_qs](https://crates.io/crates/serde_qs).

In this PR I instead manually serialize the params. Definitely open to ways to improve this.


inklesspen:
It seems I was over-eager to send this out; the builder works perfectly well, but the response type is incorrect at the moment.

inklesspen:
Ok, I think this is ready for review now.

There are three new types for the response: `MessageSearchOutcome`, `MessageSearchNotIndexed`, and `MessageSearchResults`. They're currently located in `src/model/guild/mod.rs`; I'm pretty sure that's **not** the right place for them, but I don't know where they should go.

In particular `MessageSearchNotIndexed` is sort-of-but-not-really an error response; it contains a code and message the same way `DiscordJsonError` does, but it also contains a `retry_after` value.

> You should retry the request after the timeframe specified in the `retry_after` field. If the `retry_after` field is 0, you should retry the request after a short delay.

In theory we could implement retry logic in the builder's execute method and then always return the `MessageSearchResults`, but that seems like it would be a lot of special-case code.

Also open to suggestions about type naming. The client method is `search_guild_messages`. The response types all say "search". But the builder is currently `MessageQuery` because I had this feeling that "query" was better as a standalone noun there. Still, for consistency's sake, they should probably all be the same, right?

meditationmind:
I'll take a closer look at this later when I have some time, but just a quick note... I think generally we would also want to add a method to `GuildId`, e.g., `GuildId::search_messages`.

inklesspen:
Should `GuildId::search_messages` take the builder? or the slice of param-pairs?

meditationmind:
It would take the builder. It basically just executes the builder for the lower-level `Http::search_guild_messages` method using `self` for the `GuildId`, similar to `GuildId::create_channel`, for example.

inklesspen:
Added it.

src/model/guild/mod.rs
```diff
@@ -1308,6 +1309,81 @@ pub struct IncidentsData {
...
+pub enum MessageSearchOutcome {
+    NotIndexed(MessageSearchNotIndexed),
+    Results(MessageSearchResults),
+}
+
+fn duration_via_f64_secs<'de, D>(deserializer: D) -> Result<Duration, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    f64::deserialize(deserializer).map(Duration::from_secs_f64)
```
meditationmind:
```suggestion
    u64::deserialize(deserializer).map(Duration::from_secs)
```
Unlike the rate limit headers, this `retry_after` is not a float, so probably better to use the int directly via `Duration::from_secs`. Adjust fn name accordingly.

src/model/guild/mod.rs
```diff
@@ -1308,6 +1309,81 @@ pub struct IncidentsData {
...
+#[derive(Clone, Debug, Deserialize)]
+pub struct MessageSearchNotIndexed {
+    // I know this is not great. Probably the best resolution will be to move these types
+    // elsewhere. But where?
+    #[cfg(feature = "http")]
+    pub code: crate::http::JsonErrorCode,
+    #[cfg(not(feature = "http"))]
+    pub code: u32,
+    pub message: FixedString,
+    pub documents_indexed: Option<u32>,
```
meditationmind:
```suggestion
    pub documents_indexed: u32,
```
This one is required. ([spec](https://github.com/discord/discord-api-spec/blob/main/specs/openapi.json#L37752))

src/builder/search_messages.rs
```diff
@@ -0,0 +1,576 @@
...
+#[cfg_attr(not(feature = "http"), expect(dead_code))] // Fields only read in the execute function.
+pub struct MessageQuery<'a> {
+    limit: Option<NonMaxU8>,
+    offset: Option<NonMaxU16>,
+    max_id: Option<MessageId>,
+    min_id: Option<MessageId>,
+    slop: Option<NonMaxU8>,
+    content: Option<Cow<'a, str>>,
+    channel_ids: Cow<'a, [GenericChannelId]>,
+    author_types: Cow<'a, [AuthorType]>,
```
meditationmind:
```suggestion
    author_types: Cow<'a, [AuthorType]>,
    author_ids: Cow<'a, [UserId]>,
```
`author_id` param and setters are missing.

src/builder/search_messages.rs
```diff
@@ -0,0 +1,576 @@
...
+    mention_user_ids: Cow<'a, [UserId]>,
+    mention_role_ids: Cow<'a, [RoleId]>,
+    mention_everyone: Option<bool>,
+    replied_to_user_ids: Cow<'a, [UserId]>,
+    replied_to_message_ids: Cow<'a, [MessageId]>,
+    pinned: Option<bool>,
```
meditationmind: `pinned` param is included, but setter is missing.

src/builder/search_messages.rs
```diff
@@ -0,0 +1,576 @@
...
+        add_author_type, add_author_types, author_type, author_types, "author type", "author types");
+    sequence_setters!(
+        mention_user_ids, user_id, user_ids, UserId, 'a,
+        add_user_mention, add_user_mentions, user_mention, user_mentions, "user mentioned", "users mentioned");
+    sequence_setters!(
+        mention_role_ids, role_id, role_ids, RoleId, 'a,
+        add_role_mention, add_role_mentions, role_mention, role_mentions, "role mentioned", "roles mentioned");
+
+    pub fn mention_everyone(mut self) -> Self {
+        self.mention_everyone = Some(true);
```
meditationmind:
```suggestion
    pub fn mention_everyone(mut self, mention_everyone: bool) -> Self {
        self.mention_everyone = Some(mention_everyone);
```
We generally only use one setter for each `bool` like this, rather than separate setters for `true`/`false`. So drop `no_mention_everyone` and `no_include_nsfw`. Also helps to avoid potential confusion from mistakenly using both methods..

src/builder/search_messages.rs
```diff
@@ -0,0 +1,576 @@
...
+    sequence_setters!(
+        attachment_extensions, attachment_extension, attachment_extensions, &'a str, 'a,
+        add_attachment_extension, add_attachment_extensions, attachment_extension, attachment_extensions,
+        "file extension of an attachment in the message", "file extension of an attachment in the message");
+
+    pub fn sort_ascending(mut self) -> Self {
+        self.sort_by = Some(SearchSortMode::Timestamp);
+        self.sort_order = Some(SearchSortOrder::Ascending);
+        self
+    }
```
meditationmind:
```suggestion
    pub fn sort(mut self, sort: Sort) -> Self {
        match sort {
            Sort::Ascending => {
                self.sort_by = Some(SearchSortMode::Timestamp);
                self.sort_order = Some(SearchSortOrder::Ascending);
            },
            Sort::Descending => {
                self.sort_by = Some(SearchSortMode::Timestamp);
                self.sort_order = Some(SearchSortOrder::Descending);
            },
            Sort::ByRelevance => {
                self.sort_by = Some(SearchSortMode::Relevance);
                self.sort_order = None;
            },
        }
        self
    }
```
I like the idea of handling the combinations for the user, but considering the number of methods on this builder, I'd personally go for something like this. Again, this would also prevent potential confusion caused by using multiple `sort_` methods, e.g., `query.sort_ascending().sort_by_relevance()`.

src/builder/search_messages.rs
```diff
@@ -0,0 +1,576 @@
...
+        author_types, author_type, author_types, AuthorType, 'a,
+        add_author_type, add_author_types, author_type, author_types, "author type", "author types");
+    sequence_setters!(
+        mention_user_ids, user_id, user_ids, UserId, 'a,
+        add_user_mention, add_user_mentions, user_mention, user_mentions, "user mentioned", "users mentioned");
+    sequence_setters!(
+        mention_role_ids, role_id, role_ids, RoleId, 'a,
+        add_role_mention, add_role_mentions, role_mention, role_mentions, "role mentioned", "roles mentioned");
+
+    pub fn mention_everyone(mut self) -> Self {
```
meditationmind: `bool` and sort setters are all missing documentation.

src/builder/search_messages.rs
```diff
@@ -0,0 +1,576 @@
...
+}
+
+// I'm not super happy with this macro, but I don't think it can be meaningfully improved without switching to proc macros or https://crates.io/crates/pastey.
+// And I _do_ think it's better than just inlining all this stuff.
+macro_rules! sequence_setters {
+    ($field: ident, $singular: ident, $plural: ident, $t: ty, $l: lifetime,
+        $add_singular: ident, $add_plural: ident,
+        $set_singular: ident, $set_plural: ident,
+        $singular_desc: expr, $plural_desc: expr) => {
+        #[doc = "Add a "]
```
meditationmind:
```suggestion
        #[doc = "Add "]
```
Small nitpick/peeve, but I would remove the "a" here and in L76 and handle the articles below in the invocations instead. Otherwise, you end up with "a author", "a embed", etc.

meditationmind (changes requested):
This is a big one so I'll need to do several passes, but to start, a few issues below. Still thinking about the new types and their locations, so I'll come back to that in a later pass.

src/model/guild/mod.rs
inklesspen: Huh, is that spec doc reliable? The type isn't actually documented in [the docs](https://docs.discord.com/developers/resources/message#search-guild-messages)…

src/model/guild/mod.rs
inklesspen: Weird that it's required there but not in the successful response. I'll make the change.

src/model/guild/mod.rs
meditationmind: Yeah, that's the official spec. I had to check there since they didn't document the types for the error-like response. `from_secs_f64` works, so we could leave it, but `Duration` wraps a `u64` so `from_secs` is better if you can use it. (Practically speaking, pretty much 0% chance Discord will ever give us something that makes the f64 method panic.)

inklesspen: Done.

src/builder/search_messages.rs
inklesspen: Done.

src/model/guild/mod.rs
inklesspen: Done.

src/builder/search_messages.rs
inklesspen: Done.

src/builder/search_messages.rs
inklesspen: Done.

src/builder/search_messages.rs
inklesspen: Done.

src/builder/search_messages.rs
inklesspen: Done.

src/builder/search_messages.rs
inklesspen: Done.

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+                    map.entry(message.channel_id).or_default().push(message);
+                    map
+                });
+            for (channel_id, channel_messages) in by_channel {
+                cache.fill_message_cache(channel_id, channel_messages.into_iter());
+            }
+        }
+
+        Ok(outcome)
+    }
```
meditationmind:
```suggestion
    ) -> Result<MessageSearchResults> {
        // We have to retain ownership of any Cow::Owned variants in the param pairs.
        let cow_params = self.into_param_pairs();
        let params: Vec<(&str, &str)> =
            cow_params.iter().map(|(key, val)| (*key, val.as_ref())).collect();

        let http = cache_http.http();
        loop {
            match http.search_guild_messages(guild_id, Some(params.as_slice())).await? {
                MessageSearchOutcome::NotIndexed(not_indexed) => {
                    let duration = std::time::Duration::from_secs(not_indexed.retry_after.max(1));
                    tracing::debug!("Index not yet available, retrying in {duration:?}s");
                    tokio::time::sleep(duration).await;
                },
                MessageSearchOutcome::Results(results) => {
                    #[cfg(feature = "cache")]
                    if let Some(cache) = cache_http.cache()
                        && should_cache == ShouldCache::Yes
                    {
                        let by_channel: HashMap<GenericChannelId, Vec<Message>> = results
                            .messages
                            .iter()
                            .cloned()
                            .fold(HashMap::new(), |mut map, message| {
                                map.entry(message.channel_id).or_default().push(message);
                                map
                            });
                        for (channel_id, channel_messages) in by_channel {
                            cache.fill_message_cache(channel_id, channel_messages.into_iter());
                        }
                    }
                    return Ok(results);
                },
            }
        }
    }
```
I think this should work. Unfortunately, I haven't been able to trigger the index not available condition for testing, but the success condition works fine. If we go with this, `MessageSearchNotIndexed::retry_after` should be changed to a `u64` and `GuildId::search_messages` should also return `MessageSearchResults`.

src/http/client.rs
```diff
@@ -4564,6 +4564,34 @@ impl Http {
         .await
     }
 
+    /// Returns [`Message`]s in a [`Guild`] according to various search criteria.
```
meditationmind:
```suggestion
    /// Returns [`Message`]s in a [`Guild`] according to various search criteria.
    ///
    /// **Note**: Requires the [Message Content] privileged intent and the [Read Message History]
    /// permission.
    ///
    /// [Message Content]: GatewayIntents::MESSAGE_CONTENT
    /// [Read Message History]: Permissions::READ_MESSAGE_HISTORY
```

src/model/guild/guild_id.rs
```diff
@@ -1728,6 +1730,30 @@ impl GuildId {
...
+    /// be grouped by channel before being added to the cache.
+    ///
+    /// **Note**: If the user does not have the [Read Message History] permission, returns an
+    /// outcome with an empty [`Vec`].
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::Http`] if the current user lacks permission.
+    ///
+    /// [Read Message History]: Permissions::READ_MESSAGE_HISTORY
```
meditationmind:
```suggestion
    /// **Note**: Requires the [Message Content] privileged intent and the [Read Message History]
    /// permission.
    ///
    /// # Errors
    ///
    /// Returns [`Error::Http`] if the current user lacks permission.
    ///
    /// [Message Content]: GatewayIntents::MESSAGE_CONTENT
    /// [Read Message History]: Permissions::READ_MESSAGE_HISTORY
```

I get "missing access" without read history, and docs also say it requires the permission.

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+    /// be grouped by channel before being added to the cache.
+    ///
+    /// **Note**: If the user does not have the [Read Message History] permission, returns an
+    /// outcome with an empty [`Vec`].
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::Http`] if the current user lacks permission.
+    ///
+    /// [Read Message History]: Permissions::READ_MESSAGE_HISTORY
```
meditationmind:
```suggestion
    /// **Note**: Requires the [Message Content] privileged intent and the [Read Message History]
    /// permission.
    ///
    /// # Errors
    ///
    /// Returns [`Error::Http`] if the current user lacks permission.
    ///
    /// [Message Content]: GatewayIntents::MESSAGE_CONTENT
    /// [Read Message History]: Permissions::READ_MESSAGE_HISTORY
```

meditationmind (commented):
(Personal opinions below. Feel free to disagree.)

> They're currently located in `src/model/guild/mod.rs`...

This seems fine to me, since the search is specific to the guild.

> But the builder is currently `MessageQuery` because I had this feeling that "query" was better as a standalone noun there. Still, for consistency's sake, they should probably all be the same, right?

I'm fine with either and I do like conciseness, but how about `MessageSearchQuery`? That way all the types begin with `MessageSearch` so it's clear what this query belongs to.

> In theory we could implement retry logic in the builder's execute method...

After thinking about it, I think I prefer this. Feels weird to handle rate limiting on everything else and then suddenly leave it up to the user to handle retries here. We can leave the `Http` method as is to give users the option to handle outcomes on their own, but handle the retry for them in the higher-level builder functions. Seems pretty straightforward to implement. See below.

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+    include_nsfw: Option<bool>,
+}
+
+// I'm not super happy with this macro, but I don't think it can be meaningfully improved without switching to proc macros or https://crates.io/crates/pastey.
+// And I _do_ think it's better than just inlining all this stuff.
+macro_rules! sequence_setters {
+    ($field: ident, $singular: ident, $plural: ident, $t: ty, $l: lifetime,
+        $add_singular: ident, $add_plural: ident,
+        $set_singular: ident, $set_plural: ident,
+        $singular_desc: expr, $plural_desc: expr) => {
```
meditationmind:
```suggestion
        $singular_desc: expr, $plural_desc: expr$(, $max_desc: expr)?) => {
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+        #[doc = "Add "]
+        #[doc = $singular_desc]
+        #[doc = " to the search.\n\n**Note**: This will keep all existing "]
```
meditationmind:
```suggestion
        #[doc = " to the search filter.\n\n"]
        $( #[doc = concat!("Maximum of ", $max_desc, " ", $plural_desc, " per search.\n\n")] )?
        #[doc = "**Note**: This will keep all existing "]
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+        #[doc = " to replace existing "]
+        #[doc = concat!($plural_desc, ".\n")]
+        pub fn $add_singular(mut self, $singular: $t) -> Self {
+            self.$field.to_mut().push($singular);
+            self
+        }
+
+        #[doc = "Add multiple "]
+        #[doc = $plural_desc]
+        #[doc = " to the search.\n\n**Note**: This will keep all existing "]
```
meditationmind:
```suggestion
        #[doc = " to the search filter.\n\n"]
        $( #[doc = concat!("Maximum of ", $max_desc, " ", $plural_desc, " per search.\n\n")] )?
        #[doc = "**Note**: This will keep all existing "]
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+        #[doc = " to replace existing "]
+        #[doc = concat!($plural_desc, ".\n")]
+        pub fn $add_plural(mut self, $plural: impl IntoIterator<Item = $t>) -> Self {
+            self.$field.to_mut().extend($plural);
+            self
+        }
+
+        #[doc = "Set "]
+        #[doc = $singular_desc]
+        #[doc = " in the search.\n\n**Note**: This will replace all existing "]
```
meditationmind:
```suggestion
        #[doc = " in the search filter.\n\n"]
        $( #[doc = concat!("Maximum of ", $max_desc, " ", $plural_desc, " per search.\n\n")] )?
        #[doc = "**Note**: This will replace all existing "]
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+        #[doc = concat!("[`Self::", stringify!($add_singular), "()`]")]
+        #[doc = " to keep existing "]
+        #[doc = concat!($plural_desc, ".\n")]
+        pub fn $set_singular(self, $singular: $t) -> Self {
+            self.$set_plural(vec![$singular])
+        }
+
+        #[doc = "Set multiple "]
+        #[doc = $plural_desc]
+        #[doc = " in the search.\n\n**Note**: This will replace all existing "]
```
meditationmind:
```suggestion
        #[doc = " in the search filter.\n\n"]
        $( #[doc = concat!("Maximum of ", $max_desc, " ", $plural_desc, " per search.\n\n")] )?
        #[doc = "**Note**: This will replace all existing "]
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+    /// **Note**: This field is capped to 100 due to a Discord limitation. If an amount larger than
+    /// 100 is supplied, it will be truncated.
+    pub fn slop(mut self, slop: u8) -> Self {
+        self.slop = NonMaxU8::new(slop.min(100));
+        self
+    }
+
+    sequence_setters!(
+        channel_ids, channel_id, channel_ids, GenericChannelId, 'a,
+        add_channel_id, add_channel_ids, channel_id, channel_ids, "a channel/thread", "channels/threads");
```
meditationmind:
```suggestion
        add_channel_id, add_channel_ids, channel_id, channel_ids, "a channel/thread", "channels/threads", "500");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    sequence_setters!(
+        author_types, author_type, author_types, AuthorType, 'a,
+        add_author_type, add_author_types, author_type, author_types, "an author type", "author types");
+    sequence_setters!(
+        author_ids, user_id, user_ids, UserId, 'a,
+        add_author_id, add_author_ids, author_id, author_ids, "an author", "authors");
```
meditationmind:
```suggestion
        add_author_id, add_author_ids, author_id, author_ids, "an author", "authors", "100");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+    sequence_setters!(
+        mention_user_ids, user_id, user_ids, UserId, 'a,
+        add_user_mention, add_user_mentions, user_mention, user_mentions, "a user mentioned", "users mentioned");
```
meditationmind:
```suggestion
        add_user_mention, add_user_mentions, user_mention, user_mentions, "a user mention", "user mentions", "100");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+    sequence_setters!(
+        mention_role_ids, role_id, role_ids, RoleId, 'a,
+        add_role_mention, add_role_mentions, role_mention, role_mentions, "a role mentioned", "roles mentioned");
```
meditationmind:
```suggestion
        add_role_mention, add_role_mentions, role_mention, role_mentions, "a role mention", "role mentions", "100");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    /// Filter messages that do or do not mention `@everyone`.
+    pub fn mention_everyone(mut self, mention_everyone: bool) -> Self {
+        self.mention_everyone = Some(mention_everyone);
+        self
+    }
+
+    sequence_setters!(
+        replied_to_user_ids, user_replied_to, users_replied_to, UserId, 'a,
+        add_user_replied_to, add_users_replied_to, user_replied_to, users_replied_to, "a user replied to", "users replied to");
```
meditationmind:
```suggestion
        add_user_replied_to, add_users_replied_to, user_replied_to, users_replied_to, "a user replied to", "users replied to", "100");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    sequence_setters!(
+        replied_to_message_ids, message_replied_to, message_replied_to, MessageId, 'a,
+        add_message_replied_to, add_messages_replied_to, message_replied_to, messages_replied_to, "a message replied to", "messages replied to");
```
meditationmind:
```suggestion
        add_message_replied_to, add_messages_replied_to, message_replied_to, messages_replied_to, "a message replied to", "messages replied to", "100");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+        self
+    }
+
+    sequence_setters!(
+        has, has_item, has_items, SearchHas, 'a,
+        add_message_has_item, add_message_has_items, message_has_item, message_has_items, "an item the message has", "items the message has");
+
+    sequence_setters!(
+        embed_types, embed_type, embed_types, SearchEmbed, 'a,
+        add_embed_type, add_embed_types, embed_type, embed_types, "an embed type the message has", "embed types the message has");
```
meditationmind:
```suggestion
        add_embed_type, add_embed_types, embed_type, embed_types, "an embed type", "embed types");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    sequence_setters!(
+        embed_providers, embed_provider, embed_providers, &'a str, 'a,
+        add_embed_provider, add_embed_providers, embed_provider, embed_providers,
+        "an embed provider (case-sensitive) the message has", "embed providers (case-sensitive) the message has");
```
meditationmind:
```suggestion
        "an embed provider (case-sensitive)", "embed providers (case-sensitive)", "256 chars each and 100 total");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    sequence_setters!(
+        link_hostnames, link_hostname, link_hostnames, &'a str, 'a,
+        add_link_hostname, add_link_hostnames, link_hostname, link_hostnames,
+        "a hostname of a link in the message", "hostnames of a link in the message");
```
meditationmind:
```suggestion
        "a link hostname", "link hostnames", "256 chars each and 100 total");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    sequence_setters!(
+        attachment_filenames, attachment_filename, attachment_filenames, &'a str, 'a,
+        add_attachment_filename, add_attachment_filenames, attachment_filename, attachment_filenames,
+        "a file name of an attachment in the message", "file names of an attachment in the message");
```
meditationmind:
```suggestion
        "an attachment filename", "attachment filenames", "1024 chars each and 100");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    sequence_setters!(
+        attachment_extensions, attachment_extension, attachment_extensions, &'a str, 'a,
+        add_attachment_extension, add_attachment_extensions, attachment_extension, attachment_extensions,
+        "a file extension of an attachment in the message", "file extension of an attachment in the message");
```
meditationmind:
```suggestion
        "an attachment file extension", "attachment file extensions", "256 chars each and 100 total");
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+        if let Some(min_id) = self.min_id {
+            params.push(("min_id", Cow::from(min_id.to_string())));
+        }
+        if let Some(slop) = self.slop {
+            params.push(("slop", Cow::from(slop.to_string())));
+        }
+        if let Some(content) = self.content {
+            params.push(("content", content));
+        }
+        for channel_id in self.channel_ids.iter() {
```
meditationmind:
```suggestion
        for channel_id in self.channel_ids.iter().take(500) {
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+            params.push(("channel_id", Cow::from(channel_id.to_string())));
+        }
+        for author_type in self.author_types.iter() {
+            params.push(("author_type", Cow::from(Into::<&'static str>::into(author_type))));
+        }
+        for user_id in self.author_ids.iter() {
```
meditationmind:
```suggestion
        for user_id in self.author_ids.iter().take(100) {
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+            params.push(("author_id", Cow::from(user_id.to_string())));
+        }
+        for user_id in self.mention_user_ids.iter() {
```
meditationmind:
```suggestion
        for user_id in self.mention_user_ids.iter().take(100) {
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+            params.push(("mentions", Cow::from(user_id.to_string())));
+        }
+        for role_id in self.mention_role_ids.iter() {
```
meditationmind:
```suggestion
        for role_id in self.mention_role_ids.iter().take(100) {
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+            params.push(("mentions_role_id", Cow::from(role_id.to_string())));
+        }
+        if let Some(mention_everyone) = self.mention_everyone.map(boolean_value) {
+            params.push(("mention_everyone", Cow::from(mention_everyone)));
+        }
+        for user_id in self.replied_to_user_ids.iter() {
```
meditationmind:
```suggestion
        for user_id in self.replied_to_user_ids.iter().take(100) {
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+            params.push(("replied_to_user_id", Cow::from(user_id.to_string())));
+        }
+        for message_id in self.replied_to_message_ids.iter() {
```
meditationmind:
```suggestion
        for message_id in self.replied_to_message_ids.iter().take(100) {
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+        if let Some(pinned) = self.pinned.map(boolean_value) {
+            params.push(("pinned", Cow::from(pinned)));
+        }
+        for has in self.has.iter() {
+            params.push(("has", Cow::from(Into::<&'static str>::into(has))));
+        }
+        for embed_type in self.embed_types.iter() {
+            params.push(("embed_type", Cow::from(Into::<&'static str>::into(embed_type))));
+        }
+        for embed_provider in self.embed_providers.iter() {
```
meditationmind:
```suggestion
        for embed_provider in self.embed_providers.iter().take(100) {
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+            params.push(("embed_provider", Cow::Borrowed(*embed_provider)));
+        }
+        for link_hostname in self.link_hostnames.iter() {
```
meditationmind:
```suggestion
        for link_hostname in self.link_hostnames.iter().take(100) {
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+            params.push(("link_hostname", Cow::Borrowed(*link_hostname)));
+        }
+        for attachment_filename in self.attachment_filenames.iter() {
```
meditationmind:
```suggestion
        for attachment_filename in self.attachment_filenames.iter().take(100) {
```

src/builder/search_messages.rs
```diff
@@ -0,0 +1,596 @@
...
+            params.push(("attachment_filename", Cow::Borrowed(*attachment_filename)));
+        }
+        for attachment_extension in self.attachment_extensions.iter() {
```
meditationmind:
```suggestion
        for attachment_extension in self.attachment_extensions.iter().take(100) {
```

meditationmind (changes requested):
One final pass with some suggestions to add maximums to the macro-generated docs and cap the params accordingly. Also adjusted some of the phrasing slightly to reduce awkwardness.

PR #3611, opened 2026-09-22
Add Interaction::user

mibac138:
Adds a small helper method which was useful for me.


KiloOscarSix (approved):
LGTM 

As the topic is open, one thing the serenity team might want to consider is the bounds on helper/convenience methods inside the core crate

